PR #1, opened 2026-08-23
feat(typecheck-tests-in-the-acceptance-gate): Add a real `typecheck` th…

wighawag:
Task: `work/tasks/done/typecheck-tests-in-the-acceptance-gate.md`

Done. `pnpm typecheck` is real, covers `test/`, passes with **zero suppressions**, and the full `dorfl.json` gate now runs end to end through `pnpm test` for the first time.

## Premise check (first, as instructed)

- `grep -rn '"typecheck"' package.json packages/*/package.json platforms/*/package.json` → nothing. Confirmed absent.
- `dorfl.json` `verify` still names `pnpm build && pnpm typecheck && pnpm test`. Confirmed.
- Measured baseline with an ad-hoc config: **96 errors** across `packages/` + `platforms/`, matching the observation note exactly (core 58, server 13, processor-sqlite 7, utils 6, js-processor 5, cf-worker 4, state-store-sqlite 3).

## What I built

**Per-package `tsconfig.typecheck.json`** (14 of them) extending each package's build `tsconfig.json` with `noEmit`, `declaration/declarationMap/sourceMap: false`, `rootDir: "."`, and `include: ["src/**/*.ts", "test/**/*.ts"]`. The build tsconfig is untouched, so nothing new is emitted. Per-package `"typecheck": "tsc -p tsconfig.typecheck.json"`, plus a root `typecheck` that fans out exactly like `build`/`test` (`ldenv pnpm --filter './packages/*' --filter './platforms/*' typecheck` → 14 of 22 projects, `examples/` excluded).

## The 96 errors, all fixed at the root cause

| what | count | fix |
| --- | --: | --- |
| missing `.js` extensions | 23 | added, in 12 `packages/core/test/*.ts` files — core was the only package that deviated from the `src/` convention |
| `@types/node` out of scope | 21 | `@types/node` devDependency added to `@etherfold/core`; wired via `"types": ["node"]` in the typecheck config of `core`, `server`, `state-store-sqlite` (see Decisions — TS 6 forced this) |
| implicitly-`any` callback params | 32 | all disappeared once inference was repaired; none annotated away |
| undeclared deps | 4 | `abitype` devDependency in `js-processor`; `cloudflare:test` wired via `@cloudflare/vitest-pool-workers/types` + a new `platforms/cf-worker/test/env.d.ts` merging `Cloudflare.Env` |
| genuine mismatches | 14 | one production type fix in `@etherfold/core`, the rest narrowed on the test's own side |
| cf-worker `Request`/constructable | 2 | `test/utils.ts` rewritten (this also **removed** two `as never` casts) |

The one production fix: `createAction<T, U>` in `packages/core/src/internal/utils/promises.ts` used `U extends undefined ? ... : ...`. `U` is naked, so the conditional **distributes**: `createAction<string, boolean>` produced a *union of two signatures* rather than one signature over the union. That union has no single call signature, so the executor's params fell to implicit `any` and `next(...)` demanded the intersection (`never`). Changed to the non-distributive `[U] extends [undefined]`. `indexer.d.ts` is byte-identical after the change; only the internal module's `.d.ts` moves.

## Acceptance criteria

- **`pnpm typecheck` exists, fans out, exits 0** — verified, exit 0.
- **It typechecks `test/`** — verified per package: I appended `const x: number = "not a number";` to a test file in **all 12 packages that have tests** and each one failed with exit 2 and exactly that error; each was then restored byte-for-byte.
- **`pnpm build` still emits only `src/`** — checked, not assumed: for every package, the set of `dist/**/*.js` maps 1:1 onto `src/**/*.ts`. No test file in any `dist/`. (`packages/fs-cache/dist/test.js` is pre-existing, from `src/test.ts`.)
- **Full gate passes end to end** — ran the literal `verify` string from `dorfl.json` in a throwaway clone with everything committed (the untracked-changeset caveat, below): **exit 0**, reaching `pnpm test` and 396 tests / 47 files green.
- **No suppression, widening, or strictness relaxation** — `grep` over the whole diff for `@ts-ignore` / `@ts-expect-error` / `@ts-nocheck` / `as any`: **zero added**. The repo-wide census of those directives in `packages/*/{src,test}` and `platforms/*/{src,test}` is also **zero**. `strict` untouched, `moduleResolution` untouched. Net casts: **−2** (`as never` × 2 removed in cf-worker, one documented `as` added in a named helper — see Decisions).
- **`.js` extensions in test imports** — done, and re-verified after a mid-task accident (I used `git checkout --` to undo an experiment and clobbered two files; both were re-applied and re-verified).
- **Missing devDependencies added** — `@types/node` → `core`; `abitype` → `js-processor`. `pnpm-lock.yaml` moves by 6 lines, no new resolutions.
- **Changeset** — included: one `patch` for `@etherfold/core`. Everything else is tooling-only or test-only.

## Decisions

**Typecheck configs are a separate per-package file, not a `references`/`--build` graph.** Each `tsconfig.typecheck.json` `extends` its package's build `tsconfig.json` and overrides `noEmit: true`, `declaration/declarationMap/sourceMap: false`, `rootDir: "."`, `include: ["src/**/*.ts", "test/**/*.ts"]`. Alternative considered and rejected: project references / `tsc --build`, which the task explicitly told me not to introduce and which would change how `build` resolves. Touches: any new package must add this file plus the `typecheck` script, or it is silently uncovered by the root fan-out (the root script does not fail on a package that has no `typecheck` — pnpm skips it).

**`@types/node` had to be pinned in a `types` array, contrary to the task's stated preference — because TypeScript 6.0.3 no longer auto-includes `@types/*`.** The task says "prefer [a devDependency] over pinning a `types` array, which would narrow what those packages can see". On this toolchain that is not achievable: I verified in an isolated scratch project that `tsc 6.0.3` with `@types/node` installed and no `types` field still reports `TS2591: Cannot find name 'process'`, and that adding `"types": ["node"]` fixes it. This also explains why every package in this repo that touches node builtins *already* pins `types: ["node"]`. So both are needed: the devDependency for resolution, and the `types` entry for inclusion. Since nothing is auto-included under TS 6, pinning `types` only *adds*, it no longer narrows. Touches: any future package whose tests use node builtins needs both halves, not just the dependency.

**Node types are granted in the TYPECHECK config, not the build config.** `core`, `server` and `state-store-sqlite` have `src/` that is deliberately platform-agnostic and only their `test/` reaches for `node:` builtins. Putting `"types": ["node"]` in the build `tsconfig.json` would let `src/` silently start using node builtins in a package that must run in a browser. Alternative considered: add it to the build config (simpler, one place) — rejected for that reason. `js-processor` turned out to need no node types at all (nothing in it imports `node:`), so I removed the `@types/node` I had speculatively added there; only `abitype` remains.

**Build → typecheck ordering is documented in two places a reader will actually meet it.** Cross-package types resolve through each package's `dist/`, so `pnpm typecheck` alone on a clean clone looks broken. Recorded (a) as the root `package.json` scripts section header, following the repo's existing `"---- SECTION ----": ""` separator convention: `"------- TYPECHECK (src + test; RUN AFTER build: cross-package types resolve through dist/) -------"`, and (b) as a new bullet in `CONTEXT.md` `## Conventions` explaining why `test/` is checked by nothing else. Alternative considered: JSONC comments inside each `tsconfig.typecheck.json` — rejected because no other JSON file in this repo carries comments and `*.json` is prettier-ignored, so it would be an unenforced one-off.

**`createAction`'s distributive conditional was fixed in `@etherfold/core` src rather than worked around in the test.** This is the only production change. It is a real latent defect that has been in the published package since it was written, and the test that exposes it (`createAction<string, boolean>`) has been there, unchecked, all along. Alternative considered: narrow the test to a non-union `U` — rejected, that would hide a real bug. It ships with a `patch` changeset (`.changeset/fix-core-create-action-union-arg.md`) even though `indexer.d.ts` is byte-identical, because the change lands in a published package.

**cf-worker: a test-scoped `Cloudflare.Env` declaration, not a `wrangler types` step.** `Cloudflare.Env` ships empty and is designed to be merged into. `platforms/cf-worker/test/env.d.ts` merges `CloudflareEnv` plus the test-only `TEST_MIGRATIONS` binding that `vitest.config.ts` supplies via miniflare. It lives in `test/` rather than beside the worker precisely because `TEST_MIGRATIONS` exists only under test. Alternative considered: run `wrangler types` and commit a generated `worker-configuration.d.ts` — rejected as a new generated-file convention (a standing per-change obligation) for one binding. Touches: if the worker gains a binding, this file needs it too, or `src/env.ts` does and this inherits it. Same package: `test/utils.ts` dropped `export const IncomingRequest = Request<unknown, IncomingRequestCfProperties>` (an instantiation expression that TS rejects here because `lib: ["ESNext","dom"]` in `tsconfig.base.json` merges lib.dom's non-generic `Request` over workers-types'), replacing it with a plain `new Request(url, init)`; the properly-typed `env` let two `as never` casts go.

**`packages/utils`: one named, documented `asImported()` seam instead of scattered casts.** `SAMPLE_CONTRACTS` is now properly typed `ContractData<[]>[]`, which fixes five of the six errors with no cast at all. The sixth is `it('throws when the factory returns nothing')`: it deliberately builds a module whose factory lies, which the `ProcessorModule` type by construction forbids — and that guard exists exactly because `loadProcessorModule` gets its value from a dynamic `import()` of foreign JS, which nothing checks. Rather than a `@ts-expect-error` or a bare inline cast, there is a single named helper carrying that rationale in its JSDoc, used once. Alternative considered: a `@ts-expect-error` at the call site — rejected because the situation is not "TypeScript is wrong", it is "this value genuinely comes from an unchecked boundary", and the helper says so.

**Residual `@ts-expect-error`: ZERO.** None was needed. No error required a change to published API surface, so the escape hatch the task allowed went unused.

**Changeset scope: one `@etherfold/core` patch, nothing else.** By `CONTEXT.md`'s rule only public-API changes need one, and everything outside core is tooling or tests. I verified the gate's `pnpm changeset status --since=main` is satisfied by this: its check is coarse (`changedPackages.length > 0 && changesets.length === 0`), not per-package, and with the core changeset it reports 10 packages to bump at patch (dependents via `updateInternalDependencies: "patch"`) and exits 0. **One caveat for the runner:** `changeset status` reads changesets via `git diff` against the merge base, which does **not** see untracked files, so running the gate on my uncommitted tree fails at that step. It passes once the changeset file is committed — I confirmed this by reproducing the whole gate in a throwaway clone with everything committed (exit 0). Nothing to do; just don't be surprised by a pre-commit red there.

**Two follow-ups, captured as observations rather than acted on.** `work/notes/observations/vitest-config-is-typechecked-by-nothing.md`: `platforms/cf-worker/vitest.config.ts` is the one `.ts` file at a package root, so it falls outside both `src/**` and `test/**` and is still checked by nothing; adding `*.ts` to the include would catch it but is outside this task's stated scope. `work/notes/observations/two-abitype-instances-in-the-tree.md`: `core` resolves `abitype@1.2.4` while its `viem@2.52.0` pins `1.2.3`, so both are in the tree and `core` re-exports abitype types publicly — the classic "two different types with this name exist" generator for a consumer. It is *not* what caused the errors here (that was the mapped-type inference failure) and nothing is broken today, which is why it is an observation and not a task.

**`examples/` is correctly out of scope and would not be cheap.** Checked rather than assumed: **no example has a `test/` directory at all**, and their `test` scripts are `eis run ...`, not vitest. So there is nothing in `examples/` that the typecheck gap applies to; the svelte ones would need `svelte-check`, not `tsc`. Extending scope there would add cost and cover nothing this task exists for.


wighawag:

Gate-2 was off for this run, so this is the only review layer above the acceptance gate. Three criteria said "check rather than assume", so I verified those independently in a throwaway clone rather than trusting the report.

### Criteria

| # | criterion | verdict |
| --- | --- | --- |
| 1 | `pnpm typecheck` exists, fans out across `packages/`+`platforms/` like `build`/`test`, exits 0 | ✅ verified independently, exit 0 |
| 2 | Covers `test/**`; a deliberate error in a test file fails it | ✅ **verified by injection** (see below) |
| 3 | `pnpm build` still emits only `src/`, no test file in any `dist/` | ✅ verified independently, zero test artifacts |
| 4 | Full gate passes end to end, including `pnpm test` | ✅ reached `pnpm test` for the first time; all suites green |
| 5 | No suppression, widening, or relaxed strictness | ✅ zero `@ts-expect-error`, no `strict`/`moduleResolution` touched |
| 6 | Test relative imports use `.js`, matching `src/` | ✅ |
| 7 | Missing `@types/node` and undeclared test deps added | ✅ |
| 8 | Changeset where the published surface changed | ✅ one `@etherfold/core` patch, scope justified |

### Criterion 2 is the load-bearing one, so I did not take it on trust

Injected `const __gate3_probe: number = "a string, deliberately wrong";` into `packages/state-store-sqlite/test/schema.test.ts` and re-ran:

```
packages/state-store-sqlite typecheck: test/schema.test.ts(111,7): error TS2322: Type 'string' is not assignable to type 'number'.
```

It catches it. The typecheck is real, not decoration. Also confirmed every package with a `test/` directory has a `typecheck` script, so nothing is silently uncovered by the root fan-out.

### The one deviation from the task, accepted

The task said to prefer a devDependency over pinning a `types` array. This pins `"types": ["node"]` anyway — but the justification is verified rather than asserted: TypeScript 6.0.3 no longer auto-includes `@types/*`, confirmed in an isolated scratch project, which is also why every package here that touches node builtins already pins it. Under TS 6 pinning only *adds*, so the stated concern (narrowing) no longer applies. Correct call, and granting node types in the **typecheck** config rather than the build config is a better decision than the task asked for: it keeps `core`/`server`/`state-store-sqlite` `src/` from silently acquiring node builtins in packages that must run in a browser.

### The one cast, accepted

`asImported()` in `packages/utils/test/processorSetup.test.ts` is a single-use, named, JSDoc'd helper modelling a genuine unchecked boundary (`loadProcessorModule` receives its value from a dynamic `import()` of foreign JS, and the guard under test exists precisely because that shape can lie). That is the "narrow the test's own types / honest boundary" option the task permitted, not an `as any` backdoor. Five of the six errors in that file were fixed with no cast at all by typing `SAMPLE_CONTRACTS` properly.

### Beyond the ask

The new typecheck immediately earned its keep by finding a **real latent bug in a published package**: `createAction<T, U>` used the naked-parameter form `U extends undefined ? …`, which distributes, so a union `U` (e.g. `boolean`) produced a *union of signatures* instead of one signature taking the union — collapsing the executor's parameters to implicit `any` and making `next(...)` demand `never`. `test/promises.test.ts` had been calling `createAction<string, boolean>` since it was written and nothing checked it. Fixed with the non-distributive `[U] extends [undefined]`, type-level only, no runtime change, with an explanatory comment and an accurate changeset. Fixing it rather than narrowing the test to hide it is exactly right.

Two follow-ups were captured as observations rather than scope-crept: `vitest-config-is-typechecked-by-nothing` (the one package-root `.ts` still outside both globs) and `two-abitype-instances-in-the-tree` (1.2.4 direct vs 1.2.3 under viem; explicitly noted as *not* the cause of anything fixed here). Both correctly filed, neither acted on.

`examples/` was checked rather than assumed out of scope: no example has a `test/` directory and their `test` scripts are `eis run`, not vitest.

Merging.

PR #2, opened 2026-08-23
feat(portable-mutation-context-seam): Lift MutationContext out of proce…

wighawag:
Task: `work/tasks/done/portable-mutation-context-seam.md`

The full acceptance gate is green (`format:check && changeset status && build && typecheck && test`, exit 0).

## What I did

I continued from the prior attempt's branch rather than restarting. That branch was a false red: it was bounced only because `pnpm typecheck` did not exist, so the gate short-circuited before `pnpm test`. With the real typecheck now on `main`, three genuine problems surfaced, and I fixed those plus reviewed the existing work.

**1. The deprecated `SQLProcessor` alias silently broke ABI inference** (`packages/processor-sqlite/src/types.ts`). 34 typecheck errors across `test/lifecycle.test.ts` and `test/version.test.ts`, all of the form "`SQLProcessor<TestABI>` is not assignable to `EntityProcessor<Abi, undefined>`". The cause is not variance: `ABI` was never *inferred*, it fell back to its `Abi` constraint. `EntityProcessor` is an intersection whose handler half is a mapped type with remapped keys (`on${EventName}`), which offers no inference site, so TypeScript relies on its shortcut for a source and target sharing an alias *symbol*. `export type SQLProcessor<ABI, C> = EntityProcessor<ABI, C>` declares a new symbol and the shortcut misses; `export type {EntityProcessor as SQLProcessor}` resolves to the same symbol and it fires. Confirmed with a minimal repro before changing anything, and this was a real published-API regression the prior attempt introduced: any user with `const p: SQLProcessor<typeof abi>` would have had `new VersionedStateEventProcessor(db, p)` rejected. Fixed by making it a re-export, with a JSDoc at the site spelling out why it must not be "cleaned up" back into a type alias.

**2. Pinned that fix.** `packages/processor-sqlite/test/authoring-types.test.ts` (new) asserts at compile time that the deprecated name still infers, through both the constructor and `fromSQLProcessor`. Verified it actually bites: reverting the export to a `type =` alias produces 5 errors in that file specifically (39 in the package), and restoring it goes green.

**3. The two new packages were invisible to the typecheck fan-out**, exactly as the handoff warned. Added `tsconfig.typecheck.json` + a `typecheck` script to `packages/state-store` and `packages/processor-entities`, copied from the sibling pattern. That immediately caught two implicit-`any` handler parameters in `processor-entities/test/two-backends.test.ts`, from an inline processor literal (same non-inferrable mapped type). Fixed by hoisting it to an annotated `const`, matching the style the sibling suite already uses.

No `as any`, no `@ts-ignore`, no `strict` relaxation, no public type widened to silence a call site.

**4. Reviewed the inherited work** against every acceptance criterion; all are met and the reasoning in the moved doc comments survived the move intact. One inaccuracy in the prior attempt's own ADR-0018: its naming table lists `@etherfold/processor-js`, which does not exist (it ships as `@etherfold/js-processor`). Footnoted it honestly rather than silently rewriting the table, since the underlying drift is between ADR-0016 and ADR-0017 and is out of scope. Captured that drift in `work/notes/observations/adr-0016-rename-to-processor-js-never-happened.md`.

Final state: 16 `processor-entities` tests, 26 `state-store`, 71 `state-store-sqlite`, 64 `processor-sqlite`, whole monorepo green, working tree carries only the intended changes.

## Decisions

Carried over from the prior attempt on this branch (reviewed and endorsed, restated here because the task body on `main` does not carry them):

- **The seam is TWO packages, not one: `@etherfold/state-store` (contract) and `@etherfold/processor-entities` (ABI-typed authoring).** The `on<EventName>` map needs `LogEvent`/abitype from `@etherfold/core`, and core depends on viem. A single package would force `@etherfold/state-store-sqlite` either to depend transitively on the whole indexer core (inverting ADR-0016's stated direction and its "imports nothing but remote-sql and named-logs" test) or to keep a duplicate, structurally-compatible copy of `EntityDeclaration`/`Mutation`/`BlockPointer`. Both alternatives rejected; the split keeps every definition singular and the store a primitive. Touches ADR-0016's dependency assertion, `state-store-sqlite`'s leakage test, and every future backend package (`light-store-behind-the-seam`, `indexeddb-row-backend-browser-default`), which now depends on `@etherfold/state-store` alone. Recorded as **ADR-0018** (`docs/adr/0018-the-storage-seam-is-two-packages.md`).
- **Names follow ADR-0014's role-first axis with the trailing slot filled by a non-engine.** `state-store` = the role with no backend, i.e. the contract; `processor-entities` = a processor whose state lives in declared entities. `entity-processor` and `processor-api` were rejected for breaking the ordering rule that keeps one role's variants adjacent in a scope listing.
- **What moved vs. what stayed.** Moved to the seam: entity declaration + `FieldType` + `normalizeEntity`/`idValues`, `EntityId`, `Mutation`, `BlockPointer`/`BlockUpdate`, `normalizeBlockHash`/`normalizeBlockTimestamp`, `MutationContext` + staging, and (to `processor-entities`) the handler types, `forkPoint`, `groupByBlock`, `blockPointer`, and the revert-then-apply loop. Stayed SQL-specific: `Statement`, DDL, batching, `queryCurrent`/`queryAsOf` (caller-supplied SQL), block addressing by hash/timestamp, `NoSuchBlockError`, and the `_sync` cursor. Every moved symbol is re-exported from its old package, so no existing import broke.
- **`StateStore.getAsOf` takes a resolved block NUMBER, not a `BlockAddress`.** Keeps `NoSuchBlockError` from having to be duplicated (two classes of one name would break `instanceof` across packages, ADR-0015's family) and leaves `retention-capability-and-refusal` free to decide where the refusal family lives. Alternative considered: moving addressing to the seam now, rejected as out of scope and premature. Touches `retention-capability-and-refusal` and `state-store-conformance-suite`.
- **The interface is named `StateStore`, not `StateBackend`.** ADR-0014 already uses "backend" to mean the storage ENGINE named in a package (`sqlite`), so naming the role-level interface `StateStore` avoids re-meaning it. Prose keeps "backend" for the engine.
- **`MemoryStateStore` ships inside the contract package** rather than as a third `state-store-memory` package: it is the contract's executable definition and the second backend a portability test needs, not a production backend.
- **`update` was added to `MutationContext`** (spec decision 3 and acceptance criterion 4 require it; it did not exist). It strips to declared fields before spreading, so a backend's version columns cannot travel back in as values, pinned by a test on both backends.
- **`ColumnType` → `FieldType`,** with `ColumnType` kept as a deprecated alias in `state-store-sqlite`: the four values are the intersection of what all backends can hold, not SQLite storage classes, and the declaration already calls them `fields`. Touches `tagged-bigint-codec-across-storage-adapters`, which now has a backend-neutral type to evolve.
- **Identifier validation moved to the seam and now applies to every backend.** This makes "valid declaration" a fact about the declaration rather than the deployment, which is half the fix for the finding's `index`-keyword defect; the other half remains `entity-identifier-sql-keyword` (unchanged, untouched).

New in this run:

- **A deprecated generic type alias at a package boundary must be a RE-EXPORT, not a `type X<A> = Y<A>` declaration.** `SQLProcessor` is now `export type {EntityProcessor as SQLProcessor}`. The alternative (the obvious `type SQLProcessor<ABI, C> = EntityProcessor<ABI, C>`, which is what the prior attempt wrote) is not equivalent: it declares a new alias symbol, TypeScript's alias-symbol inference shortcut misses, and `ABI` silently falls back to its `Abi` constraint at every generic call site, because `EntityProcessor`'s handler half is a key-remapped mapped type with no inference site. The user-visible damage was a published deprecated type that compiles on its own but is rejected by the very constructor it exists to feed, with an error naming handler variance and not the cause. This touches the public surface of `@etherfold/processor-sqlite` and the same trap applies to any future compat alias of `EntityProcessor`, `EventHandlers` or `MutationContext`, so it is documented at the export site and pinned by `packages/processor-sqlite/test/authoring-types.test.ts`. Considered and rejected: annotating the test fixtures with `EntityProcessor` instead (hides the regression from users while making the suite green), and dropping `SQLProcessor` entirely (permitted by the spec's Out of Scope, but it discards a working compat path for no gain now that the re-export form works).
- **ADR-0018's naming table keeps `@etherfold/processor-js` and footnotes it, rather than being corrected to the shipped `@etherfold/js-processor`.** The row is ADR-0016's name for that package, not what it publishes under, because ADR-0017's rename carried the old word order over. Correcting the row would make the table's own adjacency argument false; deleting it would hide a live inconsistency between two accepted ADRs. Touches ADR-0016 and ADR-0017, neither of which I amended; the drift is recorded in `work/notes/observations/adr-0016-rename-to-processor-js-never-happened.md` for someone to either finish the rename or amend ADR-0016.


wighawag:

Gate-2 is off for this drive, so this is the only review layer above the acceptance gate. This branch is the continuation of the one bounced by the phantom-`typecheck` false red; it correctly built on the kept tip rather than restarting (`processor-sqlite` 62 → 64 tests, 7 → 8 files).

### Criteria

| # | criterion | verdict |
| --- | --- | --- |
| 1 | One processor, unmodified, on two backends; same declarations, handlers, state | ✅ `two-backends.test.ts` |
| 2 | Authoring types live in a non-SQL-named package; `processor-sqlite` consumes them | ✅ `@etherfold/state-store` + `@etherfold/processor-entities` |
| 3 | Capabilities as data, reachable before any read | ✅ getter, asserted on both backends |
| 4 | `set` whole row, `update` sugar; unlisted field NULLed by `set`, survives `update` | ✅ pinned at the seam *and* at the two-backend level |
| 5 | Read-your-writes tested at the seam, not only in the SQLite impl | ✅ dedicated suite in `state-store` |
| 6 | Existing suites still pass unchanged in meaning | ✅ additions only, see below |
| 7 | Tests in `test/`, vitest, changeset per package whose surface moved | ✅ one changeset, all four packages |

### The constraints I was watching, all held

- **Retention has exactly one unit.** `Retention` is `revert-only | {kind:'window', blocks:number} | unbounded`. No duration, no "last N updates" kind. The only occurrences of "time"/"updates" in the module are the doc comment explaining *why* they are excluded, carrying the median-429-blocks sparsity result and the wall-clock-vs-chain-progress argument to where a future reader meets them.
- **The SQLite store does not claim a window it cannot enforce.** `{retention: {kind: 'unbounded'}, asOf: true}`, with a comment naming `retention-capability-and-refusal` and `prune-versions-outside-retention-window` as what makes the report load-bearing. Exactly right while pruning does not exist.
- Prefix listing and the typed refusal are correctly *absent*; they are later tasks and the seam leaves room for both.

### Criterion 6 deserves specific note, because it was handled unusually well

The only modification to an existing test is `no-platform-leakage.test.ts`, which had to gain `@etherfold/state-store` in its allow-list. An allow-list extension is normally where a guarantee quietly erodes. Here the hole it opens is closed in the same diff by a **new** assertion that the seam package declares zero dependencies, so "a storage primitive must not drag in `@etherfold/core`/viem" is preserved *transitively* rather than weakened, with the reasoning added to the file header. Everything else in `processor-sqlite/test` and `state-store-sqlite/test` is a pure addition.

### The `SQLProcessor` change is a root-cause fix, not a widening

I checked this closely, because "make the new typecheck happy" is exactly where a public type gets widened to silence a call site. It is the opposite. The errors reported `SQLProcessor<Abi, …>` — `ABI` collapsed to its *constraint* — and the diagnosis is precise: `EntityProcessor`'s handler half is a mapped type with remapped keys (`on${EventName}`), which offers no inference site, so TypeScript relies on its shortcut for a source and target sharing an alias **symbol**. A re-export specifier preserves that symbol; `export type X<A> = Y<A>` mints a new one, the shortcut misses, and inference silently falls back. Replacing the alias declaration with `export type {EntityProcessor as SQLProcessor}` restores inference, the reasoning is recorded in a "do NOT rewrite this as…" comment, and `authoring-types.test.ts` pins it against regression. Net production `src` change beyond the two new packages: none.

### Other things done right

- Both new packages ship `tsconfig.typecheck.json` + a `typecheck` script, so the root fan-out actually covers them rather than silently skipping them.
- `two-backends.test.ts` includes `produces the state the handlers describe, so "the same" is not "the same wrong"` — an independent oracle guarding against both backends agreeing on a wrong answer. That is a sharper test than the criterion asked for.
- The reorg counter-decrease case is already exercised on both backends, ahead of the conformance suite that owns it.
- ADR-0018 justifies the two-package split against ADR-0016's dependency-direction rule, and explains why `MemoryStateStore` ships inside the contract package.
- Filed `adr-0016-rename-to-processor-js-never-happened.md` as an observation rather than scope-creeping into a rename: ADR-0016 and ADR-0017 name the same package differently and the shipped name is the one ADR-0016 argues against. Correctly captured, not acted on.

Merging.

PR #3, opened 2026-08-23
feat(retention-capability-and-refusal): Retention declared in blocks, w…

wighawag:
Task: `work/tasks/done/retention-capability-and-refusal.md`

All checks green. Here is the summary.

## What I did

- `src/retention.ts` (new): `RetentionSetting` (`'revert-only' | 'unbounded' | {blocks: N}`), `resolveRetention(setting, {finalityDepth})` (validates: floor, integer, no durations, no updates), `retainedRange(retention, tip)`, `retentionWithoutPruning(requested)`, and `assertRetained(capabilities, requested, tipThunk)` — the one refusal every backend shares.
- `src/errors.ts` (new): `BlockUnavailableError` (family base) and `BlockNotRetainedError` carrying `requested`, `retained`, `reason` (`outside-window` | `no-historical-reads`) and `retention`.
- `MemoryStateStore` takes `{retention, finalityDepth}`, reports what it provides, and refuses as-of reads it does not cover.

**`@etherfold/state-store-sqlite`**
- `NoSuchBlockError` now extends `BlockUnavailableError` (joins the family rather than reinventing it); both errors re-exported from this package.
- `VersionedStateStoreOptions` gains `retention` / `finalityDepth`; `capabilities` reports what is provided; `getAsOf` and `queryAsOf` guard through `assertRetained`; `latestBlockStatement()` gives the tip, read only when a window is claimed.

**`@etherfold/processor-sqlite`**
- `VersionedStateProcessorOptions` (was `Record<string, never>`) carries retention through to the store; `fromSQLProcessor(processor, options)`; `load` reconciles the declared floor with `streamConfig.finality`; `VersionedStateView.capabilities` so a consumer discovers history at startup.

**Tests** (written before the production code at the seam): `state-store/test/retention.test.ts` (+ `test/utils/windowed-store.ts`), rewritten `state-store-sqlite/test/capabilities.test.ts`, new `processor-sqlite/test/retention.test.ts`, and cross-backend capability cases in `processor-entities/test/two-backends.test.ts`. Full monorepo: `format:check`, `build`, `typecheck`, `test` all green (57 / 81 / 71 / 24 in the touched packages). Changeset added for the four packages; `docs/adr/0019-...` written; CONTEXT.md glossary updated. No pruning was implemented; the SQLite store still reports `unbounded`.

(Note for the runner: the task body is at `work/tasks/backlog/retention-capability-and-refusal.md`, not `ready/`. I did not move it.)

## Decisions

- **Default retention is `unbounded`, and there is no numeric default window.** No shipped store prunes, so `unbounded` is the only report true of them, and a default window would be both a claim nothing enforces and — at any size that reads as generous — nearly empty in updates (median 429 blocks between event-bearing blocks; 64 blocks holds one). Alternative considered: defaulting to the finality depth, rejected because it silently turns every existing store into one that refuses almost all history. Touches: every deployment that does not set retention (behaviour unchanged from today), and `prune-versions-outside-retention-window`, which inherits this default.
- **A window is spelled `{blocks: N}`; a bare number is refused, and `finalityDepth` is REQUIRED alongside a window.** The key names the unit at every call site, which is the cheapest defence against the "N blocks = N updates" trap, and a window is only meaningful next to the floor it must not go under. Alternative: accept `retention: 128` — rejected (a bare number names no unit) — and defaulting the floor to core's `finality: 17` — rejected, because a store silently assuming the core's stream default would validate against a number the deployment never stated. Touches every future backend's options type.
- **The refusal is `BlockNotRetainedError` under a new shared base `BlockUnavailableError`, with `NoSuchBlockError` re-parented onto it.** ADR-0015 settled the argument once; a retention boundary is the second way a historical read fails, so it is a member of that family, not a new one. The base lives at the seam because every backend throws the retention member and two classes of one name in two packages would break `instanceof`. Alternatives: a `reason` on `NoSuchBlockError` (rejected: the address resolved fine, so the name would lie) and a discriminated result (rejected for the same reason ADR-0015 rejected it — it changes every read's return type). Carries `requested`, `retained` (`{from, to}` block numbers, `undefined` for `revert-only`), `reason`, `retention`. Touches `state-store-conformance-suite`, `prune-versions-outside-retention-window`, `light-store-behind-the-seam`, and any HTTP read layer that must map it to a status.
- **A store that cannot prune ACCEPTS a window, validates it, and reports `unbounded` (with a warning on SQLite), rather than throwing.** Acceptance criterion 1 requires a deployment to be able to set a block count; refusing it at configuration would make that unreachable until pruning lands, and the report-vs-setting distinction is exactly what the criterion describes. Understating retention is the safe direction (a caller relies on less history than exists). Alternative: throw "this store cannot provide a window" — louder about the storage expectation, rejected because it makes a correct-in-intent config fail at boot. Touches `prune-versions-outside-retention-window`, which flips the downgrade off.
- **`revert-only` is honoured by BOTH shipped stores, including SQLite.** It is enforceable with no pruning at all (refuse every as-of read, keep reverting), and it is how the refusal a patch-log backend will produce is exercised before that backend exists. Consequence worth naming: a SQLite store set to `revert-only` refuses history while still physically holding it — the report is about what a caller may rely on, not about bytes on disk.
- **A new refusal at `load`: the store's declared `finalityDepth` vs the stream's `finality`.** The window is validated at construction against the depth the deployment declared, but the depth a reorg actually reaches comes from `UsedStreamConfig.finality`, so the two can disagree and the disagreement is silent until a deep reorg. It raises only for a deployment that opted into retention (the default states no floor). Alternative: warn instead of raise — rejected, the failure it guards against is unrecoverable state corruption. Coherence note: `finality` already means the stream's reorg window in this repo; the store option is deliberately named `finalityDepth` ("the depth this retention protects") and reconciled with the stream's number rather than becoming a second, independent meaning.
- **No "last N updates" resolver was added.** Nothing asks for one today, and the honest implementation needs an indexed "blocks that changed something" lookup the seam does not expose (`bounded-id-prefix-listing` is the read side that would make it cheap). Adding it now would either force a new backend method or a full scan. ADR-0019 records the shape it must take when it arrives: a helper above the seam returning a block number, invisible below it.
- **The window refusal is exercised through test doubles, not a shipped configuration**, since no shipped store may honestly claim a window yet. The cross-backend capability cases live in `processor-entities/test/two-backends.test.ts` for now; `state-store-conformance-suite` should absorb them into the suite it exports rather than duplicating them.
- **ADR-0019** (`docs/adr/0019-retention-is-block-numbers-and-an-unretained-read-is-refused.md`) records the one-unit rule, the 429-block trap, the refusal family, and report-what-you-provide. CONTEXT.md's `retention` entry was extended and a `BlockUnavailableError` entry added, so the new vocabulary is in the glossary rather than only in code.


wighawag:

This task carries three of the four constraints this drive was told to guard hardest, so it got the closest reading. All three hold, and two are enforced more strongly than the criteria required.

### Criteria

| # | criterion | verdict |
| --- | --- | --- |
| 1 | Deployment sets a block count / `unbounded` / `revert-only`; backend reports what it provides, before any read | ✅ |
| 2 | A window below the finality depth is rejected at config time, naming both numbers | ✅ names both, explains the floor, says what to set |
| 3 | Out-of-window as-of read throws a typed `NoSuchBlockError`-family error carrying requested block + retained range; never the tip, never `undefined` | ✅ `BlockNotRetainedError` |
| 4 | A backend that cannot answer as-of reports `revert-only`, refuses every historical read, `revertTo` still works | ✅ including a read at its own tip |
| 5 | No API accepts a duration; no retention kind in updates; a test pins that only a block number reaches a backend | ✅ actively refused, not merely absent |
| 6 | SQLite reports `unbounded` and claims no window while pruning is unimplemented | ✅ plus a startup warning |
| 7 | Conformance suite capability cases pass on both stores | ✅ handled honestly — see below |
| 8 | Tests in `test/`, vitest, changeset | ✅ |

### The three hard constraints

**One unit, block numbers.** `Retention` remains exactly `revert-only | {kind:'window', blocks:number} | unbounded`. Durations and update-counts are not simply absent from the type: `parseRetention` **refuses them loudly**, on every spelling (`{seconds}`, `{days}`, `{updates}`, and a bare number, which names no unit), each with a message explaining why. `refuses every duration, on every spelling` and `refuses a count of updates, which is not a distance in blocks` pin it. Stronger than the criterion asked for.

**A read the backend cannot serve is a typed error.** `BlockNotRetainedError` carries `requested`, `retained` (`{from,to}`, or `undefined` for `revert-only`), a `reason` discriminating `outside-window` from `no-historical-reads`, and the `retention` itself. It is documented as deliberately neither the tip nor `undefined`, with the reasoning that `undefined` would read as "absent then", an ordinary answer a caller acts on normally. `never answers a historical read from the tip, which is the whole point` is a named test. Spelling `retained` as `undefined` rather than an empty range, so nobody does arithmetic on a boundary that does not exist, is a good call.

The family is genuinely unified rather than parallel: `NoSuchBlockError` is **re-parented** onto a new shared `BlockUnavailableError` base, which lives at the seam precisely because two classes of one name in two packages would break `instanceof` across the boundary. That is the right place for the right reason.

**SQLite claims no window.** `retentionWithoutPruning` downgrades a requested window to `unbounded`, with the reasoning recorded where it is applied: understating retention is the safe direction, because a caller relies on *less* history than exists, whereas claiming the window would promise a bound nothing enforces. A window is still accepted and validated (so criterion 1 stays reachable), and SQLite logs a warning making the gap explicit at startup rather than at the call. The helper is named and documented so it "disappears from a store the day that store prunes", which hands `prune-versions-outside-retention-window` a clean seam.

### Criterion 7 was handled honestly

The conformance suite is `state-store-conformance-suite`, which has not been built yet, so this criterion was not fully reachable. Rather than fudging it, the cross-backend capability cases were placed in `processor-entities/test/two-backends.test.ts` with an explicit forward-note that the suite should **absorb** them rather than duplicate. Recorded here so it can be enforced: I will check at that task's Gate-3 that they move rather than fork.

### Scope expansion, reviewed and accepted

A new refusal at `load` reconciles the store's declared `finalityDepth` against the stream's `UsedStreamConfig.finality`. It is beyond the literal criteria, but it is the same defect class criterion 2 exists for (a floor that is silently wrong), the failure it guards is unrecoverable state corruption rather than a bad answer, and it raises only for a deployment that opted into retention, so default deployments are unaffected. The naming asymmetry (`finality` for the stream's window, `finalityDepth` for the depth retention protects) is called out and reconciled rather than left as two independent meanings.

No "last N updates" resolver was added, which the task left to the builder's judgement. Correctly not added as a retention kind, and ADR-0019 records the shape it must take if it ever arrives: a helper above the seam returning a block number, invisible below it.

ADR-0019 carries the one-unit rule, the 429-block trap, the refusal family and report-what-you-provide, and `CONTEXT.md`'s glossary gained `BlockUnavailableError`, so the vocabulary is discoverable outside the code.

Merging.

PR #4, opened 2026-08-23
feat(state-store-conformance-suite): One conformance suite every backen…

wighawag:
Task: `work/tasks/done/state-store-conformance-suite.md`

Everything is green (`format:check`, `build`, `typecheck`, `test`), and the working tree holds only this task's changes.

## What I did

**Checked the premises first.** The spec is at `work/specs/tasked/one-processor-everywhere.md`, `portable-mutation-context-seam` landed (ADR-0018), and so did `retention-capability-and-refusal` (ADR-0019) after this task was written. So the seam (`StateStore`, `MemoryStateStore`, `createMutationContext`) and the capability report (`StateStoreCapabilities`, `retainedRange`, `BlockNotRetainedError`) are exactly where the task assumes, plus a typed refusal the task could lean on. No drift.

**New package `@etherfold/state-store-conformance`** (`packages/state-store-conformance/`), the suite itself:

- `stateStoreConformanceCases(factory)` builds the case list, choosing cases from the backend's own `capabilities` (read from a probe store, before `migrate`, as a caller would).
- `runStateStoreConformance(factory)` runs them framework-free and reports which failed.
- `describeStateStoreConformance(label, factory)` registers each case as its own vitest `it`. Adding a backend is a factory plus one awaited call.
- Five groups, all external behaviour, nothing touching a table/statement/version column: versioned reads, as-of reads against the CLAIM, reorg revert, read-your-writes within a block, block atomicity.

**The reorg case is built as the finding's shape**: `computedPoints` accumulated through the mutation context (read, add, write), 6 → 12, `revertTo` the fork, and it must read **6**, then resume from 6 (`+3 = 9`, not 15). It runs on every backend under every claim.

**Lying backends are proven to fail** (`test/the-suite-catches.test.ts`): a store claiming a 60-block window it does not honour, one answering every as-of read from the tip, and one accepting a revert without undoing state. Each is a decorator (a subclass could not lie, since the honest store guards `getAsOf` against `this.capabilities`), and the test asserts the honest store fails nothing while each liar fails the specific cases that name its lie.

**Two backends, six runs, 104 cases each**: `MemoryStateStore` (unbounded / 60-block window / revert-only) in the conformance package, and `VersionedStateStore` on a real libSQL database (same three) via `packages/state-store-sqlite/test/conformance.test.ts`.

**Duplicates moved, not copied.** Trimmed `state-store-sqlite/test/{as-of,revert,capabilities}.test.ts` to what only a versioned-row backend can be asked (partial unique index, raw-row invariants, `queryAsOf`/`queryCurrent`, the hash axis, `NoSuchBlockError`); trimmed `state-store/test/{memory-store,retention}.test.ts` to representation, block lookup, declaration validation and setting-to-claim mapping; trimmed the capability block in `processor-entities/test/two-backends.test.ts` to what the suite cannot see (the option spelling). Extracted the SQLite `WindowedStore` double into `test/utils/`.

Also: `docs/adr/0020-…`, a changeset, README sections in both store packages, and a `conformance suite` glossary entry in `CONTEXT.md`.

## Decisions

- **The suite is a fourth package (`@etherfold/state-store-conformance`), not a module or subpath of `@etherfold/state-store`.** ADR-0018 makes the seam's *zero dependencies* load-bearing, and `state-store-sqlite`'s leakage test asserts that emptiness; a suite needs an assertion library, so hosting it in the seam would either add a dependency there or hide one behind an optional peer that npm auto-installs anyway. Alternatives considered and rejected: a subpath export of the seam (breaks the emptiness assertion), and copies in each backend's `test/` (the drift the suite exists to prevent). Consequence, and the one visible cost: `MemoryStateStore`'s conformance run lives in the conformance package, because `state-store` cannot depend back on it. Touches every future backend (`light-store-behind-the-seam`, `indexeddb-row-backend-browser-default`), `promote-stratagems-conformance-workload`, and ADR-0018's package map. Recorded as **ADR-0020**.
- **The name fills ADR-0014's trailing slot with a non-engine, as ADR-0018 already did.** `state-store-conformance` = the state-store role's conformance suite, sorting adjacent to `state-store` / `state-store-sqlite`. `conformance-state-store` was rejected for breaking the ordering rule and reading as a backend named "conformance".
- **Cases are DATA (`{group, name, run}`), with vitest registration as a thin adapter.** This is what makes the lying-backend requirement satisfiable: a suite that has already reported itself to a runner can be run but not asserted on, and nesting a second vitest process would make the proof slower than the suite and dependent on a reporter format. Alternative rejected: `describe`/`it` at import time. Cost: a top-level `await` in each backend's test file (the case list depends on the claim, and a claim can only be read from a store); vitest is still a peer dependency, since the assertions are its `expect`. What the case list buys is independence from the RUNNER, not from the assertion library.
- **The suite tests a backend against its OWN claim, and says where it decides.** `claimedDepth` / `answersHistoryOverLadder` gate the history cases: a store claiming a window shorter than the suite's 8-block ladder is asked only the refusal cases. Testing a backend against a capability it never claimed would fail honest backends; the risk in the other direction (a tiny claimed window being asked less) is why the gate is one named, documented function rather than scattered `if`s. Touches `prune-versions-outside-retention-window`, which will make a real store able to claim a window.
- **The refusal is asserted as `BlockNotRetainedError`, not merely as "it threw".** That makes the error family part of the backend contract: a caller must be able to tell a retention boundary from a reorg (`NoSuchBlockError`) and both from an absent entity (`undefined`). This raises the bar for a third-party backend, so it is recorded rather than buried.
- **`applyBlock` sharp edges are contract, not implementation accident.** The suite requires that re-applying a block raises, that a second hash at a recorded height raises, that a block with no mutations is still recorded, and that a rejected mutation leaves the height free. Both shipped backends already behave this way and `two-backends.test.ts` had pinned them as shared meaning; making them suite cases means a future backend must honour them too. "The block was not recorded" is observed by re-applying the height, never by asking for a block table, which the seam deliberately does not expose.
- **Changeset lists `state-store`, `state-store-sqlite` and `processor-entities` as `patch`** although only their tests changed: `changeset status --since=main` treats any file in a package as a change and would otherwise go red at land time.


wighawag:

| # | criterion | verdict |
| --- | --- | --- |
| 1 | One exported suite taking a backend factory | ✅ `runStateStoreConformance(factory)` + a vitest adapter |
| 2 | Runs against two backends in CI, both green | ✅ memory + SQLite conformance entry points |
| 3 | Reorg case asserts a counter DECREASING, on every backend | ✅ a suite case, so it runs on both by construction |
| 4 | Suite reads declared capabilities and tests behaviour against the CLAIM | ✅ `src/cases/declared-capabilities.ts` |
| 5 | A deliberately-lying backend FAILS the suite | ✅ three liars, see below |
| 6 | SQLite tests duplicating a shared case are MOVED, not copied | ✅ verified by audit |
| 7 | Tests in `test/`, vitest, changeset | ✅ |

### Criterion 5 is the one that makes the rest worth having, and it is done better than specified

Three liars, each a real failure mode rather than an invented one: `LyingWindowStore` (claims a window, answers outside it), `AmnesiacStore` (claims full history, serves every historical read from the tip — the exact failure this whole design exists to prevent), and `StickyCounterStore` (accepts the revert, keeps the state, so an accumulated counter never comes back down).

Two design decisions here are sharper than the criterion required:

- **The lies are decorators, not subclasses**, and the reason given is correct: the honest store's refusal is not a method a subclass can forget, because it guards `getAsOf` against `this.capabilities`. A store whose report and behaviour genuinely disagree can therefore only be built by wrapping. A subclass-based liar would have been a weaker test that happened to pass.
- **`runStateStoreConformance` is a plain function over the case list rather than a test runner**, so a failure is a value to assert on instead of a red run to interpret. That is what makes the meta-test expressible at all.

It also includes the control case that is usually omitted — `passes an honest backend, so a failure below means something` — and asserts the suite reports *why* a case failed, not merely that it did.

### Criterion 6, audited rather than assumed

Existing suites shrank by ~378 lines and the shared cases now live in `state-store-conformance/src/cases/`. What remains in `state-store-sqlite/test/revert.test.ts` is `revertTo, in the rows`: versions dropped, dead blocks removed from the canonical table, exactly one live version per key, tables emptied without touching the schema. That is storage mechanics appropriate to the backend package, not a second copy of the behavioural case, and the file header says so explicitly and points at the conformance package. Behaviour in the shared suite, row mechanics in the backend, with a pointer — the right split.

### The forward-note from `retention-capability-and-refusal` was honoured

That task placed its cross-backend capability cases in `processor-entities/test/two-backends.test.ts` and asked that this suite **absorb** them rather than duplicate. Checked: `refuses every historical read when it declares 'revert-only'` is gone from `two-backends.test.ts`, replaced by a narrower processor-level assertion, and the capability behaviour now lives in the suite. Absorbed, not forked.

Merging.

PR #5, opened 2026-08-23
feat(bounded-id-prefix-listing): Give MutationContext a bounded listing…

wighawag:
Task: `work/tasks/done/bounded-id-prefix-listing.md`

All checks green (`format:check`, `build`, `typecheck`, `test`: every package passes; state-store-sqlite 221 tests, conformance 150, processor-entities 35).

## What I did

**Checked the premise first.** `portable-mutation-context-seam` landed as `@etherfold/state-store` + `@etherfold/processor-entities`, and `state-store/src/store.ts` explicitly reserved the listing for this task ("It is not here because the bound is a design decision that task owns"). The seam is exactly where the task assumes. No drift; built it.

- New `src/listing.ts`: the whole concept in one file — `EntityIdPrefix`, `Listing<T>`, `prefixValues` (validation), `hasIdPrefix`, `compareIds`, `assertListingLimit`, `boundedListing`.
- `MutationContext.list(entity, prefix, limit)` merges the block's staging area into the store's range scan (read-your-writes), with a fetch budget that accounts for staged deletes.
- `StateStore` gained `listCurrent` / `listAsOf`; `MemoryStateStore` implements both.

**SQLite backend** — `listCurrentStatement` / `listAsOfStatement` built as plain data; the store's `listAsOf` resolves a `BlockAddress` and honours retention. `test/listing.test.ts` pins the statement shape *and* reads the access path back out of `EXPLAIN QUERY PLAN` (one `SEARCH ... USING INDEX placement_history (epoch=?)`, no temp b-tree, no table scan).

**Conformance** — new `bounded id-prefix listing` group (12 cases incl. as-of and revert), plus the two listing cases the read-your-writes group's README had promised; `AmnesiacStore` now lies about `listAsOf` too, so those cases are proven to go red.

**The contortion, modelled properly** — `processor-entities/test/ordered-children.test.ts` rebuilds the finding's real window of seven placements (evict the oldest, cascade-delete everything nested under it) on **both** backends with no stored array, no CSV index and no count, including the all-in-one-block burst case, and a guard asserting the model declares no index/count field.

Docs: ADR-0021, CONTEXT.md glossary (`listing`, updated `StateStore`/`MutationContext`), four package READMEs, one changeset covering all four changed packages. One out-of-scope signal captured in `work/notes/observations/staged-get-returns-a-partial-row.md`.

## Decisions

- **Ordering is LEXICOGRAPHIC over the stringified id, not numeric** (`'10'` before `'9'`). Chosen because that is the order a key-prefix range scan produces for free on every backend, and because id columns carry no declared type (`fields` excludes them), so a numeric order would need type information the declaration does not have and would diverge from the index's own order. Alternative considered: sort numerically when a value parses as a number — rejected, it makes the seam's order differ from the backend's and is unstable across mixed values. Consequence, documented in the authoring README and ADR: a numeric child key must be fixed-width/zero-padded, and arrival order is a modelling answer (`(blockNumber, logIndex)`), never a parameter. Touches: `promote-stratagems-conformance-workload` (the port's keys), every future backend's ordering.
- **Truncation is REPORTED, not inferred: `list` returns `{rows, truncated}`,** and every backend fetches `limit + 1`. Alternatives considered: return a bare array and let the caller compare `rows.length` to the limit (rejected: an exact fill is indistinguishable from a cut-off one, and a cascade delete that guesses wrong leaves orphans silently — the same "plausible wrong answer" this seam refuses elsewhere), and throw when the set exceeds the limit (rejected: a handler legitimately lists a bounded head of a large set). Touches every caller's return-type shape and the conformance contract.
- **An EMPTY prefix is REFUSED; a listing must name at least the first id column.** This is the one place I let a measurement decide: `EXPLAIN QUERY PLAN` shows that unanchored (`WHERE _upper IS NULL ORDER BY <id> LIMIT ?`) SQLite abandons the id index and adds `USE TEMP B-TREE FOR ORDER BY`, so "the whole entity, bounded" would quietly break the one property the shape exists to guarantee. Alternative: allow it and accept a sort on that path — rejected as the surface's own claim becoming conditional. Consequence: "the first N rows of a table" is not expressible at the handler seam; it belongs above it. New user-visible refusal.
- **The limit is a required POSITIONAL `number`, not an options bag,** and a limit that is not a whole number ≥ 1 is a new runtime error. An options bag is an extension point by convention, and this surface's whole point is that it has none; positional leaves nowhere to hang a `where`/`orderBy`/`offset`. Pinned by two `@ts-expect-error` assertions that `pnpm typecheck` runs.
- **Names: `list` on `MutationContext`, `listCurrent` / `listAsOf` on `StateStore`.** Coherence check against CONTEXT.md: "list" meant nothing yet; `listCurrent`/`listAsOf` mirror `getCurrent`/`getAsOf` (same axis, same retention refusal) and stay deliberately distinct from `queryCurrent`/`queryAsOf`, which keep their meaning as the SQL-taking, above-the-seam surface. Both are REQUIRED of every backend, so `light-store-behind-the-seam` and `indexeddb-row-backend-browser-default` must implement them (an `IDBKeyRange.bound` cursor and a patch-log walk respectively).
- **A row staged in the current block is COMPLETED before it is listed** (id columns added, unlisted declared fields as `null`), so a listing's rows have one shape whether they came from the store or from the block. This makes `list` disagree with `get`, which still answers a staged key with only the values that were written; I did not change `get` (out of scope) and captured the asymmetry as an observation instead.
- **`VersionedStateStore.listAsOf` widens `at` to a `BlockAddress`** exactly as `getAsOf` already does, so hash/height/timestamp all work and the seam still only sees a resolved number. Follows the ADR-0018 decision rather than re-opening it.
- **The conformance fixture set gained a fourth entity, `placement` (`['epoch','position','playerIndex']`).** Every backend factory is handed it, which is a (small) cost to every future backend; a three-column id is what makes a PREFIX testable at more than one length.
- **Recorded as ADR-0021** (`docs/adr/0021-the-handler-seams-only-set-read-is-a-bounded-id-prefix-listing.md`): hard to reverse (public authoring API of two published packages), surprising without context (a read surface with no `where` looks like an oversight), and a real trade-off (questions pushed into modelling or above the seam).


wighawag:

| # | criterion | verdict |
| --- | --- | --- |
| 1 | Listing takes entity + id prefix + required limit, returns ascending id order | ✅ `list(entity, prefix, limit)` |
| 2 | Omitting the limit is a TYPE error; a bad prefix errors naming entity and columns | ✅ required positional |
| 3 | No way to pass a predicate, a sort or an offset | ✅ verified across the seam types |
| 4 | Read-your-writes: two children written, one deleted, later event in the same block sees the survivors | ✅ |
| 5 | SQLite is a single indexed range scan, asserted by pinning the statement shape | ✅ exceeded, see below |
| 6 | As-of listing returns children live at that block | ✅ |
| 7 | A test models ordered bounded children per the modelling rule, eviction with no hand-maintained index | ✅ both contortions dissolved |
| 8 | Tests, vitest, changeset | ✅ |

### The bound is genuinely by construction

This was the constraint I was watching hardest: a predicate parameter added "for flexibility" would have been a block rather than a nit. The seam is `list(entity, prefix, limit)` and a grep across the seam types finds no `where`, `orderBy`, `offset`, `filter`, `sort` or `predicate` anywhere. There is a test named `has nowhere to put a predicate, a sort or an offset`, which asserts inexpressibility rather than discouragement — exactly the framing the task asked for.

`queryCurrent` / `queryAsOf` are untouched and stay the server-side surface that may take SQL; the asymmetry is documented where a reader meets it, so the two tiers do not blur.

### Criterion 5 is exceeded

The pinned statement is an equality on the leading id columns under the validity predicate, with exactly one `ORDER BY` (asserted to be the declared id, not caller-supplied) and a `LIMIT`. Beyond pinning the SQL text, it asserts the **query plan**: `rides the id index at the tip, with no temp b-tree for the ordering`, plus the two-column seek and the as-of variant. That turns "single indexed range scan" from a claim into a verified fact, which is what "asserted, not assumed" was asking for.

Two details worth calling out as good judgement rather than incidental:

- **The staging merge over-fetches by exactly `staged.size`.** A listing cannot fall through to the store the way `get` does: rows deleted earlier in the block may occupy slots in the store's answer, so asking for exactly `limit` could come back short. Asking for `limit + staged.size` cannot, because that is the most the staging area can subtract. The reasoning is recorded next to the code.
- **`truncated` is a fact, not a guess.** The store fetches `limit + 1` so a truncated listing is distinguishable from an exact one, which answers the open question the task posed about truncation.

### Observation filed, correctly not fixed here

`staged-get-returns-a-partial-row`: `get` served from the staging area returns only what the handler passed to `set`, while `get` served from the store returns the whole row, so an id column or an unlisted declared field is `undefined` or present depending on *when in the block* the row was written. This task made `list` consistent (one row shape whatever the origin) and recorded that `get` now disagrees, rather than quietly changing `get`. That is the right call — a timing-dependent field presence is a real latent bug, and it deserves its own task rather than being absorbed here. I am carrying it into the end-of-drive triage as task-worthy.

Merging.

PR #6, opened 2026-08-23
feat(prune-versions-outside-retention-window): Actually prune versions…

wighawag:
Task: `work/tasks/done/prune-versions-outside-retention-window.md`

Implementation complete, all repo gates green.

## What I built

**`packages/state-store` (the seam)**
- `retentionFloor(retention, tip, finalityDepth?)`: the one comparison every backend prunes on. It is deliberately *identical* to `retainedRange(...).from`, so the block a read is refused below and the block a version is deleted at cannot drift apart.
- `pruneBudget`, `PruneOptions {maxVersions?}`, `PruneReport {tip, floor, versionsDeleted, complete}`.
- `StateStore.prune(options?)` added to the interface (eighth verb), with the full "when it runs and why not in the write path" reasoning on it.
- `MemoryStateStore` now prunes for real (reference implementation, not a stub) and therefore reports the window it was configured with.
- `retentionWithoutPruning` removed: it existed only to express "this store cannot enforce what it was asked".

**`packages/state-store-sqlite`**
- `prunableVersionsStatement` / `dropVersionsStatement` in `statements.ts` (plain data, house style). Predicate: `_upper IS NOT NULL AND _upper <= floor`, ordered by `_upper`, riding the existing `<table>_upper` index.
- `BatchBounds.maxRowsPerStatement` (default 500): the existing bounds cap a *batch* and say nothing about one statement touching unbounded rows, which is exactly what a hosted backend rejects.
- `VersionedStateStore.prune()`: select bounded row ids → delete that explicit set → repeat; exact count, exact `complete` (one bounded probe only when a budget stopped the pass).
- A configured window is now **reported** instead of downgraded to `unbounded`.

**`packages/processor-sqlite`**: `VersionedStateEventProcessor.prune()` (on the processor, not the read-only view: it is a write), so a deployment that configured retention through the processor can actually reach the enforcement.

**`packages/state-store-conformance`**: new `pruning, and what must survive it` case group, claim-driven like the rest; the windowed subjects on both backends are now real configuration instead of test doubles (`WindowedStore` / `WindowedMemoryStore` deleted).

**Docs**: ADR-0022, ADR-0019's superseded consequence amended, `CONTEXT.md` glossary entry for `prune` (disambiguating it from the core's `unconfirmedBlocks` prune, `History`'s patch prune, and the CLI `snapshot-prune-script`), both package READMEs, and a changeset.

**Tests written first**, starting with the dangerous case: `packages/state-store-sqlite/test/prune.test.ts` (18 cases, all asserting version COUNTS, never bytes), plus memory-store prune cases and the shared conformance group. Repo totals: build clean, `typecheck` 0 errors, 879 tests passing, `format:check` clean. (`pnpm changeset status --since=main` currently reports "no changesets found" *only* because my changeset file is untracked; changesets diffs tracked files against the merge-base, so it resolves the moment the runner commits — verified against `origin/work/task-retention-capability-and-refusal`, whose committed changeset is detected by the same command.)

One bookkeeping note, not drift: the task body is at `work/tasks/backlog/prune-versions-outside-retention-window.md`, not `work/tasks/ready/` as the prompt said. Content matched the spec exactly, so I proceeded; I touched neither.

## Decisions

- **Pruning runs when the HOST calls `prune()`; never as a side effect of a write.** A prune plus `VACUUM` measured 1.1 s at 62,553 versions while a block on the same stream carries a median of 7 mutations, so folding it into `applyBlock` would stall whichever block crossed a threshold, for work that block did not cause, and would have the store choose a maintenance cadence on behalf of a browser tab, a backfilling CLI and a long-running server alike. Alternatives considered: amortised-inside-`applyBlock` (rejected: the store invents a cadence nobody chose, and the stall lands on an arbitrary block) and a background timer owned by the store (rejected: a store that schedules its own work is unstoppable from the host and untestable without fake timers). Both are expressible *on* this verb instead. Named cost, recorded in ADR-0022: a deployment that sets a window and never prunes is bounded in what it ANSWERS and unbounded in what it HOLDS. Touches `VersionedStateEventProcessor.prune`, `one-processor-cli-and-split-server` (which will decide the CLI/server schedule), and `indexeddb-row-backend-browser-default` / `light-store-behind-the-seam` (which must implement the verb).
- **Bounded per request by an explicit list of row ids, not by predicate**, via a new `BatchBounds.maxRowsPerStatement` (default 500). `DELETE ... WHERE _upper <= ?` is one small statement that deletes unbounded rows: the exact shape that runs on a local file and is rejected by a hosted backend capping rows-written or wall-clock per request. `remote-sql`'s result carries rows and no affected-row count, so a blind bounded DELETE could report neither what it did nor whether it had finished; selecting first costs one extra round-trip per chunk and buys an exact count and an exact `complete`. Ordered by `_upper` so a budgeted pass drops the OLDEST first and a partially pruned store converges from the far end. Touches every deployment's `bounds` option (a new key, defaulted).
- **`prune` goes on the SEAM, not on the SQLite class.** Retention is a seam-level declared capability, so its enforcement must be reachable by whatever schedules it; a backend-only method would leave a host holding a `StateStore` unable to bound it. Consequence: `MemoryStateStore` implements it for real rather than stubbing, which also keeps "one retention spelling means the same thing on both backends" true (`processor-entities/test/two-backends.test.ts`). Alternative: a backend-only method plus a `StateStore & {prune}` feature-test at call sites — rejected as an optional-method surface callers must probe. Touches `light-store-behind-the-seam` and `indexeddb-row-backend-browser-default`, which now owe a `prune`.
- **A `revert-only` store prunes to its declared `finalityDepth`; with no depth declared it prunes nothing.** `revert-only` already *means* "kept only as long as reorg revert needs them" (`capabilities.ts`), so this implements the declared meaning rather than adding one; and leaving `revert-only` — the tightest retention — as the only unbounded-storage kind would have been absurd. It introduces no new refusal and changes no default: an undeclared depth states no floor and gets a no-op, because a store deleting against a number nobody wrote down is worse than one that keeps rows. Alternative: prune only `{kind:'window'}` and tell users to spell the tightest window as `{blocks: F}` — rejected because that spelling also *answers* history, so "refuse history AND bound storage" would have been inexpressible. Touches `RetentionOptions.finalityDepth`, which is now load-bearing for a second kind.
- **A window is now REPORTED even before the host has pruned.** The window is enforced on reads from the moment it is configured (`assertRetained`), so the report never promises history that is gone — it only ever refuses history that is still physically present, which is the same safe direction a `revert-only` SQLite store already took (recorded in `retention-capability-and-refusal`). Alternative: report `unbounded` until a prune has actually run — rejected as a report that changes under the caller's feet and cannot be read at startup, which is the whole point of story 7. Touches ADR-0019's superseded consequence bullet, `processor-sqlite`'s reported capabilities, and the conformance suite's claim-driven case selection.
- **The `_blocks` table is not pruned, and `prune` does not `VACUUM`.** Dropping a block row would turn `BlockNotRetainedError` ("that block is fine, its state is outside what I keep") into `NoSuchBlockError` ("never indexed, or reorged out") — a worse answer, and a wrong one for a consumer that pinned the hash; a block row is three columns against a version count two orders of magnitude larger. `VACUUM` cannot run inside a transaction (the only write surface `remote-sql` exposes), rewrites the whole file, and is unavailable on some backends; SQLite reuses the freed pages, so the file stops growing without shrinking, and reclaiming bytes is the operator's call. Pinned by a test so neither is "tidied up" later. Consequence to name: retention bounds VERSIONS, and `_blocks` still grows slowly and unboundedly at one row per event-bearing block.
- **Coherence: `prune` is a third use of an existing word, at a new layer.** The core already prunes `unconfirmedBlocks` past the finality window and `History` prunes reverse-patches by block distance — same meaning ("drop what is outside the retention window"), different object. The backlog's `snapshot-prune-script` prunes orphaned CLI snapshot *files*, a genuinely different object. I reused the word rather than inventing one because the meaning matches, and added a `CONTEXT.md` glossary entry that names all four so the overlap is documented rather than discovered.


wighawag:

| # | criterion | verdict |
| --- | --- | --- |
| 1 | A window of N physically drops versions closed below `tip - N`, proven by a count | ✅ |
| 2 | The live version survives regardless of age | ✅ two tests, see below |
| 3 | Reads inside the window still correct; outside gives the typed refusal | ✅ |
| 4 | `revertTo` still works to the finality depth after a prune | ✅ including the counter-back-DOWN case |
| 5 | Pruning an `unbounded` store is a no-op, not an error | ✅ |
| 6 | When pruning runs is an explicit documented choice, not a side effect of a write | ✅ verified not called from `applyBlock` |
| 7 | With a window configured the report stops saying `unbounded` | ✅ |
| 8 | Tests assert on version COUNTS, not reported bytes | ✅ |
| 9 | Tests, vitest, changeset | ✅ |

### Criterion 2, the one that destroys state when it is got wrong

`keeps an entity written once at a block far below the floor and never touched again` and `keeps it even when its own version is the only row left of that entity`. The complement is tested too — `drops the closing version of a DELETED entity, which has no live row to protect` — so the rule is "protect the live version", not "protect the newest row", which is the distinction a naive delete-by-age collapses.

The boundary is handled with the same care: `leaves the version that is still valid AT the floor, because a read there must answer` makes the floor inclusive, and the prune is idempotent at a fixed tip.

### The detail I did not ask for and am glad of

`keeps the block table, so an old hash still resolves and is REFUSED rather than unknown`. Pruning the block table alongside the versions would have been the obvious economy and would have silently collapsed the ADR-0015 / ADR-0019 distinction: an old hash would come back as `NoSuchBlockError` ("that block never existed or was reorged out") instead of `BlockNotRetainedError` ("that block is real, I no longer keep its versions"). Those are different news and the whole error family exists to keep them apart. Preserving it under pruning is exactly right.

### Criterion 6 was decided rather than smuggled

Pruning is not invoked from `applyBlock`; it is `prune({maxVersions})` on a schedule the host owns, with the reasoning recorded where the method lives: a prune plus `VACUUM` measured 1.1 seconds at 62,553 versions, against a stream whose blocks carry a median of 7 mutations, so folding it into the write path would stall one arbitrary block by a second. The pass is bounded per request (`never names more rows in one DELETE than the configured bound`, honouring the real per-request limits on remote backends) and supports a budget so a host can amortise it, with `stops at a budget and says so`.

### The downgrade helper disappeared as designed

`retention-capability-and-refusal` introduced `retentionWithoutPruning` and documented that it "disappears from a store the day that store prunes". It has: SQLite now reports `resolveRetention(...)` directly, so a configured window is reported as a window because it is now enforced. The two tasks fit together the way they said they would.

Merging.

PR #7, opened 2026-08-23
feat(light-store-behind-the-seam): The patch-based light store behind t…

wighawag:
Task: `work/tasks/done/light-store-behind-the-seam.md`

Built and green. Summary of the work:

## What I built

**New package `packages/state-store-patch` (`@etherfold/state-store-patch`)** — the light `StateStore`: current state as a plain object, history as immer reverse patches, reorg revert by backwards replay.

- `src/store.ts` — `PatchStateStore`. `applyBlock` validates every mutation before touching state, then `produceWithPatches` gives the next state plus the reversal that undoes the block. `revertTo` replays reversals highest-block-first, checking the whole range *before* replaying anything. `prune` drops reversals at or below `tip - finalityDepth` (the seam's `retentionFloor`), host-scheduled per ADR-0022. `listCurrent` is the seam's bounded id-prefix listing as a sorted walk. `getAsOf` / `listAsOf` **read nothing**: they call `assertRetained` and then throw `BlockNotRetainedError`, so there is no tip value in scope to return.
- `src/errors.ts` — `RevertBeyondPatchHistoryError`, deliberately outside the `BlockUnavailableError` read family.
- `src/immer.ts` — `enablePatches()` once, as `@etherfold/js-processor` does.
- Tests (vitest, 118 in-package): `conformance.test.ts` (the shared suite under two configurations, all green against its `revert-only` claim), `sparse-stream.test.ts` (event-bearing blocks at the measured 429-block median: a prune at finality 64 leaves exactly one block's reversals; revert to the full depth makes the accumulated counter go 24 → 18; a deeper revert is refused atomically; a never-pruned store replays correctly at every depth), `historical-reads.test.ts` (behavioural refusals at every depth plus a structural guard that the as-of methods touch no state), `capabilities.test.ts`, `stays-light.test.ts` (only `immer` + the seam).
- `README.md` documenting the measured `revert-only` result, the revert refusal, and memory-only.

**`packages/processor-entities`** — `test/patch-backend.test.ts`: the same imported processor produces identical state on the patch store, on SQLite (real libSQL) and on memory; a retracted block makes the counter decrease; the historical read the other two answer is refused here; a retraction deeper than the surviving patches is refused with the state untouched. `test/ordered-children.test.ts` now also runs the patch backend (the listing + in-block read-your-writes workload). Added as a devDependency only, so no store package depends on the indexer core (ADR-0016).

**Also:** `docs/adr/0023-...md`, `.changeset/light-store-behind-the-seam.md`, and one clause in `CONTEXT.md`'s `StateStore` entry naming the new backend.

Gate: `format:check`, `build`, `typecheck`, `test`, `build:examples` all green. (`changeset status --since=main` reports "no changesets found" only because the changeset file is still untracked; it is `git diff`-based and will see it once the runner commits.)

## Decisions

**Package name and placement: a new package `@etherfold/state-store-patch`, class `PatchStateStore`.** It cannot live in `@etherfold/state-store`, which declares *no* dependencies (asserted by `state-store-sqlite/test/no-platform-leakage.test.ts`) and this store needs immer. Name read off ADR-0014/0016's axis (role first, backend last), with the trailing slot naming the *representation*: `state-store-memory` was rejected because `MemoryStateStore` already means "in-memory versioned rows" in the seam package, and two "memory" stores would re-mean an existing term. The spec's phrase "the light store" is kept as prose in the README/ADR rather than as a package name, since "light" names a feeling and "patch" names the mechanism. Touches: the naming axis in ADR-0014/0016; the future `indexeddb-row-backend-browser-default` package sits beside it under the same rule.

**Persistence: memory-only, and reload is an empty store.** State and the patch log both go with the process. Alternatives considered: (a) serialise the whole state on save, which is the incumbent `keepStateOnIndexedDB` and belongs to the `KeepState` seam *above* storage; (b) row-level persistence, which is `indexeddb-row-backend-browser-default`'s job and a different backend. Doing either here would fork a persistence strategy inside the backend whose whole claim is that it is the cheap one. Consequence, stated rather than discovered: a host using this store must re-index from a cached stream (or hydrate from a snapshot) on load. Touches: `indexeddb-row-backend-browser-default`, the `KeepState` seam, and whoever wires the browser default. Recorded in ADR-0023.

**`durability: 'memory-only'` is reported on this backend's own capability object, not added to `StateStoreCapabilities`.** The acceptance criterion asks the capability report to say so; adding a field to the shared type would make every backend fill in a constant for one implementation's benefit and would break existing exact-equality assertions on `capabilities` in `processor-entities/test/two-backends.test.ts`. So `PatchStateStoreCapabilities = StateStoreCapabilities & {durability: 'memory-only'}`: structurally still a `StateStore`, and the extra word sits where a caller already looks at startup. Alternative if a second backend ever needs it: promote `durability` to the seam. New named concept, checked against the glossary: it does not overlap `retention` (how far back history reaches) — it is whether anything survives the process.

**A revert deeper than the surviving reverse patches is a typed error (`RevertBeyondPatchHistoryError`), atomic, and NOT a member of `BlockUnavailableError`.** Alternatives: revert as far as possible and report the depth reached (rejected: a half-undone reorg is the write-path twin of a tip read served as history, and keeps exactly the counter a reorged-out block raised), or keep every reversal forever (rejected: that is an unbounded patch log, and it is not what `revert-only` means). Not filed under `BlockUnavailableError` because that family is about a *read* the store cannot answer, where the caller can carry on with the tip; here the state is known-ahead-of-canonical and a `catch` written for a refused history query must not swallow it. Touches: any host wiring this backend needs a re-index path — `applyEventStream` in `@etherfold/processor-entities` will surface it from `revertTo`, so `one-processor-cli-and-split-server` and the browser wiring inherit it. Recorded in ADR-0023, with `retainedReversals()` exposed so a host can see the real depth (one block on a sparse stream) before it needs it.

**`retention` is accepted only as `'revert-only'`; a window or `'unbounded'` throws at construction.** The seam's doc says a store reports what it *does* rather than what it was told, which would suggest downgrading silently; I refuse instead, because a deployment that wrote `{blocks: 128}` believes it can read history and should find out where it configured it. Alternative considered: accept and report `revert-only` (rejected as a quiet demotion). This introduces a new refusal, so it is recorded here; it touches how a generic deployment passes one retention spelling to any backend.

**`PruneReport.versionsDeleted` counts blocks whose reversals were dropped.** This backend has no versions; one block's reverse patches are its unit of history, and `maxVersions` budgets the same unit. Documented at the method. Touches the shared `PruneReport` type's reading on a non-versioned backend.

**The cross-backend equality test lives in `processor-entities/test/`, not in the new package.** That is where "one processor, several backends" already lives, and it keeps the store package free of a devDependency on the indexer core (viem), preserving ADR-0016's direction even in the test graph.


wighawag:

| # | criterion | verdict |
| --- | --- | --- |
| 1 | One processor runs unchanged here and on SQLite, equal state | ✅ |
| 2 | Reports `revert-only`; conformance capability cases pass | ✅ |
| 3 | `revertTo` correct to the finality depth on a SPARSE stream, counter DOWN | ✅ |
| 4 | Every as-of read refused; no code path returns the tip | ✅ **structurally**, see below |
| 5 | The id-prefix listing works here, with read-your-writes | ✅ |
| 6 | Persistence is an explicit recorded decision | ✅ memory-only, in the report and ADR-0023 |
| 7 | Tests, vitest, changeset | ✅ |

### Criterion 4 is satisfied in the strongest available form

The behavioural tests are there (`refuses getAsOf at every depth, including the tip and the block just applied`, and `refuses even for an entity that never existed, because the CAPABILITY is missing, not the row` — a nice distinction). But the part that earns the approval is `the as-of methods cannot read state, by construction`, which asserts of `getAsOf` and `listAsOf` that their bodies reference no stored state, contain `throw new BlockNotRetainedError`, and contain no `return` at all.

A source-text assertion is usually a smell. Here it is the right instrument, because the property is a *negative* structural one that no behavioural test can express: there is no tip value in scope to fall back to, so the failure mode cannot be reintroduced tomorrow by someone adding a well-meaning fallback. Given that answering a historical read from the tip is the single failure this whole design exists to prevent, making it structurally impossible rather than merely absent is the correct level of paranoia.

### Honest beyond the criterion

The store **refuses** a window *and* refuses `unbounded`, rather than accepting them and reporting less. A patch log can do neither, so a configuration stating one is a mistake at the place it was configured. That is a better answer than silently downgrading, and it is the mirror image of the SQLite decision in `retention-capability-and-refusal` (which downgraded because it genuinely *kept* more than it claimed; here the store would keep *less*).

`REFUSES a revert whose reverse patches have been pruned, and changes nothing` is the other good one: an atomic refusal rather than a partial rewind, and `names the depth it could not reach, so the message is actionable rather than "failed"`.

The sparse-stream suite reproduces the finding's result directly — `keeps exactly ONE block of reversals, the tip, because the gap dwarfs the depth` — so the capability is withdrawn by sparsity in the tests for the same reason it is withdrawn in the design, rather than a dense fixture making as-of look available.

### Seam discipline held

`durability: 'memory-only'` is added as a package-local `PatchStateStoreCapabilities` extending the seam's type, not by widening `StateStoreCapabilities`. The seam is untouched and no other backend needed changing, which is right: a field only one representation can meaningfully answer does not belong in the shared contract. `stays-light.test.ts` pins that the package imports nothing but immer and the seam.

Merging.

PR #8, opened 2026-08-23
feat(indexeddb-row-backend-browser-default): The row-level IndexedDB ba…

wighawag:
Task: `work/tasks/done/indexeddb-row-backend-browser-default.md`

All acceptance criteria are met and the repo's checks are green. Here is what landed.

## What I built

**`packages/state-store-indexeddb`** (`@etherfold/state-store-indexeddb`, class `IndexedDBStateStore`) — versioned rows in IndexedDB behind the `StateStore` seam. Three object stores (`current`, `versions`, `blocks`), the entity name in the KEY rather than a store per entity, one block = one transaction. Access paths: tip read is one `get`; as-of is one backwards cursor over `bound([entity,...id], [entity,...id, at])`; the listing is one `IDBKeyRange.bound([entity,...prefix], [entity,...prefix, []])` cursor; `revertTo` is two index range scans (`lower` = what the fork opened, `upper` = what it closed); `prune` walks the `upper` index, where a live version cannot appear at all because `null` is not a valid IndexedDB key.

**Browser default** — `createBrowserStateStore(entities, config?)` in `@etherfold/browser`. IndexedDB when nothing is said; `{backend: (entities) => …}` selects any other `StateStore`. A test runs the same declarations and mutations through IndexedDB, the patch store and the memory store with no processor change.

**ADR-0024**, `docs/adr/0024-indexeddb-is-the-browser-default-until-four-things-are-true-at-once.md`: the decision as a CONDITION (the four things that must all hold for wasm SQLite to win), the five overturning criteria, the measured table, and pointers to `work/notes/findings/sqlite-in-the-browser.md` and `docs/spikes/sqlite-in-the-browser/results/`. It states plainly that this is not a speed-up over `keepStateOnIndexedDB` (2.0 ms/block against 45.6) and what row-level writes buy instead.

## Evidence

- Node (in the gate): 190 tests in the new package, including the shared conformance suite under all three retention claims via `fake-indexeddb`, the listing/prune/revert access paths asserted by recording what reached IndexedDB, cold start, two connections to one database, and the same processor agreeing with `MemoryStateStore`.
- Real browsers (`pnpm --filter @etherfold/state-store-indexeddb test:browser`, `playwright-browser-harness`): **Chromium, Firefox, WebKit — 162 shared conformance cases each, 0 failures**; the same `EntityProcessor` in a tab producing the same rows (and version bounds) as node, with the accumulated counter 5 → 4 → 5 across a retraction; the listing's key range read off the engine itself (4 records walked out of 200 rows); persistence across a real reload; **four tabs on one database, 4/4, zero mismatches** on all three engines. Raw output committed to `docs/spikes/indexeddb-row-backend-browser-default/results/` with a README.
- `pnpm format:check`, `pnpm build`, `pnpm typecheck` (browser specs and the playwright config included) and `pnpm test` all green; changeset added. (`pnpm changeset status --since=main` cannot see an untracked changeset file — it diffs against git — so it goes green once the runner commits.)

## Decisions

**Package name and placement: `@etherfold/state-store-indexeddb`, class `IndexedDBStateStore`.** Read off ADR-0014/0016's axis (role first, backend last) with the trailing slot naming the substrate, and ADR-0018 already reserved "a production browser backend is a package of its own". Runtime dependencies are the seam and nothing else, asserted by `test/stays-a-primitive.test.ts`. Touches: the naming axis; it sits beside `state-store-patch` under the same rule.

**The "browser default" selection surface lives in `@etherfold/browser` as `createBrowserStateStore`, not in the store package and not in `@etherfold/processor-entities`.** A default is only meaningful at the layer where the choice is made: putting it in the IndexedDB package would make it pointless (a host choosing otherwise would never call it), and putting it in `processor-entities` would make the backend-neutral authoring package depend on a backend, inverting the naming axis. Alternatives considered: a factory registry (rejected: it would force the browser package to know every store), and no surface at all (rejected: the criterion asks for a configurable default). Touches: `@etherfold/browser` gains `@etherfold/state-store` and `@etherfold/state-store-indexeddb` as runtime dependencies and a minor version; whatever later wires an `EventProcessor` for the browser should build on this function rather than a second chooser.

**Coherence check on the new module's placement: `src/storage/state-store/` next to the existing `src/storage/state/`.** Both write to IndexedDB and they are different seams, so the collision is real: `storage/state/OnIndexedDB.ts` is a `KeepState` keeper for the whole-state blob, `storage/state-store/` builds a `StateStore` for entity rows. I kept the glossary's own words (`KeepState` vs `StateStore`) rather than inventing a third term, and the difference is spelled out at the top of `BrowserStateStore.ts` and in ADR-0024. Alternative considered: `storage/entities/` (rejected: "entity declaration" already means the author-facing `{name, id, fields}`, so it would re-mean an existing term).

**The config is a discriminated union: `{backend?: 'indexeddb', databaseName?, retention?, finalityDepth?}` or `{backend: factory}`.** The options configure the DEFAULT backend, so a custom factory (which has already been configured by its host) cannot be passed alongside them. Alternative: one flat object with options ignored for custom backends (rejected: a silently-ignored `retention` is exactly the kind of quiet wrongness this seam exists to prevent).

**User-visible default: `databaseName` defaults to `'etherfold-state'`.** A required name is more honest about collisions; a default is what makes the common single-indexer case one line. I chose the default and documented, on both the option and the README, that the name IS the identity of the state (several tabs of one app share it; two unrelated indexers must not). Touches anyone running two processors in one origin.

**New refusal: using the store where there is no `indexedDB` throws a named error** naming the alternatives (`state-store-sqlite` on a server, `MemoryStateStore` or `fake-indexeddb` in a test, or an injected factory) rather than a `TypeError` on `undefined.open`. It is raised lazily at first use, so importing or constructing in node is still fine.

**Three deliberate departures from the measured prototype (`docs/spikes/sqlite-in-the-browser/src/store/idb.ts`), recorded in ADR-0024.** (a) `versions` carries a `lower` and an `upper` index, so `revertTo` is two range scans and needs no per-block undo journal — the prototype's journal grows with every mutation ever applied and would have to be pruned, which would silently make deep reverts no-ops. (b) `prune` walks the `upper` index instead of scanning every version (the prototype's full scan took 6.3 s at 62,553 versions). (c) No tombstone versions: a delete only CLOSES the live version, which is what the seam's own model says and what memory/SQL do. The cost is write-time index maintenance the 45.6 ms/block figure did not pay, and **I did not re-measure** (the task says not to re-litigate the measurements); the ADR says so explicitly and files it under overturning criterion 1 ("a tuned IndexedDB backend").

**The entity name is part of the key; the schema is fixed at version 1.** An object store per entity would make "the processor declares one more entity" an IndexedDB version change, and an upgrade transaction can be BLOCKED by another open tab — which would break the multi-tab property this backend is chosen for. Touches any future backend feature that wants a new object store: it inherits the upgrade/blocked problem.

**The tip is read from the database on every use, never cached.** A retention window is a distance from the tip, and another tab (or a reload) can move it, so a cached tip would refuse reads that are inside the window or answer ones that are not. Only a windowed store pays the lookup, because `assertRetained` takes it as a thunk. Asserted in `test/persistence.test.ts`.

**Browser tests are a separate `test:browser` script and are NOT in the acceptance gate.** They need three Playwright browser binaries a clean CI checkout does not have, and a gate that cannot run in CI is a gate that gets skipped. The compensation is that the SAME shared suite runs in the gate under `fake-indexeddb`, and the browser run's output is committed under `docs/spikes/indexeddb-row-backend-browser-default/results/` so the ADR points at an observation. A `vitest.config.ts` in the package keeps vitest from collecting the Playwright `.spec.ts` files.

**The cross-backend processor-equality test lives in THIS package, not in `@etherfold/processor-entities`** — a deliberate deviation from the precedent `light-store-behind-the-seam` set. The browser harness needs `@etherfold/processor-entities` as a devDependency here (running the real processor in a tab is the point), so adding this store to that package's test graph makes the two packages cyclic and pnpm warns on every install. I verified that, then reverted it. The chain is still closed: this package asserts IndexedDB ≡ `MemoryStateStore` on one processor, and `processor-entities/test/two-backends.test.ts` asserts `MemoryStateStore` ≡ `@etherfold/state-store-sqlite`.

**Two mutations of one business key in one block resolve to the last of them.** A version here is keyed by `(id, lower)`, so a block opens at most one version per key, where the SQL backend keeps both (its version identity is a surrogate row id) and the extra one is a zero-width version no read can return. The two backends therefore ANSWER identically and differ only in what they store; `MutationContext` coalesces per key anyway, so it is only reachable by calling `applyBlock` directly. Documented at the method rather than turned into a new refusal.

**Small doc updates outside the new package, because they named this work as future:** `packages/state-store/src/memory.ts` and its README pointed at the task slug for "the browser answer" and now name the package; `CONTEXT.md`'s `StateStore` glossary entry lists the new backend.


wighawag:

| # | criterion | verdict |
| --- | --- | --- |
| 1 | Implements the seam, passes the conformance suite in a real browser on Chromium, Firefox, WebKit | ✅ evidence committed |
| 2 | Reports retention honestly; prunes or reports `unbounded` | ✅ |
| 3 | The prefix listing is an `IDBKeyRange` cursor, not a full scan with a filter — asserted | ✅ with a negative control |
| 4 | Browser default, selectable by configuration, touching no processor code | ✅ |
| 5 | Four tabs, one database, zero row mismatches | ✅ on all three engines |
| 6 | An ADR recording the decision as a CONDITION, with the overturning criteria | ✅ ADR-0024 |
| 7 | Browser tests in the existing harness style | ✅ |
| 8 | Tests, changeset | ✅ |

### The ADR does the job it was written for

ADR-0024 states the four things that would all have to hold at once for wasm SQLite to win, each with the measurement behind it, then the five overturning criteria, with the numbers and pointers to both the finding and the raw results. It opens by naming its own purpose: "State it as a preference and the first person to read a wasm-SQLite benchmark reverses it. State it as a condition and they can check." That is the failure mode the criterion existed to prevent, and it is addressed head-on rather than incidentally.

It also carries the honesty constraint without softening it: "It is not a speed-up", with the 20x throughput loss at 4,072 live rows stated plainly and the four things row-level writes buy that the blob cannot do at any speed.

### The consequence I most wanted to see, and did

> **The shipped store is not byte-for-byte the measured prototype** … closed versions carry an `upper` index and versions carry a `lower` index … Two indexes cost write time that the 45.6 ms/block figure did not pay, and **that has not been re-measured**: the numbers above are the prototype's, the ranking they establish is between STORAGE ENGINES, and a re-measurement belongs with the tuning work in overturning criterion 1.

This is the right handling of a standing instruction not to re-litigate the finding's measurements. It uses the numbers for the ranking they legitimately establish, states exactly where the shipped artifact diverges from the measured one, and declines to imply the shipped store inherits the prototype's figures. A weaker diff would have quoted 45.6 ms/block as though it were this backend's number.

### Evidence checked, not taken on trust

Multi-tab: 4 tabs, `mismatches=0`, `errors=0` on Chromium, Firefox **and** WebKit — the exact case both SQLite VFSs fail at open. Conformance ran in all three real engines with genuine user agents (Chrome 151.0.7922.34, matching the finding's version), across both `unbounded` and `window-128` configurations.

Criterion 3 is asserted with a negative control rather than a positive-only check: `walks the children of the prefix and NOT the 200 rows around them`, plus `stops at the limit plus one, so 'truncated' costs one record and not the rest`. That distinguishes a genuine range scan from a scan-and-filter, which a test that merely observed correct results would not.

### Two design calls worth recording

- **The entity name is part of the key, not the name of an object store.** Creating an object store needs a version change, and an upgrade transaction can be *blocked by another open tab* — so a store-per-entity would turn "declare one more entity" into a migration a second tab can stall. Given criterion 5 is specifically about multi-tab, avoiding that interaction is well judged.
- **The browser run is deliberately outside the acceptance gate**, since it needs three browser binaries a clean CI checkout lacks; `pnpm test` runs the same suite under `fake-indexeddb` and the real-engine evidence is committed. Stated in the ADR rather than left as a surprise.

Merging.

PR #9, opened 2026-08-23
feat(query-surface-from-entity-declarations): Generate the read surface…

wighawag:
Task: `work/tasks/done/query-surface-from-entity-declarations.md`

All checks are green (`format:check`, `build`, `typecheck`, `test` across every package). Here is what I did.

## What landed

**`@etherfold/state-store` (new, zero-dependency):**
- `src/declarations.ts` — `declareEntities` (identity function that keeps a declaration's literal types) plus the type-level derivation: `EntityRow`, `EntityIdOf`, `EntityPrefixOf` (a leading run of the declared id columns, at least one), `FieldValue`, `IdColumnsOf`.
- `src/read-surface.ts` — `createReadSurface(store, declarations)`: per declared entity, the seam's four reads (`getCurrent` / `getAsOf` / `listCurrent` / `listAsOf`), typed off the declaration, rows projected to the declared columns. Also `declaredRow` (the shared projection) and `AsOfAddress<S>`.

**`@etherfold/state-store-sqlite`:** `src/query-surface.ts` — `createQuerySurface(store, declarations)`: the same four reads (delegating to `createReadSurface`, not re-implemented) plus `queryCurrent` / `queryAsOf`, which keep their caller-supplied SQL and now return typed, projected rows. The tier asymmetry (bounded by construction because a handler runs once per event on every backend, vs. predicates where a query planner exists) is stated in both module docs, both READMEs and the changeset.

**`@etherfold/processor-entities`:** re-exports `declareEntities` / `createReadSurface` and its types, so an author imports the write and read halves of one declaration from one place.

**Tests (written before the code, all vitest):**
- `packages/state-store/test/read-surface.test.ts` — behaviour on `MemoryStateStore` plus six `@ts-expect-error` type-level pins (`pnpm typecheck` is what runs them, and it errors on an *unused* expect-error, so the pins cannot rot into no-ops).
- `packages/processor-entities/test/read-surface.test.ts` — one reader function, written once, run unchanged against memory, SQLite and the patch store; the patch store's `revert-only` refusal propagates as `BlockNotRetainedError`.
- `packages/state-store-sqlite/test/query-surface.test.ts` — the predicate tier, the three address axes, `NoSuchBlockError` on an unresolvable hash on both tiers, retention refusal, and the pin that `list*` still has nowhere to hang a predicate.

**Docs:** `docs/adr/0025-...` (the `uint256` answer), the `## Conventions` changeset (`.changeset/query-surface-from-entity-declarations.md`, minor on all three packages), a `read surface` glossary entry in `CONTEXT.md`, and README sections in the three packages.

**Signals captured (not fixed):** four short notes in `work/notes/observations/` — `VersionedStateView` exposes no listing; a non-numeric `at` answers `undefined` instead of refusing; pending changesets name pre-rename packages so a plain `changeset status` fails; and the committed `pnpm-lock.yaml` carried two devDeps no manifest declares (that last one is why `pnpm-lock.yaml` shows in this diff: a plain `pnpm install` rewrites it, and leaving it stale would keep `--frozen-lockfile` broken).

## Decisions

- **A `uint256` is NOT decoded, and that is the explicit answer, not a shrug.** The surface hands back the declared storage class and nothing more, so a u256 stored as decimal `text` comes back as `string | null` and the consumer calls `BigInt()`. The only rule available to a decoder is "text that parses as digits is a BigInt", which is exactly the irreducible ambiguity `tagged-bigint-codec-across-storage-adapters` exists to remove elsewhere, and it would make the two halves of one declaration disagree (the store wrote a string, the reader returned something else). Decoding therefore needs the DECLARATION to carry more than it does: a semantic type or codec beside the storage class, which every backend must agree on, with equality and ordering defined. That is a storage-side change with a conformance obligation, so it is recorded in **ADR-0025** and left to the bigint-codec task rather than invented in a read layer. Alternative considered and rejected: add `'uint256'` to `FieldType` here (touches every backend's DDL, the conformance suite and the write path, i.e. a different task). Touches: `tagged-bigint-codec-across-storage-adapters`, any future GraphQL scalar mapping.
- **The as-of parameter is whatever the STORE takes (`AsOfAddress<S> = Parameters<S['getAsOf']>[2]`).** Over a plain `StateStore` it is a block number; over `@etherfold/state-store-sqlite` it is a height, `{hash}` or `{timestamp}`, because the type is read off that store's own signature. Alternatives considered: (a) move `BlockAddress` / `parseBlockAddress` / `NoSuchBlockError` down to the seam so every surface takes an address — rejected as a separate, larger refactor that re-homes a published class; (b) accept an address everywhere and invent a new refusal ("this store cannot address by hash") — rejected because that is a new capability and a new error family at the wrong layer, duplicating what the store's signature already says; (c) numbers only — rejected because ADR-0015's error could then never reach a consumer of the generated surface. Consequence: a hash against a store with no addressing layer is a compile error, which is the only place it can be caught (at run time such a read answers a plausible `undefined`, see the observation note). Touches: any future backend that adds an addressing layer gets address-typed reads for free; any future portable addressing task supersedes this.
- **Rows are PROJECTED to the declared columns, not cast.** The surface returns the id columns and every declared field (unlisted → `null`, because a version is a whole row) and drops `_lower` / `_upper`. A cast would make the row type a claim; the projection makes it true, and stops a read being spread straight back into a write. Alternative: return the raw row and type it loosely — rejected, it hands a consumer state the declaration does not describe. User-visible: `Object.keys(row)` differs from `store.getCurrent`'s. A consumer that genuinely wants the version columns uses the store directly.
- **Fields are typed `| null`.** `set` writes a whole row, so an unlisted declared field IS null on the common partial-write path; a non-nullable field type would be a type that lies there. Cost: consumers write `?? 0` / `!`.
- **`createReadSurface` REFUSES a declaration that disagrees with the store's** (unknown entity, or a different id/field shape), throwing at construction and naming both shapes. This is a new user-visible error. The declarations must be passed (a `ReadonlyMap<string, NormalizedEntity>` has no literal types to derive from), so the redundancy exists; checking it turns "two descriptions that could drift" into "one description, verified once". Alternative: trust the caller — rejected, a stale copy would project a column the store does not have to `null`, a plausible wrong answer.
- **Naming, checked against `CONTEXT.md`.** `createReadSurface` / `createQuerySurface`, deliberately NOT `StateView`: `VersionedStateView` (`@etherfold/processor-sqlite`) already means "the untyped, entity-name-string read handle `process()` hands back", and re-meaning it would make one term mean two things. "Query surface" is used here in the glossary's existing sense (the richer, SQL-taking layer above the seam), which is narrower than the spec's looser prose use of the phrase for the whole generated layer; the glossary entry I added states both tiers so the term has one meaning. Touches: a later GraphQL package consumes `declareEntities` + these two factories.
- **Placement: the portable tier lives in `@etherfold/state-store`, the predicate tier in `@etherfold/state-store-sqlite`.** The seam package owns the declaration, has zero dependencies, and is already depended on by every backend, so a consumer can read state without pulling in `@etherfold/core`; `@etherfold/processor-entities` re-exports it for authoring ergonomics. Alternatives considered: a new `@etherfold/state-view` package (a publish surface for ~250 lines that only re-exports the seam's own vocabulary) and putting it in `processor-entities` (drags the indexer core into any read-only consumer). Note this sits ABOVE the seam while living in the seam's package: it is not part of `StateStore`, and the module doc and README say so.


wighawag:

| # | criterion | verdict |
| --- | --- | --- |
| 1 | Typed read surface from declarations: by id (current + as-of), list by prefix with a limit | ✅ |
| 2 | Types derived from the declaration; renaming breaks compilation; a type-level test pins it | ✅ and now genuinely enforced |
| 3 | Unresolvable address → error; out-of-retention → typed refusal; neither becomes `undefined` | ✅ |
| 4 | Works against at least two backends, unchanged | ✅ |
| 5 | The server-side richer query form stays available and distinct, with the reason stated | ✅ portable tier in the seam, predicate tier in `state-store-sqlite` |
| 6 | Ships no GraphQL; states how the decided Yoga/Pothos stack would consume it | ✅ |
| 7 | The `uint256` question answered explicitly, not cast away | ✅ ADR-0025 |
| 8 | Tests, vitest, changeset | ✅ |

### Criterion 2 is the reason this task exists, and it is now actually enforceable

The type-level assertions are `@ts-expect-error` lines — inverted assertions that fail to compile if the error they expect stops happening — covering an undeclared field, a wrong id column, an unknown entity, a missing limit, a non-leading prefix, a predicate/sort/offset, and a block address the store cannot resolve.

The file says why in as many words: *"`pnpm typecheck` is what runs these assertions (vitest strips types without checking them), which is why each one is a `@ts-expect-error`."* Worth stating plainly: before the gate repair earlier in this drive, this criterion would have been decoration — the assertions would have compiled to nothing and passed regardless. These are assertions that an error *occurs*, which is the opposite of the suppression I forbade elsewhere.

Two of them defend constraints from earlier tasks at the new layer: `the limit is required, here as at the seam: a default bound is a bound nobody chose`, and `there is nowhere to hang a predicate, a sort or an offset`. The bounded-listing decision is now enforced at both tiers rather than only where it was introduced.

### Criterion 3, handled at the only layer that can catch it

A hash passed to a store with no addressing layer is a **compile** error, because at run time it would be compared against block numbers, match no version, and come back as an ordinary `undefined` — indistinguishable from "the entity was absent then". The reasoning is recorded next to the assertion, and the run-time hole it implies was filed as an observation rather than patched here.

### The `uint256` answer is a decision, not a dodge

Not decoded: the surface returns the declared storage class, so a u256 stored as decimal `text` comes back `string | null`. The reason given is the right one — the only rule available to a decoder is "text that parses as digits is a BigInt", which is the irreducible ambiguity `tagged-bigint-codec-across-storage-adapters` exists to remove, and decoding would make the two halves of one declaration disagree. Recorded in ADR-0025 and handed to that task rather than invented in a read layer. Adding `'uint256'` to `FieldType` was considered and correctly rejected as a different task (it touches every backend's DDL, the write path and the conformance suite).

### Observations filed, two of which matter beyond this task

- `as-of-read-does-not-check-its-block-number`: a non-number `at` passes `assertRetained` and then matches no version, returning `undefined` rather than refusing — the exact undefined-as-if-absent failure mode, reachable from JS or a cast. Correctly not fixed here, and correctly located: a run-time guard belongs at the seam beside `assertRetained`, covering every backend at once. Carrying to triage as task-worthy.
- `stale-changesets-name-packages-that-no-longer-exist`: pending changesets still name pre-`etherfold` packages, so a plain `changeset status` (or `version`, or a release) errors, while the gate's `--since=main` hides it. That is a live landmine in the release path and it names the task that will meet it.

Merging.

PR #10, opened 2026-08-23
feat(one-processor-cli-and-split-server): The same processor and declar…

wighawag:
Task: `work/tasks/done/one-processor-cli-and-split-server.md`

Done. Everything is green: `format:check`, `build`, `typecheck`, `test` across the whole monorepo.

## What I did

**First, I checked the task against reality.** The spec is at `work/specs/tasked/one-processor-everywhere.md` (not `proposed/`), `portable-mutation-context-seam` is in `work/tasks/done/`, and ADR-0003 through 0006 plus ADR-0016 read as the task assumes. One premise needed resolving rather than stopping: the split deployment does not exist as deployables yet (`agnostic-log-fetcher` and `ingest-wire-receiving-side` are still in `work/tasks/backlog/`, and CONTEXT.md says so). The task is explicitly framed as a guard test built *before* the seam can close, and every library piece needed to assemble the split shape exists, so I built the split deployment in miniature out of those pieces rather than stopping.

**The test:** `packages/processor-sqlite/test/deployment-shapes.test.ts` (16 tests). One `EntityProcessor` (one `version`, one set of entity declarations, imported from the shared fixtures module) runs two ways over one captured chain:

- **single-process** — one `EthereumIndexer` doing fetch, stream-building and processing, driven by `indexMore()` against a replayed chain. The CLI shape.
- **split** — a stateless log-fetcher (built out of `captureStream`) pushing contiguous ranges across a JSON wire to an indexer-server that hosts the stream-builder (`EthereumIndexer.feed`) and the processor, and has no chain at all.

Both run against two storage backends (versioned rows in real libSQL; versioned rows in a `Map`), so four states must agree. The input is a stream fixture captured once and serialized once; every run re-parses the same text. The reorg replaces block 104 with a branch carrying fewer events, so the counter goes **down** 5 → 4, token 3 vanishes and token 2 goes back to what the confirmed block wrote.

**How the boundary is made checkable** (the deliverable the task asked me to state): four encodings, the first load-bearing — the indexer-server half is constructed with a provider that **throws on every JSON-RPC method**, naming the boundary. Because the same processor and the same core run both ways, a convenience added on the single-process side is exercised again on the split side, where it cannot be answered. I verified the guard actually bites by temporarily adding `indexer.load()` to the server half: it fails with the boundary message. The other three: the wire is JSON and asserted lossless; the envelope is asserted to be ADR-0004's with no `removed` markers and no `unconfirmedBlocks`; and the receiver is authoritative about the cursor, with a mis-started batch refused and nothing applied.

**Production change:** `EthereumIndexer.expectedFromBlock` (`packages/core/src/indexer.ts`), plus its own unit tests in `packages/core/test/feed.test.ts`. Changeset added.

**Observation captured** (outside scope, not fixed): `work/notes/observations/indexer-load-needs-a-chain-so-the-server-half-cannot-call-it.md`.

## Decisions

**1. `EthereumIndexer.expectedFromBlock` is new public surface on `@etherfold/core`.** ADR-0004 makes the receiver authoritative about where the next batch starts, and `generateStreamToAppend` already *enforces* it internally (it throws naming the value) — but there was no way to **ask**. Without it the split shape cannot exist at all: the stateless fetcher would have to compute the cursor itself, which is exactly the state ADR-0004 says it must not hold, and my test would have had to duplicate the `getFromBlock` formula (which `feed.test.ts` already does inline three times, and which would then silently drift). Alternatives considered: exporting the internal `getFromBlock` (worse — it hands out a free function that needs a `LastSync` the caller does not own), or duplicating the formula in the test (rejected: a duplicated formula cannot detect the seam closing). Named with ADR-0004's own word, so it does not fork a concept. **Touches** `ingest-wire-receiving-side` (backlog), which will need exactly this number for its `409` payload — I have added the getter, not the endpoint or its status semantics.

**2. The wire in this test carries DECODED `LogEvent`s, not raw logs.** ADR-0004 says `logs[]` without settling which. `EthereumIndexer.feed` — the receiving primitive the ADRs point at — takes decoded events, and `LogEventFetcher` is internal to `@etherfold/core`, so a receiver has no public way to decode. `captureStream` already decodes at fetch time for the same reason. Alternative: carry raw logs and decode server-side (would need a new public decoder, which is not this task's to add). **Touches** `ingest-wire-receiving-side`, which owns the real envelope; my `WireBatch` type is deliberately test-local and says so, so it does not pre-empt that decision.

**3. The generic `EventProcessor` over an arbitrary `StateStore` is a TEST-LOCAL harness, deliberately not published.** The backend-swap criterion needs the same processor driven by the core over a non-SQL store, and the only published bridge is `VersionedStateEventProcessor`, which is bound to `RemoteSQL`. Publishing a general one requires deciding **where a backend-neutral processor keeps its `LastSync` cursor**, and ADR-0016 makes that a property of where its state lives — a load-bearing, hard-to-reverse decision that belongs to the task wiring a browser deployment, not to this one. So the harness is `applyEventStream` plus the lifecycle calls, persists no cursor, and is documented as such at the definition site. **Coherence note:** it is a third thing implementing `EventProcessor` over entity declarations; it stays in `test/` precisely so the vocabulary does not gain a second published meaning. The published `VersionedStateEventProcessor` is still run on both shapes and compared on its own, so the equality is not only between harnesses.

**4. `replayStream` is not used, although the task's prompt names it.** It builds an `ExistingStream` — the kept-stream cache that sits in *front* of a fetch — and only the single-process shape has one; the split shape's fetcher needs a chain to fetch *from*, and routing both shapes through `feed()` would have made them the same code path and proved nothing. So the fixture is replayed one level lower, as the provider the chain-facing half talks to: `captureStream` once, `serializeStreamFixture` once, `parseStreamFixture` per run. The criterion ("both runs see identical bytes") is met; the named function is not the right seam for it.


wighawag:

| # | criterion | verdict |
| --- | --- | --- |
| 1 | One processor + declarations run under the single-process CLI and the split path, same state | ✅ |
| 2 | Comparison uses a replayed fixture, so both runs see identical bytes | ✅ |
| 3 | Switching backend is a config change touching no processor code, demonstrated | ✅ `makes the storage backend the only difference between two runs of the same processor` |
| 4 | The seam boundary is checkable, not merely claimed | ✅ see below |
| 5 | Reorg behaves the same on both paths, counter decreasing | ✅ |
| 6 | Tests, vitest, changeset | ✅ |

### Criterion 4 was the point of the task, and it is properly encoded

"The boundary is intact" is made checkable by **construction rather than assertion**: the split deployment's processing half is built with a provider that throws on every method, so any chain call it made would fail the run outright. Three things make that a real check rather than a gesture:

- `arms that refusal rather than merely declaring it` verifies the throwing provider actually throws, so the guard cannot be silently vacuous.
- The contrast assertion requires the single-process shape to have made chain calls (`> 0`), so the test cannot pass because nothing ran.
- The wire itself is pinned: `sends the ADR-0004 envelope across the wire and nothing else`, `carries nothing across the wire that is not JSON`, and `keeps the RECEIVER authoritative about the cursor`.

`compares the two shapes on the PUBLISHED processor, not only on a harness` is a good extra: it stops the comparison being run against a strawman that is easier to keep portable than the real thing.

### The one production change is additive and load-bearing

`EthereumIndexer.expectedFromBlock` is a new getter and nothing else changes. It is exactly what ADR-0004 needs to be implementable: a stateless log-fetcher holds no cursor, so it must ask the side that does, and without exposing this the sender would have to compute the cursor itself — which is precisely the state a stateless component must not hold. It delegates to the existing `getFromBlock` rather than reimplementing, and documents the subtle part: it is **not** `lastToBlock + 1`, it reaches back over the unconfirmed window so a reorg is detectable at all. `feed.test.ts` is +43/-0, a pure extension.

### Honest disclosure rather than a hidden gap

`indexer-load-needs-a-chain-so-the-server-half-cannot-call-it`: `EthereumIndexer.load()` opens with `eth_chainId` before reaching `processor.load()`, so the half that hosts the stream-builder and processor — which by ADR-0003 has no provider — cannot use it. The test works around it by calling `processor.load()` directly, and the note says plainly what that workaround skips (the kept-stream resume and the `indexerMatches` discard path) rather than letting the green run imply full coverage. It routes the fix to `ingest-wire-receiving-side`, which owns the real ingestion endpoint. That is the right call and the right destination.

Merging.

PR #11, opened 2026-08-23
feat(entity-identifier-sql-keyword): An entity column named `index` pas…

wighawag:
Task: `work/tasks/done/entity-identifier-sql-keyword.md`

All green: `pnpm format:check`, `build`, `typecheck`, and 1,342 tests across 17 packages.

## What I did

**Confirmed the defect first (it had drifted, but only in location).** The task pointed at `packages/state-store-sqlite/src/internal/identifiers.ts`; that module no longer exists. The shape-only guard moved to the seam at `packages/state-store/src/entities.ts` (`normalizeEntity`, applied by all four backends), and its own JSDoc named this task as "the remaining half of that hole". The defect still reproduced exactly as described: `{name: 'placementPlayer', id: ['epoch','position','index'], ...}` passed validation and then threw `SQLITE_ERROR: near "index": syntax error` at `migrate()`, as did `order` as an entity name and `default` as a field.

**Test-first.** `packages/state-store-sqlite/test/identifiers.test.ts` went red (13 failures) before any production change. Then the cross-backend property: a new shared conformance fixture `RESERVED` (entity `order`, id columns `group` + `index`, fields `select`/`table`/`where`/`default`/`references`/`primary`) added to `CONFORMANCE_ENTITIES`, plus a new case group `a declaration means the same thing on every backend` (`packages/state-store-conformance/src/cases/portable-declarations.ts`). With those in, the memory / patch / IndexedDB backends stayed green and **the entire SQLite conformance run went red (174 of 180)** — the defect in its strongest form.

**The fix: quote, don't reject.** New `packages/state-store-sqlite/src/identifiers.ts` (`quoted`, `quotedList`), applied to every identifier that came from a declaration in `ddl.ts`, `statements.ts` and `store.ts` (entity name, id columns, fields, and the index names derived from the entity name). The store's own fixed names (`_lower`, `_upper`, `_rowid`, `_blocks`) stay bare, so "quoted" reads as "this name came from outside".

Also: updated the now-stale JSDoc in `state-store/src/entities.ts` and the `state-store-sqlite` README, updated four existing tests that pin SQL text to the quoted form, and added `.changeset/entity-identifier-sql-keyword.md`.

**Acceptance criteria:** the keyword declaration migrates and round-trips (not accepted-then-fatal); cross-backend agreement has its own conformance group asserting both directions (accepted-everywhere and refused-at-declaration-everywhere); all nine keywords are covered as entity name, key column and data field; the `_` prefix and the shape rule reject exactly what they rejected before (asserted in both packages); no compatibility break, which the changeset states explicitly.

**Out-of-scope signal captured:** `work/notes/observations/entity-names-differing-only-in-case-collide-on-sqlite.md` — `token` and `Token` are two entities at the seam but one table on SQLite (verified: one table created, `getCurrent('Token', ...)` returns the `token` row). Same shape of defect, different cause, not fixed by quoting. Not touched. (The stale-changeset issue I hit while checking `changeset status` is already recorded in `stale-changesets-name-packages-that-no-longer-exist.md`.)

## Decisions

**Chose QUOTE over REJECT, and did not do both.** The task offered either. I quoted, in `@etherfold/state-store-sqlite` only. Why: the property that makes this a defect is that a declaration's legality must be a fact about the *declaration*, not about the deployment, and quoting is the option that makes this backend accept *exactly* what the shared seam accepts. Rejecting would have pushed one engine's reserved-word list into `@etherfold/state-store`, which every backend shares including the ones with no SQL in them, and would re-open the same hole the day SQLite adds a keyword or a second SQL backend arrives with a different list. Doing both would combine the compatibility break of rejecting with none of its benefit. **This is NOT a source-compatibility break**: every declaration legal before is still legal, and the ones that were fatal on SQLite now work. The rejected alternative *would* have been a break for anyone already declaring such a column. Touches: `@etherfold/state-store`'s validator (deliberately left shape-only, with the rationale now written there) and any future SQL backend, which inherits the obligation to quote rather than a list to copy.

**Quoted only DECLARATION-derived identifiers, not the store's own.** `_lower`, `_upper`, `_rowid`, `_blocks` and its three columns stay bare. Alternative considered: quote every identifier uniformly, which is a simpler rule to state but churns the fixed-schema DDL and every existing test that pins it, for names this package chose itself and knows are safe. The rule as written is legible at a glance ("quoted = came from outside"), and it is documented in `src/identifiers.ts`.

**Put the cross-backend fixture in `CONFORMANCE_ENTITIES` rather than in a case-local declaration set.** That makes every existing case migrate it, every revert sweep it and every prune consider it, so the keyword identifiers reach every statement a backend emits rather than only its DDL. Cost: it is a real (minor) change to the shared conformance surface, so a backend outside this repo that upgrades and does not quote will see its whole run go red rather than one case. That is deliberate and is stated in the changeset; it is also exactly what happened here, which is how I know the case has teeth.

**Exported `quoted` / `quotedList` from `@etherfold/state-store-sqlite`.** New public API surface (additive). The reason is the caller-supplied-SQL tier: `queryCurrent({where: 'default > ?'})` is the caller's SQL to write and therefore the caller's to quote, and without an exported helper every consumer with a keyword column hand-rolls one. Alternative considered: keep it internal (matching the `internal/` placement the original task assumed). I rejected that because the package exports every other module in `src/` and because the query tier genuinely needs it. Touches: anyone writing `QueryOptions.where` / `orderBy`.

**Did not add an ADR.** The choice is easy to reverse (rejecting keywords could be layered on later without undoing the quoting) and the trade-off is fully written at the choice site plus the changeset, so it does not clear the ADR gate.


wighawag:

| # | criterion | verdict |
| --- | --- | --- |
| 1 | A column named `index` either migrates and round-trips, or is refused at declaration time | ✅ quoted, so it migrates |
| 2 | Behaviour is the SAME across backends, with its own test | ✅ tested in both directions |
| 3 | A spread of SQL keywords, in a key column and a data field | ✅ an entity named `order`, in the shared fixture |
| 4 | The `_` prefix rule still rejects what it rejected | ✅ |
| 5 | If rejection was chosen, note the compatibility break | n/a — quoting is not a break, and that is stated |
| 6 | Tests, vitest, changeset | ✅ |

### The premise drifted and was handled correctly

The task named `packages/state-store-sqlite/src/internal/identifiers.ts`, which no longer exists: `portable-mutation-context-seam` moved the validator to `@etherfold/state-store/src/declarations.ts` earlier in this drive. The defect itself still reproduced (`ddl.ts` still interpolated `${column} TEXT NOT NULL` bare), and the builder confirmed that before fixing rather than building on the stale path.

More importantly, the move **changes which fix is right**, and the reasoning here is better than my own. I had expected the move to strengthen the "reject reserved words" option, since the validator is now shared. The opposite is true and the decision says why: the seam is shared by backends with **no SQL in them**, so pushing SQLite's reserved-word list into it would impose one engine's vocabulary on every store, and would re-open the hole the day SQLite adds a keyword or a second SQL backend arrives with a different list. Quoting makes this backend accept exactly what the shared seam accepts, which is the property that makes a declaration's legality a fact about the declaration rather than about the deployment. That is the actual defect, stated precisely.

It is also correctly identified as **not** a source-compatibility break: everything legal before is still legal, and what was fatal on SQLite now works. Rejecting would have been the break.

### Details that show care

- **Only declaration-derived identifiers are quoted.** `_lower`, `_upper`, `_rowid`, `_blocks` stay bare, under a rule that is legible at a glance: quoted means it came from outside. Quoting uniformly would have churned the fixed-schema DDL and every test pinning it, for names the package chose itself.
- **The keyword fixture went into `CONFORMANCE_ENTITIES`, not a case-local declaration.** So keyword identifiers flow through every statement a backend emits — migrate, revert, prune, listing — rather than only the DDL where the bug was first seen. The cost is stated honestly in the changeset: an external backend that upgrades without quoting sees its whole run go red rather than one case. "which is how I know the case has teeth" is the right way to know.
- **`to"ken` is in the cross-backend case.** That is the one I most wanted to see given the fix is quoting: an embedded double quote is exactly how a quoted identifier gets broken out of. The seam's shape guard already refuses it, so the case pins that a name refused by one backend is refused by all — agreement tested in the refusal direction as well as the acceptance direction, which is what criterion 2 actually asks for.
- **`quoted` / `quotedList` exported deliberately**, because `queryCurrent({where: 'default > ?'})` is the caller's SQL to write and therefore the caller's to quote; without a helper every consumer with a keyword column hand-rolls one.
- **No ADR, with the gate checked rather than skipped**: the choice is easy to reverse (rejection could be layered on later without undoing quoting) and the trade-off is written at the choice site.

Merging.

PR #12, opened 2026-08-23
feat(declaration-legality-is-one-rule-everywhere): A declaration is leg…

wighawag:
Task: `work/tasks/done/declaration-legality-is-one-rule-everywhere.md`

Done. `pnpm build && pnpm typecheck && pnpm test` is green (all 27 workspace projects), `pnpm format:check` clean.

## What I did

**The seam (`@etherfold/state-store/src/entities.ts`)** gained one rule: two identifiers that differ only in case are ONE identifier. Enforced at declaration time, on every backend, at all three levels — entity names (`normalizeEntities`), and id columns + fields together (`normalizeEntity`), since they are the columns of one row. A repeated id column (`id: ['id','id']`, previously accepted then `duplicate column name` at `migrate()`) is caught by the same pass. The message names both spellings and says to rename one. The module note now spells out why case belongs here when a keyword list does not: a keyword list is one engine's *vocabulary* and quoting erases it; case is *identity*, the backends disagree about it, and no DDL makes them agree, so it is a portability rule.

**`@etherfold/state-store-sqlite`** got two fixes for defects I found in the same class:
- Derived index names moved into the store's `_` namespace (`_token_open`, …). An index and a table share one namespace in SQLite, so `token` + `token_open` was `SQLITE_ERROR: there is already an index named token_open` here and two ordinary entities everywhere else. Fixed *by construction*, not by a new refusal.
- `sqlite_`-prefixed entity names are refused in the constructor (`assertStorableEntityNames`), because SQLite refuses them however quoted. Columns are untouched — SQLite allows those.

**`@etherfold/state-store-conformance`** carries every rule: the four case cases, a non-ASCII case, and `DECLARATION_PROBES` — legally-shaped names that strain some engine, each asserting only "refused when the store is CONSTRUCTED, or stored and read back", never accepted-then-fatal-at-`migrate()`. That is the shape that lets a genuinely engine-specific limit live in a *shared* suite.

Tests: `packages/state-store/test/declaration-identity.test.ts`, `packages/state-store-sqlite/test/reserved-names.test.ts`, plus the conformance cases (which run on memory ×3, patch, IndexedDB, SQLite ×3). Changeset added. Three existing SQLite tests had index-name literals updated (`token_open` → `_token_open`); no assertion changed meaning.

I also recorded one out-of-scope signal at `work/notes/observations/id-values-containing-nul-collide-on-the-string-keyed-backends.md` (verified: an id *value* containing U+0000 collapses two business keys into one on the memory and patch stores; same shape, value level, not reachable by a declaration-time rule). The note this task closes, `entity-names-differing-only-in-case-collide-on-sqlite.md`, is now resolved but I left it in place, since moving/dropping work/ items is the runner's.

## Decisions

**Audit outcome, per candidate.**

- *Names differing only in case* (the known instance) → **refused everywhere, at the seam.** No spelling of the DDL makes SQLite keep `token` and `Token` apart, so the answer cannot be a backend's. Touches: any existing declaration with two such names (a source break, named in the changeset).
- *Identifier length* → **legal everywhere, left alone.** Measured, not assumed: libSQL creates a 2,000-character table name plus its four derived indexes and round-trips a row through it; the other three backends hold names as ordinary JS strings, and IndexedDB puts the entity name in an array key with no limit. Inventing a seam limit would refuse declarations no engine objects to. Pinned by two probes (200-char entity name; 200-char id column and field) so a future backend that *does* impose one is caught. Alternative considered and rejected: a conservative 63-char limit borrowed from Postgres, which would be a fourth engine's rule imported into a repo that has no Postgres backend.
- *Non-ASCII / NFC-NFD collisions* → **already refused everywhere; no new rule.** `IDENTIFIER` is ASCII-only, so `café` and `cafe`+U+0301 are both rejected as identifiers and the case-shaped collision is unreachable in another alphabet. I made this load-bearing rather than incidental: it is *why* `fold` can be `toLowerCase` (never `toLocaleLowerCase`) and be exact, with no dotted-I or eszett to get wrong. Written at the choice site and pinned by a conformance case, so a future loosening of the regex fails a test rather than silently re-opening the hole.
- *Collision with the store's reserved space, beyond `_`* → **two real instances, answered differently.** (a) A backend's own DERIVED names: SQLite's `${entity}_open` index shared a namespace with entity tables. Fixed by construction (prefix `_`), so the declaration became legal everywhere rather than refused. I rejected the alternative — refusing an entity whose name equals another entity's derived index name — because it makes a declaration's legality depend on which *other* entities sit beside it, and leaks this store's naming scheme into the shared surface. (b) SQLite's own `sqlite_` namespace: unfixable by construction, so refused, in the SQLite package, at construction time. **This is deliberately NOT a seam rule**, for exactly the reason `entity-identifier-sql-keyword` kept the keyword list out: it is one engine's namespace and means nothing to a key-value store. Touches: `@etherfold/state-store-sqlite` only; a `sqlite_` entity stays legal on memory/patch/IndexedDB.
- *All of it at entity, id-column and field level* → the case rule applies at all three. The `sqlite_` rule applies to entity names **only**, because a `sqlite_` column is legal in SQLite (verified) and refusing one would be the same defect pointing the other way: this backend narrowing the seam for no engine reason. That non-refusal has its own probe, so it cannot be "tidied up" into a blanket prefix ban later.

**How a backend-specific limit lives in a shared suite.** `DECLARATION_PROBES` asserts a disjunction — the factory throws, *or* migrate + write + read-back succeeds — rather than a fixed outcome per backend. This is the one shape that lets the suite forbid the actual defect (accepted-then-fatal, accepted-then-divergent) without the suite having to know which engine refuses what. Alternative considered: per-backend expected-refusal lists in the suite, rejected because it puts every engine's vocabulary in the shared package, which is the thing this whole line of work is avoiding. Touches: every backend that adds this version of the suite; a backend that dies at `migrate()` on any probe now goes red.

**Renaming the derived indexes is a break for a stored DATABASE, not for a declaration.** An existing database re-migrates cleanly (`IF NOT EXISTS`) and keeps its old `token_open` etc. as redundant duplicates of the new `_token_open`; the unique-open-version invariant is enforced by both, so nothing is wrong, only wasteful. Named in the changeset. Alternative considered: emitting `DROP INDEX` for the old names, rejected because this package has no migration-versioning concept and inventing one to tidy indexes is a much bigger change than the defect warrants.

**"Probe" reused rather than a new word.** The conformance package already uses "probe" for a store used to read capabilities; `DECLARATION_PROBES` is a declaration used to interrogate a backend's limits. Same sense (a thing used to interrogate), so I reused it instead of coining a second term. Flagging it because it is the kind of near-miss the coherence check exists for.

**No ADR.** Both halves are easy to reverse (the case rule is one function; the index prefix is one template string) and the trade-offs are written at the choice sites plus the changeset, so neither clears the ADR gate. ADR-0018's "the rule is set by the strictest backend and applied uniformly" already covers the seam-vs-backend split this used.


wighawag:

All criteria met, and the audit was done properly rather than reduced to the one known instance.

**The seam did not gain a keyword list** (checked: zero SQL vocabulary in `declarations.ts`), which was the constraint carried over from `entity-identifier-sql-keyword`. The distinction is stated where a reader meets it: a case-insensitive *uniqueness* rule is a portability rule and belongs at the seam; one engine's reserved words are not and do not.

**The audit found things I did not anticipate, and answered them differently on their merits:**

- *Length* → left legal, **measured rather than assumed**: libSQL round-trips a 2,000-character table name plus its derived indexes. Importing Postgres's 63-char limit into a repo with no Postgres backend was considered and rejected. Probes pin it so a future backend that *does* impose a limit is caught.
- *Non-ASCII / NFC-NFD* → already unreachable, and made **load-bearing rather than incidental**: because `IDENTIFIER` is ASCII-only, `toLowerCase` (never `toLocaleLowerCase`) is exact, with no dotted-I or eszett to get wrong. Pinned, so loosening the regex later fails a test instead of silently re-opening a case-folding hole.
- *Reserved space* → **two real instances, answered oppositely and correctly**. SQLite's own derived index names (`${entity}_open`) shared a namespace with entity tables: a genuine latent bug, fixed **by construction** with a `_` prefix so the declaration becomes legal everywhere rather than refused. Refusing an entity whose name collided with another entity's derived index was rejected for the right reason — it would make a declaration's legality depend on which *other* entities sit beside it. Meanwhile `sqlite_` is unfixable by construction, so it is refused **in the SQLite package only**, and a `sqlite_` entity stays legal elsewhere.
- *Level* → the case rule applies at all three; `sqlite_` applies to entity names only, because a `sqlite_` **column** is legal in SQLite (verified) and refusing one would be "the same defect pointing the other way". That non-refusal has its own probe so it cannot be tidied into a blanket ban later.

**`DECLARATION_PROBES` is the right shape.** It asserts a disjunction — the factory throws, *or* migrate + write + read-back succeeds — so the shared suite forbids the actual defect (accepted-then-fatal, accepted-then-divergent) without having to encode each engine's vocabulary. Per-backend expected-refusal lists were considered and rejected, which is the same instinct that kept the keyword list out of the seam.

**The index rename is honestly characterised** as a break for a stored database rather than for a declaration: an existing DB re-migrates cleanly and keeps the old indexes as redundant duplicates, the unique-open-version invariant is enforced by both, and `DROP INDEX` was rejected because this package has no migration-versioning concept. Named in the changeset.

**New finding, correctly scoped out:** `id-values-containing-nul-collide-on-the-string-keyed-backends` — `entityKey` joins with `\u0000`, so `{x:'a\u0000b', y:'c'}` and `{x:'a', y:'b\u0000c'}` become one row on the memory and patch backends while SQLite and IndexedDB keep them apart. Same defect shape as this task's, one level down at the VALUE level, where a declaration-time refusal cannot reach it. Verified, not fixed here, with candidate fixes named. That is exactly the right call.

One nit, not blocking: `entity-names-differing-only-in-case-collide-on-sqlite.md` is now fixed but the note is still there. I will sweep it after merge.

Merging.

PR #13, opened 2026-08-23
feat(a-read-that-cannot-be-served-refuses): Close the two remaining pat…

wighawag:
Task: `work/tasks/done/a-read-that-cannot-be-served-refuses.md`

All checks green (`pnpm build && pnpm typecheck && pnpm test`, plus `prettier --check`). Both holes are closed.

## What I did

**Bug 1 — an as-of read now checks that its block number is a block number.**

- `packages/state-store/src/blocks.ts`: added `isBlockNumber(value)` (whole, non-negative number) and `assertBlockNumber(at)`. It lives in `blocks.ts` because that module already owns the numeric/textual contract of `BlockPointer` for every backend.
- `packages/state-store/src/errors.ts`: added `InvalidBlockNumberError extends TypeError`, carrying `received`, with an optional message override. Documented as *deliberately not* a `BlockUnavailableError`; the family's doc now names it as the non-member.
- `packages/state-store/src/retention.ts`: `assertRetained` calls `assertBlockNumber(requested)` **first**, before the store's claim is consulted. That is the one call every backend taking a resolved number already makes, so memory, patch and IndexedDB inherit the guard on both `getAsOf` and `listAsOf` with no per-backend copy.
- `packages/state-store-sqlite/src/blocks.ts`: `assertHeight` now uses the seam's `isBlockNumber` and throws the same `InvalidBlockNumberError` (message unchanged, still naming the address). The addressing layer is otherwise untouched: height, `{hash}`, `{timestamp}` resolve as before, and `NoSuchBlockError` still answers an address that resolves to no recorded block.

**Bug 2 — `MutationContext.get` answers with a whole row.** `get` now builds a staged row via `stagedRow`, the construction `list` already used, so id columns and unlisted declared fields (as NULL) are present whether the row was written earlier in this block or in an earlier one. One function, two readers, no second implementation.

**Tests** (written before the production code at each seam):

- `packages/state-store/test/as-of-takes-a-block-number.test.ts` (new): what counts as a block number, the error being a `TypeError` outside the family, the guard preceding the retention refusal, and the observation's exact case (`MemoryStateStore.getAsOf(..., {hash: '0x64'})`) refusing instead of answering `undefined` — plus that a real block with an absent entity still answers `undefined`.
- `packages/state-store/test/mutation-context.test.ts`: staged `get` returns `{id, owner, transferCount: null}`, has the same declared columns as a read one block later, and agrees with `list` row-for-row.
- `packages/state-store/test/read-surface.test.ts`: the test that previously *asserted* the buggy `undefined` now asserts the run-time refusal (its compile-time point is kept).
- `packages/state-store-sqlite/test/block-addressing.test.ts`: a malformed height is not a `BlockUnavailableError`, and a new group pins that the seam guard did not narrow this layer (all three axes still answer; `NoSuchBlockError` still thrown; `resolveBlockNumber` still soft).
- Conformance (`@etherfold/state-store-conformance`): `versioned reads` gains "an `at` that is not a block number is REFUSED, not answered as absent" (asked of every backend, using `{number: '100'}` so it is equally a caller bug on a backend that resolves a richer address), and `read-your-writes within a block` gains the whole-row shape case plus `get`/`list` agreement. New `declaredColumns` fixture strips the store's `_` namespace so shapes compare across backends.

**Standing conventions**: `.changeset/a-read-that-cannot-be-served-refuses.md` added (`state-store` minor, `state-store-conformance` minor, `state-store-sqlite` patch). `CONTEXT.md` glossary updated in two places (the `BlockUnavailableError` entry names the deliberate non-member; the `MutationContext` entry states the one-shape rule).

## Decisions

- **A non-number `at` is `InvalidBlockNumberError extends TypeError`, outside the `BlockUnavailableError` family.** Every member of that family is a fact about the STORE (the address resolved to no block; the versions are outside retention), and a caller acts on one by re-pinning or widening retention. A non-number `at` is a fact about the CALL: no store configuration makes it answerable, so it is a programmer error, and a `catch (e) { if (e instanceof BlockUnavailableError) ... }` written for "my historical read did not happen" must not swallow it. Alternatives considered: (a) `BlockNotRetainedError` — wrong, the block is not outside a window, it is not a block; (b) a new member of the family — it would make "catch the family" mean two different kinds of news; (c) a bare `Error`, matching the repo's other caller-bug throws (`assertListingLimit`, `resolveRetention`) — rejected because the conformance suite has to assert a TYPE, not a message, or a backend could satisfy it by throwing anything. It touches: `@etherfold/state-store`'s public error surface, and `@etherfold/state-store-sqlite`'s height axis (below).
- **The guard sits INSIDE `assertRetained`, as its first statement, rather than as a separate call each backend makes.** "Beside `assertRetained`" as a sibling function would have meant three call sites to add and to keep. Placing it first also fixes WHICH refusal a caller gets: a `revert-only` store would otherwise report `{hash: '0x64'}` as "not retained" and send its owner off to widen a window that was never the problem. Consequence, and it is intended: `listAsOf` inherits the same guard, not just `getAsOf`.
- **`@etherfold/state-store-sqlite`'s HEIGHT axis now throws the seam's `InvalidBlockNumberError` instead of a bare `Error`** (same message, so existing assertions on `/block address/i` still hold). Reason: it is the identical caller bug on the other side of an addressing layer, and it is what lets ONE conformance case be honest for every backend, including one whose `getAsOf` legitimately takes a richer address. The malformed-SHAPE branch (`{}`, `{height: 101}`) keeps its plain `Error`: "this names no axis I have" is an addressing-layer concern that has no meaning at the seam. Touches: anyone catching that specific throw — it is still an `Error`, now also a `TypeError`.
- **The conformance case uses `{number: '100'}` as its non-block value** rather than the observation's `{hash: '0x64'}`. A hash is a LEGAL address on `@etherfold/state-store-sqlite`, which runs this same suite, so asserting a refusal for it would fail an honest backend; `{number: '100'}` is a caller bug on both sides of an addressing layer. The hash case is pinned where it belongs, in `@etherfold/state-store`'s own tests.
- **No ADR.** ADR-0015 already settles the family and its shape rule; this adds a non-member and a guard, whose rationale lives in the class doc, the changeset and the `CONTEXT.md` glossary entry. Say the word if you would rather it be an ADR.


wighawag:

Both holes closed, and the two judgement calls I flagged were made well.

**The error type is right, and for the right reason.** `InvalidBlockNumberError extends TypeError`, deliberately **outside** the `BlockUnavailableError` family: every member of that family is a fact about the STORE (the address resolved to nothing; the versions are outside retention), and a caller acts on one by re-pinning or widening retention. A non-number `at` is a fact about the CALL — no store configuration makes it answerable. The decisive argument is the one I would have wanted: a `catch (e) { if (e instanceof BlockUnavailableError) }` written for "my historical read did not happen" **must not swallow a programmer error**. Putting it in the family would have corrupted the family's meaning. A bare `Error` was also rejected correctly, because the conformance suite has to assert a TYPE, not a message, or a backend satisfies it by throwing anything.

**Placing the guard inside `assertRetained` fixes more than it was asked to.** As its first statement, it also determines *which* refusal a caller gets: without it, a `revert-only` store would report `{hash: '0x64'}` as "not retained" and send its owner off to widen a window that was never the problem. Wrong-refusal is its own bug and this closes it. `listAsOf` inheriting the same guard is a stated, intended consequence rather than an accident.

**The subtlety I most wanted checked was caught.** The conformance case uses `{number: '100'}`, not the observation's `{hash: '0x64'}` — because a hash is a **legal** address on `@etherfold/state-store-sqlite`, which runs this same suite, so asserting a refusal for it would have failed an honest backend. A careless implementation would have copied the observation verbatim and made the shared suite reject a correct store. The hash case is pinned where it belongs, in the seam package's own tests.

Verified alongside: SQLite still accepts and resolves height, `{hash}` and `{timestamp}` (unchanged), its height axis now throws the seam's error with the same message so existing `/block address/i` assertions still hold, and the malformed-*shape* branch keeps a plain `Error` because "this names no axis I have" is an addressing-layer concern with no meaning at the seam. That distinction is finer than the task asked for and is correct.

**`get` and `list` now share one path.** `stagedRow` is used by both (`mutation-context.ts:114` and `:160`), so the row shape cannot drift apart again — which was the actual requirement, not merely making `get` correct today.

On the ADR question you raised in the decisions: I agree it does not clear the gate. ADR-0015 settles the family and its shape rule; this adds a non-member plus a guard, and the rationale is at the class doc, the changeset and the glossary. No ADR needed.

Merging.

PR #14, opened 2026-08-23
feat(clear-the-release-path-and-archive-the-old-names): Clear the chang…

wighawag:
Task: `work/tasks/done/clear-the-release-path-and-archive-the-old-names.md`

## What changed

**1. Reproduced the breakage, then found the task's premise was wrong.** `pnpm changeset status` (no `--since`) exited 1 on `Found changeset bigint-literal-guard-and-hash-prefix for package ethereum-indexer which is not in the workspace`; `--since=main` exited 0, which is why the gate never saw it.

The task says the stale changesets are "unconsumed release intents for versions long since shipped", so deleting them is the honest fix. **They are not.** The last release that consumed changesets was 2024-08-29 (`git log -- '*/CHANGELOG.md'`); every one of the 54 pending files was added between 2026-05-31 and 2026-08-23. I spot-verified that the features they describe are in the source and in no `CHANGELOG.md` (`parseLogBlockTimestamp` in `packages/core/src/internal/decoding/LogEventFetcher.ts`, `dispose()` in `packages/browser/src/IndexerState.ts`; neither word appears in either changelog). The source observation note says it correctly: "the unreleased changeset backlog". ADR-0017's "the 37 files in `.changeset/` ... record what was published" is the error that propagated. So I took the task's own conditional branch ("check before sweeping ... rewrite rather than delete") and applied it to the finding that **the exception is the whole set**. See the Decisions block.

**2. Swept `.changeset/` (37 files touched).** 33 rewritten to the ADR-0017 mapping (`ethereum-indexer` → `@etherfold/core`, `-cli` → `etherfold`, `@ethereum-indexer/*` → `@etherfold/*`, and so on), plus prose inside the bodies so the published changelog does not name packages that no longer exist. `ethereum-indexer-db-utils` / `ethereum-indexer-server` were dropped from the 5 multi-package changesets that named them (ADR-0010's own precedent for a retiring package). 4 files that named **only** `ethereum-indexer-server` were deleted, their bodies transcribed verbatim into `archive/ethereum-indexer-server/CHANGELOG.md` under `## Unreleased, and never will be`.

**3. Archived, not deleted.** `packages/ethereum-indexer-db-utils` and `packages/ethereum-indexer-server` → `archive/`, outside the `pnpm-workspace.yaml` globs. `pnpm build|typecheck|test` scope went from 19 to 17 projects; the lockfile no longer mentions either. `archive/README.md` states what they are, why they are kept, that nothing builds or publishes them, which ADR retired them, that their manifests no longer resolve, and that npm is explicitly out of scope. `archive/` added to `.prettierignore`; the `db-utils` and `old-server` panes dropped from `zellij.kdl`; path notes added to the two `docs/reviews/` files that read against the moved code.

**4. Something live DID import them — reported.** Four examples (`event-processor-bleeps`, `-conquest-eth`, `-conquest-fplay`, `-nfts`) carried `"ethereum-indexer-server": "workspace:*"` as a devDependency for the `eis` bin, used by their `eis` and `test` scripts. A `workspace:*` on an out-of-workspace package fails `pnpm install` outright, so the move required unhooking them. No `.ts` source imports either package. Details in Decisions.

**5. ADR-0016 / ADR-0017 resolved by amending ADR-0016**, with a dated amendment section arguing it on the merits (the package's public vocabulary is literally `JSProcessor` / `fromJSProcessor`, so `processor-js` would just move the incoherence into the API), and explicitly noting that renaming was affordable and refused anyway. Both observation notes deleted, since neither is true any more.

**6. Test-first.** `packages/core/test/pendingChangesets.test.ts` (beside the existing workspace-level `publishedTypeDependencies.test.ts`) asserts every package named in `.changeset/*.md` front matter is in the workspace, and that no changeset names zero packages. Written first: 33 failing / 21 passing before the sweep, 50 passing after. This closes the gate's `--since=main` blind spot permanently.

**Gate:** `format:check` + `changeset status --since=main` + `build` + `typecheck` + `test` all green; plain `pnpm changeset status` exits 0; `pnpm build:examples` still green.

No changeset added for this change: no published package's surface changed, and a package *removal* cannot be expressed as a changeset (a changeset naming an absent package is exactly the failure being fixed) — ADR-0010 already records that consequence.

## Decisions

**Rewrote all ~30 stale changesets instead of deleting them; the task said delete.** The task's premise ("release intents for versions already shipped under retired names") is false: nothing has been released since 2024-08-29, and all 54 pending changesets were written in 2026, describing work that is in the tree and in no changelog (ESM-only `tsc` build, viem v2, `dispose()`, CLI exit codes, `blockTimestamp` off the log, the published-type-dependency fix). Deleting them would have discarded the entire relaunch changelog for ~18 months of work. The task's own instruction covers this ("if any describes a change that genuinely has NOT shipped, that one is rewritten") — the finding is simply that the exception is 100% of the set, which inverts its headline. I proceeded rather than stopped because the direction is information-preserving and cheap to reverse (`rm` on 33 text files, nothing published yet), whereas deleting first would be the expensive mistake. Alternative considered: delete as literally instructed and note the loss; rejected because "re-announce old changes as new" cannot apply to changes that were never announced. **Touches:** the relaunch changelog every `@etherfold/*` package will ship, and the version bumps `publish-etherfold-and-deprecate-old-names` will publish (core 0.6.21→0.7.0, browser 0.7.7→0.8.0, `etherfold` 0.6.30→0.7.0, and so on, rather than a bare rename note). **If the human wanted the deletion, this is the one thing to reverse.**

**Deleted 4 changesets rather than rewriting them: the ones that named only `ethereum-indexer-server`** (`fix-server-feed-route`, `server-indexing-robustness`, `server-route-errors-and-auth`, `server-use-shared-processor-setup`). Their package is archived and will never be published again, so there is no release for them to land in, and rewriting them onto `@etherfold/server` would attribute Koa/PouchDB work to a different codebase. To avoid losing the notes I transcribed all four bodies into `archive/ethereum-indexer-server/CHANGELOG.md` under `## Unreleased, and never will be`, with a one-line equivalent in `archive/ethereum-indexer-db-utils/CHANGELOG.md`. Alternative: delete outright; rejected as inconsistent with "archive, do not delete".

**Four examples DID depend on the archived server, so I unhooked them.** `event-processor-{bleeps,conquest-eth,conquest-fplay,nfts}` each had `"ethereum-indexer-server": "workspace:*"` plus `"eis": "eis"` and `"test": "eis run -p ./dist/index.js --disableCache"`. I removed the devDependency and both scripts. Not repointed at `etherfold`: the old `eis run` is a long-running server with a PouchDB store, the new `etherfold index` is a one-shot write-to-file, so there is no drop-in replacement, and `etherfold serve` does not yet take a processor path. These `test` scripts were manual runs against a live node, never invoked by the gate (root `test` filters `./packages/*` and `./platforms/*` only, and `typecheck-tests-in-the-acceptance-gate` already recorded that no example has a vitest suite). **Touches:** anyone who ran `pnpm --filter event-processor-nfts test` loses that command; a follow-up giving the examples a smoke test against `etherfold serve` would be legitimate work, but it is not this task.

**Settled ADR-0016 vs ADR-0017 by amending ADR-0016, not by renaming to `@etherfold/processor-js`.** Nothing is published under `@etherfold/*` yet (the rename changeset is still pending), so the rename would have cost zero npm churn; the amendment states that explicitly and refuses on the merits instead. The reason: the package's public API is the type `JSProcessor` and the factory `fromJSProcessor`, named in every processor file anyone writes. `@etherfold/processor-js` exporting `JSProcessor` re-creates the same mismatch one layer down, and fixing *that* means renaming the most-used authoring surface in the project — a real breaking change bought with sort order. ADR-0016's own consequence ("the class inside is named for the storage model, not the backend") assumes the backend does not appear in the API; for this package it does. `processor-sqlite` is unaffected (its export is `VersionedStateEventProcessor`). Alternative: finish the rename; rejected also because it would have invalidated `publish-etherfold-and-deprecate-old-names`'s step-3 runbook, which I must not edit. The amendment marks the exception as an exception, not a precedent: new `processor-<backend>` packages still follow the rule. **Touches:** ADR-0017's mapping table (now confirmed correct as written) and the sibling publish task (now unchanged).

**Added `archive/` to `.prettierignore`.** The criterion only asked for not built / typechecked / tested / versioned / published, so this goes slightly further: it stops a future prettier upgrade rewriting code nobody will read a diff of. Reversible in one line if the repo prefers uniform formatting.

**Added the guard test in `packages/core/test/`, not a new workspace-level harness.** It is not a core test; it reads `.changeset/` and the workspace manifests. It lives there because `publishedTypeDependencies.test.ts` already established that seam ("there is no workspace-level test harness") and inventing a second convention for the second such test is worse than reusing the first. **Touches:** if a workspace-level harness is ever created, both files move together.


wighawag:

**This task's headline instruction was wrong, the builder caught it, and it was right to override me.**

I wrote "delete the stale changesets — they are unconsumed release intents for versions already shipped under retired names." That premise is false and I verified the refutation independently:

- newest release tag is **2024-08-29**; `@etherfold/core` is still `0.6.21` in both `package.json` and its CHANGELOG
- the "stale" changesets were added **2026-06-01**, nearly two years after that release
- `packages/core/CHANGELOG.md` has **zero** mention of the ESM-only `tsc` build, which is in the tree

So they describe work that shipped in no release at all. Deleting them would have discarded roughly 18 months of relaunch changelog. The builder used the task's own escape hatch ("if any describes a change that genuinely has NOT shipped, rewrite that one"), found the exception is 100% of the set, and correctly noted that inverts the headline. Proceeding rather than stopping was the right call too: rewriting is information-preserving and reversible with `rm` on 33 text files, while deleting first is the unrecoverable direction. Flagging it as "the one thing to reverse if the human wanted deletion" is exactly how a contradicted instruction should be handled.

**The consequence is real and I am surfacing it rather than burying it in an approval:** version bumps become `core 0.6.21→0.7.0`, `browser 0.7.7→0.8.0`, `etherfold 0.6.30→0.7.0` and so on, shipping a genuine relaunch changelog, instead of a bare rename note.

**Criteria, verified rather than trusted:**

- Plain `pnpm changeset status` (no `--since`) **exits 0** on this branch — I ran it in a throwaway clone. That was the blocked release path.
- `archive/` holds both retired packages and is outside the `pnpm-workspace.yaml` globs; `pnpm --filter './archive/*'` matches no projects, so install/build/typecheck/test/changeset genuinely do not see it.
- The README states what is there, why, that nothing is built or published, and — a good honest touch — that the manifests still carry `workspace:*` deps that no longer resolve, so git history is the record of when they last worked.
- Both resolved observations deleted.

**The four judgement calls all land well.** Changesets naming only `ethereum-indexer-server` were deleted rather than rewritten (its package will never publish again, and repointing them at `@etherfold/server` would attribute Koa/PouchDB work to a different codebase) — with their bodies transcribed into `archive/ethereum-indexer-server/CHANGELOG.md` so the notes survive. Four examples *did* depend on the archived server: reported **and** unhooked, with a clear statement of why they were not repointed (`eis run` is a long-running PouchDB server; `etherfold index` is one-shot; `etherfold serve` takes no processor path yet).

**ADR-0016 was amended rather than obeyed, on the merits.** The argument is better than the rule: the package's public API is `JSProcessor` / `fromJSProcessor`, so `@etherfold/processor-js` exporting `JSProcessor` re-creates the same mismatch one layer down, and fixing *that* means renaming the most-used authoring surface in the project. ADR-0016's own consequence assumes the backend does not appear in the API; here it does. Marked as an exception rather than a precedent, with `processor-sqlite` explicitly unaffected. It also declined to edit the sibling publish task's runbook, which was the right boundary.

One thing for the human rather than a blocker: `pnpm --filter event-processor-<x> test` no longer exists for four examples. Those were manual runs against a live node, never invoked by the gate.

Merging.

PR #15, opened 2026-08-23
feat(three-small-gaps-left-by-the-seam-drive): Three small independent…

wighawag:
Task: `work/tasks/done/three-small-gaps-left-by-the-seam-drive.md`

All three gaps are fixed and the full gate is green.

## What I did

**(1) Package-root `.ts` files are typechecked.** Every package's `tsconfig.typecheck.json` `include` went from a list of directories (`src/**/*.ts`, `test/**/*.ts`, plus per-package one-offs) to `["**/*.ts"]`, still excluding `node_modules`/`dist` (17 files). That pulls in the three package-root configs that existed: `platforms/cf-worker/vitest.config.ts` and `packages/state-store-indexeddb/vitest.config.ts` (both previously checked by nothing) and `playwright.config.ts` (previously covered only because someone remembered to name it). Verified by breaking `platforms/cf-worker/vitest.config.ts` on purpose (`const deliberateBreak: number = __dirname;`): `pnpm typecheck` went red naming the file, and green again on restore. `CONTEXT.md`'s conventions bullet and the root `package.json` typecheck section header now describe the wider include.

Added `packages/core/test/typecheckCoverage.test.ts`: for every package the root script fans out over, it asserts a `typecheck` script exists (pnpm *skips* a package without one, silently), a `tsconfig.typecheck.json` exists, and its include is exactly `["**/*.ts"]`. Proved it goes red by narrowing `packages/utils`' include back to the old pair.

**(2) `VersionedStateView` forwards `listCurrent` / `listAsOf`.** Test first (`packages/processor-sqlite/test/view-listing.test.ts`, red with `view.listAsOf is not a function`), then the forwarding in `src/view.ts`. Assertions go through the handle `process()` actually returns, over a composite-id entity so `{owner}` is a real prefix: tip listing, truncation, as-of by height and by hash, and the `NoSuchBlockError` refusal. `@etherfold/state-store-sqlite` now re-exports `EntityIdPrefix` and `Listing` (its own `listCurrent` signature already named them, but the package did not export them). Handle stays untyped, as specified.

**(3) abitype collapsed to one copy.** Root `pnpm.overrides` pins `abitype: 1.2.4`. Verified by inspecting the installed tree after `rm -rf node_modules && pnpm install --frozen-lockfile`: exactly one `abitype` directory on disk. `packages/core/test/abitypeIdentity.test.ts` pins it: type-level mutual-assignability aliases for `Abi` / `AbiEvent` / `AbiParameter` / the register-sensitive `AbiParameterToPrimitiveType<{type:'uint256'}>`, a viem-parsed event flowing into a `core`-typed slot, and a resolution check comparing the `abitype` version reached from `core` with the one reached from inside `viem`. Proved that last case red by removing the override and reinstalling (`expected '1.2.3' to be '1.2.4'`).

Deleted all three observation notes (all three are now false). Added `.changeset/view-listing-on-the-read-handle.md` (processor-sqlite minor, state-store-sqlite minor).

Gate: `pnpm format:check`, `pnpm build`, `pnpm typecheck`, `pnpm test` all green (16 packages, cf-worker's workers-pool suite included). `changeset status --since=main` only counts changesets git can see in a diff, so it passes once the new changeset is staged/committed by the runner.

One note on inputs: the task body is at `work/tasks/backlog/three-small-gaps-left-by-the-seam-drive.md`, not `ready/` as the prompt said, and it has no `spec:` field, so there was no source spec to read. I did not move it.

## Decisions

**Root-level `.ts` files are picked up by widening each package's typecheck include to `["**/*.ts"]`, not by adding `*.ts` to the existing list and not via a shared root base.** Why: a list of directories is the thing that has to be remembered, and this repo already has the scar — `packages/state-store-indexeddb` had `browser/**/*.ts` and `playwright.config.ts` hand-added to its include and `vitest.config.ts` forgotten. `**/*.ts` makes the rule statable ("every TypeScript file in the package except build output and dependencies") and covers a new root config *or* a new directory the day it appears. Alternatives considered: (a) adding `"*.ts"` to each include, which fixes root files only and leaves the same forget-a-directory hole; (b) hoisting the include into a shared root `tsconfig.typecheck.base.json` that packages extend, rejected because TypeScript resolves `include` relative to the file it is WRITTEN in, so a base-level `**/*.ts` would mean the repo root, not the package: a base can dedupe the compilerOptions but cannot carry the one line that must be remembered. Touches: every package's `tsconfig.typecheck.json`, plus `CONTEXT.md`'s conventions bullet and the root `package.json` typecheck section header, which both described the old globs.

**"A new package inherits automatically" is enforced by a test rather than by config inheritance.** TypeScript gives no way for a package to inherit an include, so a new package still authors its own `tsconfig.typecheck.json`; the previous task (`typecheck-tests-in-the-acceptance-gate`) recorded that a package with no `typecheck` script is skipped silently by the root fan-out. Rather than leave that as a second thing to remember, `packages/core/test/typecheckCoverage.test.ts` asserts both halves for every package in the fan-out. Alternative considered: leaving it, since the criterion is literally about root files. Touches: any future package or platform must carry the script + the config with the canonical include, or the core suite goes red naming it; it lives in core's suite for the same reason `publishedTypeDependencies.test.ts` does (there is no workspace-level harness).

**The abitype override pins UP to `1.2.4`, viem's copy moving, rather than down to viem's exact `1.2.3`.** viem 2.52.0 depends on `abitype@1.2.3` exactly, while `@etherfold/core` and `@etherfold/js-processor` declare `^1.2.4`. Pinning down would make the workspace resolve something *older* than what a consumer installing `@etherfold/core` from npm gets (overrides are workspace-local and do not propagate), so the local tree would stop reflecting the published one, which is the opposite of what the check is for. Pinning up gives viem a patch of its own dependency and keeps local and published resolution identical. The gate is green, including every viem-touching suite, so the conditional "back it out and record a negative result" branch did not fire. Touches: root `package.json` and `pnpm-lock.yaml`; the pin can be dropped the day viem itself moves to 1.2.4, and `packages/core/test/abitypeIdentity.test.ts` is what will say so.

**The type-level assignability check is not the whole guard, and the file says so.** The criterion asked for a type-level pin that viem-derived and `core`-derived ABI types stay mutually assignable. I verified what it actually detects: with the override removed and both copies installed, `pnpm typecheck` still passed, because abitype's types are structural and the two copies were structurally identical. So the type-level aliases catch the copies *diverging* (a later abitype reshaping `Abi`, or a `Register` augmentation merging into one copy only) and not the second copy existing. I added a resolution-based case in the same file that compares the `abitype` version reached from `core` with the one reached from inside `viem`, which is what actually fails on a second copy (proved red). Alternative considered: scanning `node_modules/.pnpm` for `abitype@*` entries, rejected because it hard-codes pnpm's layout and breaks under a different linker, and because pnpm leaves unpruned directories behind after an override change (it did here) so the scan would report copies that nothing resolves.


wighawag:

All three gaps closed, all three observations deleted, and two of the three decisions are better than the task asked for.

**(1) The typecheck include went to `**/*.ts`, not `*.ts`** — and the reason is the right one: a *list of directories* is the thing that has to be remembered, and this repo already carries the scar. `packages/state-store-indexeddb` had `browser/**/*.ts` and `playwright.config.ts` hand-added to its include and `vitest.config.ts` forgotten, which is the bug being fixed. `**/*.ts` makes the rule statable and covers a new directory as well as a new root config.

The rejection of a shared root base is technically correct rather than a preference: TypeScript resolves `include` relative to the file it is *written in*, so a base-level `**/*.ts` would mean the repo root, not the package. A base can dedupe `compilerOptions` but cannot carry the one line that matters.

And since TypeScript cannot make a package inherit an include, "a new package inherits automatically" is enforced by a **test** (`typecheckCoverage.test.ts`) asserting every package in the fan-out carries both the `typecheck` script and the canonical include. That closes the silent-skip hole `typecheck-tests-in-the-acceptance-gate` recorded but left open, which is more than this task required.

**(2) `VersionedStateView` forwards `listCurrent` / `listAsOf`**, exercised through the handle `process()` returns, with the handle's untyped character preserved.

**(3) The abitype work is the best part, because it doubted itself.** The task asked for a type-level pin that viem-derived and `core`-derived ABI types stay mutually assignable. That was implemented — and then *checked*: with the override removed and both copies installed, `pnpm typecheck` **still passed**, because abitype's types are structural and the two copies were structurally identical. So the type-level aliases catch the copies *diverging* (a later reshape of `Abi`, a `Register` augmentation merging into one copy only), not the second copy *existing*. A resolution-based case was added that compares the abitype version reached from `core` against the one reached from inside `viem`, which does fail on a duplicate, and it was proved red.

That is the failure mode I care most about in a guard task: a check that passes for the wrong reason and is trusted anyway. Verifying what the guard actually detects, finding it detects something else, and adding the one that works is exactly right. Scanning `node_modules/.pnpm` was rejected for good reasons too — it hard-codes pnpm's layout, and pnpm leaves unpruned directories behind after an override change (it did here), so the scan would report copies nothing resolves.

The override direction is also right: pinned **up** to 1.2.4 rather than down to viem's exact 1.2.3, because overrides are workspace-local and do not propagate, so pinning down would make the local tree resolve something *older* than a consumer installing `@etherfold/core` from npm gets — local would stop reflecting published, which is the opposite of what the check exists for. The conditional back-out branch did not fire; the gate is green including every viem-touching suite, and the pin can be dropped when viem itself moves.

Merging.

PR #16, opened 2026-08-24
feat(promote-stratagems-conformance-workload): Promote the spike's capt…

wighawag:
Task: `work/tasks/done/promote-stratagems-conformance-workload.md`

I reviewed the prior attempt's work on this branch, verified it end to end, and found it complete. No further code changes were needed.

## What is on the branch (reviewed, not re-done)

`git diff origin/main...work/task-promote-stratagems-conformance-workload` — 57 files:

- **`packages/conformance-workload-stratagems/`** (new, `private`): `src/entities.ts`, `processor.ts`, `project.ts`, `replay.ts`, `oracle.ts`, `fixtures.ts`, `workload.ts`, `stratagems-contract.ts`; `test/workload.test.ts` (fast), `test/alpha1.test.ts` (heavy + the named revert case), `test/oracle.test.ts`, `test/utils/backends.ts`; `run/regenerate-golden-state.ts`; `fixtures/` + `fixtures/README.md`; `vendor/stratagems/` with its GPL-3.0 note.
- **`docs/adr/0026-the-heavy-conformance-workload-is-a-private-package.md`**, a changeset for `@etherfold/state-store-conformance`, `CONTEXT.md` gaining a **conformance workload** glossary entry, and the spike README / finding / scripts re-pointed at the new home.

## Verification I ran

| check | result |
| --- | --- |
| `sha256` of all four fixtures vs `origin/main` | **identical** — they were moved, not regenerated (`63559834…`, `8edce428…`, `633985b3…`, `f2e7d89d…`) |
| `pnpm format:check` | green |
| `pnpm changeset status --since=main` | `@etherfold/state-store-conformance` patch |
| `pnpm build` / `pnpm typecheck` | green (incl. the new package) |
| `pnpm test` (fast tier) | green, 25s total; workload package 8 passed / 6 skipped |
| `CI=1 pnpm test` (heavy tier) | green, 30s total; workload package **14 passed**, incl. alpha1 on memory / sqlite / patch and the revert case |
| `pnpm --filter …stratagems test:full` | green |
| working tree | clean, no stray untracked files |

Acceptance criteria spot-checks: the port carries no CSV index, no `playerCount`, no order singleton (asserted as a regression guard in `workload.test.ts`, and children are keyed by `(blockNumber, logIndex)` arrival, fixed-width-padded so lexicographic order is numeric); the revert case is named `reverting to block 13,364,821 makes the evil owner's computedPoints DECREASE from 12 to 6`; the alpha1 stream is still gzipped with its in-file `provenance` (`omittedFields`, chain head, block range) plus `fixtures/README.md`; both deployments are labelled with the `base` vs `alpha1` trap called out first; the GPL-3.0 note sits in `vendor/stratagems/README.md`, the package `license` field and ADR-0026.

One small documentation nuance I deliberately left alone: the fixtures README's trap paragraph says the abandoned deployment "saw 45 logs", while its table (and the committed fixture) say 42 events over 9 blocks. Both numbers come from `work/notes/findings/sqlite-in-the-browser.md`, which uses 45 for the deployment and 42 for the captured fixture; the capture window starts at 11,681,917, which explains the gap. Not introduced by this task, and correcting the finding is out of scope.

The block below is unchanged from what the prior attempt already recorded in the done record; I re-emit it verbatim because it is still the accurate rationale for what is on the branch.

## Decisions

**Where the promoted artifacts live: a new private workspace package, `packages/conformance-workload-stratagems/`, with the fixtures and the vendored oracle MOVED (not copied) out of `docs/spikes/`.** `docs/spikes/` is an evidence store for a finished measurement; a conformance fixture is production test material that must not rot. A copy would have meant 1.3 MB duplicated and two goldens that can silently diverge. The spike's `src/port/` deliberately STAYS behind, because it is the exhibit the finding's contortion list describes, and the promoted port is a rewrite of it: the spike README, the finding and the spike's scripts all now point at the new home. Alternatives considered: inside `@etherfold/state-store-conformance` (refused: `processor-entities` closes a workspace dependency cycle, and the licence), or `examples/` (refused by the spec: a workload is not a teaching example). Touches: `@etherfold/state-store-conformance`'s README and doc comment (hence the changeset), and every spike script that read a fixture. Recorded as **ADR-0026**.

**The fast/full split.** Fast, every invocation (about 2s): the `stratagems-base` smoke replay on all FOUR backends, the model-shape regression guard, and the oracle re-deriving BOTH golden states. Full (`CI` set, or `test:full`, about 18s): the launched game's 31,332 events on memory, sqlite and patch, plus the named revert case. Opt-in only (`test:all-backends`): the launched game on IndexedDB, because on `fake-indexeddb` it costs about half an hour and degrades quadratically, where the same backend measured 45.6 ms/block on real Chromium. So the heavy IndexedDB coverage is a property of the SHIM, not of the backend, and the honest route to it is the real engine. Alternative considered and rejected: putting it in CI anyway (it would be switched off by the next person to wait for it). Touches: any CI workflow that wants the heavy tier must set `CI` (it does) or call `test:full`.

**The oracle now re-derives BOTH golden states on every invocation, not just the small one.** The inherited code justified oracle-on-small-fixture-only with "running it over 31,332 events would pay the oracle's runtime to learn nothing". I measured it: 1.2s, because the original processor is in-memory immer with no store underneath. That justification was false, so the expensive fixture's oracle would have been the one nobody re-ran. `run/regenerate-golden-state.ts` remains a script for the one thing that actually differs: it WRITES.

**`packages/conformance-workload-stratagems/package.json` declares `"license": "GPL-3.0-only"`, the only package here that does not say MIT.** ADR-0026 refuses to publish the derived port under MIT because that would misstate what it is; leaving the manifest at MIT restates exactly that misstatement in the file a reader and every licence scanner check first. It costs nothing (the package is `private`), and the odd field is one click from `vendor/stratagems/README.md`. Alternative considered: leave it MIT and rely on prose (rejected: the prose is what the manifest contradicts). Touches: nothing in the release path, since changesets and `pnpm publish` skip private packages.


wighawag:

| # | criterion | verdict |
| --- | --- | --- |
| 1 | Suite runs the real captured stream, state equals the golden after canonicalisation, on every backend | ✅ |
| 2 | `base` is the fast smoke case; full `alpha1` available and runs in CI | ✅ |
| 3 | Port rewritten onto the idiomatic model and STILL byte-identical | ✅ |
| 4 | Reorg on the real stream, revert to 13,364,821, `computedPoints` 12 → 6, named | ✅ on three backends |
| 5 | Fixture stays gzipped, provenance travels | ✅ 634,488 bytes |
| 6 | Both deployments labelled so the `base`/`alpha1` trap cannot repeat | ✅ |
| 7 | GPL-3.0 vs MIT note travels with the vendored oracle | ✅ and see below |
| 8 | Artifact home decided, spike folder coherent | ✅ ADR-0026 |
| 9 | Tests, changeset | ✅ |

### The BLOCK condition did not fire

Both golden states are byte-identical to the pre-move baselines I recorded before the first attempt: `alpha1.state.json` `sha256:63559834774f0f53`, `base.state.json` `sha256:633985b3d27ec360`. Matched, not regenerated.

I also ran the heavy tier the fast loop skips (`STRATAGEMS_WORKLOAD=full`): **14/14**, including `lands on the state the original JSProcessor computed from the same stream` on memory, sqlite and patch, and the named `reverting to block 13,364,821 makes the evil owner's computedPoints DECREASE from 12 to 6` on all three. The rewrite is real, pinned by `declares no stored array, no CSV index, no count and no remembered order` and `keys its ordered children by arrival, fixed-width, so the id order is the numeric one`.

### Three decisions better than the task asked for

**The oracle now re-derives BOTH golden states on every invocation.** The inherited code justified oracle-on-the-small-fixture-only with "running it over 31,332 events would pay the oracle's runtime to learn nothing". That was *measured* rather than accepted: 1.2s, because the original processor is in-memory immer with no store under it. The justification was false, and the consequence of believing it was that the expensive fixture's oracle would have been the one nobody re-ran. This materially strengthens the guarantee: the golden state is continuously re-derived from the vendored original `JSProcessor`, so a regression in the port *and* a corrupted golden file are both caught, rather than the port being compared to a file that could rot.

**Artifacts MOVED, not copied**, with the right reason: a copy is 1.3 MB duplicated and two goldens that can silently diverge. The spike's `src/port/` deliberately stays behind, because it is the exhibit the finding's contortion list describes and the promoted port is a rewrite of it. The spike README, the finding and the spike scripts were all repointed (29,380 deletions, 44 insertions).

**IndexedDB's heavy tier is opt-in, and the reason is a real finding rather than a shrug.** On `fake-indexeddb` the launched-game replay costs about half an hour and degrades quadratically, where the same backend measured 45.6 ms/block on real Chromium. So that cost is a property of the SHIM, not of the backend, and the honest route to heavy IndexedDB coverage is the real engine. Putting it in CI anyway was considered and rejected because "it would be switched off by the next person to wait for it", which is correct about how such tests actually die.

### Licence handling

`package.json` declares `"license": "GPL-3.0-only"` — the only package here that is not MIT — because the port derives from GPL stratagems, and leaving the manifest at MIT would restate the misstatement in the first file a reader or a licence scanner checks. The package is `private`, so changesets and `pnpm publish` skip it and the release path is untouched.

### Note on how this landed

The build itself succeeded on the previous attempt; the gate bounced at `prepare` on `ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION` for a two-hour-old `dorfl@0.13.1`, which was an environment fault of mine, not a defect in this work. That is fixed on main by pinning pnpm to 10.28.1, and this run recovered and integrated the kept commit rather than rebuilding.

Merging.

PR #17, opened 2026-08-24
feat(backend-neutral-entity-event-processor): An EventProcessor that ru…

wighawag:
Task: `work/tasks/done/backend-neutral-entity-event-processor.md`

Done. The gate is green: `pnpm build && pnpm typecheck && pnpm test` all pass, and `prettier --check` is clean.

## What I built

**The seam gained a cursor port** (`packages/state-store/src/cursor.ts`, `store.ts`): `readCursor` / `writeCursor` / `clearCursor` over an opaque string under a caller-chosen key, plus an optional third argument to `applyBlock` that writes the cursor **in the block's own transaction**. Implemented on all four backends: SQLite (a new fixed `_cursor (key, value)` table, the upsert appended to the block's `batch([...])`), IndexedDB (a new `cursors` object store inside the block's transaction; package schema version 1 → 2, additive and `contains`-guarded), the patch store and `MemoryStateStore` (a map, written after the last thing that can refuse). `@etherfold/state-store` still declares zero dependencies and `no-platform-leakage.test.ts` passes unchanged.

**`EntityEventProcessor`** (`@etherfold/processor-entities`) is the shell: the store is injected, and it owns `load` / `process` / `reset` / `clear` / `prune`, the version hash, the code fingerprint and the retention reconciliation. `EntityStateView` is its read handle: the seam's four reads plus the capability report, with no `queryCurrent` / `queryAsOf` on the type at all.

**`VersionedStateEventProcessor` is now a thin SQLite flavour of it** — build the store from a `RemoteSQL`, substitute `VersionedStateView`, delegate everything else. Revert-then-apply exists once.

**The crash window is closed, and per block rather than per `process` call.** One `process` carries many blocks and each is its own transaction, so writing the stream's cursor with the last block alone would have left the same window open for the whole run. Each block is applied with the cursor describing *it* (`syncedThrough` truncates `lastToBlock` and the unconfirmed window; the observed chain tip is not truncated), and the last carries the stream's own. A stream with no blocks moves the cursor through `writeCursor`, where a crash costs a re-scan and not a wedge. `work/notes/observations/sync-cursor-write-is-not-atomic-with-the-block-it-describes.md` is deleted.

## Acceptance criteria

All ten met. New tests: `state-store-conformance/src/cases/sync-cursor.ts` (10 cases, run by all four backends; the suite's own `the-suite-catches.test.ts` gained a `PrematureCursorStore` so the "never ahead" case is proven not to be decoration), `processor-entities/test/entity-event-processor.test.ts` (one processor, four backends, same state; cursor round-trip through a genuine reopen per backend; reorg with a counter that goes back DOWN on all four), `processor-entities/test/read-handle-tier.test.ts` (`@ts-expect-error` guards, so `pnpm typecheck` goes red the day a SQL method appears on the neutral handle, plus a runtime absence check), `processor-sqlite/test/cursor-atomicity.test.ts` (a real libSQL database that dies after the Nth block batch; asserts store tip and cursor are the same number at every crash point, and that the next run resumes and finishes), and `conformance-workload-stratagems/test/through-the-processor.test.ts`.

**Criterion 9 (the stratagems workload):** it drives through the component fine and it was not a big change. I did *not* replace `replayIntoStore` with it, because that function keeps the id ledger the golden-state projection needs and the seam has no unanchored "list everything" read by design (ADR-0021). The new test runs the fixture both ways on every backend and asserts they land on the same state, and that it is still the golden one.

**On the two forward constraints (snapshot bootstrap):**
- *Contents and cursor settable together as one unit* — **supported today**. `applyBlock(snapshotBlock, rows, cursor)` is exactly that, atomically, on every backend, and there is a conformance case for it. One caveat for the bootstrap task: a snapshot bigger than a hosted backend's per-request bound is sent as one oversized batch with a warning (a block is one atomic unit), and `applyBlocks` does not yet take a cursor, so a multi-block install would need that.
- *Reporting a retention floor it did not compute* — **needs more.** Retention today is entirely tip-relative (`RetentionSetting` is `'revert-only' | 'unbounded' | {blocks: N}`, and `retentionFloor` is `tip - N`). A store holding only a snapshot's current rows needs an **absolute** floor at the snapshot block, which is a new retention shape plus its capability report plus its conformance cases. Nothing here forecloses it; nothing here provides it.

Docs: ADR-0027, `CONTEXT.md` glossary (three entries), five package READMEs, and two changesets.

*(Minor note, not drift: the item was at `work/tasks/backlog/` and its spec at `work/specs/tasked/`, not the `ready/` paths in my brief. Content matched, so I proceeded.)*

## Decisions

**The sync cursor's home is the storage seam, as an opaque string, written with its block.** Chosen (the task decided it; building it confirmed it). Alternatives considered and rejected: a new narrow keeper interface the deployment wires (a second thing to misconfigure, and *no atomicity*, which is the whole point); a reserved entity through the existing seam (the cursor would become versioned, revertible and prunable — destroyed by exactly the operations it must survive); making replay idempotent instead of atomic (weakens a check that catches a real caller bug). Touches: every `StateStore` implementation now and forever, `@etherfold/processor-sqlite`'s public surface, and `bootstrap-an-entity-store-from-a-snapshot`. Recorded as **ADR-0027**.

**The cursor is written per BLOCK, not per `process` call.** The task said "the same transaction as the block it describes", which literally reads either way. Writing only with the last block would close the narrow window the observation described and leave the wider one open: a crash after block 3 of 5 still puts state ahead of the cursor and still wedges. So `applyEventStream` derives an intermediate cursor per block (`syncedThrough`: truncate `lastToBlock` and the unconfirmed window to that block; leave `latestBlock` and `context` alone) and gives the last block the stream's own `lastSync`, which is safe because the stream covered every block up to `lastToBlock`. Alternative rejected: making the whole stream plus cursor one transaction, which needs a batched seam verb and would force an unbounded backfill into a single request. Cost: one extra serialization per block, small in practice because `unconfirmedBlocks` is empty during backfill and streams near the tip are short. Touches: `applyEventStream`'s signature (the cursor argument is optional, so existing callers are unaffected) and any future processor over the seam.

**COHERENCE: the port's argument is a `key`, not a "context key".** The task's wording was "under a context key", but `context` already means `ContextIdentifier` in this repo (source/config/processor hashes, validated by the core). Keying the cursor by *that* would break the core's discard path, which only runs when `load` returns something — a context-keyed cursor answers "nothing stored" after a processor upgrade and the new run indexes on top of the old processor's rows. So the seam's key names *which* cursor, the neutral processor writes one fixed `SYNC_CURSOR_KEY = 'lastSync'`, and the context travels inside the value. Both `cursor.ts` files say so explicitly, and the glossary carries the distinction. Touches: anyone later tempted to make the cursor multi-tenant.

**SQLite's cursor table is `_cursor (key, value)`, not the old `_sync (id, lastSync)`.** A storage primitive should not have a column called `lastSync`. Keeping the old shape would have made existing databases resume, which I weighed and dropped: nothing is published under `@etherfold` yet (all packages are `0.0.0`, and `publish-etherfold-and-deprecate-old-names` is still in the backlog), and the spec puts migrating existing processors out of scope. An existing database reads as never-synced and re-indexes. Touches: `@etherfold/processor-sqlite`'s public surface — `SYNC_TABLE`, `SYNC_ROW_ID`, `SYNC_SCHEMA_DDL`, `readLastSync`, `writeLastSyncStatement` and `deleteLastSyncStatement` are **removed** (the changeset names them); `serializeLastSync` / `deserializeLastSync` survive as re-exports. Two assertions in `processor-sqlite`'s tests were retargeted from `_sync` to `_cursor`, meaning unchanged.

**`assertProcessorVersion` is called in both the wrapper and the inner processor.** Not redundancy: the message names the class the caller actually constructed, and a deployment told to fix `EntityEventProcessor` would be hunting a class it never wrote. Touches: `processor-sqlite/test/version.test.ts`, which asserts exactly that.

**The cursor is not reverted and not pruned.** `revertTo` leaves it alone on every backend — how far the *caller* got is not entity state, and the caller moves it when it applies the canonical branch, which arrives in the same stream as the retraction. Asserted in the conformance suite so a backend cannot quietly decide otherwise.

**IndexedDB's schema version moved 1 → 2.** `keys.ts` previously said the schema "stays version 1 forever". That claim was about *declarations* (a processor gaining an entity is still not a migration and still cannot be blocked by another tab), and I rewrote the comment to say so rather than deleting it. The upgrade is additive and `contains`-guarded. Touches: any existing browser database, which gains `cursors` and keeps every row.


wighawag:
**APPROVE ✅** — Gate-3 (conductor's diff-vs-criteria review). Gate-2 was off for this drive, so this is the only review layer above the acceptance gate; I read the diff rather than the report.

## The four constraints I was asked to check hardest

**1. The cursor is an opaque string, and the seam still declares nothing.** ✅ `CursorWrite` is `{key: string, value: string}` and `readCursor`/`writeCursor`/`clearCursor` are all `string`. `packages/state-store/package.json` still has `"dependencies": {}` on this branch, `state-store-sqlite` still imports only `@etherfold/state-store` + `named-logs` + `remote-sql`, and `no-platform-leakage.test.ts` is **byte-unchanged** (`git diff --stat` on it is empty) rather than adjusted to keep passing. ADR-0016's direction is intact: the codec (`serializeLastSync`/`deserializeLastSync`) sits in `processor-entities`, and the store genuinely does not know what a `LastSync` is.

**2. The cursor is written in the block's own transaction, and the observation is gone.** ✅ SQLite appends the cursor statement to the same `db.batch([...])` as the block; IndexedDB adds `CURSORS` to the block's `tx` scope and `put`s inside it. Memory and patch have no transaction and say so explicitly at the call site, writing the cursor **last, after every point that can still refuse** — which is the "must say so rather than quietly leave the gap" the task demanded, not a silent gap. `work/notes/observations/sync-cursor-write-is-not-atomic-with-the-block-it-describes.md` is deleted; the only surviving mentions are historical prose in ADR-0027, the crash test's docblock and the done task body, which is correct.

The conformance suite asserts the invariant two ways ("a refused block leaves it where it was", "a re-applied height leaves it where it was") on all four backends, and `the-suite-catches.test.ts` gained a `PrematureCursorStore` that writes the cursor first — so the new group is **proven to catch** the drift rather than being decoration. That is the check I most expected to find missing.

**3. The neutral read handle has no SQL tier.** ✅ `queryCurrent`/`queryAsOf` are absent from `EntityStateView` entirely — no stub, no throw. `read-handle-tier.test.ts` guards them with `@ts-expect-error`, and `tsconfig.typecheck.json` includes `**/*.ts`, so `pnpm typecheck` (in the gate) really does go red the day a SQL method appears. The test also exercises the four seam reads first, so it cannot pass vacuously against an empty handle, and a second case proves the methods are absent at run time rather than typed away.

**4. Revert-then-apply exists once.** ✅ `VersionedStateEventProcessor` is a genuine delegate: it holds an `inner: EntityEventProcessor` and forwards `load`/`process`/`reset`/`clear`/`prune`/`getVersionHash`/`getCodeFingerprint`. It contains no `applyEventStream` call and no `revertTo` of its own. The only thing it keeps is store construction from `RemoteSQL` and substituting `VersionedStateView` for the SQL tier, which is exactly the "thin SQLite flavour" the task asked for.

## The rest of the criteria

Same processor on four backends with equal state, cursor round-trip through a **genuine** reopen (a new store over the same libSQL handle / the same IndexedDB database name after a real `close()`; memory and patch reuse the instance and the file says plainly why that is the honest half of the claim for a `memory-only` backend), reorg with a counter going back down on all four, a real libSQL crash test that kills the database after the Nth batch and asserts tip and cursor agree at every crash point, the stratagems workload driven through the component, and two changesets. Criterion 9's honest "I did not replace `replayIntoStore`, because it keeps the id ledger the golden projection needs and the seam has no unanchored list read by design" is the right call, not a dodge.

## Judgement calls I checked rather than took on trust

**Per-block cursor instead of per-`process`.** The task's wording admits either reading. Writing only with the last block would have closed the narrow window and left the wide one open — crash after block 3 of 5 and the cursor is still pre-stream, which is the same wedge. `syncedThrough` truncates `lastToBlock` and cuts `unconfirmedBlocks` to the prefix at or below the block, and deliberately does **not** lower `latestBlock` (an observed tip, not progress) or touch `context`. That is correct: lowering `latestBlock` would widen the re-fetch window, and an unconfirmed block above the resume point would hide the blocks between them.

**The key is a name, not a `ContextIdentifier`.** Rightly caught. A context-keyed cursor would answer "nothing stored" after a processor upgrade, `load` would return `undefined`, the core's discard branch (which only runs inside `if (loaded)`) would never fire, and the new run would index on top of the old processor's rows. Keying it would have been the plausible-looking wrong answer.

**Cursor not reverted and not pruned**, asserted in the suite so a backend cannot quietly decide otherwise. Consistent with rejecting the reserved-entity design.

## Carried forward, not blocking

- `applyBlocks` does not take a cursor, so a multi-block snapshot install would need it. Self-reported, and it lands on `bootstrap-an-entity-store-from-a-snapshot`.
- **Retention floor from a snapshot is explicitly NOT provided** — `RetentionSetting` is still entirely tip-relative. The report says so plainly instead of implying the forward constraint was satisfied. That is the hard criterion of task 3 and I will hold it there.
- `_sync (id, lastSync)` → `_cursor (key, value)` makes an existing database read as never-synced and re-index. Documented, weighed, and legitimate while nothing is published under `@etherfold` and migration is out of the spec's scope.

Merging.

PR #18, opened 2026-08-24
feat(index-in-the-browser-with-a-chosen-backend): Index in a browser ta…

wighawag:
Task: `work/tasks/done/index-in-the-browser-with-a-chosen-backend.md`

Done. Everything is green: `pnpm format:check`, `pnpm build`, `pnpm typecheck`, `pnpm test` (full monorepo), `pnpm build:examples`, plus the browser run (`pnpm --filter @etherfold/browser test:browser`, 12 tests × Chromium/Firefox/WebKit), plus the example driven against Ethereum mainnet for real.

## What I did

**The join (`packages/browser/src/IndexerState.ts`).** `createIndexerState` now takes either processor kind, discriminated by a tag the caller writes (`ProcessorKind = 'js-object' | 'entities'`, `TaggedProcessor`, `IndexerStateProcessor`). A bare `EventProcessorWithInitialState` still means `'js-object'`, so every shipped call site is untouched. The free-form path creates its initial state; the entity path reads its store through `processor.state` (the handle that already exists). A `keepState` on the entity path is refused with a message naming the store. `updateProcessor` takes the same union; `options.createIndexer` now receives the processor as `EventProcessor` (the only widening, called out in the changeset). The doc comment on the hook says which path persists what, and why the two mechanisms are different rather than redundant.

**Tests.** `test/entityIndexing.test.ts` (expected state on the default; the same processor on IndexedDB / patch / memory; reorg with a decreasing counter on two backends; reload continuity by *fetched ranges*, since both outcomes land on the same rows; patch-store start-over plus its `durability: 'memory-only'` claim). `test/processorKinds.test.ts` pins the type errors with `@ts-expect-error` (typecheck runs those). `test/bundlesForABrowser.test.ts` esbuilds the package with `platform: 'browser'`. `browser/` holds the shared workload plus playwright-browser-harness specs, including a real page reload; `browser/workload.ts` is the *same object* both runners drive.

**The example (`examples/event-processor-nfts`).** `src/entities.ts` is the entity-path processor; `browser/` is an app that indexes BAYC off mainnet with no server. `pnpm --filter event-processor-nfts browser`. Verified end to end in a headless Chromium against the live chain: it indexes, and on reload prints `resumed: the first fetch of this page load asked for block 25825393, not 25823410`. Switching the one line to `PatchStateStore` was verified too, and the page then reports `survives a reload: NO (memory-only, ADR-0023)` and re-indexes. README + a root-README pointer.

**One blocker fixed on the way:** `@etherfold/browser` could not be bundled for a browser at all (`@etherfold/utils`' barrel re-exports the CLI modules, whose top-level `node:fs`/`node:path`/`node:module` imports fail to resolve before tree-shaking helps). See the decision below.

Changeset: `.changeset/browser-index-with-an-entity-processor.md` (`@etherfold/browser` minor, `@etherfold/utils` minor). Glossary entry added to `CONTEXT.md`.

## Decisions

**The two processor kinds are discriminated by an explicit TAG the caller writes (`{kind: 'entities', processor}`), not by structure.** Chosen because both kinds are `EventProcessor`s and the only structural difference is `createInitialState` vs `state` — a sniff a wrapper, proxy or decorator can make wrong in silence, and the wrong branch does not throw, it seeds the UI store with the wrong value and indexes on. Alternatives considered: (a) a second factory `createEntityIndexerState` (the task asked for one hook that accepts either); (b) a structural union with no tag (TS would resolve it, but a runtime `'createInitialState' in p` check would still be a guess). The union is deliberately SYMMETRIC — `{kind: 'js-object', processor}` exists even though nobody has to write it — so the language has two named kinds rather than one tag plus an untagged "other". The bare processor stays the shorthand for `'js-object'`, so nothing published breaks. **Touches:** `@etherfold/browser`'s public surface, `updateProcessor`, `options.createIndexer` (widened to `EventProcessor`), and any future host that wraps the hook (e.g. `examples/web-demo`'s `createIndexeInitializer`, which needs no change). Recorded in `CONTEXT.md` as **processor kind**. No ADR: it is a naming/shape choice covered by the glossary and reversible by adding an overload, not a hard-to-reverse trade-off.

**`keepState` on the entity path is a runtime REFUSAL, not a type error.** Making it a compile error needs the options type to depend on the tag, which means overloads with an explicit return type for a hook whose return type is a large inferred object literal including `withHooks(){...this}`. The task also said the `keepState` optionality "is already the right shape and needs no change". So it throws, with a message naming `createBrowserStateStore` and the store. **Touches:** anyone passing a keeper; the existing free-form refusal (`this processor do not support "keepState" config`) is unchanged in wording and meaning.

**`@etherfold/utils` gained a `./indexer` subpath, and `@etherfold/browser` imports `contextFilenames` from it.** This is a fix to a package the task did not name, and I made it because the required deliverable is impossible without it: `@etherfold/utils`' barrel re-exports `contracts.ts` and `processorSetup.ts`, whose top-level `node:fs` / `node:path` / `node:module` imports mean `import '@etherfold/browser'` does not resolve for esbuild or vite — before tree-shaking can drop them. Alternatives rejected: copying `contextFilenames` into the browser package (it is the naming scheme a published snapshot is fetched under, so the CLI and a browser client MUST agree — a copy is a drift waiting to happen); making the node imports lazy (turns synchronous functions async). The barrel is unchanged, so nothing existing moves. **Touches:** `@etherfold/utils`' published export map (additive, in the changeset), and `bootstrap-an-entity-store-from-a-snapshot`, which will read the same filenames. Pinned by `test/bundlesForABrowser.test.ts` so it cannot regress silently.

**The example keeps BOTH processors, and the browser demo runs the entity one.** `examples/event-processor-nfts/src/index.ts` (free-form `JSProcessor`) stays because `examples/web-demo` imports `createProcessor` from it and the free-form path is not deprecated; `src/entities.ts` is the same question written against the seam. Alternative rejected: converting the example wholesale, which would break `web-demo` (out of scope). The README frames the duplication as the one place where the two authoring styles sit side by side, so the porting cost is readable in a diff. **Touches:** `examples/web-demo` (unaffected — it still builds).

**The example's start script is `browser`, not `dev`.** In this repo `dev` means "watch the sources and rebuild the package" in every package and example, and the root `start:pnpm` is `pnpm -r dev`; giving `dev` a second meaning in one example would have silently added a Vite server to that command. Documented command is `pnpm --filter event-processor-nfts browser`. **Touches:** the root `start` / `start:pnpm` scripts (deliberately not).

**The example defaults to a public RPC (`https://rpc.mevblocker.io`) and BAYC on mainnet, with `?rpc=` / `?contract=` / `?blocks=` overrides, and uses an injected wallet when there is one and no `?rpc=`.** "Against a real chain" needs a real endpoint, and a demo that requires a wallet is not one command. I verified this endpoint serves `eth_getLogs` with `blockTimestamp` (which the entity path needs to record a block) and sends CORS headers; several other free endpoints refuse historical `eth_getLogs` outright. The README says public endpoints rate-limit and come and go, and points at the override. **Touches:** nothing in the packages; it is an example default a reader can change in the URL bar.

**The example remembers its start block in `localStorage`.** `startBlock` is part of the indexing source and the source is hashed into the sync context, so a start block recomputed as `tip - 2000` on every load would make each reload a different deployment, discard the state as stale, and make the reload-continuity claim untestable by construction. Alternative rejected: a hard-coded start block (goes stale, and the demo's run length would grow without bound).

**The `resume` line in the example reports the first `eth_getLogs` range, not the state.** Resuming and re-indexing end on the same rows, so the state cannot say which happened; the fetched range can. The node and browser tests assert the same property the same way.


wighawag:

**I did not rely on the gate for the parts that matter here.** The gate's `verify` filters to `./packages/*` and `./platforms/*`, so `pnpm build`, `pnpm typecheck` and `pnpm test` never touch `examples/`, and the playwright run is deliberately outside it. The runnable example — the hard criterion of this task — is therefore entirely ungated. I built and ran it myself in a throwaway clone.

## The runnable example, verified live rather than read

`pnpm install --frozen-lockfile` (pnpm 10.28.1) is clean, so the lockfile change is consistent. `pnpm build` and `pnpm build:examples` both pass, and `web-demo` still builds, confirming the free-form example it consumes was not disturbed.

`pnpm --filter event-processor-nfts browser:build` bundles to **88.4 KB (28.9 KB gzipped)**. That is the real proof of the `@etherfold/utils/indexer` fix: before it, `import '@etherfold/browser'` could not resolve for a bundler at all.

Then I served `dist/browser` and drove it in a real headless Chromium against Ethereum mainnet:

```
FIRST LOAD
  what      0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d on chain 1, from block 25823548
  where     retention: unbounded, as-of reads: yes, survives a reload: yes
  progress  block 25825552 / 25825552 (100%)
  transfers 12
  resume    indexed from the start block (25823548): no cursor was found

RELOAD
  progress  block 25825554 / 25825554 (100%)
  resume    resumed: the first fetch of this page load asked for block 25825535, not 25823548
```

So it genuinely indexes BAYC in a tab with no server, and the reload genuinely resumes from the cursor rather than re-indexing. This is a real example, not a test wearing an example's clothes, and the task's "route to needs-attention rather than quietly downgrade this" escape hatch was not needed or taken.

I also ran the ungated browser suite (`playwright test --project=chromium`): 4/4 pass, including **`resumes from its cursor after a real page reload`** and **`reverts a reorg through the browser path, and the counter decreases`**.

## The constraints I was told to check hardest

**Discrimination is by declared type, not by field-sniffing.** ✅ `ProcessorKind = 'js-object' | 'entities'` with a `TaggedProcessor` union, and the only runtime read is `'kind' in given` — the discriminant the caller wrote. Nothing anywhere asks whether `createInitialState` or `state` is present, which was the specific failure mode to avoid: both kinds are `EventProcessor`s, so a structural guess is one a proxy or decorator makes wrong in silence, and the wrong branch does not throw, it seeds the UI with the wrong value and indexes on. `processorKinds.test.ts` pins four separate type errors with `@ts-expect-error` (entity processor bare, entity processor under the wrong tag, untagged object, unknown tag), and typecheck covers tests.

**The published `EventProcessorWithInitialState` path still works.** ✅ A bare processor still means `'js-object'`, so no shipped call site changes. `git diff --stat` on `packages/browser/test/` shows **three new files and zero modifications** — the existing tests pass unchanged in the strongest sense, having not been touched. The only widening is `options.createIndexer` receiving `EventProcessor`, called out in the changeset.

**The two persistence models are not unified, and the invariant is what is held.** ✅ The entity path stores its state once, in the store, and passes no keeper; the free-form path keeps its single-keyed `{state, lastSync}` write, which is its atomicity mechanism and is untouched. Neither was bent to fit the other. The hook's docblock states plainly which path persists what and why the mechanisms differ, which is what the task asked for so nobody unifies them wrongly later. A `keepState` on the entity path is refused with a message naming `createBrowserStateStore` rather than being silently ignored.

**`keepStateOnIndexedDB` and `createBrowserStateStore` cannot be confused at the type level.** ✅ Pinned directly, not merely documented: `@ts-expect-error a StateStore is not a KeepState: it has no save, and its clear is not that clear` and `@ts-expect-error a KeepState keeper is not a StateStore factory`.

**Backend chosen by configuration, processor untouched.** ✅ The patch-store switch really is one line, sitting commented immediately below the default in `browser/main.ts`, with the README showing the capability report changing to `survives a reload: NO (memory-only, ADR-0023)`. The memory-only start-over is made explicit through `store.capabilities` at startup rather than being discovered from a blank tab.

## Scope judgement I checked rather than waved through

`@etherfold/utils` gained an `./indexer` subpath. That is a package the task did not name, so I checked it is a genuine blocker and not creep: the barrel re-exports CLI modules whose top-level `node:fs`/`node:path`/`node:module` imports fail to resolve before tree-shaking, which made the required deliverable impossible. The fix is **purely additive** (`src/index.ts` is byte-unchanged; the export map only gains an entry), it is in the changeset, and it is pinned by `bundlesForABrowser.test.ts` so it cannot regress silently. Copying `contextFilenames` instead would have forked the naming scheme a published snapshot is fetched under — exactly the drift `bootstrap-an-entity-store-from-a-snapshot` would then trip over.

The `browser` script name (not `dev`, which means "watch and rebuild" everywhere else here and is what `pnpm -r dev` relies on) and the `localStorage` start block (the source is hashed into the sync context, so a tip-relative start block would make every reload a different deployment and render the continuity claim untestable) are both correct calls with the reasoning recorded.

Merging.

PR #19, opened 2026-08-24
feat(bootstrap-an-entity-store-from-a-snapshot): Bootstrap an entity st…

wighawag:
Task: `work/tasks/done/bootstrap-an-entity-store-from-a-snapshot.md`

Everything is green. Here is what I built.

## What landed

**`@etherfold/state-store` — the seam owns the honesty** (`src/snapshot.ts`)
- `StateSnapshot` = `{format, processor, savedAt, takenAt, cursor, rows}`, shaped deliberately like the CLI's free-form envelope; `SnapshotHead` is the same minus `rows` (the cheap file a client compares mirrors on).
- `openSnapshotAware(store)` → `SnapshotAwareStateStore`, a thin decorator over ANY backend. `.bootstrap(snapshot, {processor})` installs rows + cursor as ONE `applyBlock`, after recording the **snapshot origin** under a second cursor-port key so a reload is as honest as the first run.
- A bootstrapped store reports a **window floored at the snapshot's block** (intersected with what the deployment configured), refuses as-of reads below it with `BlockNotRetainedError`, and refuses a revert reaching under it with `RevertBeyondSnapshotError`, changing nothing. New errors: `SnapshotProcessorMismatchError` (names both versions), `SnapshotFormatError`.

**`@etherfold/processor-entities` — the client side** (`src/snapshot.ts`)
- `bootstrapFromSnapshot` / `openAndBootstrap`: mirrors, most-advanced-wins, prefer-local, fail over on error; declines a snapshot taken inside the reorg window when given a `finalityDepth`; `createSnapshot` is the minimal producer and says so.

**`@etherfold/state-store-conformance`** — a `bootstrapping from a snapshot` group, claim-selected, so every backend inherits the trap's answer (runs green on memory, sqlite, indexeddb, patch).

**Also**: ADR-0028, a `## Conventions`-required changeset, CONTEXT.md glossary entry for **bootstrap**, README sections in the three packages, a doc section in `createBrowserStateStore`, and the measurement + script at `docs/spikes/bootstrap-an-entity-store-from-a-snapshot/`.

Tests: 17 (state-store), 20 (processor-entities, incl. all four backends' reload), 3 (browser, through the hook with a fake chain), 12 per backend in conformance. `pnpm format:check && pnpm build && pnpm typecheck && pnpm test` all green. `pnpm changeset status --since=main` still reports "no changesets found" only because my changeset file is untracked (changesets uses `git diff` against the merge base); it resolves the moment the runner commits.

## Decisions

**A snapshot contains CURRENT ROWS ONLY.** Measured, not assumed (`docs/spikes/bootstrap-an-entity-store-from-a-snapshot/`, the launched stratagems workload): a current-rows snapshot is **45.6 KB gzipped = 7% of the 619.6 KB gzipped stream**, while every version in the same encoding is **304.6 KB = 49% of the stream** and 6.7x the current-rows one. A client willing to pay 305 KB is most of the way to paying 620 KB for the stream, which it can replay into a store that then holds history it verified itself. Decisively, a full-history snapshot is not expressible through the seam at all: a version is what applying a block PRODUCES, and a write surface that could set `_lower`/`_upper` could manufacture states no sequence of blocks could reach. Alternatives considered: rows plus a retained window (rejected on the numbers and on the new write surface it would need); rows plus the tail of the stream (rejected as a second format for something `replayStream` already does). Touches: what a consumer's capability report inherits, and therefore the publishing spec. Recorded as **ADR-0028**.

**The floor is expressed as a WINDOW, not a new `Retention` kind.** A bootstrapped store's retained range is `[snapshotBlock, tip]`, i.e. a window of `tip - snapshotBlock` blocks, and intersecting it with a configured window is `min` of two distances — so `retentionFloor`, `retainedRange`, `assertRetained` and the conformance suite's `claimedDepth` all keep working untouched. Alternative considered and rejected: a fourth kind (`{kind: 'from-block', from: N}`), which is exact but re-means "retention" (what a deployment SETS) to also cover "where history begins", and would force every backend, the patch store and the suite to learn it. **Cost, recorded because a reviewer should see it:** a window is a distance from a tip and the seam has no verb reporting one, so the handle tracks the highest block it has seen. Between a reload and the first `applyBlock` of that session it therefore *under*-claims (never over-claims) and corrects itself immediately. Touches: `StateStoreCapabilities` consumers.

**Reorg below the floor: refuse loudly AND decline the snapshot that invites it.** The task allowed either. Refusal is the consumer-side half that cannot be delegated (there are no versions under the snapshot to reopen at any price, and a partly-undone reorg is a plausible wrong state — the same reasoning `RevertBeyondPatchHistoryError` records). Prevention is the producer's job and unverifiable after the fact, so `bootstrapFromSnapshot` does the half it can: given `finalityDepth`, a candidate taken within that many blocks of the tip its producer had observed is not adopted. `RevertBeyondSnapshotError` is deliberately NOT a `BlockUnavailableError` (that family is the read path). Touches: `EntityEventProcessor.process` now has a reachable throw on a deep reorg; a host's answer is to re-bootstrap.

**COHERENCE: the snapshot origin lives at the CURSOR PORT, under a second key.** The port is documented as "how far a processor has got", so a second key holding something else is a mild widening. The port is really a keyed opaque-string slot the store never interprets, never versions, never reverts and never prunes — exactly the durability a history floor needs — and the conformance suite already asserts two keys are kept apart. The alternative (a new `readMeta`/`writeMeta` port) would be a second slot with identical semantics that every backend implements twice. Touches: every `StateStore` implementation (no code change; a reserved key name), and anyone later tempted to key cursors by context.

**COHERENCE: "snapshot" now names two shapes, one meaning.** The free-form path's snapshot is the whole state blob in the CLI file envelope; this is its entity-path counterpart. I reused the word rather than inventing one (CONTEXT.md already says etherfold "hydrat[es] clients from pre-computed snapshots"), and the glossary entry now states both shapes and what only the entity one owes: a history floor.

**Install ORDER is marker-first, and it is not arbitrary.** The seam carries one cursor slot per `applyBlock` and the sync cursor has it, so the origin marker is a separate write. Marker-first leaves a crash with a floor over an empty store (over-cautious, self-healing on re-run); marker-last leaves rows with no floor, which is the exact failure this task exists to prevent. Same principle as the memory store writing its cursor last: where atomicity is unavailable, the order must be the safe one.

**Two deliberate departures from the free-form `fetch` behaviour, both strengthenings.** (1) Failover walks every remaining candidate in descending order, where `keepStateOnIndexedDB` tries the winner and exactly one more (its source says `// TODO more than 2`). (2) A snapshot from another processor version is not a candidate at all; the free-form keeper does not check. Also **not carried over**: the `{prefix}` location form, which builds filenames from a `ProcessorContext` via `@etherfold/utils/indexer`. The entity envelope carries `takenAt` and `processor` at the top level, so selection needs no filename convention; the object form is `{url, head?}` instead, where `head` is the same envelope without `rows`. A published mirror layout is the publishing spec's decision, and inventing one here would pre-empt it.

**The mirror logic lives in `@etherfold/processor-entities`, not `@etherfold/browser`.** "Is local already ahead" means reading `lastToBlock` out of a stored cursor, and the cursor is an opaque string at the seam (ADR-0027) whose codec is the entity runtime's. `@etherfold/browser` deliberately types the entity path structurally so it depends on no processor package, and putting this there would invert that. Browser gains documentation only (patch changeset). Touches: where a browser author looks — hence the pointer added to `createBrowserStateStore`'s docs.

**`openSnapshotAware` migrates the store itself.** It must read the marker, and a store that has not been migrated has nothing to read from (SQLite raises `no such table: _cursor`). Making it a migrating call means it REPLACES the `migrate()` a host would otherwise call rather than adding a step whose omission throws. Alternative rejected: documenting "migrate first", which is a rule that gets broken once and fails at the boundary furthest from where it was broken.

**`revertTo(-1)` is a wipe, not a reorg, and drops the floor.** `EntityEventProcessor.reset()`/`clear()` is that call and had to keep working; a store with no rows has no snapshot-derived history to be honest about, so it goes back to claiming what it was configured to keep. `0 <= keepUpTo < origin` is the refused case.


wighawag:

## The failure mode this task exists to prevent

**A bootstrapped store never reports `unbounded`.** ✅ And the test proves it by contrast rather than in isolation, which is the right shape:

```ts
const fresh = new MemoryStateStore([TOKEN, ACCOUNT]);
expect(fresh.capabilities.retention).toEqual({kind: 'unbounded'});   // it WOULD have lied
const store = await bootstrapped(fresh);
expect(store.capabilities.retention).toEqual({kind: 'window', blocks: 0});
```

The floor is derived from the snapshot, not from configuration, and it is **persisted** at a second cursor-port key — so the trap does not return on reload. `is recovered by a second handle over the same storage, rather than reverting to 'unbounded'` covers exactly that, and the conformance suite carries `never reports 'unbounded' over rows whose history it never received` so a future backend inherits the answer instead of rediscovering the trap in a tab.

An as-of read below the floor is the typed refusal, asserted structurally rather than by message: `BlockNotRetainedError` with `requested` and `retained: {from: TAKEN_AT, to: TAKEN_AT}`. Refused for `listAsOf` too, not only the point read, and **answered at the floor itself**, which is the block the rows are the state as of.

## The judgement call I checked hardest

Expressing the floor as a **window** (a distance from a moving tip) rather than a fourth `Retention` kind is the decision that could have been quietly wrong, because a window whose tip estimate is stale could over-claim. It cannot here:

- `blocks` is recomputed on every read as `knownTip - origin`, so `floor = tip - blocks = origin` exactly. The floor is an absolute block, not a decaying distance.
- `assertReadable` passes the *same* `knownTip` to `assertRetained` that `capabilities` used, so the refusal boundary and the advertised claim are computed from one number rather than two that can drift apart.
- After a reload before the first `applyBlock`, `knownTip = origin`, so `blocks = 0`. A caller combining that with the real chain tip computes a floor *above* the true one — it believes it has **less** history than it holds. Under-claiming, never over-claiming, and it corrects itself on the first applied block.

The cost is recorded in the code rather than glossed, which is what let me check it. `prune` is delegated whole on the same reasoning (the handle's floor is never lower than the store's own, so the store can only keep more than the report promises — the safe direction for a deletion).

## The rest of the criteria

**One unit.** ✅ `applyBlock(snapshot.takenAt, snapshot.rows, snapshot.cursor)` — rows and cursor in a single atomic call, which is the property task 1 was required to leave open and did.

**Marker-first ordering, and it really is self-healing.** ✅ The seam carries one cursor slot per `applyBlock` and the sync cursor has it, so the marker is a separate write and a crash window exists. Marker-first leaves a floor over an empty store (over-cautious); marker-last would leave rows with no floor, the exact failure this module prevents. I checked the recovery claim rather than taking it: `bootstrap()` does not gate on an existing marker, and the client side decides "already synced" from the **sync** cursor, which a crashed install never wrote — so a re-run genuinely re-bootstraps. Same principle as the memory store writing its cursor last.

**Different processor version refused, naming both.** ✅ `SnapshotProcessorMismatchError` carries `expected`, `found` and `takenAt`, tested at both layers, and a mismatched mirror is not merely rejected on adoption but excluded as a candidate.

**Revert below the floor refused loudly and changes nothing.** ✅ `RevertBeyondSnapshotError`, with `changes nothing, so a host that catches it still holds the state it had` asserting the state and the origin are both intact afterwards. Reverting down to the snapshot block itself is allowed (it is a block the store holds), and `revertTo(-1)` is correctly treated as a wipe that drops the floor with the rows. Deliberately not a `BlockUnavailableError`, since that family is the read path and a `catch` written for a refused query must not swallow this.

Both halves of the reorg answer are present rather than one: refusal on the consumer side (which cannot be delegated), plus declining a candidate taken inside the finality window on the producer side (which is the half that can be prevented).

**Mirror parity.** ✅ Most-advanced-wins, prefer-local, fail over on unreachable — plus two departures that are strengthenings and are stated: failover walks *every* remaining candidate in descending order (the free-form keeper tries the winner and one more, its source says `// TODO more than 2`), and a foreign-processor snapshot is not a candidate at all. The `{prefix}` location form is deliberately not carried over, with reasoning: the envelope carries `takenAt` and `processor` at the top level so selection needs no filename convention, and a published mirror layout belongs to the publishing spec rather than being pre-empted here. The task asked for exactly this kind of "say which and why", and it got it.

**Snapshot contents, measured.** ✅ Current rows only, and the numbers in the ADR match the committed results file: 45.6 KB gzipped current rows (7% of the 619.6 KB stream) against 304.6 KB for every version (49%). The decisive argument is the structural one, not the size one: a full-history snapshot is **not expressible through the seam at all**, since a version is what applying a block *produces*, and a write surface that could set `_lower`/`_upper` could manufacture states no sequence of blocks could reach. That is the right reason to reject it.

**Scope held.** ✅ `createSnapshot` is the minimal test producer and says so; publishing as a first-class artifact is left to its own spec.

## Noted, not blocking

The spike re-derives the workload as 4,046 live rows / 24,759 versions where `sqlite-in-the-browser.md` recorded 4,072 / 29,393. That is not a re-litigation of the finding — it is a different script over the same fixture, and most likely reflects the re-modelled port that landed with `promote-stratagems-conformance-workload`. The decision rests on the ratios (6.7x, 7%, 49%), which are robust to it. I am carrying this to the drive's triage as an observation rather than treating it as a defect.

Merging.

PR #20, opened 2026-08-25
feat(ingest-wire-receiving-side): Log ingestion endpoint, authoritative…

wighawag:
Task: `work/tasks/done/ingest-wire-receiving-side.md`

Done. Everything is green: `pnpm format:check && pnpm build && pnpm typecheck && pnpm test` all pass, and the working tree holds only this task's changes.

## What I built

**Reality check first.** `docs/adr/0004` reads exactly as the task assumes. Both blockers landed: `sql-backed-event-processor` as `@etherfold/processor-sqlite` (`VersionedStateEventProcessor`, now a thin flavour of `EntityEventProcessor`), and `agnostic-server-skeleton` as `@etherfold/server`. `ReorgCause` (`ca6f981`) is in `generateStreamToAppend` and returns `{cause, blockNumber, blockHash}`. No drift; I proceeded.

**`StreamBuilder` (`packages/core/src/streamBuilder.ts`)** — the receiving side, chain-free. It reads the persisted cursor through `processor.load()` on every call, validates the batch's `{source, config}`, checks the envelope, then hands the batch to `generateStreamToAppend` and drives the processor. This is what the observation note `indexer-load-needs-a-chain-so-the-server-half-cannot-call-it.md` asked this task to resolve: `EthereumIndexer.load()` opens with `eth_chainId`, so the processor-hosting half could never use it.

**Typed refusals (`packages/core/src/errors.ts`)** — `generateStreamToAppend` now throws `UnexpectedFromBlockError` carrying `expectedFromBlock` instead of an `Error` whose message had to be parsed. Same rule, same message, one place, so the wire and the engine cannot drift. Plus `WireContextMismatchError` and `InvalidBatchError`.

**`GET` / `POST /ingest` (`packages/server/src/api/ingest.ts`)** behind an `INGEST_TOKEN` bearer guard applied to the *path*, so a later route inherits it. `409` (+ `expectedFromBlock`) is the only resumable refusal; `400` for the ones no block number fixes; `501` when the server hosts no processor. Reorg counters land in `Meta` and are reported on `/status` as `reorgs: {absence, contradiction, last}`.

**Tests.** 19 new in `packages/core/test/streamBuilder.test.ts` (the contract, including "derives the SAME stream the engine derives" against `EthereumIndexer.feed`) and 19 in `packages/server/test/ingest.test.ts` (the whole sequence — apply, re-send, gap, context mismatch, malformed payload, contradiction reorg, absence reorg, counters, auth — against a real `VersionedStateEventProcessor` over local libSQL). Both suites verified load-bearing by mutation: nulling `assertContext` and the token check turns 12 of them red.

Changeset: `.changeset/ingest-wire-receiving-side.md` (core minor, server minor, processor-entities patch). `changeset status --since=main` reports "no changesets found" only because the file is still untracked — it reads `git diff --name-only <divergedAt>`, so it becomes visible the moment the runner commits.

## Decisions

- **`StreamBuilder` lives in `@etherfold/core`, not in `@etherfold/server`.** The two primitives it must not duplicate (`generateStreamToAppend`, `getFromBlock`) plus `defaultFromBlockOf` and `indexerMatches` are engine logic; a copy in the server package would be the "second, parallel mechanism" ADR-0004 exists to avoid. Alternative considered: export the two primitives from core and write the receive loop in the server, which would have duplicated the source-hash/fresh-cursor/discard logic. Touches: `agnostic-log-fetcher` (its counterpart), `server-platform-adapters` (hosts wire it in), and `indexer-server-feed` (ADR-0006's stored emission stream belongs to this same object when it is built). I also moved `indexerMatches` and the `defaultFromBlock` derivation out of `EthereumIndexer` into `internal/engine/utils.ts` so both shapes read one implementation. `@etherfold/server` gains `@etherfold/core` as a runtime dependency; `platformAgnostic.test.ts`'s exhaustive dependency list is updated with the reason.
- **The wire carries DECODED `LogEvent`s, not the `eth_getLogs` JSON-RPC shape.** ADR-0004 says "raw logs", which I read as *raw as opposed to reorg-annotated* (no `removed`, no `unconfirmedBlocks`), not *undecoded*. The receiver's primitive takes decoded events, `captureStream` already produces them, and the landed `processor-sqlite/test/deployment-shapes.test.ts` already pins this exact shape crossing a wire. Alternative: ship undecoded logs and decode server-side, which would move `LogEventFetcher`'s decoding (currently provider-bound) to the receiver for no gain, since both halves hold the ABI anyway. Cost: `args` restates what `data`/`topics` encode, so payloads are larger; both are still carried, which ADR-0006 will need. **Touches `agnostic-log-fetcher` directly** — I documented it on the `WireBatch` type, and ADR-0004's wording is worth a one-line clarification by a human.
- **Auth fails CLOSED on a missing `INGEST_TOKEN`: every caller gets `401`.** The retired server generated a key at boot and printed it to stdout, which `docs/reviews/server-cli-batch.md` LOW-1 flagged; that is not repeated. Alternatives: `503` "misconfigured" (more accurate about *why*, but weakens the invariant "an unauthenticated caller cannot advance the cursor" into two codes) or allowing anyone when unset (rejected outright). This is a user-visible default and a new env key. The compare is a hand-written constant-time XOR loop, because `node:crypto` is forbidden in this package (a test asserts it).
- **A context mismatch is `400`, not `409`.** `409` is the one and only RESUMABLE refusal: it carries `expectedFromBlock` and the sender's whole recovery is to re-send from there. A foreign `{source, config}` is a misconfiguration no block number fixes, and a sender that treated the two alike would retry forever. Malformed envelopes join it at `400`. Touches `agnostic-log-fetcher`'s retry policy.
- **A new refusal that ADR-0004 does not name: `InvalidBatchError`.** It rejects a range that is not a range, a log outside `[fromBlock, toBlock]`, and a log already marked `removed`. The last two are the enforcement side of two ADR-0004 invariants that would otherwise be silently violated: `groupLogsPerBlock` *drops* `removed` events, so a sender shipping retractions would never learn they went nowhere. Alternative: accept and ignore, which is how a truncating fetcher stays invisible.
- **Reorg counters live in the DATABASE (`Meta`), not in process memory, and are reported on `/status`.** ADR-0004 asks for absence-driven reverts to be "logged loudly and counted"; what an operator needs is a *rate*, and a counter that resets when a Worker isolate is recycled measures the isolate. This deliberately differs from `lastError`, which is in memory for the opposite reason (it often records a failure to reach the database). It is written after the batch, best-effort, and a failure to count never fails an ingestion that succeeded. It does **not** affect `healthy`: an absence-driven revert is a signal to investigate, not a fault that should get the server restarted. `platforms/cf-worker/test/status.test.ts` pins the status shape and is updated.
- **`getIngestion` is OPTIONAL, and a server without one answers `501`.** `/status` and `/admin/setup` are useful on a processor-less server and every existing host builds one that way. Alternative: not registering the route (a `404`, which reads as "wrong URL") or making it required (breaks `platforms/*` and every existing caller). Touches `server-platform-adapters`, which owns wiring a processor into the Node and Worker hosts — I deliberately did **not** add a pass-through to `startServer`, since a pass-through without processor-module loading is half a feature and that task owns it.
- **`GET /ingest` can WRITE.** Reading `expectedFromBlock` reconciles a persisted cursor belonging to another source/config/processor version by calling `processor.clear()`, exactly as `EthereumIndexer.load()` does. Alternative: report a number derived from state the next `POST` is about to wipe, so the read and the write disagree. Nothing is lost that was not already invalid. Documented at the choice site.
- **The wire uses the TAGGED BigInt codec (`{__bigint__: "…"}`), and that codec now lives once in `@etherfold/core`.** A decoded log's `args` hold BigInts, `JSON.stringify` throws on those, and the `"123n"` suffix pair (`bnReplacer`/`bnReviver`) would revive a contract-emitted string ending in `n` as a number — a failure mode `packages/core/src/utils/bigint.ts` already documents. Rather than add a second identical tag codec, I moved `@etherfold/processor-entities`' private copy in `cursor.ts` onto the new `taggedBnReplacer`/`taggedBnReviver`. Coherence note: core now carries two BigInt JSON conventions on purpose — the suffix one describes data this repo's storage adapters have already persisted and cannot be changed without rewriting it; both are documented against each other. Touches `@etherfold/processor-entities` (internals only, no API change) and `agnostic-log-fetcher`, which must use `serializeWireBatch`.
- **Coherence: `StreamBuilder` claims the glossary's `stream-builder` name while owning only half of it.** `CONTEXT.md` defines a stream-builder as owning `expectedFromBlock`, reorg derivation *and* the stored emission stream. The stored stream is ADR-0006 / the `indexer-server-feed` spec and is not built. I took the name rather than inventing a second one, because the missing half belongs to this same object when it arrives; `CONTEXT.md` is updated to say so explicitly, along with new entries for `reorg cause` and "the cursor is the idempotency key".


wighawag:

## The two invariants this task exists to protect

**The cursor rule lives in ONE place.** ✅ `generateStreamToAppend` throws `UnexpectedFromBlockError(expectedFromBlock, receivedFromBlock)` and the route maps that to `409` carrying `expectedFromBlock`. The wire does not re-implement the check, so it cannot drift from the engine's — which was the specific risk, since a second copy would be a resumption protocol that disagrees with the thing it resumes. Turning the message into a typed error rather than parsing a string is the right call and closes that off properly.

**No reorg information crosses the wire, and it is ENFORCED rather than documented.** ✅ `assertWellFormed` refuses a log marked `removed`, with a message that says why: *"the receiver derives every retraction, so a sender that ships them holds reorg logic it must not have."* That matters more than it looks — `groupLogsPerBlock` silently *drops* `removed` events, so without this check a sender shipping retractions would have them go nowhere and never learn. It also refuses a non-range, a `toBlock` above the reported tip, and a log outside the claimed range, which is the enforcement side of "completeness is an invariant, not a flag".

## The rest of the criteria

Batch at `expectedFromBlock` applies and advances state and cursor together; anything else is refused with nothing applied; a re-sent batch fails the check and is corrected rather than double-applied; a context mismatch is a distinct loud error; reorgs are derived server-side from raw logs alone and reach the same result as the engine — pinned by `derives the SAME stream the engine derives` against `EthereumIndexer.feed`, which is the test that makes "one mechanism, not two" checkable rather than asserted; and absence-derived reverts are surfaced separately from contradiction-derived ones via `reorg.cause`.

The refusal ORDER is right and reasoned: identity, then envelope, then cursor — so a batch for another source is never told to resume from a block number that means nothing to it.

## Judgement calls I checked rather than accepted

**A context mismatch is `400`, not `409`.** Correct, and the reasoning is the important part: `409` is the one resumable refusal and its whole contract is "re-send from `expectedFromBlock`". A foreign `{source, config}` is a misconfiguration no block number fixes, so a sender treating them alike would retry forever.

**Auth fails closed on a missing `INGEST_TOKEN`.** Right default, and it explicitly does not repeat the retired server's generate-a-key-and-print-it behaviour that `docs/reviews/server-cli-batch.md` LOW-1 flagged.

**`StreamBuilder` in `@etherfold/core` rather than `@etherfold/server`.** Right: `generateStreamToAppend`, `getFromBlock`, `defaultFromBlockOf` and `indexerMatches` are engine logic, and a copy in the server package is precisely the "second, parallel mechanism" ADR-0004 exists to prevent.

**The wire carries decoded `LogEvent`s.** ADR-0004 says "raw logs"; reading that as *raw as opposed to reorg-annotated* rather than *undecoded* is defensible, the landed `deployment-shapes.test.ts` already pins this shape crossing a wire, and the cost (larger payloads) is stated. The report is right that ADR-0004's wording deserves a one-line human clarification — flagging it rather than silently reinterpreting it is the correct handling.

## Carried, not blocking

**`GET /ingest` can write.** Reading `expectedFromBlock` reconciles a cursor belonging to a foreign context by calling `processor.clear()`. The justification holds — the alternative reports a number derived from state the next `POST` is about to wipe, so the read and write disagree — and it is idempotent and behind a bearer token. But a `GET` with a side effect is still a shape that surprises proxies, prefetchers and retry-happy clients, and it is documented only at the choice site. Worth a second look when `agnostic-log-fetcher` is built, since that is the caller.

**Scope reached into `tagged-bigint-codec-across-storage-adapters`.** The tagged codec moved from `processor-entities/src/cursor.ts` into `@etherfold/core`, which is literally that backlog task's first acceptance criterion. I accept it: the wire genuinely needs a BigInt codec, a *second* tagged copy would violate the same principle the task exists to enforce, and it was flagged openly rather than slipped in — including the deliberate coherence note that core now carries two conventions, documented against each other. The task stays valid (the CLI, both browser adapters, the fs adapter, the stream fixture and the stratagems workload are still on the `"123n"` suffix), but its premises have now moved a third time, and I will update it again.

## One thing I fixed rather than blocked on

`work/notes/observations/indexer-load-needs-a-chain-so-the-server-half-cannot-call-it.md` names *this task* as its resolver, and the PR body claims to resolve it — but the note was still in the tree. I verified the claim is true rather than taking it: `StreamBuilder.currentLastSync()` calls `processor.load()` directly with no chain call, and it calls `indexerMatches`, which is exactly the discard path the note recorded as being skipped by the workaround. The signal is carried by the code now, so the note discharges by deletion per the work contract. Pushed to this branch as `67b0010` so it lands in the same change rather than as stale residue.

Merging.

PR #21, opened 2026-08-25
feat(tagged-bigint-codec-across-storage-adapters): Move every storage a…

wighawag:
Task: `work/tasks/done/tagged-bigint-codec-across-storage-adapters.md`

The drift check passed (codec already in core, `processor-entities` migrated, `simple_hash` prefixing with `h`, the two archived consumers still in `archive/`, the two unnamed ones present), so I built the task.

## What I did

**One codec, in the core.** `packages/core/src/utils/bigint.ts` now holds only `taggedBnReplacer` / `taggedBnReviver`. `bnReplacer`, `bnReviver` and `isBigIntLiteral` are deleted from the workspace, along with `@etherfold/cli`'s `src/utils/bn.ts` and `@etherfold/fs`'s `src/utils/json.ts`, and `bnReviver` is gone from `@etherfold/browser`'s public surface.

**Every adapter moved onto it**: the CLI's snapshot keeper, `keepStateOnIndexedDB` (its remote reads only — the local half is structured clone, which stores a BigInt as a BigInt), `keepStateOnLocalStorage`, the fs keeper, the core's stream fixture, and the stratagems workload's `canonical()`. Two tests that still spoke the old convention (`processor-sqlite`'s wire crossing, `js-processor`'s memory keeper) now use the real codec.

**Committed artifacts regenerated, and proved unchanged.** Both stream fixtures were re-encoded (format 1 → 2) and both golden states re-rendered. `docs/spikes/tagged-bigint-codec-across-storage-adapters/` holds the migration (which refuses to convert any value whose declared ABI type is not an integer — all 57,846 are `uint64`/`uint256`/`uint112`/`uint104`) and the proof: the new goldens re-rendered in the OLD encoding are byte-identical to `b40298e`. `test:full` (31,332 events) passes.

**Tests per adapter**, all asserting `typeof`, with a real BigInt and `"123n"`/`"0n"`/`"-5n"` look-alikes in one `LastSync` beside `context` digests: `cli/test/keepState.test.ts`, `fs/test/keepStateOnFile.test.ts`, `browser/test/keepStateBigInts.test.ts` (both keepers, local and remote), `core/test/streamFixture.test.ts`, `core/test/bigint.test.ts`.

Gate: `format:check`, `changeset status --since=main` (verified against a throwaway index, since the changesets are untracked until the runner commits them), `build`, `typecheck`, `test` all green. `@etherfold/state-store` still declares no dependencies and `no-platform-leakage.test.ts` passes.

**Out of scope, captured not fixed:** `STRATAGEMS_WORKLOAD=all` still times out on IndexedDB (626s of store writes; the fixture parse is 0.7s of that) — that is the pre-existing quadratic `fake-indexeddb` cost already recorded in `work/notes/observations/fake-indexeddb-write-cost-grows-quadratically.md`. Two new notes: `keeper-version-marker-is-written-and-never-read.md` and `snapshot-size-measured-under-the-retired-bigint-encoding.md`.

## Decisions

**1. The reader does not accept the legacy suffix form at all, and there is no fallback.** The ambiguity being removed lives in the reader, so a reader that still honours `"123n"` still has it. A `"123n"` string is now simply a string, everywhere. Alternatives considered: (a) accept both with an end date — rejected, it keeps the defect for the whole window and there are effectively no users; (b) translate on read — rejected, the translation *is* the guess; (c) refuse `/^-?\d+n$/` value-by-value — rejected outright, it would refuse legitimate event data, a worse failure than the one being fixed. Touches: every storage adapter, and anyone holding data written by a previous build.

**2. Where a persisted artifact carries a FORMAT number, the number was bumped and its reader refuses an unrecognised one; where it carries none, nothing was invented.** `STREAM_FIXTURE_FORMAT` 1 → 2 (`parseStreamFixture` throws, naming the file) and **the CLI's `SNAPSHOT_FORMAT` 1 → 2, which is a USER-VISIBLE behaviour change**: a snapshot at format 1 or in the bare pre-envelope form is now logged and treated as absent, i.e. a cold start, where it used to be read. That is the point — its BigInts cannot be recovered by this reader, so resuming would resume from state whose every `uint256` had become a string, and re-indexing is the CLI's existing recovery for a snapshot it cannot read. It removes the "still reads a legacy bare snapshot" behaviour that `cli/test/keepState.test.ts` pinned. The format-less artifacts (`keepStateOnFile`'s blob, `keepStateOnLocalStorage`'s, a bare published `lastSync` file) were deliberately NOT given an envelope: they are caches whose recovery is a re-index and whose `clear()` is one call, and adding persisted shape to reject data nobody holds is more machinery than the problem. Alternative considered: a `__FORMAT__` marker on all three keepers — rejected as new persisted surface for no live user; noted that the already-written-and-never-read `__VERSION__` is where such a check would naturally go (observation filed). Touches: `etherfold`'s resume behaviour, `@etherfold/fs`, `@etherfold/browser`.

**3. A KNOWN gap left open, written down at the call site rather than closed.** `keepStateOnIndexedDB` reads a remote snapshot published by the CLI's keeper, and that file *does* carry a format number — but the browser package does not read it, so a legacy published snapshot is installed rather than refused the way the CLI refuses the same bytes locally. Closing it requires `SNAPSHOT_FORMAT` to live where both packages can see it (`@etherfold/browser` does not depend on `etherfold`), which is a seam decision, not a line of code. Recorded in `OnIndexedDB.ts` and in ADR-0029's consequences. Touches: a future task, and anyone publishing snapshots.

**4. The two committed artifacts are REGENERATED, not read through a compatibility path**, and the migration's one unavoidable guess is *checked* rather than asserted. Reading a format-1 file necessarily makes the guess format 1 could not avoid; a fixture carries its own ABIs, so the migration refuses to write unless every value it converts is declared an integer. The proof that no meaning moved is separate and stronger than a diff: new golden → old rendering → byte-compare against `b40298e`. Alternative considered: a compatibility reader for format 1 — rejected, it is exactly the fallback decision 1 rules out.

**5. The stratagems `canonical()` renderer moved onto the tag too, which is a WRITE-side instance of the same defect.** Under `"123n"`, `123n` and the string `"123n"` render identically, so two states that genuinely differ compare equal and the oracle reports agreement it never established. Cost: 138 BigInts in a 24,067-line file, so the readability of the golden diff is essentially unaffected. Touches: the golden state files (regenerated) and `run/regenerate-golden-state.ts`.

**6. `simple_hash` keeps rendering BigInts as `"123n"`, and keeps its `h` prefix.** Those bytes exist to be hashed and are never parsed back, so there is no guess to make; changing either would change every digest ever persisted. The prefix is now belt-and-braces rather than load-bearing, and the `fp-` fingerprint prefix likewise. Documented in `hash.ts` and `fingerprint.ts`.

**7. Frozen spike scripts: re-pointed where they WRITE a shared artifact, left alone where they only measure.** `docs/spikes/sqlite-in-the-browser/run/verify-port.ts` writes the committed golden state, so leaving it on `bnReplacer` would silently un-migrate that file on the next run; it and its two siblings now use `taggedBnReplacer` (none of their committed result JSONs contains a legacy literal, so nothing on disk changed). `measure-snapshot-size.ts` keeps its local legacy replacer so its committed byte measurements stay reproducible; only its now-false comment was corrected, with the drift filed as an observation.

**8. Wrote `docs/adr/0029-one-bigint-convention-and-it-identifies-rather-than-guesses.md`.** It meets the gate: hard to reverse (format numbers bumped and consumed, committed artifacts re-encoded), surprising without context (a snapshot that used to resume now cold starts), and a real trade-off (ambiguity in the reader vs. a forced re-index). It also states explicitly what is still NOT decided: the HTTP wire format for a decoded `LogEvent`, which left this task with the archived server and gets its own ADR the first time such a surface exists.


wighawag:

## The criterion I added to this task, verified by running it

The re-scope added a criterion the original task did not have: the stream fixture and the golden state are **committed files in the encoding being changed**, so regenerate-vs-compatibility had to be a decision rather than a discovery. It was decided (regenerate, no compatibility reader) and — more importantly — *proved*, and I re-ran the proof in a clean clone rather than trusting the report:

```
stratagems-base.state.json:   IDENTICAL to b40298e once re-rendered (2 BigInts)
stratagems-alpha1.state.json: IDENTICAL to b40298e once re-rendered (138 BigInts)
```

And the other half, which the proof alone does not cover — that the re-encoded *stream* still replays to those goldens:

```
STRATAGEMS_WORKLOAD=full  →  4 files, 22 tests passed  (31,332 events)
```

The proof's DIRECTION is the part that makes it worth anything, and the script says so itself: new is projected onto old, never the reverse, because *"projecting old onto new would have to guess which `123n` strings were BigInts, which is the very ambiguity being removed, and a check that has to guess proves nothing."* That is exactly right, and it is the difference between a proof and a restatement.

## The ambiguity is removed, not relocated

`bnReplacer`, `bnReviver` and `isBigIntLiteral` are gone from the workspace — I grepped, and the only survivors are CHANGELOG history and a deliberate local copy inside `processorFingerprint.test.ts`, which keeps the old decoder to assert the `fp-` prefix keeps a fingerprint off the ambiguous shape for readers outside this repo. That assertion outliving the bug it was written for is good judgement, not drift.

**No fallback in any reader** is the right call and the reasoning is the task's own: the ambiguity lives in the reader, so a reader that still honours `"123n"` still has it. Accepting both with an end date would have kept the defect for the whole window.

`@etherfold/state-store` still declares `"dependencies": {}` and `no-platform-leakage.test.ts` is byte-unchanged. The ADR's catch that `describeValue` renders `123n` in an ERROR MESSAGE — JavaScript's own literal syntax for a human, not a codec — is a distinction worth making rather than over-applying the rule to it.

## Judgement calls I checked rather than accepted

**Format numbers bumped where they exist, none invented where they do not.** `STREAM_FIXTURE_FORMAT` and the CLI's `SNAPSHOT_FORMAT` both go 1 → 2 and their readers refuse an unrecognised one; the format-less keeper blobs deliberately get no new envelope. The reasoning holds: those are caches whose recovery is a re-index, and adding persisted shape to reject data nobody holds is machinery for its own sake.

**The CLI snapshot at format 1 now cold-starts rather than resuming.** User-visible and correctly called out as such. It is the honest behaviour: its BigInts cannot be recovered by this reader, so resuming would resume from state whose every `uint256` had silently become a string — which is the failure being fixed, arriving through the recovery path.

**`simple_hash` keeps `"123n"` and its `h` prefix.** Correct: those bytes are hashed and never parsed back, so there is no guess to make, and changing either would move every digest ever persisted.

**The stratagems `canonical()` renderer moved too**, which the report rightly identifies as a WRITE-side instance of the same defect: under the suffix, `123n` and the string `"123n"` render identically, so two genuinely different states compare equal and the oracle reports agreement it never established. Catching that the equality oracle itself was ambiguous is the sharpest observation in the PR.

## One thing I corrected rather than blocked on

ADR-0029's consequences said *"`@etherfold/server`'s `/ingest` carries RAW logs (ADR-0004) and no surface here serves [a decoded `LogEvent`]"*. That is wrong: `WireBatch.logs` is `LogEvent<ABI>[]`, and `ingest-wire-receiving-side` explicitly recorded reading ADR-0004's "raw logs" as *raw as opposed to reorg-annotated* rather than *undecoded*. The sentence also contradicted its own next clause, which notes `serializeWireBatch` already uses this codec.

It matters more than a typo because **`agnostic-log-fetcher` is the sender on that exact wire and is the next task on the board**; an ADR is the durable record, and this one would have read as licence to ship undecoded logs. Corrected on the branch as `99fdcc1`: the ingest direction is settled, and only a surface that SERVES decoded events in a response body remains open.

## Carried, not blocking

**The remote-snapshot gap (decision 3) is the one real residue.** `keepStateOnIndexedDB` reads a snapshot published by the CLI's keeper; that file carries a format number and the browser does not read it, so a legacy published snapshot is **installed rather than refused**, where the CLI refuses the same bytes locally. It comes back with its BigInts as `"123n"` strings — silently wrong data, which is this task's own failure mode surviving in one corner.

I accept it here: it is written down at the call site and in the ADR rather than overlooked, closing it needs `SNAPSHOT_FORMAT` to live where both packages can see it (a seam decision, not a line of code), and the spec puts migrating live consumers out of scope. But it stops being theoretical the moment `publish-etherfold-and-deprecate-old-names` runs, and a call-site comment is findable only by someone already in that file. It is task-worthy, and I am recording it as such rather than authoring it.

Merging.

PR #24, opened 2026-08-28
feat(handler-types-do-not-lie-when-one-name-covers-two-events): Handler…

wighawag:
Task: `work/tasks/done/handler-types-do-not-lie-when-one-name-covers-two-events.md`

The prior attempt's work is on this branch and complete. I reviewed it and re-verified it end to end rather than restarting.

## What I found on the branch (commit `7fb32c0`)

- `packages/js-processor/src/processor/types.ts` and `packages/processor-entities/src/types.ts`: `InputValues` is now distributive (`T extends AbiEvent ? {...} : never`), each with a JSDoc explaining why the conditional is load-bearing.
- `packages/js-processor/test/handlerArgs.test.ts` and `packages/processor-entities/test/handler-args.test.ts`: both directions asserted, `@ts-expect-error` in the `processorKinds.test.ts` style, plus an `IsExactly` type-identity pin and a vitest run-time half.
- `.changeset/handler-args-are-a-union-when-one-name-covers-two-events.md`: `minor` on both packages.
- No signature-keyed handler names were added (correctly deferred).

## What I verified myself

1. **The reality check the prompt demands, in reverse.** I temporarily reverted `InputValues` to the non-distributive form in both packages and re-ran the typecheck. Both packages went red, each with exactly two errors: the union identity assertion failing (`Type 'true' is not assignable to type 'false'`) and `error TS2578: Unused '@ts-expect-error' directive` on the un-narrowed `event.args.memo` line. That is the premise confirmed still true against current code (the merge does compile without the fix), and it proves the tests actually bite rather than passing vacuously. I restored both files byte-for-byte; the tree is clean.
2. **No third copy.** `packages/processor-sqlite/src/types.ts` only re-exports `InputValues`/`InputNames` from `@etherfold/processor-entities`, so the two changed copies are all of them.
3. **`noUnusedLocals` is not set** in `tsconfig.base.json`, so the unused `const lie` cannot be what satisfies the `@ts-expect-error`; the property access is.
4. **Full gate green**, in the `dorfl.json` `verify` order: `pnpm format:check` clean, `pnpm changeset status --since=main` reports `@etherfold/js-processor` + `@etherfold/processor-entities` at minor (and `@etherfold/processor-sqlite` at patch via internal deps), `pnpm build`, `pnpm typecheck` (all packages, platforms and examples), `pnpm test` all pass.

Every acceptance criterion is met. I made no source change, because the correct change was already there and re-doing it would only churn it.

## One thing for the human, outside my scope

The prior attempt's commit already includes the runner's own done-move: on this branch the task body is at `work/tasks/done/handler-types-do-not-lie-when-one-name-covers-two-events.md` (it is still at `work/tasks/ready/...` on `origin/main`), with the prior `## Decisions` block already transcribed into it. So this requeue happened *after* a successful completion commit, and the runner's next done-move has nothing left in `ready/` to move. I did not touch it, per the no-git-state rule, but it may need the runner (or you) to reconcile the item state rather than a rebuild.

## Decisions

None new. The three decisions standing behind this change (both packages bump `minor`; each test file carries a vitest half driving a two-version processor through a direct `process()` stream; the `examples/web-demo` typecheck-needs-build observation captured rather than fixed) were made by the prior attempt and are already recorded verbatim in the `## Decisions` section of the task body on this branch. I re-checked all three against the code and still agree with them, so I am deliberately not re-emitting them here to avoid duplicating that section in the done record.


wighawag:

Gate-1 (acceptance) green. Gate-2 off by repo config. Reviewed the diff against the task's six acceptance criteria and its one drift note.

**Per criterion**

1. **`InputValues` distributes in BOTH packages** — met. `packages/js-processor/src/processor/types.ts` and `packages/processor-entities/src/types.ts` each take the `T extends AbiEvent ? {...} : never` form, and each carries a doc comment explaining that the conditional is load-bearing rather than decorative.
2. **Un-narrowed read does not compile** — met, and independently verified rather than taken on trust. Reverting `InputValues` to the non-distributive form in a clean clone yields `test/handlerArgs.test.ts(102,3): error TS2578: Unused '@ts-expect-error' directive`, i.e. the guarded line starts compiling exactly when the fix is removed. The assertion bites.
3. **`in`-narrowing gives the version-specific field and the shared ones** — met. The `if ('memo' in event.args)` branch reads both `memo` and the shared `id`.
4. **Single-version processor compiles exactly as before** — met, and this is the strongest part. Rather than assuming it, `singleVersionArgsAreUnchanged` pins the type with an `IsExactly` identity check against a spelled-out object type, so a widening to `any`, to an optional, or to a union with `never` fails here instead of leaking downstream. Reverting the fix trips it: `TS2322: Type 'true' is not assignable to type 'false'`. Identity rather than assignability is the right instrument, since a union and a merge are assignable in one direction.
5. **Assertions run under `pnpm typecheck` via `@ts-expect-error`** — met, in the `processorKinds.test.ts` style, with both files documenting why vitest cannot make this assertion (esbuild strips types without checking them).
6. **Changeset covers both packages** — met, `minor` for `@etherfold/js-processor` and `@etherfold/processor-entities`.

**Drift note honoured.** No signature-keyed handler names (`on['Transfer(address,address,uint256,bytes)']`). Handler keys stay NAME-based, and the changeset records the deferral as an alias that could be added later without removing the union.

**Note on the bundled observation.** `work/notes/observations/examples-typecheck-needs-examples-built-first.md` records the repo-wide gate defect found while building this task. It was accurate when written and is now resolved by c88d9c8, which points the private example packages' types at `src/` and adds the type-only imports that `verbatimModuleSyntax` then requires. Landing the note as written, since observations are append-only history; the resolution is recorded separately.

Merging.

PR #25, opened 2026-08-28
feat(an-event-is-never-silently-dropped-from-the-fetch-filter): An even…

wighawag:
Task: `work/tasks/done/an-event-is-never-silently-dropped-from-the-fetch-filter.md`

All gates green. Summary of the work:

**Verified the defect first** (it reproduced exactly as the task described): with two contracts declaring same-named events with different inputs, only ONE topic0 reached `eth_getLogs` (`0xddf252…`, the other absent, no error thrown); with the same address declared twice, construction threw `two events with same name but different inputs`. 15 of the 18 new tests failed before the change.

**Production change** (`packages/core/src/internal/decoding/LogEventFetcher.ts`):
- `deleteDuplicateEvents` lost its `failOnIdenticalNameButDifferentInputs` flag and is now keyed on the canonical event **signature** (so on `topic0`) instead of the name. Different topic0 keeps both; same topic0 with an identical decoding shape collapses silently; same topic0 with a different decoding shape throws at construction, naming both declarations and the colliding topic0.
- The same call is now made on **every** ABI list (each per-address list and the global list), so the verdict cannot depend on `parseAllEventsIrrespectiveOfAddresses`.
- Topics are computed per event via `toEventSelector` instead of `encodeEventTopics({abi, eventName})`, which selected by name and so could only ever yield one topic per name (that alone would have kept the second version out of the filter). Anonymous events still contribute no topic, as before.
- `_nameToTopic` became `_topicsPerEventName` (a list), so a name-keyed argument filter applies to every topic0 that name covers.

**Tests** (`packages/core/test/fetchFilter.test.ts`, 18 cases) assert on the topics the fetcher **requests**, via a recording provider, including an explicit regression case for the silent-drop path, the upgrade pair at a single block, refusal in four ABI arrangements × both paths, and the surviving legitimate de-duplication.

**Also**: `.changeset/an-event-is-never-silently-dropped-from-the-fetch-filter.md` (`@etherfold/core: minor`), `docs/adr/0031-duplicate-events-are-keyed-on-topic0-and-only-a-topic0-collision-is-refused.md`, and one out-of-scope observation note (`work/notes/observations/sqlite-conformance-times-out-under-load.md`): four `state-store-sqlite` conformance cases hit vitest's 5s default under full-suite load and pass in 3.17s when re-run alone; that package does not depend on `@etherfold/core`.

`pnpm build`, `pnpm typecheck`, `pnpm format:check` all pass; every package's tests pass (core 306, browser 60, stratagems workload 16, state-store-sqlite 273 in isolation).

## Decisions

**Accept both events when their topic0s differ; refuse only a true topic0 collision.** The task left the verdict open between "accept both" and "refuse, naming them". I chose accept-both because topic0 is what a log carries: `Transfer(address,address,uint256)` and `Transfer(address,address,uint256,bytes)` are distinguishable on the wire with no ambiguity to resolve, and at an upgrade block both legitimately occur (the upgrade transaction sits mid-block). Refuse-both was the alternative, and it would make a contract that ever changed an event signature un-indexable at all, which is exactly what `abi-versions-are-block-ranged` needs to stop being true. Touches: `abi-versions-are-block-ranged` (this is its unblock), and the public behaviour of `LogEventFetcher` for anyone whose ABI previously lost an event. Recorded as ADR-0031.

**Declarations are compared on their DECODING shape, not with a whole-object equality.** The old per-address path used `deepEqual(event.inputs, ...)`, which includes `internalType`. Since the refusal now stops the indexer starting on both paths (where the global path previously just truncated), an `internalType` difference between two compilations of the genuinely same event would newly break a working deployment. So the comparison is on parameter names, types, `indexed` flags, tuple components and `anonymous`, with a missing parameter name normalised to `''`. Alternative considered: keep whole-object equality and accept the false refusals as "loud is safe" — rejected because a loud refusal that is wrong is still wrong, and `internalType` cannot change what a log decodes to. Touches: any source assembled from ABIs produced by different compilations. In ADR-0031.

**A name-keyed argument filter now applies to EVERY topic0 that name covers.** `LogParseConfig.filters` is documented as keyed by event name, and a name can now cover several topic0s. Previously `_nameToTopic` was last-write-wins, so one version got the filter and the other fell into the shared, unfiltered request — the same declared filter meaning two different things for two versions of one event. The alternative, filtering only one topic0, is not expressible as a user intent anyone would have written. Touches: `parseConfig.filters` (user-visible: with an upgrade pair, a configured filter now produces one request per version rather than one filtered plus one unfiltered). In ADR-0031's consequences.

**The refusal is a plain `Error`, not a new exported error class.** `packages/core/src/errors.ts` is an explicitly framed family about wire refusals and retryability (every member carries `retryable`), and this is a construction-time programming error, alongside the existing plain `duplicate topics found` throw in the same file. Alternative considered: an `AmbiguousEventABIError` in `errors.ts` — rejected as expanding the public error surface and re-meaning that family for something that never crosses the wire. The message carries what a caller needs (both declarations and the topic0), and the tests assert on it.


wighawag:

Gate-1 (acceptance) green. Gate-2 off by repo config. Reviewed against the task's criteria as RE-SCOPED on 2026-08-28, including the two added when this task grew to cover the upgrade case.

**Per criterion**

1. **Verdict identical with and without `parseAllEventsIrrespectiveOfAddresses`** — met, and structurally rather than by matching two branches. The flag parameter is gone; one `deleteDuplicateEvents` runs over every list, per-address and global alike. The refusal test is parameterised across both paths.
2. **No surviving event absent from the REQUESTED topics** — met, and asserted at the right place. `topicsRequested` drives a real `getLogEvents` through a recording provider and reads `topics[0]` off the captured `eth_getLogs` calls, so the assertion is on what the node was asked, not on what the fetcher accepted. That was the whole point: what it accepted was never the thing that was wrong.
3. **One loud rule for same-named events with different inputs** — met. Keyed on the canonical signature, so on topic0: different topic0 keeps both, identical declaration collapses, shared topic0 with different definitions refuses.
4. **Ambiguous ABI refused at construction, naming the events** — met. The message carries both rendered declarations, the colliding topic0, and a remedy, and it degrades correctly for the anonymous case where there is no topic0 to name.
5. **Legitimate de-duplication still works** — met, covered on both the global and per-address lists.
6. **Tests in the repo's vitest style, with a regression test for the silent-drop path specifically** — met, and the regression test is explicitly labelled and commented with the failure it pins.
7. **Changeset** — met, `minor` on `@etherfold/core`, and it documents the two consequential side effects rather than hiding them: a name-keyed filter list now applies to every topic0 that name covers, and declarations are compared on decoding shape rather than whole-object equality.
8. **Two versions of one contract's event both kept, both topic0s requested** — met, parameterised across both parse paths.
9. **Both versions requestable at a single block** — met, and the test says why: the upgrade transaction sits mid-block, so a transaction before it still fires the old event.

**Beyond the criteria, and worth calling out**

The root cause found here is deeper than the task described. The task blamed the `splice` in `deleteDuplicateEvents`. The real second half is that `encodeEventTopics` selects an ABI item by NAME, so even without the splice, two events sharing a name resolve to whichever was found first and the other's topic0 could never enter the filter. Moving to a per-event `toEventSelector` closes that too. The task would have been only half fixed otherwise.

`decodingShapeOf` excluding `internalType` is the right call and is justified in ADR-0031: two compilations of the same event routinely disagree on it, and this refusal stops the indexer starting, so it must be about the wire and nothing else.

ADR-0031 records the decision and the rejected alternatives. Numbering does not collide; 0030 was the latest.

**One out-of-scope finding, not blocking.** An anonymous event has no topic0, so `topic0Of` returns `undefined` and it is skipped for the filter. A lone anonymous event is therefore absent from the requested topics, which superficially reads like the failure this task exists to prevent. Not blocked, because it is not a regression (`encodeEventTopics` omitted it before as well) and it is not fixable, since there is no topic to filter on. But "cannot be filtered" and "is silently not fetched" should probably not be the same thing, and refusing an anonymous event loudly at construction may be the honest answer. Captured as a finding for separate triage rather than widened into this change.

**Unblocks `abi-versions-are-block-ranged`**, which is blocked by this task: with no per-range buckets, a source carrying two versions of one event could not be constructed at all until this landed.

Merging.

PR #26, opened 2026-08-29
feat(abi-versions-are-block-ranged): An ABI event is live over block ra…

wighawag:
Task: `work/tasks/done/abi-versions-are-block-ranged.md`

The gate is fully green and the working tree is clean. Here is what I found and did.

## What this attempt was

This was a continuation, and the prior attempt's work was already complete and already done-moved (commit `5506976`). It was bounced **twice for reasons unrelated to it**: `core`'s base36 digest sweep and `server`'s `sql2ts` round-trip timing out at vitest's 5s default on a loaded machine (see `work/questions/task-abi-versions-are-block-ranged.md`, Q2/Q3). The fix for that class of failure landed on `main` as `c0d694f` ("stop the acceptance gate assuming an idle machine", 60s `testTimeout`/`hookTimeout` in every package, ADR-0032... actually ADR recorded there). The branch is now `0` behind `origin/main`, so it already carries that fix.

So my job reduced to: re-verify the premises, review the landed work against every acceptance criterion, and run the gate. **No source change was needed.**

## Premise re-check (all still true on the current tip)

- `defaultFromBlockOf` (`packages/core/src/internal/engine/utils.ts`) still minimises across contracts and reads `startBlock` alone; no range field reaches it.
- `TODO handle history (in reverse order)` is gone from both `reinit` call sites, replaced by `sourceHashesOf(...)`; `ContractData.history` is removed and nothing references it.
- `ContextIdentifier.source` is still `{startBlock, hash}[]`, and a source declaring no range still produces exactly `[{startBlock: 0, hash: simple_hash(source)}]` — byte-identical to before, so persisted contexts stay readable with no migration.
- `indexerMatches` is preserved as the boolean surface, now delegating to `sourceInvalidationOf`, which names `invalidFromBlock` + `reason` (the forward-pointer for stream branching).

## Acceptance criteria, spot-checked against the tests

- Inclusive `firstBlock`/`lastBlock` on `RangedAbiEvent`, exported via `RangedAbi` from `@etherfold/core` (`packages/core/src/types.ts`), deliberately not `startBlock`.
- The one-block upgrade overlap (`A.lastBlock = b`, `B.firstBlock = b`) is accepted and both topics are requested at `b` — `packages/core/test/fetchFilter.test.ts`.
- `[A@a, B@b, A@c]` normalisation, gap refusal (naming event + uncovered span), `lastBlock < firstBlock` and non-integer refusals — `packages/core/test/eventRanges.test.ts` (345 lines).
- The headline "nothing is re-fetched" is asserted on `chain.ranges` (the ranges the fake node was asked for), with `{stateDiscarded}` as the second reading — `packages/browser/test/eventRanges.test.ts`, including a reload case proving the kept state adopted the new entries.
- Decoding has no block axis; topic0 collision still refused identically with and without `parseAllEventsIrrespectiveOfAddresses`.
- Both asymmetries documented at the declaration site (`RangedAbiEvent` JSDoc), in `CONTEXT.md`, and in `docs/adr/0032-...`.
- Changeset `.changeset/an-abi-event-is-live-over-block-ranges.md` covers the public source-type change (`@etherfold/core` minor, `@etherfold/browser` patch); `pnpm changeset status --since=main` is green.

## Gate result

`pnpm format:check && pnpm changeset status --since=main && pnpm build && pnpm typecheck && pnpm test` → `GATE_EXIT=0`. All 19 test packages pass (core 339 passed, browser 66, server 47, state-store-sqlite 402, ...); the single "skipped" in core is `publishedTypeDependencies`'s conditional runner, not a disabled guard. `git status --porcelain` is empty.

One note: I ran `pnpm install --frozen-lockfile` because this worktree had no `node_modules`; it wrote nothing tracked.

No new decisions were made in this attempt. The prior attempt's `## Decisions` block (ranges on ABI event entries rather than a per-contract version list; `ContractData.history` removed; a removed entry below the cursor now invalidates; two extra construction refusals; `wireContextOf` deliberately not ranged; ordering ties broken by hash) is already transcribed verbatim into `work/tasks/done/abi-versions-are-block-ranged.md` and stands unchanged.


wighawag:

Gate-1 (acceptance) green. Gate-2 off by repo config. Reviewed against the task as RE-SCOPED and then SPLIT on 2026-08-28, so the fetch narrowing is deliberately absent and belongs to `a-range-requests-only-the-events-it-can-contain`.

**The design constraints that were the point of the re-scope, all held**

- **No block axis in decoding.** `parse()` still selects by address and decodes by topic0; the only `blockNumber` in `LogEventFetcher` populates the event object. This was the single thing the rejected model got wrong, and it stayed rejected.
- **`firstBlock`/`lastBlock` are not `startBlock`** and never reach `defaultFromBlockOf`, which still minimises over contract `startBlock` alone. Pinned by its own test rather than left to inspection.
- **The one-block overlap at an upgrade is preserved**, with a test named for it. `A.lastBlock = b` and `B.firstBlock = b` is accepted and not collapsed, which is the off-by-one that would otherwise silently lose every pre-upgrade log in block `b`.
- **Normalisation is exactly as specified**: an open-ended occurrence absorbs the rest from the minimum `firstBlock`, otherwise the union, and merely touching ranges join since the coverage has no hole.
- **A gap is refused**, naming the event, the address, the uncovered span, and both remedies. The reasoning is recorded where it belongs: a hole is a span nobody requests, so afterwards "the chain had none" is indistinguishable from "we never asked".

**Invalidation on the normalised form, which was the subtle one**

`sourceInvalidationOf` computes over normalised ranges, so `[A@a, B@b, A@c]` produces coverage byte-identical to `[A@a, B@b]`. The test asserts it on the RANGES the fake chain was asked for, not on the resulting state, which is the only assertion that can tell a resume from a re-index. That matters because a generator usually cannot tell an upgrade from a rollback and will just append; without this, a redundant append would shift the by-index comparison and re-index the world.

**The forward-pointer is honoured, not just acknowledged.** The decision is now `{valid, invalidFromBlock, reason}`, with `indexerMatches` a thin wrapper reading `.valid`. So the decision can name the block it starts at, even though the only action today is a full discard. That is precisely the seam `work/notes/ideas/a-stream-branches-instead-of-being-discarded.md` needs, and it keeps branching a refinement rather than a rewrite.

**Comparing on decoding shape rather than whole-object equality** is carried over correctly from ADR-0031, with the same justification: `internalType` is a Solidity-side annotation two compilations disagree about, and discarding a user's entire indexed history over it would be indefensible.

**One fixup I made on the branch**, and it was my fault, not the builder's. The ADR was numbered 0032, which collided with `0032-the-acceptance-gate-does-not-assume-an-idle-machine` that I landed on `main` while this branch was in flight. Mine was second in authorship but first to `main`, so I renumbered this one to **0033** and updated its two references (`CONTEXT.md` and the changeset). The 19 `vitest.config.ts` files and `bigint.test.ts` still correctly mean 0032. Committed on this branch rather than a parallel one. Worth noting the gate does not validate ADR numbering, so nothing caught this.

**Scope discipline held.** The fetch filter is untouched, so every range still carries every topic: wasteful, correct, and no regression. The changeset says so explicitly rather than implying the saving already exists.

Merging.

PR #27, opened 2026-08-29
feat(a-range-requests-only-the-events-it-can-contain): A block range re…

wighawag:
Task: `work/tasks/done/a-range-requests-only-the-events-it-can-contain.md`

Done. Both premises re-checked against the code before building, and both hold.

**Premise 1 — `abi-versions-are-block-ranged` has landed.** `work/tasks/done/abi-versions-are-block-ranged.md`, with `packages/core/src/internal/engine/eventRanges.ts` carrying inclusive `firstBlock` / optional inclusive `lastBlock`, the open-ended-absorbs / union normalisation, and the gap refusal.

**Premise 2 — the request shape is as described.** `generateLogRequestForTopicsAndFiltersCombinations` (`internal/engine/ethereum.ts`) returns one request holding every topic when `filters` is absent, and one request per (topic × filter) otherwise, run sequentially by `getLogsWithVariousFilters`. Now pinned by a test rather than by reading.

## What I did

Test first (`packages/core/test/fetchFilter.test.ts`, the prior art), then the production code. Six new assertions were red before the change and green after.

- **`packages/core/src/internal/engine/eventRanges.ts`** — `requestableRangesPerTopic(contracts)` gives the fetch filter its reading of the ranges: normalised per contract, then keyed by `topic0` and **unioned across contracts**; empty when nothing declares a range. `canOccurIn(ranges, fromBlock, toBlock)` is the inclusive intersection test. `liveEventsOf` is unchanged in behaviour; it and the new function now share one `liveEventsIn`, differing only in the default first block.
- **`packages/core/src/internal/engine/RangeLogFetcher.ts`** — `topicsThatCanOccurIn(fromBlock, toBlock)` filters the topics on the range **actually requested** (after the adaptive clamp). When nothing declared can occur there, no `eth_getLogs` call is made at all.
- **`packages/core/src/internal/decoding/LogEventFetcher.ts`** — computes the map from `contractsData` and passes it down.
- Docs: ADR-0033's stale "until the follow-up narrowing lands" tail replaced with what actually happens, plus the rejected `startBlock` option; `CONTEXT.md` glossary entry; `RangedAbiEvent` JSDoc; a changeset.

Measured here: request COUNT. Under an argument filter, the range `100..200` of an upgraded source issues **1** request where it issued **2**. The node's own `logsBloom` screening work is *not* measured here and is labelled as such in the changeset and the ADR.

Gate green: `pnpm build && pnpm typecheck && pnpm test` (core 358 passed), `prettier --check` clean, working tree carries only the intended changes.

One thing I hit and did not fix: `publishedTypeDependencies.test.ts` read `from "we never asked"` out of my JSDoc prose as an import specifier. That is already captured in `work/notes/observations/published-type-scanner-reads-import-specifiers-out-of-jsdoc-prose.md`, so I reworded the comment rather than adding a duplicate note.

## Decisions

**1. An event that DECLARES no range is open-ended from block 0 in the fetch filter, not from its contract's `startBlock`.** The invalidation reading (`liveEventsOf`) keeps defaulting to `startBlock`; the fetch reading defaults to 0. Alternative considered: reuse `liveEventsOf` as-is (fewer requests, one code path). Rejected because `startBlock` is minimised across contracts by `defaultFromBlockOf`, so a range below one contract's `startBlock` is still fetched with that contract's address in the filter; narrowing on it would mean that adding a range to ONE event silently changes what an UNRELATED event fetches, which is exactly the invisible omission this task must not reintroduce. Touches: `liveEventsOf`/`sourceHashesOf` (unchanged, but now deliberately different), the meaning of `ContractData.startBlock`, and any later task reading ranges. Recorded in ADR-0033 (Considered options) and at the choice site.

**2. A range no declared event can occur in issues NO `eth_getLogs` call at all.** Alternative: send the request with an empty topic list. Rejected because a node reads an empty topic position as a wildcard, so that request asks for *every* log. User-visible: such a range now costs zero round trips (and `RangeLogFetcher` treats it as a zero-log range for its adaptive sizing, so it widens the next span, same as an empty answer). Touches: anything counting requests, and the fetcher's block-count adaptation.

**3. Narrowing is computed on the range actually REQUESTED, after `RangeLogFetcher` clamps `toBlock`.** Narrowing a smaller range can only ever request a subset, so this is the safe order; computing it on the caller's asked-for range would be wider but would also make the assertion in a test depend on `numBlocksToFetchAtStart`. Touches: nothing else; the test helper sets a wide `numBlocksToFetchAtStart` where a test needs the range it asked for.

**4. Ranges are unioned per `topic0` ACROSS contracts, and that union is NOT re-normalised.** The topic filter of an `eth_getLogs` request is global to the request while a range is declared per contract, so the union is the only safe reading. Deliberately not re-normalised, because `normalizeRanges` refuses a GAP: a hole between contract A's `lastBlock` and contract B's `firstBlock` is not a hole in an event's coverage (two contracts have independent lifetimes, as CONTEXT.md already says), so refusing it would reject a legitimate source. Consequence, pinned by a test: inside such a hole both contracts declare the event dead and the topic is not requested. Touches: the gap-refusal rule from `abi-versions-are-block-ranged`, which still applies per contract.

**5. Extended ADR-0033 instead of writing ADR-0034.** Its opening sentence and closing paragraph both forward-referenced this work ("— later —", "until the follow-up narrowing lands"), so they were made false by this change and had to be edited anyway; the block-0 default lives there as a rejected option next to the existing "Reusing `startBlock`" one. Alternative: a new ADR pointing back at it. Touches: `docs/adr/0033`, `CONTEXT.md`.

**6. `RangeLogFetcher` gained a fifth constructor parameter** (`topicBlockRanges`, defaulting to an empty map) rather than a new key on `LogFetcherConfig`. The config type is user-facing through `types.ts`, and this is computed by `LogEventFetcher`, never supplied by a deployment. `LogEventFetcher` is its only subclass.


wighawag:

Gate-1 (acceptance) green on the first pass. Gate-2 off by repo config.

This is the one task in the set that can REMOVE a topic from a request, so it is the one that could reintroduce the exact failure `an-event-is-never-silently-dropped-from-the-fetch-filter` was built to eliminate. I reviewed it primarily against that.

**The safety property holds by construction, not by test coverage alone**

- `canOccurIn` returns `true` for `undefined` ranges, so an event that declared nothing is live everywhere and can never be narrowed away.
- `lastBlock ?? Number.POSITIVE_INFINITY` means an open-ended event is never dropped at any height.
- `topicsThatCanOccurIn` returns `null` when no ranges exist at all, and the call site is `narrowedTopics ?? this.eventNameTopics`, so a source declaring no range takes today's path unchanged rather than a narrowed path that happens to agree.

Every omission therefore traces to a DECLARED range. Nothing is inferred from observed logs or a first-appearance block, which was the explicit prohibition.

**The trap I was most worried about is handled, and handled deliberately**

`asks for NOTHING when no declared event can occur in the range` could have been the worst possible bug: an empty topics array is read by a node as a WILDCARD, so "narrow to nothing" would have inverted into "fetch every log on chain", and with a merged ABI naming no address it would have been every log on the network. The implementation short-circuits to `logs = []` with no provider call, the comment names the hazard, and the test asserts zero requests rather than an empty request. Both branches, filtered and unfiltered, are covered.

**Per criterion:** below `firstBlock` and above `lastBlock` both narrow, on both request shapes; narrowing is identical with and without `parseAllEventsIrrespectiveOfAddresses`; a range crossing a boundary requests the union; the one-block upgrade overlap is preserved at the boundary block itself; an open-ended event is never dropped; a source declaring no range is unchanged topic for topic AND request for request; the request-count saving is demonstrated as strictly fewer requests under argument filters. The merged-ABI path that names no address is covered too, which is the case where a mistake would have been widest.

**Cross-contract independence is right and was not asked for.** Two contracts declaring the same event have independent lifetimes: the topic survives above one contract's `lastBlock` while another declares it open-ended, and it is never narrowed where one contract declared a range and another did not. The gap refusal is per event per address, so one address going quiet cannot punch a hole in another's coverage. That is a genuine subtlety the task did not spell out.

**Honesty split preserved.** The changeset and ADR-0033 claim the request-count saving, which is measured here and demonstrated, and continue to mark the node-side `logsBloom` cost as how nodes implement the method rather than as one of our measurements.

Merging. This completes the upgrade-indexing set.

PR #28, opened 2026-08-29
feat(invalidation-is-computed-on-what-each-thing-depends-on): Invalidat…

wighawag:
Task: `work/tasks/done/invalidation-is-computed-on-what-each-thing-depends-on.md`

Build, typecheck, tests and format all green. Working tree holds only this task's changes.

## What I did

**Verified both premises first.** `sourceHashesOf` did short-circuit to `[{startBlock: 0, hash: simple_hash(source)}]` when no range was declared, and `sourceInvalidationOf` did return one verdict that `indexerMatches` used for both the state gate (`indexer.ts:489`) and the kept-stream gate (`indexer.ts:546`, `:569`). Neither had drifted, so I built.

**1. Per-event hashing on both paths** (`packages/core/src/internal/engine/eventRanges.ts`). `sourceHashesOf` now always emits one entry per (contract, event, live range) over a block-0 skeleton entry (`chainId`, `genesisHash`, addresses, `startBlock`s), computed from `liveEventsOf` and `decodingShapeOf`. Non-event ABI members contribute to nothing; the list is sorted, so reordering an ABI produces identical bytes. `sourceDeclaresEventRanges` became dead and was removed; `contractsDeclareEventRanges` stays for the fetch filter.

**2. Two verdicts** (`internal/engine/utils.ts`). `SourceInvalidation` is now `{state, stream}`, each an `InvalidationVerdict` keeping `invalidFromBlock` and a reason. Each entry carries `hash` (address + signature + decoding shape + range) and `streamHash` (address + `topic0` + range). The stream half ignores removals (a shrunken topic set leaves a superset); the state half does not. `indexerMatches` was replaced by `stateMatches` / `streamMatches`, wired to the respective call sites in `indexer.ts` and `streamBuilder.ts`.

**3. Stored-format bridge.** The block-0 entry carries `legacyHash`, the whole-source digest the shipped code persisted, so a whole-source context compares as unchanged instead of re-indexing. A per-range context needs no bridge: `hash` is computed exactly as before and matches byte for byte. Asserted at the seam and end to end (a persisted IndexedDB context aged to the shipped shape, reloaded, resumes above the start block).

**Tests**: `packages/core/test/invalidation.test.ts` (new), `packages/browser/test/invalidation.test.ts` (new, asserting the ranges the fake node was asked for), plus updates to `packages/core/test/eventRanges.test.ts`. Each new claim was checked non-vacuous by breaking the production code and watching it go red. Also: ADR-0034, a supersession pointer on ADR-0033, CONTEXT.md glossary entries, and a changeset.

## Decisions

**The kept stream is RE-DECODED on replay, unconditionally.** The task frames the stream as "raw logs", but in this codebase a cached `LogEvent` is the raw log (`topics`, `data`, `address`) *plus* `args`/`eventName`, which is what some earlier ABI made of it. Keeping the stream across a renamed non-indexed parameter and replaying the stored `args` would rebuild the state out of keys the current ABI does not have: correct-looking and silently derived from a contract description no longer in force. So `LogEventFetcher.reparse` (new; `decodeOnto` extracted so the fetch path and the replay decode through one rule) re-decodes the stream against the source running now, before anything reaches the processor. Unconditional rather than "only when the shape moved", because a stream file carries ONE context for events appended across several sources, so it cannot answer that per event. Alternatives considered: conditional re-decode (defeated by that shared context) and not keeping the stream when the decode moved (fails the acceptance criterion, and gives up the case that motivates the split). Where a `logValues` projection dropped the raw log there is nothing to re-read, so the stream is CLEARED rather than replayed on trust. Touches `keepStream` for every deployment, and `ExistingStream` implementors, whose stored bytes are unchanged.

**Entries are compared as a SET DIFF, not element-wise by index.** By-index was workable while entries were per-range and ordered by `startBlock`; it is not now that an un-ranged source puts every event at the same block, since an insertion shifts whatever sorts after it and every shifted entry would read as a change to something already indexed. Consequence worth naming: an entry EDITED *above* the cursor now costs nothing where by-index it invalidated. That is correct (nothing above the cursor was indexed, so nothing derived from it is stale) but it is a behaviour change beyond the stated criteria. Touches every `abi-versions-are-block-ranged` behaviour, all of which still pass unchanged.

**`legacyHash` is persisted on the block-0 entry, forever.** It is the only way a stored whole-source digest can be compared at all, and a caller cannot forget to supply it. It is deliberately NOT part of the identity invalidation reads, and it is the one field that moves when a regenerated ABI gains a view function, so the "byte-identical entries" claim is about `{startBlock, hash, streamHash}`. Alternative considered: a fifth parameter to `sourceInvalidationOf`, rejected as a footgun whose failure mode is a silent re-index of every deployment.

**The stream half is entry-level, not coverage-level.** "The topic set did not grow" is decided by whether any current `streamHash` is absent from the stored set, so a NARROWED range reads as a new entry and invalidates the stream even though it is a subset. Reasoning about coverage is unreachable, not undesirable: a stored entry is an opaque digest, so a subset relation cannot be recognised. Conservative, therefore safe, and strictly better than the previous behaviour.

**`wireContextOf` is left on the whole-source hash.** It is an identity check between the two halves of a split deployment computing the same thing from the same declarations (ADR-0004), not a question about what a stored context covers. The server-side `StreamBuilder` therefore still gets the all-or-nothing answer, through the same code path via the both-sides-legacy branch. Touches `agnostic-log-fetcher` / `ingest-wire-receiving-side`: bringing the server split onto per-event invalidation is a separate, wire-format-visible change.


wighawag:

Gate-1 green on the first pass. Gate-2 off by repo config. Ships ADR-0034; no numbering collision.

**Every criterion met**, including the two that could have been faked and were not:

- **Back-compat is proved, not asserted.** `agePersistedContextsToTheShippedShape` rewrites the stored context in BOTH keepers into the shipped whole-source shape, first asserting the aged shape actually differs from what the new code writes, then reopens with the upgraded library and asserts it resumes without re-fetching. That is the trap the task named: an upgrade misreading its predecessor's bytes would silently re-index every existing deployment once, charging exactly the cost this change removes, at the moment nobody is looking.
- **The stream/state split is asserted as a PAIR** on the renamed-parameter case: no range re-fetched AND `{stateDiscarded: true}`, plus a test that the rebuild speaks the new parameter names. That is the case that could not be expressed before.

**Two things it got right that the task did not anticipate, and both are load-bearing**

1. **By-index comparison had to be abandoned.** It worked while every entry was per-range and ordered by `startBlock`, so an append landed at the end. It does not survive generalisation, because an un-ranged source puts every event at the same block, so inserting one shifts whatever sorts after it and every shifted entry reads as a change to something already indexed. Replaced with a set diff on digests, which asks the question that was always meant (which entries are new, which are gone) and is immune to position. The side effect is correct too: an entry edited ABOVE the cursor now costs nothing, because nothing above the cursor was ever indexed.

2. **A kept stream must be re-decoded, unconditionally.** A cached stream is two things: the raw log, which is what the node said and is true forever, and `args`/`eventName`, which is what SOME ABI made of it. Only the first survives a source change. Keeping the stream while discarding the state would otherwise rebuild from stale decodings. `reparse` runs before any event reaches the processor, and it is unconditional rather than conditioned on the decoding shape moving, because a stream file carries one context for events appended across several sources and cannot answer "was this decoded under the current ABI" per event. Where a `logValues` projection dropped the raw log there is nothing to re-read, so the stream is CLEARED rather than replayed on trust. That is a soundness hole the task did not mention and the review would have had to catch.

**Honest about what it did not achieve.** Deciding the stream half by topic COVERAGE — "is every (address, topic0, range) I now need inside what was already fetched" — is recorded as rejected because it is UNREACHABLE, not undesirable: a stored entry is an opaque digest, so a narrowed range cannot be recognised as a subset and reads as new. Conservative, safe, and still strictly better than invalidating on any difference. I raised the subset framing earlier in design and this correctly identifies why the general form is not available with digests.

**Scope discipline.** `legacyHash` is explicitly excluded from identity, so the byte-identity claims are about `{startBlock, hash, streamHash}` and the ADR says so rather than quietly meaning something narrower. The entry list is SORTED rather than transcribed, which is what makes a reordered ABI produce identical bytes instead of merely an identical verdict. ADR-0033's now-false consequence is marked superseded rather than left standing.

Merging.

PR #29, opened 2026-08-29
feat(the-published-type-scanner-reads-code-not-prose): The published-ty…

wighawag:
Task: `work/tasks/done/the-published-type-scanner-reads-code-not-prose.md`

Done. The gate is green (`format:check`, `build`, `typecheck`, full `pnpm test` across the workspace).

## What I did

**Reproduced it first, exactly as specified.** Put a JSDoc comment carrying `from "we never asked"` on a type emitted into `packages/core/dist/types.d.ts`, built, and got:

```
core/dist/types.d.ts imports 'we never asked', which is not declared at all
```

**Failing test before the code.** Added a `describe('the scan reads declarations, not prose')` block to `packages/core/test/publishedTypeDependencies.test.ts` pinning the reading rule: the exact regression shape (a comment carrying `from "we never asked"`), an `@example` block containing a real-looking `import ... from 'left-pad'`, a line comment, an import specifier inside a string literal, the genuine forms that must still be found, and the relative/built-in skips. 5 of the 7 were red against the old regex.

**The fix: it parses now.** `importedPackages` builds a `ts.SourceFile` and reads module specifiers out of the four positions a `.d.ts` can actually name a module in (`import`/`export ... from`, `import('x')` types, `import x = require('x')`, a dynamic `import()` call). Those are positions no comment and no string literal can occupy. TypeScript was already a devDependency here, so this cost one `createSourceFile` per file.

**The claim survives.** I appended a genuine `import type {Whatever} from 'we-never-asked'` to the emitted `dist/types.d.ts` and the suite still fails with the identical message. I also added a guard test asserting the scan of core's real emitted declarations finds something, because the failure mode of a parser-based scan is the silent one (a parse that yields nothing makes every package look clean).

**Restored the reworded prose.** `RangedAbiEvent`'s doc in `packages/core/src/types.ts` had been softened to "nothing distinguishes a chain that had no such log from a request that was never made". It now reads `nothing distinguishes "the chain had none" from "we never asked"` again, the phrase used for this failure class everywhere else (ADR-0031, ADR-0033, `eventRanges.ts`, `LogEventFetcher.ts`). That sentence is emitted verbatim into `dist/types.d.ts`, so the live gate, not just a synthetic fixture, now proves the scanner does not care.

**Changeset:** `.changeset/the-published-type-scanner-reads-code-not-prose.md`, `@etherfold/core: patch`. Required regardless (the gate runs `changeset status --since=main` and `packages/core` is touched); the only published change is that doc comment. Note it currently still reports red locally only because the changeset file is untracked, and `@changesets/git` reads added changesets via `git diff` against the merge base; it counts once the runner commits.

## What else scans emitted output by pattern-matching text

Asked for either way, so: **nothing else reads `dist/`.** This was the only gate inspecting emitted output. I checked every `test/` directory, `scripts/`, and package sources for `dist/`, `.d.ts`, `readFileSync`, `matchAll` and `toMatch(`.

But **five tests read `src/` the same way**, and I left them alone (out of scope), capturing the signal in `work/notes/observations/source-scanning-gates-still-match-text-including-comments.md`:

- `packages/state-store-sqlite/test/no-platform-leakage.test.ts` is the worst of them: `not.toMatch(/\bD1\b/)` and `not.toMatch(/cloudflare/i)` over the whole file text, so a doc comment explaining *why* this store must not name D1 would redden the gate. Same defect, same incentive.
- `packages/state-store-patch/test/stays-light.test.ts`, `packages/state-store-indexeddb/test/stays-a-primitive.test.ts` and `packages/server/test/platformAgnostic.test.ts` do the same with `\bconsole\.`, `from ['"]node:` and `\bD1Database\b`.
- `packages/state-store-patch/test/historical-reads.test.ts` slices a method body out of `src/store.ts` by string index and regexes that; a comment inside the body counts as code.
- Their `^\s*import\s+...from\s+'...'` scans are line-anchored and much safer: a JSDoc line begins ` * `, so prose does not reach them. The whole-file matchers are the exposed ones.

## Decisions

**An import specifier that is only the TEXT of a string literal is no longer reported.** The old text search counted `export type Hint = "run: import x from 'left-pad'"`; a parse cannot, and I did not add a string-literal sweep to preserve that. A string literal type is data: nothing resolves it, so a consumer lacking that package is not broken by it, and counting it is the same false-positive class as counting a comment. The alternative (parse for declarations, then additionally regex string literals) reintroduces exactly the noise this task exists to remove, and cannot tell an illustrative snippet from anything else. Acceptance criterion 3 asked for this to be recorded. Touches only this test; nothing else consumes `importedPackages`.

**The scan's COVERAGE is unchanged; only its reading changed.** A real parse could easily be widened to count `declare module 'x' {}` and triple-slash `/// <reference types="x" />`, both of which the text search never caught. I deliberately did not: widening what the gate catches is a different decision from fixing how it reads, and it can only turn packages red for reasons this task never examined (tsc emits `/// <reference types="node" />` on its own in some configurations). No dist file in the workspace currently contains either form. Recorded because a reviewer might reasonably expect "parse it properly" to imply "catch more"; it deliberately does not. Documented at the choice site in the `importedPackages` JSDoc.

### DIFF
--- a/.changeset/the-published-type-scanner-reads-code-not-prose.md
+++ b/.changeset/the-published-type-scanner-reads-code-not-prose.md
@@ -1,0 +1,13 @@
+---
+'@etherfold/core': patch
+---
+
+The published-type dependency scanner reads declarations rather than text.
+
+`packages/core/test/publishedTypeDependencies.test.ts` asserts that every package a published `.d.ts` imports from is a real dependency, which is a good claim: a type-only import is erased from the emitted `.js` but SURVIVES in the emitted `.d.ts`, so a package whose public types name `abitype` or `eip-1193` and declares neither is broken for whoever installs it. What was wrong was only HOW it read the file. It pattern-matched `from '...'` over the raw text, where a sentence is indistinguishable from a declaration, so a doc comment reading `nothing distinguishes "the chain had none" from "we never asked"` was reported as `core/dist/types.d.ts imports 'we never asked', which is not declared at all` and turned the acceptance gate red.
+
+A false positive that fails a gate is the expensive direction, and this one taught the wrong lesson: the author reworded the COMMENT to get past a test that was never about comments, in a repository whose whole documentation style is long explanatory comments. It would have recurred on every one of them.
+
+The scan now parses the declaration file with TypeScript (already a dependency here) and reads module specifiers out of the four positions a `.d.ts` can actually name a module in: `import`/`export ... from`, `import('x')` types, `import x = require('x')`, and a dynamic `import()` call. Those are positions no comment and no string literal can occupy. The claim is unchanged and a genuine undeclared import still fails with the same message.
+
+The only published change is a doc comment: `RangedAbiEvent`'s explanation of which way to err on `firstBlock` says `nothing distinguishes "the chain had none" from "we never asked"` again, which is the phrase used for this failure class everywhere else in the repo (ADR-0031, ADR-0033). Its presence in the emitted `dist/types.d.ts` is now what proves the scanner no longer cares.
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -256,10 +256,10 @@ export type UntypedWireBatch = Omit<WireBatch<Abi>, 'logs'> & {logs: unknown[]};
  *
  * - **`firstBlock` too EARLY is safe**; too LATE loses logs undetectably. The
  *   blocks between the real first occurrence and the declared one are indexed
- *   without that event in the filter, so afterwards nothing distinguishes a
- *   chain that had no such log from a request that was never made. For a proxy
- *   deployment the implementation's own deploy block is naturally safe: an
- *   implementation cannot emit before it exists.
+ *   without that event in the filter, so afterwards nothing distinguishes "the
+ *   chain had none" from "we never asked". For a proxy deployment the
+ *   implementation's own deploy block is naturally safe: an implementation
+ *   cannot emit before it exists.
  * - **`lastBlock` too LATE is safe**; too EARLY loses logs the same way, and for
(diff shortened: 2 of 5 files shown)


wighawag:
APPROVE ✅ — Gate-3

Reproduced-then-fixed, parsing rather than pattern-matching, and the prose in `packages/core/src/types.ts` is restored to what it was meant to say, which is the proof the scanner no longer cares.

Coverage is on the right axis: prose, `@example` blocks, line comments and a string literal are all ignored; every real declaration form is still found; relative specifiers and node built-ins still excluded.

**The best test here is the one nobody asked for.** `reads the real emitted declarations rather than parsing them into nothing` guards the silent failure a parser rewrite invites: a parse yielding zero imports makes every package look clean and every assertion pass. Asserting that `@etherfold/core`'s own emitted types import *something* is what stops the fix being vacuous. A validator that has never been seen to reject anything is not known to work, and this is the same discipline applied to the fix itself.

**The observation is worth more than the fix.** Asked to check whether anything else scans by text, it found four more and did not touch them: `no-platform-leakage.test.ts`, `stays-light.test.ts`, `stays-a-primitive.test.ts` and `platformAgnostic.test.ts` all match whole-file patterns like `\bD1\b`, `cloudflare`, `\bconsole\.` and `\bD1Database\b`, so a comment EXPLAINING why the store must not name D1 would redden the gate that forbids naming D1. That is a perverse incentive against documenting the very rule being enforced. Worse, `historical-reads.test.ts` slices a method body out of `src/store.ts` by string index and regexes that, so a comment inside the body counts as code.

Correctly scoped: recorded, not fixed. It is task-worthy and I will carry it forward.

Merging.

PR #30, opened 2026-08-29
feat(two-adrs-cannot-claim-the-same-number): Two ADRs cannot claim the…

wighawag:
Task: `work/tasks/done/two-adrs-cannot-claim-the-same-number.md`

The gap was real (nothing in root scripts, `dorfl.json`'s `verify`, or `.github/workflows/*` looked at `docs/adr/`; `0001`–`0034` are currently unique), so I built the check.

## What I did

**`scripts/check-adr-numbers.mjs`** (new, dependency-free, no workspace package): a pure `checkAdrFilenames(filenames)` over `docs/adr/` that reports duplicate leading numbers (naming the number and every colliding file) and malformed filenames, with a header explaining the 2026-08-28 incident it exists for. Accepts an optional directory argument so a failing case can be demonstrated without touching the real directory.

**Wired into the gate**: `"check:adr": "node scripts/check-adr-numbers.mjs"` in the root `package.json`, called from `dorfl.json`'s `verify` (right after `format:check`, before `changeset status`) and added as a CI step in `.github/workflows/ci.yml`. One line in `CONTEXT.md`'s `## Conventions` so the next ADR author knows the rule is enforced and that gaps are deliberately not checked.

**Failing case demonstrated, three ways.** Test-first: the self-check harness was written and run before the detector existed (`ReferenceError`, red). Then (a) a throwaway directory reproducing the real incident (both 0032 files plus an unnumbered file) is rejected with both names and exit 1; (b) a duplicate `0032-*.md` dropped into the real `docs/adr/` turns the gate red at the `check:adr` link (exit 1) and was then removed; (c) a mutated copy of the script with the duplicate test broken is caught by its own self-check ("the ADR filename check is BROKEN") instead of silently passing.

**Gate green** on the unmodified tree: `format:check`, `check:adr` (34 ADRs, no duplicates), `changeset status --since=main` (exit 0, no packages to bump: the changeset-free claim is confirmed, not assumed), `build`, `typecheck`, `test` all pass. Working tree holds only the four modified files plus the new script.

## Decisions

- **A malformed ADR filename is CAUGHT, not out of scope.** A `.md` in `docs/adr/` that is not `NNNN-slug.md` fails the check. Rationale: it is the same failure as a duplicate with the number missing rather than doubled, since the document cannot be cited as ADR-NNNN; and the shape is already specified by `work/protocol/ADR-FORMAT.md`, so this enforces an existing convention rather than inventing one. Alternative considered: duplicates only, which would leave `adr-about-x.md` invisible forever. What it touches: it is a NEW refusal that anyone adding a file to `docs/adr/` now hits, so it is user-visible beyond this task. It is cheap to reverse (delete one branch in the script).
- **`README.md` and non-`.md` files in `docs/adr/` are exempt.** Follows from the refusal above: without the exemption, adding a conventional index or a diagram to `docs/adr/` would fail the gate, which is a rule nobody agreed to. Alternative: an empty allowlist, rejected as a trap for a future, unrelated task.
- **A GAP in the sequence is NOT checked.** A hole is legitimate (a withdrawn ADR, a number reserved on a branch that has not merged) and, unlike a duplicate, makes no citation ambiguous: nothing points at a number that does not exist. Recorded in the script header and in `CONTEXT.md` so the omission reads as a decision rather than an oversight.
- **The failing case lives in the script as an always-on self-check, not as a vitest test.** The repo's test convention is vitest inside a package's `test/`, and this check must be root-level (no package, so no changeset) and must run before anything is installed or built. So the rejecting cases are carried as data and asserted on every invocation, including in the gate. Alternatives considered: a root vitest config (adds a root test surface and a runner dependency for one file) or putting the script in a workspace package (would require a changeset and contradicts the task's sizing). Trade-off: the self-check costs milliseconds on every gate run and prints one line of output.
- **New root script named `check:adr`, and the gate wiring goes in `dorfl.json` + CI rather than inside an existing script.** `check` in this repo already means "verify without writing" (`format:check`), so the name extends the existing vocabulary rather than re-meaning it; folding the ADR check into `format:check` would have made that script mean two things. What it touches: `dorfl.json`'s `verify` string and one new CI step, both root-level.


wighawag:
APPROVE ✅ — Gate-3

Runs where the mistake happens: `pnpm check:adr` is in `dorfl.json`'s `verify` AND in CI, so a colliding ADR cannot merge green. Root-level script, no workspace package touched, therefore no changeset, and the report says so explicitly rather than leaving it implied.

**The failing case is demonstrated, which was the criterion most likely to be skipped.** Rather than a fixture directory, the rejecting cases are carried as data and asserted on every invocation, including every gate run. The reasoning is recorded: the repo's test convention is vitest inside a package's `test/`, but this must be root-level and must run before anything is installed or built, so a root vitest config would add a test surface and a runner dependency for one file, and moving the script into a package would force a changeset and contradict the sizing. A self-check costing milliseconds is the right trade, and a validator never observed to reject anything is not known to work.

A gap in the numbering is explicitly allowed and documented as such, rather than left ambiguous.

Merging. This closes the hole that let me put a second ADR-0032 on main while a branch already had one.

PR #31, opened 2026-09-01
feat(the-indexer-and-its-stream-cache-agree-on-who-is-ahead): The index…

wighawag:
Task: `work/tasks/done/the-indexer-and-its-stream-cache-agree-on-who-is-ahead.md`

The work is complete and the gate is green. Final summary:

## What I did

**`packages/core/src/indexer.ts`** — three engine changes plus the deletions they enable:

1. **The stream is written BEFORE `process`, and a batch that was not written is not processed.** `promiseToSave` now REPORTS (`StreamWriteOutcome`) instead of swallowing; `promiseToIndex` consults it and, on a failed write, returns without processing and without moving the cursor (paced by `wait`). A failed write is retried by the next cycle re-deriving the same delta; after `streamWriteRetry.maxConsecutiveFailures` the cache is FROZEN (loudly logged) and indexing carries on without it. Out of space is the one cause that CLEARS (`isOutOfSpace`, new in `errors.ts`, read structurally like `retryable`).
2. **The state-KEPT load branch now FEEDS a stream that is ahead**, re-decoded, with the requested `fromBlock` assigned onto the fetched cursor and `this.lastSync` set *before* the block.
3. **A present-but-empty stream's cursor is ADOPTED** (current context, empty window) instead of being reached only via `feed`.

Deleted: `streamNotYetSaved` (the carry-forward that never fired) and, with its only caller gone, `createAction`'s `setContext`/`getContext`. Replaced by the inverse: `streamLastToBlock` (what is on disk) and `streamWrittenNotProcessed` (written-but-unaccepted batch).

**Tests** (all fail against HEAD's engine — verified by restoring `indexer.ts` from HEAD and re-running: 21/21 red): `packages/core/test/streamCache.test.ts`, `packages/core/test/streamLoad.test.ts`, `packages/core/test/utils/streamCacheWorld.ts`, `packages/browser/test/streamCache.test.ts` (real `keepStreamOnIndexedDB` + `keepStateOnIndexedDB` on `fake-indexeddb`, the workload's fake chain, range counting).

**Also:** `docs/adr/0038-…`, a changeset for `@etherfold/core` + `@etherfold/browser`, and one amended sentence in `CONTEXT.md`'s **hole** entry. `pnpm build && pnpm typecheck && pnpm test`, `format:check` and `check:adr` all green; tree clean.

## Decisions

**The keeper's forward-jump guard is NOT sufficient, so the ENGINE decides whether the cache may be written to — and a frozen stream is never appended to again.** The task's premise ("the guard is already the arbiter; the guard makes a wrong attempt harmless") is false, twice over. First, that guard does not exist at HEAD (it belongs to `the-stream-appends-in-segments-on-indexeddb`, which is blocked on this task), so with the shipped blob keeper nothing would refuse anything. Second, and more importantly, the guard *could not* refuse the dangerous case even once it lands: it compares the batch's `lastFromBlock` with the stored cursor, but a batch is a DELTA against the STATE's unconfirmed window while the fetch RANGE reaches back `finality` blocks every cycle — so after a freeze an append routinely *overlaps* the stored cursor (accepted) while the events it carries begin above the blocks the frozen stream never got. That is a hole the keeper cannot see. The engine holds both cursors, so `streamCanReceive()` asks the real question: the stream may be AHEAD of the state or level with it, never behind while still being written to. Both halves of the acceptance criterion fall out of that one comparison: an outage shorter than the bound heals (nothing was processed, the cursors never diverged), and one that hit the bound stops writing for good. **Alternatives considered:** implementing the guard in the blob keeper (rejected: it is the blocked task's, it would be rewritten wholesale, and it would still be insufficient); asserting the refusal only against a test-double keeper (rejected: leaves the shipped browser path holed in a real window); re-fetching the gap so the cache catches up (rejected: a second fetch range with its own cursor is a feature, not a guard, and is out of scope). **What it touches:** `the-stream-appends-in-segments-on-indexeddb` — its forward-jump guard is still wanted as the durable backstop (a second tab, a reload, a non-engine caller) and its acceptance criterion stands; it is no longer the thing that makes the freeze safe. Recorded as **ADR-0038** and as an amendment to `CONTEXT.md`'s **hole** entry.

**The bound is 3 consecutive failures, the pace is 1 second, both configurable at `ProvidedIndexerConfig.streamWriteRetry`.** Three because the browser auto-index ticks every 4s and a transient IndexedDB refusal should survive a few seconds without the app freezing its data for longer than a user notices; one second because that is already this repo's retry pace on both drivers (the CLI's `retryDelaySeconds`, the browser's `indexToLatest` catch). Deliberately NOT under `config.stream`: that object is HASHED into the stream and wire identity, so tuning a retry would have invalidated the cache it protects. **Touches:** the CLI's and browser's tip loops (they now see a cycle that resolves without progress, which the bound guarantees terminates).

**A failed cycle returns the observed `latestBlock` as a value while leaving the in-memory cursor untouched.** A driver looping on `lastToBlock !== latestBlock` would otherwise conclude it was done on the very first cycle of a fresh index (the fresh cursor says `latestBlock: 0`) and the CLI would have written an empty snapshot after a single cache hiccup. It is returned rather than assigned because `getFromBlock` reads `latestBlock === 0` as "nothing indexed yet"; writing it back would drag the next fetch below the source's start block.

**A failed write logs at ERROR, naming the count and the bound, and the freeze logs once at ERROR.** Every failure: `could not save the stream (N consecutive failure(s) of L): <err>. The batch is NOT processed and the cursor does not move…`. At the bound, once: `the cached stream is FROZEN after N consecutive failed writes. Indexing carries on WITHOUT the cache…`. Recovery logs at INFO; the decline logs once at ERROR rather than once per cycle. All through `named-logs`, never `console`.

**The feed in the state-kept branch DOES re-decode** (`logEventFetcher.reparse`), like the discarded branch. The stored `args` are what some earlier ABI made of those bytes (ADR-0034), and one decoding rule for a replayed stream is better than two even where `stateMatches` implies the decode shape did not move; it also keeps the dropped-raw-half refusal on both branches.

**Death between the two writes is simulated by a handler that throws deterministically, then dropping the indexer and keeping the stores.** With the write order flipped, a `process` that never returns is exactly a tab that went away after the stream write: the processor persists its own state inside `process()`, so no cursor is written. No mock of death, no fake crash — the browser test then builds a second `createIndexerState` over the same two IndexedDB keepers.

**A written-but-unaccepted batch the chain then reorged under is RETRACTED into the stream** (a scope call-out, one helper: `streamRemainderOf`). This path is *created* by change 1 and is not in the task text: events are on disk that the processor never applied, so the state cannot retract them (its window never held them) and a replay would apply a dead branch. Rather than clear (which the task forbids: a transient processor failure must not destroy a large cache) or leave them (silent wrong state), the divergence beyond the common prefix is re-appended flagged `removed`, which is precisely what an emission stream is for.

**`createAction`'s `C` type parameter and `setContext`/`getContext` are deleted**, not just their use. They existed only for `streamNotYetSaved`; the module is internal (not exported from `index.ts`) and had no other caller or test.


wighawag:

Reviewed the diff against every acceptance criterion of `work/tasks/done/the-indexer-and-its-stream-cache-agree-on-who-is-ahead.md`:

- **Stream written before `process`** ✓ — save precedes `processor.process` in `promiseToIndex`, asserted at instrumented stubs (`streamCache.test.ts`), not read off the source.
- **Transient failure** ✓ — `StreamWriteOutcome` reports instead of swallowing; `failed` ⇒ no process, no cursor move, paced retry; next cycle re-derives the same delta; batch held exactly once; subtree never cleared.
- **Permanent failure** ✓ — bounded at 3 consecutive failures (paced 1s), loudly logged, indexing resumes, prefix survives as a contiguous seed; reload replays it and re-fetches only the remainder, landing on from-scratch-equal state.
- **Frozen revive/refuse** ✓ — short outage accepted while cursors never diverged; past the bound the append is refused and the prefix intact.
- **No-space clears** ✓ — `isOutOfSpace` reads the flag structurally (+ standard `QuotaExceededError`), clears; every other persistent failure freezes.
- **Retraction never for an unheld event** ✓ — with the cursor held back, the window cannot advance past the stream; tested on the producing path.
- **No hot spin** ✓ — retry paced; tip-loop driver terminates.
- **Throwing processor** ✓ — `streamWrittenNotProcessed` high-water mark (the honest inversion of the deleted `streamNotYetSaved`) skips a repeated save; one copy, no clear; replay equals from-scratch.
- **Death between writes** ✓ — browser test drops the indexer, keeps the stores, reloads; nothing re-fetched, no duplicate.
- **Stream ahead of state is replayed** ✓ — state-kept branch now feeds (re-decoded), with both the `lastFromBlock` assignment and the early `this.lastSync` assignment the task warned about; range counting shows the node asked for nothing.
- **Cursor with no segments resumes and keeps advancing** ✓ — adopted directly with the current context and an empty window.
- **Reorg / overlap / retraction batching / `checkTxInclusion`** ✓ — all explicitly tested; overlap stays legal; window stays on the state side.

**The one premise correction is sound and verified:** the task prose claimed the keeper's forward-jump guard already exists as the arbiter — it does not at HEAD (the shipped keeper only filters on read), and the guard could not see the delta-vs-range distinction anyway. The engine-side `streamCanReceive` arbiter + ADR-0038 is the right placement, keeps the blocked task's guard as the durable backstop, and the CONTEXT.md "hole" entry is amended to match. Deleted `streamNotYetSaved` + the dead `createAction` context API rather than building around them — in the task's spirit of deleting what stops earning its keep.

Scope fence respected: no storage-layout changes (browser package touched only in tests). Gate green (build/typecheck/test/format/adr/changeset).

PR #32, opened 2026-09-01
feat(index-to-a-store-from-the-cli): `etherfold index` runs an entity p…

wighawag:
Task: `work/tasks/done/index-to-a-store-from-the-cli.md`

The build is complete and the full gate is green. Summary of what landed:

**`packages/cli`** — `etherfold index` now assembles the two ADR-0003 halves with the transport removed (`LogFetcher` → `createDirectIngestion` → `StreamBuilder` → processor), driven by `runFetcherLoop` plus an `AbortController`. `EthereumIndexer` is gone from the CLI (its `init`/`indexToTip` deleted, replaced by `prepareIndexing`/`run`). New `src/options.ts` owns `--store` (required, `file`|`sqlite`), `--db`, `--folder`, `--retention` and every refusal.

**`packages/utils`** — `instantiateProcessorWithKind` / `ResolvedProcessor` / `ProcessorKind`: the kind is read off the module's `createProcessor` return in the browser's own `{kind, processor}` vocabulary; untagged still means `'js-object'`. `instantiateProcessor` unwraps a `'js-object'` tag and refuses an `'entities'` module.

**`packages/core`** — exports `resolveStreamConfig`, so the store's retention floor is checked against the finality the stream actually runs with.

**`examples/event-processor-nfts`** — `src/entities.ts` is byte-identical (it is not in `git status`); a new `src/cli.ts` imports it, tags it and names what to index. README section + one documented command + `test/cli.test.ts`, which drives that exact file through the CLI into libSQL (ownership listing, transfer counter, and the ERC-20 collision counted by `handleUnparsedEvent`).

Tests: 48 in `packages/cli` (entity indexing, backend equivalence, reorg with a counter that goes 4 → 3, stop-and-resume from the store cursor, exit codes, retention + no automatic prune, refusals before any RPC, an engine check that no `EthereumIndexer` is imported or constructed), 30 in `@etherfold/utils`, 1 in the example. `pnpm format:check && pnpm check:adr && pnpm build && pnpm typecheck && pnpm test` all pass. `pnpm changeset status --since=main` currently reports "no changesets were found" only because the changeset file is still UNTRACKED (changesets reads it via `git diff`); it disappears once the runner commits, and `pnpm changeset status` (no `--since`) shows the correct plan.

Also written: `docs/adr/0039-…`, one changeset, and `work/notes/observations/contracts-data-only-modules-cannot-resolve-a-source.md` (a latent `resolveSource` bug I noticed and did not fix).

## Decisions

**1. The module declares its KIND by returning a tag: `createProcessor()` → `{kind: 'entities', processor}`; untagged means `'js-object'`.** Same two words and same shape `@etherfold/browser` takes (`ProcessorKind` / `TaggedProcessor`), read as a discriminant (`'kind' in value`) rather than sniffed, so a wrapper or proxy cannot make the guess wrong in silence. What the tag CARRIES differs on purpose: the browser is handed a store-bound runtime, a module cannot be (the store is the deployment's choice), so the `'entities'` arm carries the AUTHORING `EntityProcessor` and the CLI builds `new EntityEventProcessor(store, processor)` around it. Alternatives considered and rejected: a separate named export (reads as a tag, behaves as a sniff, a module carrying both says two things), a `processorKind` export beside `createProcessor` (two exports that can disagree), a field on the processor object (would put a host's question inside the authoring surface, and `entities.ts` had to stay unchanged). **Touches** `@etherfold/utils` (`instantiateProcessor` now refuses an `'entities'` module rather than returning a non-`EventProcessor`), `@etherfold/browser`'s vocabulary (duplicated, not imported: a module loader must not depend on a browser runtime to type an export), and story 2 of `one-command-runs-the-whole-pipeline`, which was explicitly waiting on this. Written up as **ADR-0039**.

**2. `--folder` is REFUSED with `--store sqlite`, not ignored — and `--db` and `--retention` are refused with `--store file`.** It stops being a `requiredOption` and becomes required only with `--store file`. An accepted-and-ignored flag is a deployment believing something untrue (a retention window nothing enforces, a database nothing writes). Alternative: accept and ignore, which is what "optional" would have meant. **Touches** every existing `etherfold index -p … -f …` invocation, which now needs `--store file`; the changeset says so.

**3. `--db` is REQUIRED with `--store sqlite`, with no default — deliberately UNLIKE `serve`.** `serve` defaults to `DB` then `file:./etherfold.db` inside `startServer`; the one-shot does not, so no run ever writes a database nobody named, and it is symmetric with `--folder` (which has no default either). Alternative: mirror `serve`'s default literally, which the task's "mirroring `serve`" could be read to require; I read `--db <url>` as the flag SHAPE, not the defaulting. **Touches** `serve` (unchanged) and `run`/`build`, which will have to pick one of the two conventions.

**4. Stopping at the tip: abort from `onReport` on the FIRST `progress` with `caughtUp`, or on `idle`.** A `fatal` ends the loop by itself and is re-thrown so the process exits 1. `contended` (yielded) deliberately does NOT stop: it means another sender moved the cursor, and stopping would report success having landed nothing. **Retryable failures retry indefinitely** on the host's backoff curve rather than after N attempts: bounding them would invent a policy `run` must not have, and core already applies a bounded in-cycle `RetryPolicy`. Note the practical effect is smaller than it sounds: an unreachable node fails fast during source resolution (verified: exit 1), so the unbounded path only opens after the first successful chain call. If a reviewer wants a bound, it belongs beside `build`, not here. **Touches** `build`, which the spec says reuses this driver.

**5. `resolveStreamConfig` is now public on `@etherfold/core`.** The entity store's retention window is validated against a finality depth, and the only correct number is the one the stream resolves to; restating `17` in the CLI would have pinned a value that is hashed into the wire identity and silently forked the config hash the day the default moved. Alternative: a `--finality` flag (a second opinion about one number). No `--finality` flag exists, so the CLI is always on the stream default. **Touches** `@etherfold/core`'s published surface (changeset: minor).

**6. `--retention` takes a bare number as BLOCKS, plus `revert-only` / `unbounded`, default `unbounded`.** A duration is refused naming the unit (ADR-0019). Nothing prunes inside the index loop and there is no `--prune` verb: pruning is the host's scheduled call (ADR-0022), and a prune verb is a different command's business.

**7. The root `pnpm test` filter now includes `./examples/*`.** The example is EVIDENCE for this task ("the same processor, unchanged, under the CLI"), and evidence nothing runs is a claim. Packages with no `test` script are skipped (verified: the filter exits 0 today). Alternative: put the example test in `packages/cli` importing across the workspace, which breaks `pnpm typecheck` (TS6059, `rootDir`). **Touches** the acceptance gate's scope: any future example test now runs in it.

**8. `eip-1193` moved from devDependency to dependency of `etherfold`.** `IndexingDependencies.provider` puts `EIP1193ProviderWithoutEvents` in the published `.d.ts`; core's own `publishedTypeDependencies` test caught it, and `@etherfold/core` / `@etherfold/fetcher-host` already declare it the same way.

**9. The example's CLI entry indexes ONE collection, from `NFT_CONTRACT` / `NFT_START_BLOCK`.** The browser app indexes one ACCOUNT across every address using a topic filter it passes as stream config, which a command line cannot express; "every ERC-721 transfer on mainnet" is not a demo. This also forced `"types": ["node"]` into the example's build tsconfig, since it is no longer browser-only.


wighawag:

Reviewed against every acceptance criterion of `work/tasks/done/index-to-a-store-from-the-cli.md`:

- **Entity path wired** ✓ — `--store sqlite --db` assembles `LogFetcher` → `createDirectIngestion` → `StreamBuilder` → `EntityEventProcessor` over a `VersionedStateStore`; rows equal the same processor's output on another backend.
- **The engine is asserted, not claimed** ✓ — `engine.test.ts` reads the package's own sources: no `EthereumIndexer` construction, no import, and the halves present.
- **Exit codes** ✓ — 0 at the tip (abort on first `progress`+`caughtUp` or `idle`), 1 on `fatal`, both asserted.
- **`--store` required, both arms on the same engine** ✓ — `options.ts` refuses each arm's foreign flags rather than ignoring them.
- **Kind from the module** ✓ — `instantiateProcessorWithKind` reads the `{kind, processor}` discriminant; untagged = `'js-object'` (ADR-0039); 30 utils tests.
- **Mismatch refused before any RPC** ✓ — asserted with `chain.calls` empty, matching the old keeper-less ordering.
- **`--folder` conditional** ✓ — required with `file`, refused with `sqlite`.
- **Stop and resume** ✓ — continues from the store cursor; lands where an uninterrupted run lands.
- **Reorg with a decreasing counter** ✓ — 4 → 3 through the CLI path.
- **Retention settable, nothing auto-prunes** ✓ — `prune` spy never called (ADR-0022 honoured).
- **`entities.ts` runs unchanged** ✓ — zero lines in the diff; example CLI entry, one documented command, README section, and a test driving that exact file through the CLI into libSQL.
- **Changeset** ✓ — `etherfold` / `@etherfold/utils` / `@etherfold/core`, with the breaking flag change named (cheap, nothing published).

Two non-blocking notes for the record:
1. The changeset prose "every existing CLI invocation keeps working unchanged" (about module resolution) sits one paragraph above the breaking `--store file` requirement — two true statements about different layers, but a reader could stumble. Not worth blocking.
2. The `resolveSource` `contractsData`-only bug was correctly captured as an observation (`work/notes/observations/contracts-data-only-modules-cannot-resolve-a-source.md`) rather than fixed in-place — triaged in the end-of-drive pass.

PR #33, opened 2026-09-01
feat(a-snapshot-a-client-cannot-read-is-refused-not-installed): A publi…

wighawag:
Task: `work/tasks/done/a-snapshot-a-client-cannot-read-is-refused-not-installed.md`

# Report

## Drift check (done first, as instructed)

- `retire-the-js-object-processor-path` is still in **backlog** and is **blockedBy this task** — the free-form keeper this hardens is alive, not on its way out. No drift.
- Both `SNAPSHOT_FORMAT` constants exist exactly as described: CLI's `keepState.ts` (= 2, blob) and `state-store/src/snapshot.ts` (= 1, entity). No third spelling existed. Premise holds.

## Prior attempt — reviewed and built on, not restarted

The branch already carried the placement half: `packages/core/src/snapshot.ts` (`BLOB_SNAPSHOT_FORMAT`, `isReadableBlobSnapshot`, `BlobSnapshotEnvelope`), exported from core, with the CLI's `keepState.ts` migrated to it. I kept that work and fixed what it left broken: the envelope type omitted `history` (the CLI writes `{format, processor, savedAt, lastSync, state, history}` and its build failed on `json.history`), and nothing had yet touched the browser reader, the entity-path rename, tests, changesets, or the ADR.

## What I did (test first at the seam, then production)

1. **Failing tests first**: new `packages/browser/test/remoteSnapshotFormat.test.ts` (8 tests, red against the old code — the format-1 payload was demonstrably installed with `"123n"` string args) plus a rewrite of the two `keepStateBigInts.test.ts` tests that had pinned the old behavior (un-enveloped remote bodies).
2. **`packages/browser/src/storage/state/OnIndexedDB.ts`**: every remote fetch now goes through one `fetchReadableSnapshot` helper that checks `isReadableBlobSnapshot` — an unreadable format is refused whole, logged with the location and both numbers, never translated. An unreadable mirror takes the path an unreachable one already had: a `{url}`-form mirror is refused **at selection** (its file is downloaded there anyway), so it cannot win on a block number this build can't read; a prefix-form mirror's payload is refused **after** it wins, and the winner fails over to the next mirror. Local state that is already ahead still returns early, readable remote or not. Module note rewritten: the "known gap" ADR-0029 recorded is closed, and the bare `lastSync` file's meaning is stated (below).
3. **Entity constant renamed** `SNAPSHOT_FORMAT` → `ENTITY_SNAPSHOT_FORMAT` (`@etherfold/state-store`), with usages updated in its test, `@etherfold/processor-entities` (import + re-export), and the conformance suite's `snapshot-bootstrap` case — so the two envelopes are distinguishable by name at a call site that can hold both.
4. **CLI tests** now import `BLOB_SNAPSHOT_FORMAT` from `@etherfold/core`; `core/src/utils/bigint.ts` prose updated to the new name.
5. **ADR-0040** written (placement, per-envelope naming, refusal ladder, unversioned head, rejected alternatives); ADR-0029's "gap left open knowingly" paragraph now points at it; one clause added to CONTEXT.md's **bootstrap** glossary entry (drift corrected on touch — the sentence about free-form selection stays true, since failover depth is unchanged).
6. **Changeset** for every package whose surface changed: `@etherfold/core` minor (new exports), `@etherfold/browser` patch (behavior), `@etherfold/state-store` / `@etherfold/processor-entities` / `etherfold` major (removed/renamed exports).

## Verification

`pnpm build` ✅ · `pnpm typecheck` ✅ · `pnpm test` ✅ (all 20 packages, exit 0 — includes `bundlesForABrowser.test.ts`, which still passes: the browser gained no dependency, only an import from the already-depended-on `@etherfold/core`) · `pnpm check:adr` ✅ · `pnpm format:check` ✅. Working tree contains only intended changes. One pre-existing inefficiency I noticed and did not touch is recorded in `work/notes/observations/url-form-mirror-selection-downloads-every-snapshot-twice.md`.

## Decisions

- **The blob format number lives in `@etherfold/core` as `BLOB_SNAPSHOT_FORMAT`** (with `isReadableBlobSnapshot` and the `BlobSnapshotEnvelope` type), per the task's steer — the writer (CLI) and every reader (CLI local, browser remote) import one number, and core is where the codec the number versions already lives. Alternatives rejected: a duplicate constant in the browser kept in step by attention (the outcome to avoid); `@etherfold/browser` depending on the CLI (unbundleable for a tab); re-exporting it from the CLI (two importable homes for one number). Touches: `@etherfold/core` and `etherfold` public surfaces, and the future `publish-etherfold-and-deprecate-old-names` task, which now publishes under a constant whose name says which envelope it versions.
- **One constant PER ENVELOPE, renamed as relocated**: state-store's `SNAPSHOT_FORMAT` (= 1) became `ENTITY_SNAPSHOT_FORMAT`. The two version different file shapes and revise independently, so merging them would let one envelope's revision falsely invalidate the other's, and `@etherfold/browser` (now importing both) is exactly the call site where a bare `SNAPSHOT_FORMAT` would be a coin toss. Touches: `@etherfold/state-store`, `@etherfold/processor-entities` (re-export), `@etherfold/state-store-conformance` (internal), all covered by changesets. Recorded as **ADR-0040**.
- **An unversioned bare `lastSync` file means SELECTION DATA ONLY**: it is read without a format check, deliberately. The one field used from it is `lastToBlock` (a plain number, identical under every encoding), nothing from it is ever installed, and the file that IS installed (the state file) carries the check — so a stale head can mis-order the mirrors but cannot smuggle an unreadable payload past them. The alternative — refusing the head — would make every mirror the CLI publishes unselectable (it writes the head bare by design), a guard placed where the damage is not. Not "assume it is current" in the strong sense: its claim is only ever used to order mirrors, never trusted as state. Touches: `keepStateOnIndexedDB`, and the publishing-snapshots design when it comes.
- **Failover depth preserved, not converged**: an unreadable winner fails over to exactly the next mirror (winner + one), as the path already did for an unreachable one — I did NOT adopt the entity path's walk-every-remaining-candidate, because converging the two free-form/entity behaviours is `retire-the-js-object-processor-path`'s business (it is blocked by this task and will delete this code path). Touches: that backlog task; CONTEXT.md's claim that the free-form keeper does not walk every candidate remains true. Recorded in ADR-0040's rejected-alternatives.
- **URL-form mirrors are refused at selection, prefix-form at payload** — a consequence of where each shape's bytes are already downloaded, not a new policy; both refusals log the location and both numbers.


wighawag:

Reviewed against every acceptance criterion of `work/tasks/done/a-snapshot-a-client-cannot-read-is-refused-not-installed.md`:

- **One constant per envelope, distinguishable at a call site** ✓ — `BLOB_SNAPSHOT_FORMAT` (+ `isReadableBlobSnapshot`, `BlobSnapshotEnvelope`) in `@etherfold/core` where writer and every reader import it; the entity path's renamed `ENTITY_SNAPSHOT_FORMAT`; no third `SNAPSHOT_FORMAT` spelling left exported; the CLI no longer exports a format constant of its own.
- **Unreadable format REFUSED, never installed, visibly** ✓ — one `fetchReadableSnapshot` gate on every remote fetch; refusal logs the location and BOTH numbers; no translation.
- **The defect test, asserting the TYPE** ✓ — `expect(typeof args.value).toBe('bigint')` against a format-1 payload that previously installed `"123n"` strings.
- **Failover exercised** ✓ — both mirror shapes: url-form refused at selection, prefix-form falls through to the next mirror.
- **Local ahead still wins** ✓ — over a readable-behind and an unreadable-however-far-ahead remote alike.
- **Unversioned bare `lastSync` file has a stated, tested meaning** ✓ — selection data only: it orders mirrors, nothing from it is installed, and it cannot smuggle an unreadable payload past selection. Recorded in ADR-0040.
- **Browser still bundles, no node-only dependency** ✓ — only an import from the already-depended-on core; `bundlesForABrowser.test.ts` green.
- **Changesets** ✓ — core/browser/state-store/processor-entities/cli, with the renames honestly marked major (nothing published yet, so cheap).

Noted and routed: the url-form mirror double-download inefficiency is correctly an observation (`work/notes/observations/url-form-mirror-selection-downloads-every-snapshot-twice.md`), not a fix-in-place. ADR-0029's knowingly-left-open paragraph now points at ADR-0040. The build also continued from the saved partial branch after the transient 429 bounce rather than discarding it — the keep+continue recovery working as intended.

One housekeeping note: the done record still carries the `needsAnswers: true` residue and the stale questions sidecar from the infra bounce; I am clearing those on main directly after this merge (protocol housekeeping, will appear in the drive report).

PR #34, opened 2026-09-01
feat(drop-filesystem-storage-and-rehome-the-fixture-loader): Drop files…

wighawag:
Task: `work/tasks/done/drop-filesystem-storage-and-rehome-the-fixture-loader.md`

Done. Here is what I did.

## Verifications before deleting (the task's drift checks)

- **`@etherfold/fs` consumers, re-grepped**: still exactly one *package-name* importer — `packages/conformance-workload-stratagems/src/fixtures.ts`, taking only `loadStreamFixture`. No new consumer appeared, so no STOP. `keepStreamOnFile` still has zero callers (its own definition only); the CLI still uses its own unrelated `keepState.ts`.
- **`@etherfold/fs-cache` importers, re-grepped at delete time: ZERO** in `packages/*`, `platforms/*`, `examples/*` (code and package.json). The only other reference in the repo is the pre-rename `archive/ethereum-indexer-server` tree, which is outside the `pnpm-workspace.yaml` globs and explicitly "not built, typechecked, tested, versioned or published" — left untouched.
- **One premise gap found and handled** (recorded in Decisions): five scripts in the *closed but re-runnable* spike `docs/spikes/sqlite-in-the-browser/` import the loader **by relative path** to `packages/fs/dist/index.js`, which a package-name grep does not see. They are not *new* consumers (they predate the task; the task's evidence greps were scoped to the workspace trees), so I rewired them rather than stopping.

## What I built (test first at the seam)

1. **`packages/conformance-workload-stratagems/test/fixture-file.test.ts`** — written and run *red* first (`Cannot find module '../src/fixture-file.js'`), adapted from the fs package's own fixture-file test: round-trip with BigInts, gzip-by-`.gz`-extension, and the refusal that names the file.
2. **`packages/conformance-workload-stratagems/src/fixture-file.ts`** — the new home of `loadStreamFixture` and `saveStreamFixture` (gzip-aware read/write, parse/serialize delegated to `@etherfold/core`), exported from the package index; `src/fixtures.ts` now imports it locally and the `@etherfold/fs` devDependency is gone.
3. **Deleted `packages/fs` and `packages/fs-cache`** entirely (source, tests, configs, CHANGELOGs), removed the two dev-session panes from `zellij.kdl`, and regenerated `pnpm-lock.yaml`.
4. **Rewired the five spike scripts** to the loader's new home and updated the prose that named `@etherfold/fs` (spike README, capture script, workload `fixtures/README.md`, package README layout table).
5. **Fixed the gate fallout the deletion forced**: the pending changeset `the-gate-does-not-assume-an-idle-machine.md` named `'@etherfold/fs': patch` and went stale exactly the way `packages/core/test/pendingChangesets.test.ts` exists to catch — that one line is removed (it still names 16 live packages).
6. **`docs/adr/0041-filesystem-storage-is-deleted-not-dormant.md`** — the durable record of the removal (see Decisions for why an ADR and not a changeset).
7. One observation note: `work/notes/observations/publish-task-deprecation-messages-point-at-deleted-names.md` (the backlog publish task's `npm deprecate` messages forward-point at the two names this task deleted).

**Gate**: `pnpm build`, `pnpm typecheck`, `pnpm test` all green (every suite, including the stratagems workload loading its fixtures through the relocated loader on all four backends), plus `pnpm format:check`, `pnpm check:adr` and `pnpm changeset status` green. Working tree contains only this task's changes.

## Decisions

- **Where `loadStreamFixture` went, and why.** It moved to `@etherfold/conformance-workload-stratagems` as `src/fixture-file.ts` (exported from `src/index.ts`), the task's stated obvious choice and the only consumer. It is deliberately NOT in `@etherfold/core`: core names no runtime and must not grow a `node:fs`/`node:zlib` dependency for a test-fixture loader. Alternative considered: core (rejected for exactly that reason); the CLI (rejected — the loader is test material, not a CLI capability).
- **`saveStreamFixture` moved with it, though the task named only `loadStreamFixture`.** The two halves share one file convention (gzip chosen by the `.gz` extension, indented by default), and `saveStreamFixture` has a real consumer: the spike's re-capture script, which is the documented way the committed fixtures are re-captured. Deleting it would have orphaned that path; keeping it in a deleted package was not an option. Alternative considered: deleting it and inlining ~12 lines into the capture script (rejected — duplicates the convention the fixtures' READMEs document).
- **The changeset criterion could not be satisfied as written; the removal is recorded as ADR-0041 instead.** The task asks to "ship a changeset covering the removal of BOTH published package names", but the repo forbids that form: `packages/core/test/pendingChangesets.test.ts` (landed 2026-08-23, before this task was promoted 2026-09-01) fails any pending changeset naming a package not in the workspace, and I verified empirically that a changeset naming `@etherfold/fs`/`@etherfold/fs-cache` makes both `pnpm test` and plain `pnpm changeset status` fail ("Found changeset … which is not in the workspace"). ADR-0014 already records that "changesets cannot express … the removal". Naming the only package that actually changed (`@etherfold/conformance-workload-stratagems`, `private` with `privatePackages: false`) would produce a changeset `changeset version` never consumes — a zombie of exactly the accumulated-stale-changesets shape that test and ADR-0014 exist to prevent — so I shipped no changeset and wrote the record where the repo puts removals: an ADR. The release-notes need is also vacuous: neither name was ever published, so no release ever ships them and no `npm deprecate` is owed. **This deliberately leaves acceptance criterion "Ship a changeset…" unmet; if a changeset-shaped artifact is nonetheless required, that is a human call** (the alternatives are weakening the gate test or flipping `privatePackages`, neither of which I would do unilaterally). Touches: the publish/release path and the gate.
- **The stale `'@etherfold/fs': patch` line was removed from the pending changeset `the-gate-does-not-assume-an-idle-machine.md`.** Forced, not chosen: with the package deleted, that pending release intent cannot be cut and `pnpm test` is red until the line goes (the exact staleness the pendingChangesets test documents). The changeset still names the 16 packages that genuinely gained the timeout config. Touches: that pending changeset only.
- **The five `docs/spikes/sqlite-in-the-browser/` scripts were rewired rather than left broken.** Four `run/*.ts` and `capture/capture-stratagems-base.mjs` imported the loader from `packages/fs/dist/index.js` by path — invisible to the task's name-based grep and not *new*, but they are a durable, re-runnable artifact and the capture script is the documented re-capture path for the workload's fixtures. They now import `packages/conformance-workload-stratagems/src/fixture-file.js` (the same tsx `.js`-specifier-onto-`.ts` pattern the spike already uses for the vendored oracle — verified working by loading the real base fixture through the new path). Consequence: the capture script now runs under `npx tsx` instead of plain `node` (its header, the spike's `package.json` and README updated). Alternative considered: leaving them broken with an observation note (rejected — a documented command silently failing is worse than a two-line rewiring).
- **Left alone, deliberately**: `archive/ethereum-indexer-server` (imports `@etherfold/fs-cache`; outside the workspace, not built or published); `docs/spikes/promotion-cost-of-a-two-label-stream/src/port-fs.ts` (a closed spike's self-contained exhibit whose comment "mirrors `packages/fs/src/utils/fs.ts`" is now historical); `CONTEXT.md` (its KeepState entry's "fs storage adapters" remains true — the CLI's own file keeper, `packages/cli/src/keepState.ts`, is untouched); ADRs and CHANGELOGs that mention the deleted names (historical records); `docs/reviews/todo-triage.md` (a dated review).


wighawag:

Reviewed against every acceptance criterion of `work/tasks/done/drop-filesystem-storage-and-rehome-the-fixture-loader.md`:

- **`loadStreamFixture` rehomed without a storage package** ✓ — `conformance-workload-stratagems/src/fixture-file.ts`, exported from the index, gzip-aware, `@etherfold/core` kept free of `node:fs` (the reason recorded in Decisions). `saveStreamFixture` moved with it so the documented re-capture path survives — sound.
- **`packages/fs` deleted, nothing imports it** ✓ — verified on the branch: the only remaining `@etherfold/fs` mentions in workspace trees are CHANGELOG history and provenance prose; zero imports.
- **`packages/fs-cache` deleted after re-confirming zero importers** ✓ — re-grepped at delete time (code + package.json); the `archive/` tree is outside the workspace globs and correctly left untouched.
- **The stratagems workload still loads fixtures and runs** ✓ — gate green, all four backends, plus the moved fixture-file test run red-first.
- **Bonus premise gap handled well**: the five spike scripts importing the loader by relative path (invisible to a name grep) were rewired rather than left broken or treated as a false STOP — the right call for a documented, re-runnable artifact.
- **The stale `'@etherfold/fs': patch` line in the pending changeset** — forced removal, exactly the staleness `pendingChangesets.test.ts` exists to catch.

**One criterion deliberately unmet, and I concur with the resolution while flagging it for the human:** "ship a changeset covering the removal of both names" is unmeetable as written — the repo's own gate fails any changeset naming a non-workspace package (verified empirically; ADR-0014 already records that changesets cannot express removals), so it contradicts the "gate green" criterion. The removal is recorded in ADR-0041 instead, and since neither name was ever published, no release or deprecation is owed. This is carried into the end-of-drive batch as a confirmation question, not silently accepted.

PR #35, opened 2026-09-02
feat(the-stream-appends-in-segments-on-indexeddb): The stream appends i…

wighawag:


wighawag:

Reviewed the full diff against the task's acceptance criteria. Gate-1 (acceptance) is green (`verify` SUCCESS); this is the judgement pass on top.

## Criteria

- **Core helper + three keeper operations, re-exported, changesets** — `createSegmentedStream` in `packages/core/src/stream/segments.ts`, exported from `core/src/index.ts`, changeset covers `@etherfold/core` + `@etherfold/browser`. The port is five operations, not three: the three cursor operations named in ADR-0035 are all present and every cursor move routes through them, with a scoped read and scoped delete added because a full ordered scan and a subtree delete need them on any substrate. Reads as the criterion's "injected port scoped to a stream's subtree" rather than a widening of it.
- **Array key, digest level present, `chainId` not a level** — `['stream', name, digest, ordinal|'cursor']`, placeholder `chain-${chainId}`, i.e. DERIVED from `chainId` and not the bare constant the task explicitly rejects. Test asserts the shape and that `chainId` is not a level of its own.
- **Key-range reads, not a whole-store scan** — both ranges built correctly, including the distinction that is easy to get wrong: `bound([...p,0],[...p,'cursor'],false,true)` for segments-only vs `bound([...p,0],[...p,[]])` for the subtree. Asserted by count at the instrumented seam.
- **Two chains / two names isolated** — covered, including the prefix hazard (`stream_tag_1` vs `stream_tag_10`) that the flat namespace used to carry.
- **Exactly two puts per non-empty save, nothing rewritten** — asserted at the instrumented object store, not by mocking `set`/`setMany` (which the task notes would pass vacuously).
- **One `readwrite` transaction; two keepers cannot lose a batch** — raw transaction reading the cursor and writing both records; the allocate-decision-function design is what keeps the read inside the commit. Interleaved two-keeper test asserts no ordinal written twice and no batch dropped.
- **Empty save writes only the cursor; no `unconfirmedBlocks`; `[]` returned** — all present.
- **State side still stores the window (the GUARD)** — `streamCache.test.ts:226` answers `checkTxInclusion` across a reload with NO stream keeper configured. This is the criterion that fails loudly if someone reads the window removal as global, and it is real rather than nominal.
- **Cursor from the cursor record; `clear` scoped; legacy blob DELETED and logged** — legacy detection is in `fetchFrom` as well as `clear`, which is the part that matters given the state-kept branch guards its clear behind `if (existingStreamData)`.
- **Damage clears; cursor-with-no-segments does NOT** — gap, orphan segments, unparseable segment each covered; nothing raises; the empty-cursor negative asserts the scan position keeps ADVANCING across reloads.
- **`startBlock` refusal, both halves** — cleared on a rebuild, and the paired negative (a state-kept reload does NOT clear) is asserted, which is what distinguishes comparing against the requested `fromBlock` from comparing against the source minimum.
- **Forward jump REFUSED not cleared, overlap accepted, revival** — all three, with the decline logged once via `declineReported` rather than per cycle.
- **`fetchFrom` parity, defined-result-when-empty, reorg append order, `invalidation.test.ts`** — parity asserted against a whole-blob read; `invalidation.test.ts` updated deliberately to `streamAddress(tag, chainId).cursor`, still asserting what it was written to assert.

## Drift notes / fences

- `packages/core/src/indexer.ts` **untouched** — the engine-side fence held.
- `packages/fs` untouched (its deletion had already landed; the fence was moot, per the drift correction on the task body).
- No real digest computed, no canonical pointer, no per-segment block-range metadata.

## Off-path finding

`work/notes/observations/a-replayed-reorg-reapplies-the-replacement-block.md` records a genuine ENGINE bug found while building (a replayed reorged stream applies the replacement block twice), reproduced against the shipped whole-blob shape so it pre-dates this keeper. Correctly captured as an observation rather than fixed in place. Not blocking here; flagged for tasking.

Merging.

PR #36, opened 2026-09-02
feat(d1-limits-reach-the-stores-batch-bounds): D1's per-request limits…

wighawag:
Task: `work/tasks/done/d1-limits-reach-the-stores-batch-bounds.md`

Done. The gate is green (`pnpm format:check && pnpm check:adr && pnpm changeset status --since=main && pnpm build && pnpm typecheck && pnpm test`, plus `wrangler deploy --dry-run` inside the cf-worker build).

## Drift check first (both items the prompt named)

1. **D1's limits: re-fetched, UNCHANGED.** https://developers.cloudflare.com/d1/platform/limits/ (still "Last updated Apr 21, 2026") on 2026-09-02: 100 bound parameters per query, 50 / 1,000 queries per Worker invocation (Free / Paid), 100 KB per statement. So the shipped defaults stay correct and I updated the finding's `source:` line to record the dated re-verification rather than restating numbers.
2. **The store IS reachable from a host's configuration.** `ServerOptions.getIngestion` (landed with `ingest-wire-receiving-side`) is the injection point, and `platforms/nodejs-fetcher/test/deployment.ts` is a working host doing exactly the chain store → `EntityEventProcessor` → `StreamBuilder` → `getIngestion`; `packages/cli/src/index.ts:298-320` is a second host building a store from its own configuration. So no seam had to be invented, and I did not design one.

## What I built

- **`platforms/cf-worker/src/d1.ts`** (new) — the only place in shipped code that names D1's numbers: `D1_LIMITS` (per plan, with the dated source), `D1Plan`/`DEFAULT_D1_PLAN`, `resolveD1Plan(env)`, `d1BatchBounds(plan)` → the store's `BatchBounds`, `d1PruneBudget(plan)` → a `prune({maxVersions})` budget, `createD1DB(env)` and `createD1Store(db, declarations, {plan, ...})`.
- **`wrangler.toml`** — `D1_PLAN = "free"` in `[vars]` and `[env.production.vars]` (a named env does not inherit top-level vars); `src/env.ts` types it; `worker.ts` now builds its `getDB` through `createD1DB` and documents how a deployment that hosts a processor reaches the store.
- **`test/d1-limits.test.ts`** (new, written before the source) + a `RecordingSQL` in `test/utils.ts` that records `{sql, args}` at the `remote-sql` seam: the deployment states its plan; free bounds equal `DEFAULT_BATCH_BOUNDS` exactly; Paid raises statements-per-batch to 1,000 and the store *actually* batches 120 statements in one round trip against real D1 while Free splits them; a 300-block prune against real D1 deletes >100 versions with every statement at or under 100 bound parameters (and a full round hitting exactly 100), inside one invocation's 50-query budget.
- **Notes**: the finding records that the host half landed and where; the per-INVOCATION *ingest* half is deferred as a named follow-up appended to `work/notes/observations/nothing-bounds-the-size-of-an-ingest-batch.md`; one new observation about the Worker test pool's 10 s close timeout once the package has two test files.

`packages/state-store-sqlite` is untouched — `no-platform-leakage.test.ts` and the value-pinning test in `batch.test.ts` pass unweakened.

## Decisions

**1. The adapter's limits reach the store through a store FACTORY in the host, not through a new seam.** `platforms/cf-worker` gains `createD1Store(db, declarations, {plan, ...})`, which is `new VersionedStateStore(db, declarations, {bounds: d1BatchBounds(plan)})`. The store is handed a `RemoteSQL` rather than the D1 binding so one handle serves both the store and `@etherfold/server`'s `getDB`; `createD1DB(env)` mirrors `createNodeDB` in `@etherfold/platform-nodejs`, exported for the same reason. Alternatives rejected: adding a `bounds`/store option to `ServerOptions` (that is the store-into-server seam this task is fenced off from, and would give `@etherfold/server` a store dependency), and putting a `d1` preset inside `@etherfold/state-store-sqlite` (the thing the task exists to avoid). **Touches**: nothing today's deployable runs — the shipped Worker hosts no processor (it cannot load a processor module at runtime, so a deployment bundles one), so the factory is what a deployment with a processor calls, and the adapter's tests are what exercise it against real D1. If `run`'s sibling work later gives the Worker a bundled processor, it consumes this rather than restating D1's numbers.

**2. Which plan is a deployment VARIABLE, `D1_PLAN`, defaulting to `free`, and an unknown value is REFUSED.** Nothing on a request tells a Worker its plan, and the caps differ 20x, so guessing high breaks a Free deployment in production while guessing low costs a Paid one 20x the round trips. `free` is the default because an under-configured deployment should be slow rather than rejected. The refusal (rather than a quiet fallback) is a new user-visible error: a typo in a deployment variable is exactly where a silent fallback hides the mistake. `wrangler.toml` states `free` explicitly in both the top-level and the production env, so switching plans is `wrangler deploy --var D1_PLAN:paid` or a one-line config edit, never a code change. Alternative considered: inferring the plan from a binding or from an observed rejection (not knowable, and only after a failure).

**3. The bounds are D1's documented caps EXACTLY, with no scheduling reserve folded into them.** `maxStatementsPerBatch` = queries per invocation (50 / 1,000), `maxRowsPerStatement` = 100 on both plans (bound parameters per query is not per-plan), and `maxBytesPerBatch` deliberately keeps the store's own 90,000: D1's 100 KB cap is per STATEMENT while that bound is per BATCH, so deriving one from the other would be a coincidence dressed as a rule. A bound is what the backend rejects above, so the per-invocation *reserve* is expressed once, in the prune budget (4), rather than smeared into a batch bound where it would read as a limit.

**4. I addressed the per-INVOCATION budget for PRUNE and explicitly deferred it for INGEST.** `d1PruneBudget(plan)` sizes `prune({maxVersions})` from the plan's query cap: two queries per prune round, one round drops at most `maxRowsPerStatement` versions, and half the plan's queries are reserved by default for whatever else the invocation does (overridable via `reservedQueries`). "Half" is an invented, documented default and is the one arbitrary number here; the test asserts the arithmetic can never exceed the plan's cap. The INGEST half cannot be fixed by any bound the store owns (one fold issues as many batches as the work needs), so it is recorded as a named follow-up on the existing observation `nothing-bounds-the-size-of-an-ingest-batch`, which already owns the byte-shaped version of the same root cause and names its owner (the payload-building side, not a host).

**5. I did NOT lower the shipped default and did NOT narrow its docstring.** `DEFAULT_BATCH_BOUNDS` already targets D1 Free and its docstring already states, rather than papers over, both honest caveats (the per-invocation budget is not a batch quantity, and `maxBytesPerBatch` is not what satisfies a per-statement size cap). The remaining dishonesty the criterion pointed at is the per-invocation one, and the answer to it is the host lever in (4), not a smaller number. Instead the coupling is now ASSERTED from the host side: `d1BatchBounds('free')` must equal `DEFAULT_BATCH_BOUNDS`, so if either the vendor's Free column or the default moves without the other, that test goes red.

**6. No changeset.** The only package changed is `@etherfold/platform-cf-worker`, which is `private` and excluded by `.changeset/config.json` (`privatePackages: false`); no published package's surface or behaviour changed. `pnpm changeset status --since=main` exits 0.


wighawag:

Reviewed the full diff against the task's acceptance criteria. Gate-1 (acceptance) is green (`verify` SUCCESS); this is the judgement pass on top.

## Criteria

- **D1's limits are CONFIGURATION in `platforms/cf-worker`, reaching the store's `BatchBounds`, with no provider constant in shared code** — met, and met in the strongest available form: the diff touches **zero files under `packages/state-store-sqlite`** (and no other shared package). `D1_LIMITS`/`d1BatchBounds` live in `platforms/cf-worker/src/d1.ts`, the only place a runtime is named (ADR-0003).
- **`no-platform-leakage.test.ts` still passing UNWEAKENED** — the file is **untouched**. This is the criterion most at risk of being "made to pass", and it wasn't edited at all; the wiring was shaped to satisfy it rather than the reverse.
- **A Paid deployment can raise `maxStatementsPerBatch` above the Free default** — asserted by a deployment configuring it AND the store actually batching to the raised bound *against real D1* ("makes a Paid deployment actually batch to the raised bound"), plus the negative that Free stays inside the Free cap for the same work. That is the capability the task exists to add, tested at the level the criterion asked for rather than at the type level.
- **Which PLAN is stated rather than implied, and changing it is configuration not code** — `D1_PLAN` var, `resolveD1Plan` defaulting to the TIGHTER plan when unstated, and refusing an unknown plan by name. Changing plan is a `wrangler.toml`/var change.
- **The per-INVOCATION budget is addressed or explicitly deferred** — both halves handled honestly, which is more than the criterion's minimum. The PRUNE half is genuinely bounded (`d1PruneBudget(plan)` sizes `prune`'s existing `maxVersions` budget from the plan's query cap, which is exactly what ADR-0022's host-scheduled call is for). The INGEST half is explicitly deferred, with the reason it cannot be fixed by any batch bound (`maxStatementsPerBatch` bounds one `batch()` call; the fold issues as many as the work needs) and with a named owner, recorded in `work/notes/observations/nothing-bounds-the-size-of-an-ingest-batch.md`. Not left implied.
- **The value-pinning test in `batch.test.ts` still asserts what it was written to assert** — untouched, along with the bound-parameter guard. The defaults were correctly NOT re-fixed (that half had already landed).
- **Worker adapter tests + `wrangler deploy --dry-run`** — `wrangler deploy --dry-run --env production` is the package's `build` script, so the gate's `pnpm build` covers it; the pool tests pass.
- **Changeset for every published package whose surface changed** — none required: `@etherfold/platform-cf-worker` is `private: true`, and no published package's surface changed. Satisfied rather than skipped.
- The dated finding `d1-caps-bound-parameters-per-query-at-100.md` was re-checked and updated in the same change, as the prompt required (the limits are per-plan and Cloudflare revises them).

## Off-path finding

`work/notes/observations/worker-pool-close-times-out-with-two-test-files.md` records a real, reproduced infrastructure nuisance (`@cloudflare/vitest-pool-workers` 0.22 takes a 10 s close timeout once the package has two test files; exit code stays `0`, so it costs the gate ~10 s rather than turning it red). Correctly captured as an observation rather than worked around by deleting or merging a test file. Not blocking.

Merging.

PR #37, opened 2026-09-02
feat(retire-the-js-object-processor-path): Retire the JS-object process…

wighawag:
Task: `work/tasks/done/retire-the-js-object-processor-path.md`

The work is complete and the gate is green (`pnpm format:check`, `pnpm check:adr`, `changeset status`, `pnpm build && pnpm typecheck && pnpm test`, plus `test:full` on the heavy workload and both browser example builds).

## What I did

**Drift check first.** `index-to-a-store-from-the-cli` is in `work/tasks/done/` and the CLI already ran an entity processor into SQLite before I touched it, so the premise held. The command rename (`build`/`run`) has NOT landed, so the refusal I removed was on `index`, exactly as the task predicted. `a-snapshot-a-client-cannot-read-is-refused-not-installed` had also landed, so the snapshot path's real requirements were pinned in code.

**Deleted:** `packages/js-processor` and every workspace reference (9 dependents, incl. all six examples), `@etherfold/core`'s `KeepState` family and blob snapshot envelope, `@etherfold/browser`'s state keepers and `ProcessorKind`/`TaggedProcessor` fork, `@etherfold/utils`' kind fork and `/indexer` subpath, the CLI's `--store file`/`--folder`/`keepState.ts`, and the conformance workload's oracle.

**Ported rather than dropped:** `streamCache`, `streamSegments`, `invalidation` and `reconfigure` in `@etherfold/browser` (they are stream/reconfigure tests that merely used the free-form processor as a vehicle) onto entity processors over a real `createBrowserStateStore`, with a shared `test/utils/applied.ts` whose `times` counter makes a double application visible where an idempotent row write would hide it. `packages/processor-sqlite/test/equivalence.test.ts` was dropped: it was the differential against the deleted path, and `reorg.test.ts` still pins the same numbers as literals.

**One thing I found and did not fix** (`work/notes/observations/a-rebuild-off-a-cached-stream-drops-its-retractions.md`): rebuilding from a cached stream that contains a reorg discards the stream's own `removed` markers (`groupLogsPerBlock` skips them; retractions come only from the cursor's window, which is empty on a rebuild), so both branches are applied at one height. The free-form path swallowed it; the entity store refuses it. Pinned as a characterization case that will go red when it is fixed.

## Decisions

**What survived of `KeepState`: nothing, snapshot half included.** It had exactly one caller (`JSObjectEventProcessor.keepState`), and its two masters turned out to be one master and a namesake. The entity path's bootstrap never used it: `openSnapshotAware` / `bootstrapFromSnapshot` / `StateSnapshot` / `ENTITY_SNAPSHOT_FORMAT` live at the storage seam, where a store's own transaction is, and duplicate the free-form keeper's client-side selection point for point plus two things it never did. So `KeepState`, `ExistingStateFetcher`, `StateSaver`, `AllData`, `ProcessorContext`, `EventProcessorWithInitialState`, `keepStateOnIndexedDB`, `keepStateOnLocalStorage`, `createFileKeepState` and `contextFilenames` all go. **This deletes the guard `a-snapshot-a-client-cannot-read-is-refused-not-installed` had just added** (the format refusal in `storage/state/OnIndexedDB.ts`) and the `BLOB_SNAPSHOT_FORMAT` envelope it guarded: what is lost is the free-form reader, not the RULE, which the entity reader already enforces with `SnapshotFormatError`. Alternative considered and rejected: keeping the blob envelope as a publish format with no writer and no reader. Touches ADR-0040 (its rule stands, one of its two numbered envelopes is gone) and the `bootstrap` / `KeepState` glossary entries.

**The `{kind, processor}` module tag is deleted too, superseding ADR-0039, and a module still carrying it is REFUSED.** The acceptance criteria mandate removing the browser's `TaggedProcessor` fork ("one kind, one call shape"); leaving `@etherfold/utils` with a one-valued tag would have kept "processor kind" a live concept in the glossary, which another criterion forbids. So `instantiateProcessorWithKind` / `ResolvedProcessor` / `ProcessorKind` go and `instantiateProcessor` returns the authoring object, typed by the caller. The refusal is a NEW ERROR and a deliberate one: unwrapping the retired shape would keep a second module shape alive forever, and silently accepting it would hand a store a wrapper it asks for `entities` and gets `undefined` three frames later. It is refused before any RPC call, at the same point the kind/store mismatch was. Touches `examples/event-processor-nfts/src/cli.ts` and any external processor module; ADR-0039 is marked superseded rather than deleted. Alternative considered: silently unwrapping the tag for one release (rejected: "nothing is published, do not add a compatibility shim for a caller that does not exist").

**`--store` survives with exactly one value and stays REQUIRED.** The task says `--store` loses its `file` value and that `one-command-runs-the-whole-pipeline` renames the command the flag lives on, so the flag itself is not a leftover: it is the axis a second backend arrives on, and the word an operator types is the same before and after one appears. Alternatives: dropping it (`--db` alone names the destination) or defaulting it to `sqlite`. Both are user-visible default changes, and neither is a removal, so I made the minimal one. Touches `one-command-runs-the-whole-pipeline` (which renames this command) and any documented invocation.

**Examples: `event-processor-nfts` ported, five deleted, `web-demo` deleted as collateral.** `event-processor-nfts` keeps `entities.ts` (which its browser demo and `etherfold index` already ran) and loses `src/index.ts`, `src/types.ts` and `scripts/version.mjs` (which substituted `__VERSION_HASH__` into the deleted `dist/index.js`; the entity processor carries a literal version and the README's build command is now plain `tsc`). `basic`, `event-processor-bleeps`, `event-processor-conquest-eth`, `event-processor-conquest-fplay` and `mud` are deleted. `web-demo` was not one of the six but consumed three of them and rendered a state blob as a JSON tree, which is the shape the entity path does not have, so porting it was a rewrite rather than a port. Two examples remain (`event-processor-nfts`, `browser-reference`), both entity-path end to end. The root `README.md`'s whole Usage section was the `basic` app, so it is rewritten against the entity path.

**Conformance workload: the golden is now frozen, and the vendored original stays unrun.** `src/oracle.ts`, `test/oracle.test.ts` and `run/regenerate-golden-state.ts` are deleted; `runWorkload` already compared against the committed file, so the workload's question is unchanged and the alpha1 case still passes on every backend. The vendored `js-processor.ts` is KEPT (the task forbids deleting it, and the goldens' whole claim is that it computed them), which meant vendoring the `JSProcessor` TYPE beside it as `vendor/stratagems/js-processor-type.ts` so the folder still typechecks rather than becoming a folder of `any`. Recorded in `fixtures/README.md`, the package README and `vendor/stratagems/README.md`: a diff is now triaged by READING that file, not by re-running it.

**`docs/spikes/sqlite-in-the-browser/`: kept with a marker, not pruned.** The five files importing `packages/js-processor/dist/` are left in place, and the spike README now names all five, says the import no longer resolves, says why they are kept (a measurement whose harness has been deleted is a number with no method), and says explicitly that `docs/spikes/` is not a workspace package so `pnpm typecheck` will not tell you. The `Re-running it` block marks each broken command `BROKEN`.

**Two pending changesets on `main` named the deleted package** and would have failed `changeset status` and any release (`packages/core/test/pendingChangesets.test.ts` catches this). I removed `@etherfold/js-processor` from their frontmatter and made one sentence of prose truthful; I did NOT rewrite `index-an-entity-processor-into-a-store-from-the-cli.md`, whose `--store file` examples my own changeset supersedes in the same release.


wighawag:

Reviewed the full diff (210 files, ~14.5k deletions) against the task's acceptance criteria. Gate-1 (acceptance) is green (`verify` SUCCESS); this is the judgement pass on top.

## The one that needed real scrutiny

This task's `Blocked by` section warned that the ordering created "a chance to delete a guard that had just been added" — and this diff **does** delete the format refusal that `a-snapshot-a-client-cannot-read-is-refused-not-installed` landed, along with the `BLOB_SNAPSHOT_FORMAT` envelope. I checked this rather than taking the Decisions block's word for it, and it holds:

- The deleted guard protected the **free-form blob keeper's** install path, and that entire path is gone. A guard cannot be kept on a code path that no longer exists.
- The **RULE** survives where it can still apply: `SnapshotFormatError` is enforced on the entity path (`packages/state-store/src/snapshot.ts`, `processor-entities/src/snapshot.ts`) and is **tested** — `packages/state-store/test/snapshot.test.ts:78` asserts `bootstrap({format: 99})` rejects with it. Refused, not installed.
- The criterion explicitly sanctioned this decision shape ("removed **or** reduced to whatever the SNAPSHOT path still needs — decide which, and say so in `## Decisions`"), and the warning was about the UNSERIALISED hazard; the tasks were serialised, so the judgement was made with the snapshot path's requirements pinned in code, exactly as intended.

Worth flagging downstream rather than blocking: ADR-0040 now has one of its two numbered envelopes gone, and the `a-snapshot-...` → `publish-etherfold-...` edge is moot for the blob format.

## Criteria

- **`packages/js-processor` deleted with every workspace reference** — gone; no live import remains in `packages/*`, `platforms/*` or `examples/*`. The surviving `js-processor` matches in `conformance-workload-stratagems` are to the **vendored** copy, which the scope fence protects; the agent vendored the `JSProcessor` TYPE beside it so that folder still typechecks instead of degrading to `any`.
- **The five `docs/spikes/` files handled DELIBERATELY** — chose keep-with-marker over pruning, and the marker is real: the spike README names all five, states the import no longer resolves, says why they are kept (a measurement whose harness is deleted is a number with no method), marks each command `BROKEN`, and explicitly notes `docs/spikes/` is not a workspace package so `pnpm typecheck` will not catch it. No silently dangling imports.
- **`'js-object'` kind, `TaggedProcessor` fork, bare `EventProcessorWithInitialState`** — all removed; only a historical comment remains. The agent went further and retired the one-valued module tag (superseding ADR-0039, correctly marked `status: superseded` rather than deleted), on the ground that leaving it would keep "processor kind" a live glossary concept that another criterion forbids. Reasonable, and recorded with the alternative it rejected.
- **`KeepState` family** — decided (removed entirely, snapshot half included) and justified: its two masters turned out to be one master and a namesake, since the entity path's bootstrap lives at the storage seam. Verified above.
- **Conformance workload** — oracle, its test and the regenerate script are gone; `runWorkload` still compares against the COMMITTED golden, so the workload's question is unchanged. The frozen-expectation trade is recorded in three places (`fixtures/README.md`, the package README, `vendor/stratagems/README.md`). Vendored original and all fixtures preserved, per the fence.
- **Examples** — `event-processor-nfts` ported to the entity path (`processor-entities` + `state-store-patch`, no `fromJSProcessor` anywhere), five deleted rather than left broken, `web-demo` deleted as collateral with its reason given. Two remain, both entity-path end to end. Root README's Usage section rewritten off the deleted `basic` app.
- **`CONTEXT.md` updated in the same change**; changeset added; two pre-existing pending changesets that named the deleted package were corrected so `changeset status` and the release would not fail.

## Off-path finding

`work/notes/observations/a-rebuild-off-a-cached-stream-drops-its-retractions.md` is a strong piece of work: it identifies the MECHANISM behind the symptom observed during the stream task (`promiseToFeed` derives retractions from `unconfirmedBlocks` only, and a rebuild starts from an empty window, so stored retractions are discarded), notes the entity store turns it from silently-wrong state into a hard failure, and pins it as a characterization test that will go red when it is fixed. Captured, not fixed in place. Not blocking; flagged for tasking.

Merging.

PR #38, opened 2026-09-02
feat(add-package-readmes-and-descriptions): Add descriptions + READMEs…

wighawag:
Task: `work/tasks/done/add-package-readmes-and-descriptions.md`

Done. Both blockers had landed (`retire-the-js-object-processor-path` is commit `bb86a77`; `drop-filesystem-storage-and-rehome-the-fixture-loader` is in `work/tasks/done/`), so I derived the real package set from `ls packages platforms` plus each manifest's `private` flag rather than from the task's stale role list.

## What the derived set looked like

17 packages (14 + 3 platforms), 15 non-private. Missing a `description`: `@etherfold/core`, `@etherfold/browser`, `etherfold`, `@etherfold/utils`. Missing a `README.md`: those four plus `@etherfold/server`, `@etherfold/platform-nodejs`, and the private Worker host. Nothing in `archive/` was touched. `@etherfold/js-processor`, `@etherfold/fs` and `@etherfold/fs-cache` no longer exist, as expected.

## What I wrote

- **Descriptions** (`packages/{core,browser,cli,utils}/package.json`), each verified against source: core stores nothing and drives an `EventProcessor`; browser is the engine plus observable stores; `etherfold` is a one-shot into libSQL plus `serve`; utils is the Node-side processor-module loader (the old "context/source hashing, filenames" role in the task is stale, `contextFilenames` and the `/indexer` subpath went with ADR-0037).
- **READMEs** for `packages/core`, `packages/browser`, `packages/cli`, `packages/utils`, `packages/server`, `platforms/nodejs`, `platforms/cf-worker`, in the house style of the existing ones (role, when-to-use-vs-alternatives table, a snippet taken from code that runs, related links, tests line). Every snippet's signature was checked against source (`new LogFetcher(provider, source, target, config?)`, `createBrowserStateStore(entities, config)`, `resolveProcessorAndSource<ABI, R, P>`), and every relative link resolves.
- **Status stated truthfully**: `etherfold index` is documented as a one-shot that exits at the tip and cannot be reconfigured while running, with live reconfigure named as the browser package's ability. The CLI README also says plainly that the `run`/`build`/`fetch` command set is decided and not built.
- **Consumers checked**, and they changed what I wrote: `wighawag/stratagems` still resolves `ethereum-indexer-browser@^0.7.6` / `ethereum-indexer-js-processor` (its indexer is a `fromJSProcessor` processor) and `stratagems-snapshots` has no manifest at all, so nothing claims stratagems as a current `@etherfold/*` consumer. The signal that this makes the publish task's migration step a port rather than a rename is captured in `work/notes/observations/publish-task-still-names-the-deleted-js-processor.md`.
- **Changeset** `.changeset/every-package-says-what-it-is.md`: a `patch` for every package whose directory this change touches (`browser`, `core`, `etherfold`, `utils`, `server`, `platform-nodejs`, `state-store-sqlite`); the two private packages are correctly outside the versionable set (`privatePackages: false`).
- **Optional criterion taken**: a "The packages" table in the root `README.md` mapping engine → authoring → storage backends → run client / CLI / server / split-fetch, with a line saying `archive/` is not built or published.

Gate: `pnpm format:check`, `pnpm check:adr`, `pnpm build`, `pnpm typecheck`, `pnpm test` are all green. (`pnpm changeset status --since=main` cannot see an UNTRACKED changeset file, so it reports the "changed but no changesets" error until the runner commits; it passes on the committed tree, and the check is `changedPackages > 0 && changesets === 0`.) Working tree holds only the intended changes.

## Decisions

**`etherfold` stops publishing the repo's root README; the CLI now has its own committed one.** `packages/cli/package.json`'s `prepack` copied `../../README.md` into the package and `.gitignore` listed `/packages/cli/README.md` as generated, so the CLI's npm page described the monorepo and documented none of its flags, and any README I wrote there would have been overwritten at pack time. I changed `prepack` to copy only the LICENSE and removed the ignore entry (the LICENSE mirroring is untouched). Alternatives considered: leave the CLI as the one package with no README of its own (fails the acceptance criterion), or commit a README that `prepack` silently replaces (worse than either). Touches the publish path, so `work/tasks/ready/publish-etherfold-and-deprecate-old-names.md` should expect `packages/cli/README.md` to be a source file now.

**The private Worker host got a README too.** Criterion 1 is explicitly about non-private packages, but criterion 2 says "every KEPT package", so I read `platforms/cf-worker` as in scope and wrote one that opens by saying it is a deployable and not a library, never published. No description was added or changed for either private package (both already had one). The alternative was to skip it, which would leave the one package where D1's limits are legitimately named undocumented.

**I corrected one EXISTING description rather than only filling the gaps.** `@etherfold/state-store-sqlite` called itself a "Versioned-row state store for `@etherfold/core`", but it depends on `@etherfold/state-store`, `remote-sql` and `named-logs` and on nothing else, and `test/no-platform-leakage.test.ts` asserts that; naming core there inverts ADR-0016's direction. It now says "behind the `@etherfold/state-store` seam". This puts a package outside the missing-description set into the changeset (a `patch`), which is why it is called out rather than done quietly.


wighawag:

Reviewed the diff against the task's acceptance criteria. Gate-1 (acceptance) is green (`verify` SUCCESS). Because this task's criteria are about COVERAGE, I enumerated the workspace rather than spot-checking.

## Criteria

- **Every non-private package under `packages/*` AND `platforms/*` has a `description`** — verified by enumerating all 17 manifests: 15 public + 2 private, every one carrying a `description` and a `README.md`. The set was DERIVED from the manifests rather than read off the task's own stale role list, which is what the criterion asked for (the list named three packages that no longer exist and omitted about eleven that do).
- **Every kept package has a `README.md`** — all 17, including the two private ones (harmless, and useful for the Worker host).
- **Nothing in `archive/` gains a README or description** — `archive/` is untouched (0 files in the diff).
- **No README describes an archived or deleted package as a CURRENT option** — the only surviving mentions of `@etherfold/js-processor` / `@etherfold/fs` are explicitly HISTORICAL and correct ("vendored when that package was deleted (ADR-0037) so this folder still typechecks"; "moved here when `@etherfold/fs` was deleted (ADR-0041)"). They document why an artifact is where it is, not a package to reach for.
- **Summaries verified against source, status stated truthfully** — the CLI is described as "one shot, exits at the tip", and the changeset calls out that it does not follow the chain and cannot be reconfigured while running, with live reconfigure attributed to `@etherfold/browser`. That is the specific trap the criterion named, and it was avoided rather than guessed past. `@etherfold/utils` was likewise re-described against what it actually contains now that `contextFilenames` went with the blob snapshot, instead of the stale "bag of hashing helpers".
- **A changeset covers the description changes, a `patch` per changed package** — and it is precisely scoped: exactly the 7 published packages whose published content actually changed (4 gained descriptions, 1 had a wrong one corrected, 2 gained READMEs), not a blanket bump across the workspace.
- **Docs + metadata only, no runtime behaviour changes** — confirmed: no non-test `.ts` file is touched anywhere in the diff.

## Worth calling out

- `@etherfold/state-store-sqlite`'s existing description named the wrong seam ("a state store for `@etherfold/core`"). Corrected to match what it actually depends on, which is the right call: the old wording inverted ADR-0016, and a test in that package already asserts the real dependency direction.
- The CLI's `prepack` was copying the monorepo's root README into the published package, so `etherfold`'s npm page would have described the whole repo and documented none of its own flags. Now it has a committed README of its own and `prepack` copies only the LICENSE. This is a genuine packaging defect caught by doing the work carefully, and it matters because it would have shipped on first publish.

Merging.

PR #39, opened 2026-09-02
feat(a-rebuild-replays-the-retractions-the-stream-carries)

wighawag:
Task: `work/tasks/done/a-rebuild-replays-the-retractions-the-stream-carries.md`

All acceptance criteria are met and the full gate (format, ADR numbers, build, typecheck, all 18 suites) is green.

## What I did

**Drift check first.** The task's premises all still hold against the tree: `promiseToFeed` routed both the fetch path and the kept-stream replay through `generateStreamToAppend`; `groupLogsPerBlock` still drops `removed`; the browser characterization test still passed (i.e. the bug was live), and I confirmed it flips red/green with the fix in/out. One inaccuracy, not load-bearing: the task (and a code comment) says `feed()` is "the server's import route". There is no such route: `@etherfold/server` ingests through `StreamBuilder.receive`, which never shared `feed()`. The fix is unaffected, so I proceeded and corrected the stale prose where I was already touching it (`promiseToFeed`'s comment, `packages/core/test/feed.test.ts`'s header).

**The fix** (`packages/core`):

- `generateStreamFromReplay` (`internal/engine/utils.ts`): the IN-direction counterpart of `generateStreamToAppend`. It enforces the same `expectedFromBlock` refusal, then WALKS the stored emission stream: an applied block enters the unconfirmed window and is delivered, a retracted block LEAVES it and its retraction is delivered, keyed by block HASH. The finality prune happens once at the end against `lastToBlock`, so the cursor a replay leaves behind is the one the live run held, window included. That window is also what de-duplicates the catch-up shape of a replay (kept state, stream ahead), which the fetch path used to do with `startingBlockForNewEvent`.
- `EthereumIndexer.replay(eventStream, lastSyncStored)`: the new public entry, used by both replay call sites in `promiseToLoad`. `feed()` keeps its meaning (a FETCH) and now refuses a batch carrying `removed` markers with an `InvalidBatchError` naming `replay()`.
- `groupStreamPerBlock` now groups CONSECUTIVE runs instead of keying a map over the whole list (a whole stored stream can apply a block, retract it and apply it again under one hash; the map folded the third emission into the first group).
- `groupLogsPerBlock`'s fetch-side rule is UNCHANGED, with a new doc comment at the seam saying it is the fetch rule and not the replay rule.

**Tests.** New `packages/core/test/replay.test.ts` (7 cases) exercises the replay entry directly against a fresh cursor with a stream carrying `removed` markers: same applies and reverts in the same order as the live run, same window, same `expectedFromBlock`, replacement applied exactly once, no re-application on the first following tip cycle, no-reorg replay unchanged, catch-up replay does not re-apply its window, and `feed()`'s refusal. Three of those fail without the fix. The browser characterization test is INVERTED (not deleted), keeps and updates its explanation, and now asserts an EQUALITY against the live state plus `times === 1` plus a further tip cycle; it fails with `block 104 is already recorded` when the fix is disabled.

**Standing conventions.** Changeset added (`@etherfold/core`, minor). ADR-0042 added at the next unused number (`pnpm check:adr` green). `CONTEXT.md` gains a glossary entry for the fetch/replay distinction, and the three live places that said "`generateStreamToAppend` rebuilds the window from the replayed events" now name `generateStreamFromReplay` (ADR-0035's body left as the historical record, cross-referenced from ADR-0042).

## Decisions

- **A replay gets its OWN entry rather than `generateStreamToAppend` learning to take a stream.** `EthereumIndexer.replay()` + `generateStreamFromReplay`, with `promiseToFeed` branching on a `replay` flag and sharing the delivery half (batching, one-batch retractions, `groupStreamPerBlock`) unchanged. Alternatives considered: (a) a flag on `generateStreamToAppend`, (b) SNIFFING for a `removed` marker inside it. Rejected because that function's entire contract is "a complete re-read of a range plus a window decides what changed" (by-NUMBER matching, absence-vs-contradiction cause, `startingBlockForNewEvent`), and a replay decides nothing, it obeys; two contracts in one function is how the defect arose, and sniffing makes behaviour depend on whether the data happened to contain a reorg. Touches: `EthereumIndexer.feed`, `promiseToLoad`'s two replay call sites, and any future host replaying a captured stream (`replayStream` fixtures reach it through the load path unchanged). Recorded as **ADR-0042** (`docs/adr/0042-a-replay-honours-the-verdicts-the-stream-carries.md`).
- **`feed()` now REFUSES a batch carrying `removed` markers** (`InvalidBatchError`, message naming `replay()`), instead of accepting it and silently dropping them. This is a NEW refusal and a user-visible behaviour change on a public method, so it is flagged here. Chosen because the failure mode it replaces is silent and expensive (state derived from a dead branch), and because it is the same refusal `assertWellFormed` already makes on the wire for the same reason, so it introduces no new concept. Alternative considered: leave `feed()` permissive and route a stream-shaped input to the replay path automatically; rejected as the sniffing above. Reversible at the cost of a changeset. Touches: any host calling `feed()` (today: the core/processor-sqlite tests; all green).
- **A replay reconstructs the unconfirmed window by WALKING the stream, not by filtering out its `removed` entries.** The filter is the obvious shortcut and it is wrong: it leaves BOTH branches of a reorg in the window at one height, which is a window no live run ever held, and it is what makes the next tip cycle re-apply the replacement block (the folded-in second symptom). Matching is by block HASH, not height, because a height names whichever branch won. Alternative considered and rejected on ADR-0035 grounds: make a stream keeper store `unconfirmedBlocks` again. The finality prune uses `newLastToBlock` (`generateStreamToAppend`'s carry-forward bound, the more permissive of the two it uses); the final list is sorted ascending because the next fetch's reorg loop reads it in block order.
- **`groupStreamPerBlock` groups CONSECUTIVE runs rather than keying a map over the whole list.** This changes the delivery-side helper the task told me not to regress, so it is recorded rather than buried: it is a generalisation, not a loosening. Keying `{removed, hash}` over a whole stored stream folds a re-application of one block (chain went A, then B, then back to A) into the FIRST apply group, delivering it before the retraction it undoes. Its three existing tests pass unchanged, and it is now used by both directions of the seam (the replay's walk and the delivery batching), which keeps one concept instead of two near-identical groupers.


wighawag:

Reviewed the diff against the eight acceptance criteria in `work/tasks/done/a-rebuild-replays-the-retractions-the-stream-carries.md`, and against the three scope fences in "What this is NOT". Gate-1 (acceptance gate) and CI `verify` are both green; this pass is the judgement layer on top of them, not a re-run.

## Criteria

1. **Same state as the live run, asserted as an EQUALITY** ✅ — `streamSegments.test.ts` now does `expect(rebuiltApplied).toEqual(applied)` against the state captured from the live run, plus the positive/negative branch checks (`0xb104:0` present, no `0xa104*`). Not a no-throw.
2. **Replacement block applied EXACTLY once, including the first following tip cycle** ✅ — `times === 1` on every row (which is what catches a double-apply through a write that merely looks idempotent), and a further `indexToTip` asserts the state is unchanged after it. Mirrored engine-side by "does not apply the replacement block again on the FIRST tip cycle after the replay".
3. **Same reverts in the same order, and the cursor's window matches the live one** ✅ — asserted directly, not inferred: `expect(replayed.unconfirmedBlocks).toEqual(live.lastSync.unconfirmedBlocks)` (and again in the catch-up case), on top of the same-applies-and-reverts-in-order case.
4. **Characterization test INVERTED, not deleted, comment kept and updated** ✅ — the case is renamed to assert the correct behaviour, its explanation is rewritten to record what it used to pin and why the inversion is the regression proof, and `block 104 is already recorded` survives only as prose, never as an expectation.
5. **Engine-level coverage in `packages/core`** ✅ — new `test/replay.test.ts`, 7 cases driving the replay entry directly with a `removed`-carrying stream against a fresh cursor.
6. **No-reorg replay and live `indexMore()` unchanged; suites green** ✅ — full gate + CI green. I separately confirmed every remaining `feed()` call site (`core/test/feed.test.ts`, `core/test/streamBuilder.test.ts`, `processor-sqlite` x2) passes fetch-shaped logs, so the new refusal cannot fire on an existing caller. Worth noting the task's premise that "the server's import route shares `feed()`" is inaccurate: `@etherfold/server` ingests through `StreamBuilder.receive`. The report flagged this rather than quietly relying on it, and corrected the stale prose where it was already editing. Correct handling.
7. **`groupLogsPerBlock`'s fetch-side contract intact or explicitly documented** ✅ — the code is untouched and it gains a doc comment naming it as the FETCH rule and not the replay rule, with the pointer to `generateStreamFromReplay`. The seam is now readable in both directions.
8. **Tests touch nothing shared** ✅ — per-test `freshName()` tags against the fake IndexedDB; the core suite is in-memory.

## Scope fences

- **Not fixed in the stream keeper** ✅ — the only change to `packages/core/src/stream/segments.ts` is two comments retargeted from `generateStreamToAppend` to `generateStreamFromReplay`. `unconfirmedBlocks: []` stays; ADR-0035 is not undone.
- **Characterization test not deleted** ✅ — see criterion 4.
- **Delivery side not regressed** ✅ with a note. `groupStreamPerBlock` *was* changed (map keyed by `{removed, hash}` over the whole list, to consecutive runs), which is the one place this brushes the "do not rework the out-direction" fence. I accept it: for any single generated batch the two are equivalent (a batch never interleaves two hashes), its three existing tests pass unchanged, and over a whole STORED stream the map is actively wrong (a block applied, retracted, then re-applied under the same hash folds its third emission into the first group, delivering the re-application before the retraction it undoes). It was recorded as an explicit decision rather than buried, which is the right call for a change that touches a fence.

## Design

`replay()` as its own entry rather than teaching `generateStreamToAppend` to sniff its input is the right seam: the two functions have opposite contracts (a fetch DECIDES what changed from a window; a replay OBEYS verdicts it is handed), and merging them is how the defect arose. Rebuilding the window by WALKING the stream rather than filtering out its `removed` entries is the part that actually closes the second symptom, and the reasoning is captured where the next reader will need it (ADR-0042 + the function docs). The new `feed()` refusal is a user-visible behaviour change on a public method; it is flagged as such in the decisions, carries a changeset, and replaces a silent failure mode with a loud one that names the alternative.

Merging.

PR #40, opened 2026-09-02
feat(a-stream-is-identified-by-the-digest-of-its-filter): A stream is i…

wighawag:
Task: `work/tasks/done/a-stream-is-identified-by-the-digest-of-its-filter.md`

All acceptance criteria are met and `pnpm build && pnpm typecheck && pnpm test` (plus `format:check` and `check:adr`) are green.

## What I built

**`@etherfold/core` — the identity.** New `packages/core/src/stream/identity.ts` exporting `streamDigestOf(source, streamConfig)` and `STREAM_DIGEST_LENGTH`. The digest is over the **deduplicated `streamHash` values sorted by themselves** (so the block-0 skeleton entry's chain facts are in it, and `hash`/`legacyHash` are not) plus the **resolved stream config**, hashed with `viem`'s synchronous `sha256` truncated to 128 bits and rendered as 32 fixed-length lowercase hex characters. `simple_hash`'s canonicalisation is extracted as `canonical_form` and shared, so the wide digest and the 32-bit detector cannot disagree about what "the same value" means; `simple_hash` itself is byte-for-byte unchanged.

**The plumbing.** `ExistingStream` gains an optional `setStreamConfig`, which `EthereumIndexer.reinit` calls with the config it resolved — before any other call and again on every reconfigure (see Decisions).

**`@etherfold/browser` — the address.** `placeholderStreamDigest` is deleted; `streamAddress(name, source, streamConfig)` puts the real digest in the level the placeholder held. The `<indexer-name>` level, the key ranges, the cursor placement, the legacy-blob rule and every damage rule are untouched. Nothing migrates: placeholder-era subtrees are simply unreachable, and their disposal is left to the registry task.

**Tests (written before the code at each seam).** `packages/core/test/streamIdentity.test.ts` (14): stability under a decode-only rename *while asserting the entry list really reorders*, under ABI reordering and a redundant append; dedup; movement on filter and on config; resolved-config equivalence; synchronous, 128-bit, fixed-length hex; no `crypto.subtle`, no second implementation in the package; no collision over a 2,400-source corpus. `packages/browser/test/streamIdentity.test.ts` (11): the digest occupies the level with the name level intact, decode-only reuses the stream, a filter or config change resolves elsewhere and leaves the old subtree whole, two names and two chains still isolated, a placeholder-era subtree is neither adopted nor deleted, and — the criterion-critical one — the address the keeper writes under is the config the **indexer** resolved, including after `updateIndexer`.

**Also:** a changeset for `@etherfold/core` and `@etherfold/browser`; existing browser tests re-aimed at the new address deliberately (they still assert what they were written to assert). ADR-0008's 2026-08-31 amendment already records this narrowing of ADR-0006's `{source, config}` to the fetch half, so no new ADR was needed.

Out-of-scope signal captured (not fixed): `work/notes/observations/updateindexer-hashes-the-provided-stream-config-not-the-resolved-one.md`.

## Decisions

- **How the stream CONFIG reaches a keeper: an optional `setStreamConfig` on `ExistingStream`, called by `EthereumIndexer.reinit`.** The seam hands a keeper a `source` on every call and never the config, so a keeper that addresses a stream by `{filter, config}` could not compute its own address. Alternatives: (a) a `streamConfig` option on `keepStreamOnIndexedDB` — rejected because in the browser the keeper is constructed at `createIndexerState` and the config only arrives at `init`, so the value would be duplicated at the call site and could silently disagree with the config the indexer actually runs, which is precisely the failure this criterion exists to prevent; (b) an extra parameter on all three seam methods — rejected because it would have to be threaded through `createSegmentedStream` and all five `StreamSegmentPort` operations for the same effect; (c) handing over the computed DIGEST rather than the config — rejected because ADR-0035 leaves addressing to the keeper, and a keeper driven directly still has to address something. Putting it in `reinit` (not the constructor) is what makes a reconfigure follow onto the new stream while leaving the old subtree intact. **Touches:** `ExistingStream` (the IndexedDB keeper, `replayStream`, test fakes — all of which may ignore it), `EthereumIndexer.reinit`, and the later generation tasks, which will want several stream identities per keeper rather than one.
- **A keeper nobody has told runs at `resolveStreamConfig(undefined)`.** So a keeper driven directly (a test, a tool) addresses the same stream an indexer given no `stream` config addresses, rather than a fourth thing. The alternative — refusing until told — would make the keeper throw on a path that is a local cache, which `fetchFrom`'s no-throw rule forbids. **Touches:** anything constructing a keeper outside an indexer.
- **The config enters the preimage as canonical BYTES, not as `simple_hash(config)`.** CONTEXT.md and ADR-0008 say "plus the stream CONFIG hash"; that value is 32 bits, and a collision there maps two configs onto one stream — the exact silent adoption this task forbids. The digest is never compared against a stored `context.config`, so nothing needs the two to be the same bytes, and hashing the config directly is strictly stronger. The per-entry `streamHash` values stay as they are, because the task specifies them.
- **The preimage carries a rule tag (`etherfold/stream/1`).** A later change to what enters the digest is then a different digest by construction rather than by luck. It costs nothing and it is the same "unreachable, swept later" story the placeholder now has. **Touches:** the registry task's sweep, which collects orphans from a digest redefinition.
- **`streamAddress` changes signature from `(name, chainId)` to `(name, source, streamConfig)`,** and `placeholderStreamDigest` is deleted rather than kept beside it. It is exported from `@etherfold/browser`, so this is a published-surface break; nothing is published and the conventions doc says to delete the wrong thing rather than deprecate it beside the right one. The changeset carries it.


wighawag:

Reviewed the diff against the ten acceptance criteria in `work/tasks/backlog/a-stream-is-identified-by-the-digest-of-its-filter.md` and against every scope fence in the task body and prompt. Gate-1 and CI `verify` are green; this is the judgement pass on top of them.

## Criteria

1. **Digest over deduplicated `streamHash` values sorted by themselves, plus the config, and nothing else** ✅ — `streamDigestOf` is literally `[...new Set(entries.map(e => e.streamHash))].sort()` plus the resolved config. `hash` and `legacyHash` never enter. (The `!== undefined` filter is type-narrowing only: `sourceHashesOf` sets `streamHash` on the skeleton entry and on every event entry, so nothing is silently dropped from the preimage — I checked, because a filter that quietly discarded entries would defeat the whole criterion.)
2. **Stable under a DECODE-ONLY change, with the reorder asserted** ✅ — and asserted the right way round: the test first proves the precondition (`expect(after).not.toEqual(before)` while `[...after].sort()` equals `[...before].sort()`), so it genuinely exercises the ordering trap instead of passing vacuously. This is the criterion a digest over the sorted entry list fails, and it is the one I looked hardest at.
3. **Stable under ABI reordering and a redundant appended entry** ✅ — both, plus an explicit dedup case using two anonymous events that share a `streamHash` while differing in `hash`.
4. **MOVES on a stream-config change, old stream left intact, not `parse.logValues`** ✅ — the worked examples are `alwaysFetchTimestamps` and `alwaysFetchTransactions` (plus `parse.filters`), and `parse.logValues` appears nowhere. The browser side proves the "left intact" half: after a config change the old subtree is still whole and readable at its own address.
5. **128-bit, synchronous, `viem`'s `sha256` truncated, one implementation, no collision** ✅ — value not promise, 32 hex chars, comment-stripped source asserted free of `crypto`/`subtle`/`async`/`await`/`Promise` and of `simple_hash(`, plus a package-wide scan asserting `stream/identity.ts` is the ONLY file using `sha256`. Collision corpus of 2,400 sources. The source-text assertions are a slightly blunt instrument, but they pin exactly the thing that must never regress, and no lighter mechanism would.
6. **Fixed-length rendering usable as an IndexedDB key element** ✅ — 32 lowercase hex chars, no `0x`, `STREAM_DIGEST_LENGTH` exported, exercised through a real keeper write.
7. **Real digest in the placeholder's level, name level untouched, isolation not regressed** ✅ — `placeholderStreamDigest` is deleted rather than left beside the real thing, the address SHAPE is unchanged, and "keeps two indexer NAMES and two CHAINS apart" carries the prerequisite's isolation forward. Two chains now stay apart because the block-0 skeleton entry hashes `chainId`/`genesisHash` into the digest, which is the reasoning the placeholder was written to preserve.
8. **Nothing migrates, no payload rewrite** ✅ — asserted as "a placeholder-era subtree is neither adopted nor deleted". The sweep is correctly left to the registry task.
9. **Changeset for every published package whose surface changes** ✅ — `@etherfold/core` + `@etherfold/browser`, both minor, and it documents the `ExistingStream.setStreamConfig` addition and the `streamAddress` signature change rather than just naming the feature.
10. **Gate green** ✅.

## Scope fences

No registry, canonical pointer or caps; no sweep or deletion of placeholder-era subtrees (explicitly asserted as a non-deletion); no restatement of the cursor contract; no server-side discriminator (no server package touched); no second tenancy discriminator and no composite-key type — the `<indexer-name>` level is left exactly as found.

## Two things I checked closely and accept

- **`simple_hash` was touched.** `canonical_form` is a pure extraction and `simple_hash` is byte-for-byte unchanged, which matters more than usual here: those digests are persisted, and a change would have silently invalidated stored processor versions. Sharing one canonicalisation rather than copying it is the right call — two normalisations that could disagree about "same value" would make the wide digest and the 32-bit detector fork independently.
- **`finality` enters the digest** (via the resolved config), so a finality change resolves to a different stream. Broader than the three fields the task names, but consistent rather than surprising: `streamConfigHash` already covers the whole resolved config and `sourceInvalidationOf` already invalidates the stream half from block 0 when it moves. Having the digest move too is what keeps the address and the verdict agreeing; the alternative would be exactly the "adopt logs the verdict has declared invalid" failure this task exists to prevent.

The `setStreamConfig` seam is the one real design addition, and putting it in `reinit` (where the config is RESOLVED, and which a reconfigure also goes through) rather than at the keeper's construction site is well argued: it is what makes a reconfigure follow onto the new stream while leaving the old subtree intact, and it removes the chance of an application repeating a config that disagrees with the one the indexer runs. The off-path finding about `updateIndexer` hashing the PROVIDED config was captured as an observation and NOT fixed here, which is the correct handling.

Merging.

PR #41, opened 2026-09-03
feat(generations-are-registered-and-one-pointer-is-canonical): Generati…

wighawag:
Task: `work/tasks/done/generations-are-registered-and-one-pointer-is-canonical.md`

Done. `pnpm build && pnpm typecheck && pnpm test` are green, `pnpm format:check` and `pnpm check:adr` pass, and the working tree holds only this task's changes.

## What I built

**Drift check first.** The task's blocker (`a-stream-is-identified-by-the-digest-of-its-filter`) landed at `351c585` with `streamDigestOf` real and the `<streamDigest>` address level filled, exactly as this task assumes; ADR-0008's 2026-08-31 amendment and ADR-0019 both say what the task says they say. No stale premise, so I built.

**`@etherfold/core` — `src/generation/registry.ts` (new).** The rules over a five-operation port, the same shape as `createSegmentedStream`: `read` / `commit(plan)` / `listStreamDigests` / `dropStreamSubtree` / `dropState`. `commit` takes a decision function so the cap check and the "already registered?" check happen inside the substrate's transaction (two tabs cannot both pass a cap of two). It exposes `create`, `list`, `streams`, `canonical`, `moveCanonicalTo`, `deleteGeneration`, `deleteStream`, plus `GenerationCapReachedError`, `UnknownGenerationError`, `UnknownStreamError`, `GenerationIsCanonicalError`. It never fetches, folds or opens a store: everything is exercisable with no indexer running.

**`@etherfold/browser` — `src/storage/generation/OnIndexedDB.ts` (new).** `['generation', <name>, 'entry', <streamDigest>, <processor>]` plus `['generation', <name>, 'canonical']`, hierarchical for the reason the stream address is. The sweep's digest listing is a key cursor over `['stream', <name>]` that jumps past each digest's subtree, so it costs one read per stream rather than one per segment. `BROWSER_GENERATION_CAPS` is `{maxGenerations: 2, maxStreams: 2}`. `KEYVAL_DATABASE` / `KEYVAL_OBJECT_STORE` / the memoised store moved to `src/storage/keyval.ts` (exported unchanged) so the registry and the stream keeper share one connection to the store the sweep must see; `streamAddress` is now `streamSubtree(name, digest)` plus the legacy key, and `streamsUnder(name)` is new.

**Tests.** `packages/core/test/generationRegistry.test.ts` (rules, memory port) and `packages/browser/test/generationRegistry.test.ts` (keys, the sweep against the real placeholder subtree the previous task left behind, pointer forward/back over two real `IndexedDBStateStore`s, a real database dropped on delete). `navigator.storage.estimate()` is asserted twice: a runtime spy over a full create/promote/delete/create exercise, and a source scan of both new modules. Both test files were written before the production code.

**Changeset:** `.changeset/generations-are-registered-and-one-pointer-is-canonical.md`, minor for both packages.

## Decisions

- **The FIRST generation created becomes canonical; a successor does not.** A registry that holds generations and points at none answers nothing, and every value the promotion policy will take (`on-catch-up`, `immediate`, `manual`) presupposes an incumbent before it has a question. Alternative considered: leave the pointer unset until an explicit `moveCanonicalTo`, which is purer mechanism but makes every caller special-case "first" and would make a `manual` deployment answer nothing until a human acted. **Touches** `the-promotion-policy-moves-the-canonical-pointer`: that task still owns the incumbent-versus-successor decision entirely; nothing here moves the pointer to a successor.
- **Deleting the canonical generation (or its stream) is REFUSED, not allowed with the pointer cleared.** New refusal `GenerationIsCanonicalError`: move the pointer first, which is one write. Deleting what answers reads blanks the app for a re-index, which is the outage this design exists to remove. Alternative: clear the pointer on delete, which leaves an indexer holding generations and naming none. Consequence: there is no way to empty an indexer completely through the registry; that is deliberate and can be revisited by whichever task owns tearing an indexer down.
- **Deleting a generation or stream the registry does not hold is REFUSED, not a silent no-op.** New refusals `UnknownGenerationError` / `UnknownStreamError`. The cap refusal NAMES what to delete, so a mistyped name coming back silently successful is the wrong answer to give an operator about a destructive call. **Touches** any future CLI/admin verb that wraps deletion: it must expect a throw rather than a count of zero.
- **`dropState` is INJECTED and required, not derived.** The registry does not know where a generation's state store lives and does not invent a database-naming convention: the container above `StateStore` (`the-generation-container-expands-beside-the-old-shape`) owns that, and it is inside this task's scope fence. Alternative considered: have the browser registry default to deleting an IndexedDB database named by a convention chosen here, which would fork a convention the container task must then either adopt or contradict. **Touches** tasks 4-6.
- **The record goes before the bytes on delete.** Registry commit, then `dropState`, then the stream subtree. A crash between them LEAKS storage; the other order leaves the registry claiming a generation whose state is gone. The leak is collected by the sweep on the next open, which is the same mechanism the placeholder case needs.
- **The sweep has no public entry point.** `openGenerationRegistry` runs it and exposes `swept` as a value. There is deliberately no `registry.sweep()`, so "run it on open, not on a timer" is structural rather than documentary. Cost: a caller who wants a sweep re-opens the registry.
- **Browser default caps are 2 and 2** (`BROWSER_GENERATION_CAPS`), a user-visible default: the previous generation and the new one, transiently. Core requires the caps explicitly and defaults nothing, so the server/CLI's generous values stay `the-server-and-cli-hold-generations-too`'s to set.
- **The stream `subtree` key range's lower bound widened** from ordinal `0` to the prefix itself (exclusive), so the range spans every key under a digest whatever its last element is. The sweep drops subtrees written by code that is gone, and assuming a dead digest rule's key shape is exactly the assumption that leaves bytes behind. It makes the existing `clearSubtree` strictly more complete; no behaviour any test pinned has changed.

No ADR: none of these meets the hard-to-reverse-plus-surprising bar on its own, and the model they implement is already recorded in ADR-0008's 2026-08-31 amendment and in the `CONTEXT.md` glossary entries for **generation**, **canonical pointer** and **generation cap**, whose language I reused rather than extended. The only new named concepts are the two refusals and `BROWSER_GENERATION_CAPS`.


wighawag:

Reviewed the diff against the twelve acceptance criteria in `work/tasks/backlog/generations-are-registered-and-one-pointer-is-canonical.md`, plus the body's mechanism-vs-policy fence and the vocabulary rule. Gate-1 and CI `verify` are green.

## Criteria

1. **Created, registered, identified by stream digest + processor version hash; creation takes its starting stream as an INPUT** ✅ — and the input half is asserted the way the task demanded (a real criterion, not a note): a second generation over an already-present stream leaves it untouched and issues no `dropStreamSubtree`, i.e. creation NAMES a stream rather than assuming it backfills its own. That is the seam `a-generation-can-be-seeded-from-a-published-artifact` would otherwise have to re-open.
2. **One canonical pointer, moved in a single small record write** ✅ — one `['generation', <name>, 'canonical']` record, asserted as one write, and asserted per-indexer (one name's move is not another's).
3. **The pointer moves BACK, restoring the previous generation EXACTLY, with no re-index and no fetch** ✅ — asserted twice: on the rules over the memory port, and in the browser over two real `IndexedDBStateStore`s, which is what makes story 4 credible rather than argued.
4. **A cap REFUSES, NAMES WHAT TO DELETE, evicts nothing, and everything stays readable** ✅ — for both caps. `GenerationCapReachedError` carries `cap`, `limit`, `refused`, `candidates` and `candidateStreams` as structured fields (not just prose), and the message explains WHY nothing was evicted. Both tests then assert the full prior list is intact, the canonical pointer is unmoved, and no `dropState`/`dropStreamSubtree` was issued. The "every generation but the canonical one" candidate rule is the right shape: it names everything deletable and picks nothing.
5. **`maxGenerations` is a TOTAL, not per-stream** ✅ — directly asserted, plus the complementary case that a second generation on an already-held stream is accepted under `maxStreams` because it adds no stream.
6. **Never derived from `navigator.storage.estimate()`** ✅ — belt and braces: a runtime spy across a full create/promote/delete/create exercise, and a source scan of both new modules. Given the finding this rests on (6.45 GB of headroom reported while writes were failing), the runtime spy is the assertion that matters and it is present.
7. **Deleting a generation leaves a shared stream; the stream is reaped with its last generation** ✅ — both, in core and again in the browser against a real keyspace. The record-before-bytes ordering ("a crash leaks rather than lies") is a good call and is asserted.
8. **The unregistered-subtree sweep** ✅ — this is the criterion I checked hardest, and it is done properly. The browser test writes a REAL `chain-<chainId>` placeholder subtree (the exact orphan the previous task left behind), opens the registry, and asserts the subtree is gone while the live stream is whole down to BOTH its segments and its cursor. Idempotence, "never touches another indexer NAME's subtree even under the same digest", "keyed on the registry not knowing the digest rather than on a known value", "runs on open", and "leaves the registry's own records alone" are each their own case. The keying is what makes it generalise to an orphan from any future digest-rule change, as specified.
9. **Testable with NO indexer running** ✅ — the rules run over a memory port; the browser suite drives the registry and stores directly. Nothing fetches or folds.
10. **Changeset for both published packages** ✅.
11. **Gate green** ✅.

## Fences and vocabulary

Mechanism only: the pointer moves when told, and there is an explicit test that it does NOT follow a successor on its own, so `the-promotion-policy-moves-the-canonical-pointer` still owns the whole policy question. `dropState` is INJECTED rather than derived from a database-naming convention invented here, which correctly leaves the state-store location to the container task (4-6) instead of forking a convention it would have to adopt or contradict. Caps are called caps throughout; `retention` (ADR-0019's block distance) is not overloaded.

## Two design choices I accept

- **The first generation created becomes canonical.** Slightly more than bare mechanism, but the alternative leaves a registry that holds generations and answers nothing until someone acts, and it would force every caller to special-case "first". It does not pre-empt the promotion policy, which is about successors.
- **Deleting the canonical generation (or its stream) is REFUSED rather than allowed with the pointer cleared.** Consistent with the task's whole posture that a refusal costs one operator action and a wrong destructive default costs a re-index. The stated consequence, that an indexer cannot be emptied through the registry, is a real and deliberate limit and is written down rather than discovered later.

The shared-`keyvalStore` extraction is the right call for a correctness reason and not just tidiness: a registry holding its own connection to the same database would be a second opinion about what is stored, and the sweep must see exactly what the keeper wrote. `streamAddress` is preserved beside the new `streamSubtree(name, digest)`, so last task's identity work is extended rather than reworked, and the sweep can address an orphan it cannot resolve a source for.

Merging.

PR #42, opened 2026-09-03
feat(the-invalidation-verdict-becomes-a-published-answer): The invalida…

wighawag:
Task: `work/tasks/done/the-invalidation-verdict-becomes-a-published-answer.md`

Done. The verdict is now a published, actionable answer, and the verbs behave exactly as before.

## What I built

**Core (`@etherfold/core`)**
- `packages/core/src/index.ts` — exported `SourceInvalidation`, `InvalidationVerdict`, `InvalidationReason`. The types are published; `sourceInvalidationOf` deliberately is not (the verdict is *reported*, never re-derived by a caller, for the same reason `stateDiscarded` is).
- `packages/core/src/indexer.ts` — `ReconfigureOutcome` grew `sourceInvalidation?: SourceInvalidation`. `updateIndexer` now returns the verdict it already computed (the comment that said the block "is carried no further than the log line" is gone). `resetNeeded` is still `!invalidation.state.valid`; nothing about what the verbs *do* changed.

**Tests (failing first, at the seam)**
- `packages/core/test/updateProcessor.test.ts` — three new cases went red before the production change: both halves valid on an unmoved source; the `stream-config` case naming block + reason on both halves; `updateProcessor`/`reset` carrying no source verdict. The two `toEqual({stateDiscarded})` assertions over `updateIndexer` were swept to the grown shape.
- `packages/browser/test/eventRanges.test.ts` — the **regression guard**: `streamDigestOf(SOURCE_RANGED_APPENDED_ABOVE)` ≠ `streamDigestOf(SOURCE_RANGED)` while the verdict says both halves valid, nothing is discarded, and no range below the cursor is re-fetched. Also asserts the appended-*below* case names `invalidFromBlock: 102, reason: 'entry-added'` on both halves. This file imports the type from `@etherfold/core`, so the cross-package reach is checked by `pnpm typecheck`.
- `packages/browser/test/reconfigure.test.ts` — the no-op reconfigure (`SOURCE_REDEPLOYED_SAME_ABI`) now asserts the verdict alongside the existing ranges-fetched + `stateDiscarded` pair.
- `packages/browser/browser/cut.ts` + `indexing.spec.ts` — the verdict is carried out of a real Chromium page and asserted there (ran green: 6/6 chromium).

**`stateDiscarded` sweep** (list derived by grep, not from the task file): `packages/core/src/indexer.ts` (5), `packages/core/test/updateProcessor.test.ts` (6, two updated), `packages/browser/src/IndexerState.ts` (3, unchanged — it branches on `stateDiscarded` and forwards the outcome, so the verdict already crosses), `packages/browser/browser/cut.ts` (3) + `indexing.spec.ts` (1), `packages/browser/test/{reconfigure,invalidation,eventRanges}.test.ts` (16), `packages/browser/README.md`, `docs/guide/indexing-in-a-browser-app/index.md`, `examples/browser-reference/browser/main.ts` (2, `onRedeploy` now reads the verdict and names the block in its message), plus the two existing changesets (left as history).

**Docs/conventions**: `CONTEXT.md` glossary bullet updated (verdict is published; verdict ≠ stream digest), guide + browser README updated, changeset `.changeset/the-invalidation-verdict-is-a-published-answer.md` for `@etherfold/core` and `@etherfold/browser` (both surfaces widened).

**Gate**: `pnpm build && pnpm typecheck && pnpm test` green (all 18 test-running packages), plus `pnpm format:check` and `pnpm check:adr`.

**Observation captured** (out of scope, not fixed): `work/notes/observations/context-and-adrs-still-point-at-specs-proposed-a-reconfigure-is-not-an-outage.md` — seven dead `work/specs/proposed/...` citations after the spec's status-folder move, plus a dead ideas-note link in `InvalidationVerdict`'s JSDoc.

**Drift check**: the blocking dependency `generations-are-registered-and-one-pointer-is-canonical` is in `work/tasks/done/`, `sourceInvalidationOf` is still internal with `ReorgCause`/`ReorgDetection` the only re-exports from that module, core's `exports` map is still `.` + `./package.json`, and ADR-0034's free-append behaviour is intact. No stale premise. (One path nit: the item is at `work/tasks/backlog/…`, not `work/tasks/ready/…` as the brief said; the body matches, so I built it.)

## Decisions

**`sourceInvalidation` is OPTIONAL, and absent on `updateProcessor` and `reset`.** Chosen because those two verbs ask no source question: a processor swap moves neither the fetch filter nor the decoding shape, and `reset` is a discard by fiat that also *clears* the cached stream. The alternative I considered was a non-optional field carrying the trivially-valid verdict (`{state: {valid: true}, stream: {valid: true}}`) for all three verbs, which is what literally computing it would return. I rejected it for `reset` specifically: `stream: {valid: true}` is true *of the source* but reads as "the cached stream stands" about a stream `reset` has just deleted, and the container that consumes this is exactly the caller that would act on it. `undefined` is the standard spelling of "not asked" and cannot be misread. What it touches: the later `the-generation-container-expands-beside-the-old-shape` / `every-caller-moves-onto-the-generation-container` tasks, which must handle `undefined` on those two verbs (the verb itself tells them what to do there); `@etherfold/browser`'s `IndexerState` forwards the outcome untouched, so nothing there changes.

**The verdict TYPES are exported; `sourceInvalidationOf` is NOT.** Chosen so the verdict stays *reported* rather than re-derived, which is the argument `ReconfigureOutcome`'s own doc already makes for `stateDiscarded`: a caller hashing its own source gets a second answer that can disagree with the one the core acted on, and it fails silently. The alternative was exporting the function too, which the acceptance criterion's "a real export" would also satisfy. What it touches: any later task that wants to ask the invalidation question *outside* a reconfigure would need a further, deliberate export decision rather than finding it already open.

**Field named `sourceInvalidation`, not `invalidation`.** Coherence check against `CONTEXT.md`'s glossary: `SourceInvalidation` / `sourceInvalidationOf` already name exactly this concept, so this reuses the existing language rather than forking a new one. The shorter `invalidation` would have re-meant it as "what this reconfigure invalidated", which for `reset` is a different (and larger) set than what the source comparison decided. No new concept was introduced.


wighawag:

Reviewed the diff against the seven acceptance criteria in `work/tasks/backlog/the-invalidation-verdict-becomes-a-published-answer.md`, and against this task's defining constraint: that it is ADDITIVE and lands green on its own. Gate-1 and CI `verify` are green.

## Criteria

1. **The verdict is PUBLISHED from `@etherfold/core`, reachable across the package boundary** ✅ — `SourceInvalidation` / `InvalidationVerdict` / `InvalidationReason` are exported, and the VALUE crosses on `ReconfigureOutcome.sourceInvalidation`. Keeping `sourceInvalidationOf` itself internal is the right line and is argued rather than assumed: the verdict is REPORTED for the same reason `stateDiscarded` is, and a caller re-deriving it from its own hashes gets a second answer that can silently disagree with the one core acted on. The cross-package reach is verified by the browser test importing the published type, so `pnpm typecheck` is what enforces it.
2. **`ReconfigureOutcome` grows to carry it in an actionable form** ✅ — both halves plus `invalidFromBlock` and `reason`, with the JSDoc naming what a caller actually decides with it (invalid STREAM half means the logs must come from the node again; a standing stream with an invalid STATE half is a new fold over logs already on disk).
3. **The verbs still discard exactly as today** ✅ — `resetNeeded` is still `!invalidation.state.valid` and the only production change is that the computed value is returned instead of dropped. The pre-existing `toEqual({stateDiscarded})` assertions were swept to the grown shape rather than loosened, and a new case pins that `updateProcessor`/`reset` carry no source verdict.
4. **The regression guard** ✅ — this is the criterion that mattered most and it is done exactly right. `an entry appended ABOVE the cursor is decided by the VERDICT and not by the stream digest` asserts, in one test, that `streamDigestOf` MOVES while the verdict reports both halves valid, `stateDiscarded` is false, and no range below the cursor is re-fetched. That is the pair ADR-0034 established, and it pins the specific mistake this task creates the opportunity for: now that a digest exists, reading digest inequality as a verdict would re-fetch an entire history for an upgrade that is free. The no-op reconfigure case is likewise asserted on ranges fetched AND state discarded, and the appended-BELOW case is asserted to name `invalidFromBlock: 102, reason: 'entry-added'` on both halves, which is the half `stateDiscarded` structurally cannot carry.
5. **`stateDiscarded` sweep derived rather than read off the task file** ✅ — the report enumerates the derived list with counts per file and says which were updated versus deliberately left (e.g. `IndexerState.ts` already forwards the outcome, so the verdict crosses without a change). Core, browser, the example app and the guide doc are all covered, and `examples/browser-reference` now reads the verdict and names the block in its `onRedeploy` message, which is what makes the published answer demonstrably actionable rather than merely present.
6. **Changeset for every published package whose surface changes** ✅ — `@etherfold/core` and `@etherfold/browser`, both minor, with a usage example.
7. **Gate green** ✅.

## Judgement

Making `sourceInvalidation` OPTIONAL and absent on `updateProcessor`/`reset` is the correct call, and the reasoning for `reset` is the sharp bit: reporting "both halves valid" there would be true of the source and would be READ as "the stream stands" about a stream that has just been cleared. A non-optional trivially-valid verdict would have been the easier shape and the more misleading one.

The additive discipline held: no verb changed behaviour, nothing was removed early, and the two-step is preserved for the stated reason (removing the discard belongs to the container tasks), not as an accidental leftover. The out-of-scope finding about dead `work/specs/proposed/...` citations after the spec's status-folder move was captured as an observation and NOT fixed here, which is the right handling, and its own note correctly reads it as a general problem with status moves rather than a handful of typos.

Merging.

PR #43, opened 2026-09-03
feat(the-generation-container-expands-beside-the-old-shape): EXPAND: th…

wighawag:
Task: `work/tasks/done/the-generation-container-expands-beside-the-old-shape.md`

## What landed

**`@etherfold/core`**

- `packages/core/src/indexer.ts` — the class `EthereumIndexer` is renamed to **`IndexerGeneration`** (one stream, one processor, one state IS a generation), and `EthereumIndexer` stays as an **alias to the GENERATION** (`export {IndexerGeneration as EthereumIndexer}`), never to the container. No `@deprecated` marker, per the task's Answers.
- `packages/core/src/container.ts` (new) — **`Indexer` / `openIndexer`**, the container: holds N generations in a `GenerationRegistry`, builds each from a `GenerationSpec` (`createState` → `createProcessor(state)`, supplied per generation), resolves reads through the canonical pointer via an **indirect handle** with stable identity, and applies a pointer move **only inside a state notification** (`promote` records the decision in the registry, the read path follows it at the notification). Verbs delegate to the canonical generation; nothing advances a non-canonical one.
- `packages/core/src/generation/memory.ts` (new) — the reference registry substrate (`createMemoryGenerationRegistryPort` / `openMemoryGenerationRegistry`).
- `packages/core/test/container.test.ts` (new) — the alias points at the generation and not the container; the old construction shape still loads; state-then-fold build order and the identity taken from `getVersionHash()`; only the canonical generation answers or indexes; the handle follows the pointer forwards and back; and the read-unit-of-work log (`read:A, read:A, read:A, notify:B, read:B`) proving the swap happens at the notification and not when the registry records it. Verified non-vacuous: stubbing the indirection red-lights 2 core + 1 browser test.

**`@etherfold/browser`**

- `createIndexerState` now accepts **both** shapes: the built processor (unchanged) and a `BrowserGenerationSpec` (`{createState, createProcessor, registry?}`), which builds the container and publishes its indirect handle into the `state` store.
- `packages/browser/test/generationContainer.test.ts` (new) — both shapes index to the same state; the factories are called once each with the stream context; and story 6 on real stores: a handle taken once keeps answering across a promotion, from generation B's separately-folded state (`transfers` 5 → 10 → 5 on the revert), with the same object identity.

**Also:** `docs/adr/0043-…`, a changeset for `@etherfold/core` + `@etherfold/browser`, and the CLI's source-text guard widened to match both names (see Decisions).

Nothing removed: the alias, the old factory shape and the `stateDiscarded` discard all remain; no call site in `core`, `browser`, `processor-sqlite` or `server` was touched. `pnpm -r build`, `pnpm build && pnpm typecheck && pnpm test`, `pnpm format:check` and `pnpm check:adr` are green.

## Decisions

- **A generation's factories are supplied PER GENERATION, and its state is built BEFORE its identity is known** (`createState` → `createProcessor(state)`, identity read off `processor.getVersionHash()` afterwards). The obvious shape, `createState(generationId)`, is unbuildable: the fold half of the identity needs the processor, the processor needs its state, and on a reload finding the record would need the hash it is keyed by. Alternatives considered and rejected: a caller-declared version (unwritable on the entity path, where the hash is `${version}-${simple_hash({entities, config})}`), a probe construction over a throwaway store, and a lazily-resolved store handle. Touches: the new call shape the MIGRATE batch moves every site onto, and `the-promotion-policy-moves-the-canonical-pointer`, which creates generations. Recorded as **ADR-0043**.
- **Two new refusals: `CanonicalGenerationNotHeldError` and `UnheldGenerationError`.** A registry whose canonical pointer names a generation this container has no spec for is refused at open, rather than silently promoting the one that was built (a pointer move nobody asked for) or answering reads from a non-canonical generation. Touches: the promotion-policy and shared-stream-follower tasks, which own resuming several generations across a restart.
- **New exported reference substrate in core (`createMemoryGenerationRegistryPort` / `openMemoryGenerationRegistry`).** Mirrors `MemoryStateStore` at the storage seam; it reports no stream subtrees because it stores none. Alternative was leaving the port with one real implementation and hand-rolling a memory port in every test. Touches: `@etherfold/core`'s public surface and the browser default below.
- **USER-VISIBLE DEFAULT: `createIndexerState`'s container shape defaults its registry to a MEMORY one under `BROWSER_GENERATION_CAPS`.** This hook knows no indexer NAME, and a durable registry is addressed under one, so defaulting to IndexedDB would have meant inventing a discriminator the stream address already carries; requiring a registry would have put a mandatory new argument on every site the migrate batch moves. Consequence, documented at the option: with the memory default the unregistered-subtree sweep has nothing to see, so an app that wants durable generations (and the sweep) passes `openGenerationRegistryOnIndexedDB(name, {dropState})`. Touches: `every-caller-moves-onto-the-generation-container` and `the-server-and-cli-hold-generations-too`.
- **The pointer is applied at a STATE notification only, not at a cursor (`onLastSyncUpdated`) one**, and the container publishes the INDIRECT handle to `onStateUpdated` rather than the value the generation produced — so a subscriber that keeps what it was handed keeps something that follows the pointer. Touches nothing outside the container, but it is what `the-old-indexer-shape-is-deleted` re-asserts.
- **`updateProcessor` / `updateIndexer` still reconfigure the canonical generation IN PLACE** (no behaviour change, per the scope fence), but the container now re-points its held entry's processor and drops its last published state on a discard, so the indirect handle cannot answer from a fold that was replaced. The registry record still names the fold the generation was registered with; closing that drift is the promotion policy's job, and it is flagged in the JSDoc.
- **Widened the CLI's source-text guard (`packages/cli/test/engine.test.ts`) to match `IndexerGeneration` as well as `EthereumIndexer`.** Not a call-site migration: the rename would otherwise have left that guard green and VACUOUS for the whole expand→contract window, since a CLI file could import the engine under its new name unnoticed. Verified it still bites (a probe file importing/constructing `IndexerGeneration` reds both cases). Deliberately NOT extended to `Indexer`, the container, because whether a CLI holds generations is `the-server-and-cli-hold-generations-too`'s question. Touches: the MIGRATE batch's item (c), which now finds this partly done and should still re-derive it.
- **The hook exposes no promotion surface.** Story 6 is asserted against the container directly (with real `EntityEventProcessor`s and stores) rather than by adding `generations`/`promote` to `createIndexerState`'s return, which would be surface no caller in this batch needs and which the promotion-policy task is better placed to shape. Touches: task 9, if it wants a hook-level promotion verb.


wighawag:

Reviewed against the nine acceptance criteria in `work/tasks/backlog/the-generation-container-expands-beside-the-old-shape.md`, and against the naming answer recorded in the task at `18d1570`. Gate-1 and CI `verify` are green.

## The naming decision, honoured exactly

`EthereumIndexer` is renamed to **`IndexerGeneration`**; the container is **`Indexer`**, a new export in `container.ts`; and the alias is `export {IndexerGeneration as EthereumIndexer}`, pointing at the GENERATION and never at the container. The class doc says why in the terms the decision was made in. The instruction to omit the deprecation annotation was followed AND recorded as deliberate ("Deliberately NOT marked `@deprecated` -- nothing is published, so this is internal scaffolding for one refactor rather than a compatibility promise"), which is the difference between a decision and an omission a later reader would try to "fix".

## Criteria

1. **Container is a new export, class renamed, alias to the generation, every construction site compiles untouched** ✅ — no construction site is touched anywhere, and a test asserts the alias resolves to the generation class rather than to the container.
2. **The expand covers the whole non-indirected surface across every package that reads it** ✅ — the rename is in `@etherfold/core` behind an alias, so `browser`, `processor-sqlite` and `server` need no edit at all, which is exactly what the criterion asks to be demonstrated rather than claimed.
3. **Both factory shapes accepted** ✅ — `createIndexerState` takes a BUILT processor (as before) or the FACTORIES that build one, each with its own passing test. The old shape is untouched, not reimplemented in terms of the new one.
4. **The entities-path handle is INDIRECT** ✅ — asserted in core and again in the browser: a handle held across a pointer move keeps answering from the newly canonical generation, and follows the pointer BACK, which is the mechanism story 6 will rest on.
5. **Read unit of work is the interval between notifications** ✅ — this is the criterion most easily faked and it is done properly. `moveCanonicalTo` on the registry records the decision durably and the read path deliberately does NOT move with it; the log assertion `['read:A', 'read:A', 'read:A', 'notify:B', 'read:B']` pins that every read in an interval answers from ONE generation and that the move is applied AT the notification. The companion case asserts the handle PUBLISHED to a subscriber is the indirect one, so a subscriber that keeps what it was handed keeps something that follows the pointer rather than a reference to the generation that was canonical when it was told. No scope API, no transaction handle, no timer.
6. **Green on its own with no call site migrated** ✅.
7. **Nothing removed** ✅ — alias present, old `createIndexer` option shape present, `stateDiscarded` still branched on in all four places in `IndexerState.ts`. The contract batch's work is left for the contract batch.
8. **Changeset** ✅ — core + browser, minor.
9. **Gate green** ✅.

## The one file I looked at twice

`packages/cli/test/engine.test.ts` is the only change outside core/browser, and it is NOT a migrated call site: it is the guard asserting the CLI constructs no such engine, widened to match BOTH spellings. That was necessary rather than incidental. Left matching only `EthereumIndexer`, the guard would have stayed green while silently enforcing nothing the moment a file used the new name, which is worse than no guard. It also deliberately does NOT extend to `Indexer`, on the grounds that whether a CLI holds generations is `the-server-and-cli-hold-generations-too`'s question and a grep here must not pre-answer it. Correct restraint on both halves.

ADR-0043 records the state-first build order and per-generation factories, which is a genuine "hard to reverse, surprising without context" choice and belongs in an ADR rather than a comment.

Merging.

PR #44, opened 2026-09-03
feat(every-caller-moves-onto-the-generation-container): MIGRATE: every…

wighawag:
Task: `work/tasks/done/every-caller-moves-onto-the-generation-container.md`

Done. All acceptance criteria met, full gate green.

## What I did

**Both surfaces re-derived from the tree, not read off the task body.** The task's counts were a launch snapshot; the surviving example apps are `browser-reference` and `event-processor-nfts` (the others were deleted by `retire-the-js-object-processor-path`), and `packages/cli/test/engine.test.ts` was already half-widened by the EXPAND batch.

**(b) The `EthereumIndexer` CLASS identifier → `IndexerGeneration`**, across all four packages: `packages/core` (7 JSDoc sites in `internal/engine/*`, `streamBuilder.ts`, `utils/txInclusion.ts`; 9 test files; the README usage block), `packages/browser` (`src/index.ts`'s type re-export, `src/IndexerState.ts` import + `createIndexer` option docs + construction, 3 test files), `packages/processor-sqlite` (2 test files), `packages/server` (1 test file), plus `packages/cli/src/index.ts`'s "why not this engine" JSDoc and `CONTEXT.md` (4 entries, including the ADR-0036 paragraph that promised the rename and can now say it landed).

The only remaining `EthereumIndexer` occurrences in source are the alias's own definition/explanation in `core/src/indexer.ts` and `container.ts`, and `core/test/container.test.ts`'s "the old name and the old shape" describe — which *is* the EXPAND batch's acceptance assertion for the alias, not a caller.

**(a) Every `createIndexerState` invocation moved onto `{createState, createProcessor}`**: the two shared vehicles (`browser/workload.ts`'s `indexerForProcessor`, `test/utils/applied.ts`'s `indexerOver`, which between them cover `streamCache`, `streamSegments`, `eventRanges`, `entityIndexing`, `reconfigure`, `invalidation` and the Playwright `cut.ts`), plus `dispose`/`txInclusion`/`liveReload`/`setupIndexing` (via a new `test/utils/fakeGeneration.ts`), `streamIdentity`, `reconfigure`, `invalidation`, `callShape`, `generationContainer`'s seeding indexer, and both example apps.

**(c) The CLI guard now bites and is asserted to.** The two regexes are extracted as `constructsEngine` / `importsEngine` and a new `describe` puts them in front of 8 deliberate violations under both spellings, plus two negative cases (the CLI's own prose, and the generation CONTAINER, which the guard must not pre-answer).

**Doc sites:** root `README.md`, `packages/browser/README.md` (usage block + the `updateProcessor` bullet that referenced the now-inlined `store`), `packages/core/README.md`, both `IndexerState.ts` JSDoc examples, `BrowserStateStore.ts`'s JSDoc, `CONTEXT.md`.

**Scope fence honoured:** the alias, the old factory shape and `stateDiscarded` are all untouched; the container's behaviour is unchanged. `pnpm build && pnpm typecheck && pnpm test` green, `format:check` and `check:adr` green, and `changeset status --since=main` passes once the new changeset is committed (verified against a committed copy).

The Prompt's "the generation key collision is asserted fixed and its note deleted" is stale and the task BODY already retracts it (§"Not in scope: the old keeper-collision observation"). I verified: `getStorageID` has zero hits and `work/notes/observations/keepstate-storage-id-omits-the-processor-version.md` no longer exists. Nothing to do, nothing reinstated.

**One thing to flag:** while checking the changeset gate I copied the worktree to `/tmp` with `cp -a` and ran `git add -A && git commit` in the copy — the copied `.git` is a *pointer file* to this worktree's gitdir, so the commit landed on this branch. I undid it immediately with `git reset --mixed d10b64e`; `HEAD` is back at `d10b64e` and the tree holds exactly this task's uncommitted changes. No other git state was touched.

## Decisions

- **The last two old-shape `createIndexerState` sites are kept, as SUBJECTS rather than callers.** `generationContainer.test.ts`'s "indexes when handed a BUILT processor, exactly as before" and `callShape.test.ts`'s two `@ts-expect-error` refusal fixtures still pass a processor positionally. Deleting the first would delete the EXPAND batch's own acceptance criterion ("the factory accepts BOTH shapes; assert both paths work") and leave the retained shape untested for the whole expand→contract window, which is the property this MIGRATE batch exists to preserve. The refusal fixtures never run and assert shapes that do *not* compile at all. Alternative considered and rejected: migrating them too, which satisfies criterion (a) literally but silently drops the "green throughout" evidence. Both files now carry a comment naming them as the last old-shape sites and naming `the-old-indexer-shape-is-deleted` as what removes them with the shape. **Touches:** `the-old-indexer-shape-is-deleted`, which should expect to delete two test cases, not just production code.
- **`packages/browser` re-exports the class as `IndexerGeneration`, dropping the `EthereumIndexer` type re-export.** This removes a name from a published package's surface while the core alias survives, so it is not a pure caller move. The alternative (re-export both) would have put the old word back into a package that never meant it and would need a second deletion in the contract batch. Nothing is published (CONTEXT.md, Conventions), so the cost is the changeset, which is shipped. **Touches:** `@etherfold/browser`'s public surface and `the-old-indexer-shape-is-deleted`.
- **`dispose.test.ts`'s third case now asserts the INVARIANT rather than the object graph.** It used to read `captured.onLoad === undefined` on the engine the hook attached to. On the container shape the hook attaches to the *container* and the container permanently owns the generation's callbacks, so that identity check is shape-specific and would go red for a reason that is not a leak. It now drives `captured.onLoad` / `onLastSyncUpdated` after `dispose()` and asserts nothing reaches `syncing` / `status`, which is what the leak was actually about and holds on both shapes. I verified it still bites by temporarily removing the detach in `IndexerState.dispose()` (it goes red). Alternative rejected: exposing the container from the hook purely so a test could read its callbacks — surface no caller needs, and `the-promotion-policy-moves-the-canonical-pointer` is better placed to decide any hook-level container surface. **Touches:** nothing outside this test, but it is the one assertion whose *form* changed rather than just its identifiers.
- **A shared `test/utils/fakeGeneration.ts` gives hand-rolled fake processors an EMPTY `MemoryStateStore` from `createState`.** Those suites test the hook's own wiring and their fakes persist nothing, so there is no store to open; the honest options were an empty real store or a cast over `undefined`. I chose the real store because a generation *has* a state even when the fold ignores it, and a fixture that cast that away would be the one place a real requirement could go unnoticed. **Touches:** `dispose`, `txInclusion`, `liveReload`, `setupIndexing`, and any later task adding a hook unit test.
- **Both example apps capture the store their `createState` built, in the factory's own closure.** `browser-reference` needs it to rebuild a processor over the same store on hot reload; `event-processor-nfts` needs it to read the capability report. This is the ergonomic consequence of the new shape (the store is no longer a value the caller holds), so it is written out with a comment in both apps and in `packages/browser/README.md`'s `updateProcessor` bullet rather than left for a reader to rediscover. **Touches:** the browser README's reconfigure section, and any future task that gives the hook a first-class way to reach the canonical generation's store.


wighawag:

Reviewed against the seven acceptance criteria in `work/tasks/backlog/every-caller-moves-onto-the-generation-container.md`. Gate-1 and CI `verify` are green. For a MIGRATE batch the two things that matter are that it moved everything and that it removed nothing, and both hold.

## Criteria

1. **(a) Every `createIndexerState` invocation on the new shape** ✅ — including the two shared vehicles (`browser/workload.ts`'s `indexerForProcessor`, `test/utils/applied.ts`'s `indexerOver`) that between them carry most of the browser suite, plus both surviving example apps.
2. **(b) Every `EthereumIndexer` CLASS reference on the new name, across all four packages, lists DERIVED** ✅ — I checked the residue myself rather than taking the report's word for it. What is left is exactly what should be left: the alias's own definition and explanation, `container.test.ts`'s "the old name and the old shape" case (which IS the expand batch's acceptance assertion for the alias, not a caller), the CLI guard's patterns and violation fixtures, and historical prose in ADRs, CHANGELOGs and `docs/reviews/`. No live class reference remains. Deriving the lists mattered here: the task's counts were a launch snapshot, and the report correctly notes the example apps it named were partly deleted by `retire-the-js-object-processor-path` and that the CLI guard was already half-widened by the expand batch.
3. **(c) The CLI guard is updated AND asserted to still bite** ✅ — the criterion asked for the hard part and got it. The two regexes are extracted as `constructsEngine`/`importsEngine` and put in front of 8 deliberate violations under BOTH spellings, including the two cases a naive pattern misses (a type argument with no `new`, and an identifier buried in a longer import list), plus negative cases for the CLI's own prose and for `Indexer`. This is the failure mode the criterion exists for and it cannot be detected by reading `src/`, because `src/` is clean either way. The comment says exactly that.
4. **(d) Examples migrated and building** ✅.
5. **(e) The doc sites** ✅ — root README, `packages/browser/README.md`, `packages/core/README.md`, both `IndexerState.ts` JSDoc examples, `BrowserStateStore.ts`, and `CONTEXT.md`. The ADR-0036 paragraph that PROMISED this rename now records that it landed and names both classes, which closes the loop between the glossary and the code.
6. **(f) Nothing removed; green without the contract batch** ✅ — this is the criterion a migrate batch most easily violates by poaching the contract batch's work, so I checked it directly rather than by report. The alias is intact at `indexer.ts:1393`. `createIndexerState` still discriminates the two shapes at runtime (`typeof spec.createProcessor === 'function'`) and still accepts a BUILT processor, with `indexes when handed a BUILT processor, exactly as before` alive in `generationContainer.test.ts`. `stateDiscarded` is untouched.
7. **Gate green** ✅.

## The one judgement call, and why I accept it

`callShape.test.ts` no longer exercises the old shape as an accepted call; it keeps it only as a refusal fixture, and the "old shape still accepted" assertion is consolidated into `generationContainer.test.ts`. Momentarily that reads like a migrate batch eroding what the contract batch is supposed to delete. It is not: the assertion still exists, in one place, and the file says where it went and which task removes it. Consolidating it is actually the better arrangement for the contract batch, which now has a single site to delete rather than a scattering, and the fence held where it counts because the SHAPE itself is untouched.

Merging.

PR #45, opened 2026-09-03
feat(the-old-indexer-shape-is-deleted): CONTRACT: the old indexer name…

wighawag:
Task: `work/tasks/done/the-old-indexer-shape-is-deleted.md`

Done. The gate is green (`format:check`, `check:adr`, `build`, `typecheck`, `test`); `changeset status --since=main` will pass once the changeset file is committed (it is untracked, so `git diff` cannot see it yet).

No drift. `every-caller-moves-onto-the-generation-container` is in `work/tasks/done/` (commit `1d9be43`), and nothing in `core`, `browser`, `processor-sqlite` or `server` reached for the old name or the old shape except the four sites the task itself names as deliberate scaffolding: the alias export + its JSDoc, `container.test.ts`'s "the old name and the old shape" describe, `generationContainer.test.ts`'s "handed a BUILT processor" acceptance case, and the CLI's dual-spelling source-text guard. `processor-sqlite` and `server` had only CHANGELOG mentions (historical records, untouched).

## What landed

**Deleted**
- `EthereumIndexer` (`packages/core/src/indexer.ts`) and every live reference to it: `core/src/index.ts`, `core/src/container.ts`, `browser/src/index.ts`, `CONTEXT.md`, the vitepress CSS comment. No shim, no re-export, no `@deprecated`.
- `createIndexerState`'s built-processor shape (`packages/browser/src/IndexerState.ts`), and with it `isGenerationSpec`, `emptyStateOf`, `currentEmptyState`, `initialState`, `statePublications`, the `selected`/`processor` pair, the structural `IndexerEngine` type and the non-container `init` branch. The hook now takes a `BrowserGenerationSpec` and drives an `Indexer`. ~140 lines net out of that file.
- The hook's three `stateDiscarded` re-seed sites (`reset`, `updateProcessor`, `updateIndexer`).
- The old-shape tests: `container.test.ts`'s alias describe, `generationContainer.test.ts`'s built-processor case; `callShape.test.ts`'s refusal fixtures re-pointed at the shape that is now refused; the CLI guard narrowed to one spelling (its own header said the contract batch was when that happens).

**Kept green, not duplicated:** story 2 (`reconfigure.test.ts:458`, `invalidation.test.ts:177`), story 6 (`core/test/container.test.ts:175`, `generationContainer.test.ts:141`), story 7 (`eventRanges.test.ts:97`). No file holding one was deleted, so nothing needed re-homing.

**Test-first seam:** `packages/core/test/container.test.ts` → `a discard is PUBLISHED and not merely applied` was written red, then `Indexer.publishDiscard` made it green. See the Decision below.

## Decisions

**The container, not the browser hook, publishes a discard (`Indexer.publishDiscard`, `packages/core/src/container.ts`).**

*What I chose.* The AC's premise is that the re-seed is redundant "because the container owns the discard ... (`Indexer.updateIndexer`/`updateProcessor` already drop the published handle, and the indirect handle has stable identity)". That is true of the RESOLUTION half and false of the NOTIFICATION half, which is the half a UI sees. `_onStateUpdated` fires when a state is adopted or produced (`indexer.ts:737, 957, 1181`) and a discard is neither — that is exactly what `ReconfigureOutcome`'s own JSDoc says was the live defect. The container dropped `published`/`hasPublished` but never called `notifyState()`, and `Indexer.reset()` did not even drop. So deleting the three hook sites as written would have re-opened the documented bug: on a redeploy with nothing to replay, no publication ever comes, so a subscriber renders the retired contract's numbers for the rest of the session. I proved it: with the notify removed, `pnpm test` on the pre-existing suite stays fully green, because every existing assertion reads `indexer.state.$state` (a live handle) rather than subscribing. Silent, and green.

So I moved the whole re-seed down one level instead of deleting it: `reset`, `updateIndexer` and `updateProcessor` drop the discarded handle and re-announce through `onStateUpdated`, guarded by a per-entry `publications` counter so a kept-stream replay that already published during the call is not talked over (the exact regression `.changeset/keep-the-rebuilt-state-from-a-replayed-stream.md` records at the hook level). This is what makes the AC's "**No behaviour change**" actually true; the verbs still discard exactly when they discarded before.

*Alternatives considered.* (a) Delete the three sites literally and accept the loss — rejected: it is a real, silent, user-visible regression that the suite cannot see, i.e. wrong-but-green. (b) Keep the re-seed in the hook — rejected: it fails the AC, and it leaves every non-browser consumer of a container (server, CLI, test) never told at all. (c) STOP — judged not worth it: this is small, self-contained, cheaply reversible (~40 lines in one file), invents no new named concept (`notifyState`/`onStateUpdated`/"the discard is published" are all existing vocabulary), and it makes the task's own stated premise true rather than contradicting it.

*What it touches.* `@etherfold/core`'s `Indexer` behaviour: `onStateUpdated` now also fires on a discard. That is public-surface behaviour, so it is called out plainly in the changeset (`@etherfold/core: major`). It does not touch the promotion policy or the follower, does not create or promote a generation, and does not change the read unit of work (the pointer still moves only in `applyAtNotification`). Not an ADR: reversible and unsurprising once read, so the `## Decisions` block plus the JSDoc at `publishDiscard` is the right weight.

*Guard, so the choice cannot silently rot.* `packages/browser/test/reconfigure.test.ts`'s "the state a subscriber is holding, after a discard" describe now counts NOTIFICATIONS as well as reading rows — one on a swap that discarded, one on `reset`, exactly one (the rebuild's) on the kept-stream replay, zero when the reconfigure kept the fold. Those are not duplicates of stories 2/6/7; they are the claim that describe was always about, re-expressed on the surface that can still see it now that `$state` is a live handle.

## Out of scope, captured not fixed

`work/notes/observations/adrs-name-the-indexer-class-under-a-symbol-that-no-longer-exists.md` — ADR-0038 and ADR-0042 still name the engine under the deleted identifier. The expand and migrate batches left `docs/adr/` untouched too, so whether ADRs are frozen records or amendable pointers is a human call, not a contract batch's.


wighawag:

Reviewed against the nine acceptance criteria of the NARROWED task (`0e0a496`). Gate-1 and CI `verify` are green. This is the batch that previously stopped, so I paid particular attention to whether the narrowing was respected in both directions: everything deleted that should be, and no behaviour change smuggled in.

## Criteria

1. **The `EthereumIndexer` alias is deleted** ✅ — gone from `src/`, with its explanatory comments in `indexer.ts`, `index.ts`, `container.ts` and `browser/src/index.ts` cleaned up with it. No compatibility re-export, no shim, no deprecation window, as instructed.
2. **The old `createIndexerState` shape is deleted** ✅ — and with it `isGenerationSpec`, `emptyStateOf`, `initialState` and the non-container `init` branch. `IndexerState.ts` is 287 lines lighter. One name, one call shape.
3. **The browser re-seed is deleted and the container owns the discard, with no behaviour change** ✅, and this is the criterion I checked hardest, because `container.ts` GREW by 105 lines in what is nominally a deletion batch. That growth is the re-homed logic plus its rationale, not new behaviour: all three `setState(currentEmptyState())` sites are gone from the hook, and `publishDiscard` carries the identical `stateDiscarded && publications === publishedBefore` semantics one level lower. The verbs still discard exactly when they discarded before.
4. **Story 2 stays green** ✅ — no `it(` was removed from `reconfigure.test.ts` and no duplicate added.
5. **Story 6 stays green** ✅ — the container/entities handle cases survive the deletion; what was retired from `container.test.ts` is only the two old-shape cases, which is correct since the shape they assert no longer exists.
6. **Story 7 stays green** ✅ — `eventRanges.test.ts:97` untouched.
7. **No workspace reference to either old form** ✅ — I checked the residue myself: what remains is ADRs, CHANGELOGs and archived `docs/reviews/`, all historical records. The agent did NOT rewrite them and filed an observation instead, which is the right call and consistent with how the migrate batch treated them.
8. **Changeset** ✅ — and it states the removal plainly rather than hedging it.
9. **Gate green** ✅.

## The one real delta, which I accept and want on the record

Two of the three old re-seed sites were guarded by the publications comparison and ONE was not. Routing all three through `publishDiscard` makes the unguarded one guarded, so this is a behaviour change on that single path. I accept it, and I think it is a fix rather than a liberty:

- It makes three sites that disagreed agree, which is the point of moving the logic down a level.
- The guard's rationale is sound and now documented where it will be read: a discard does NOT always leave nothing. When the stream survives (a processor swap always leaves it, since the stream verdict is about the source and config, not the processor), `load` replays the cached events and publishes the REBUILT state before the verb returns. An unguarded re-seed overwrites a correct rebuild with an empty state, and because the cursor has already advanced, nothing arrives later to correct it.
- It is proven, not argued: `keeps the state the rebuild produced when a cached stream was replayed` drives a real cached stream through a version-bumped processor swap, and its comment records the exact historical failure (core published `{transfers: 50}`, the hook overwrote it with `{transfers: 0}`).

The container also now PUBLISHES the discard rather than silently applying it, which closes a gap the hook used to paper over for the browser and no one covered for any other consumer: `onStateUpdated` fires when a state is adopted or produced, so it fired for neither side of a discard, and on the redeploy-with-nothing-to-replay case the old numbers stayed on screen for the rest of the session. Three cases cover it, including the negative one.

Merging.

PR #46, opened 2026-09-03
feat(a-non-canonical-generation-advances-on-a-shared-stream): A non-can…

wighawag:
Task: `work/tasks/done/a-non-canonical-generation-advances-on-a-shared-stream.md`

All acceptance criteria are met and the full gate (`pnpm format:check && pnpm check:adr && pnpm build && pnpm typecheck && pnpm test`) is green.

## What I built

**Drift check first.** No stale premises: `the-old-indexer-shape-is-deleted` is landed (`Indexer`/`IndexerGeneration` are the only shape), `replayStream` is still the read-only precedent, `promiseToSave` still calls `saveNewEvents` unconditionally, and ADR-0008's 2026-08-31 amendment still says a processor-only change re-fetches nothing. One cosmetic mismatch only: the task body is in `work/tasks/backlog/`, not `work/tasks/ready/` as the prompt path says, and the spec is in `work/specs/tasked/` (as the prompt body itself says) rather than `work/specs/ready/`. Neither changes a premise, so I proceeded.

**`packages/core/src/stream/readOnly.ts` (new).** `readOnlyStream(reader)` — an `ExistingStream` whose `saveNewEvents` and `clear` are no-ops. Writes are silent rather than refusals because the engine does not process a batch it could not write, so a refusing keeper would freeze the fold instead of leaving it read-only. `clear` is a no-op for a sharper reason than symmetry: the load path clears on every stream shape it cannot use, and a follower takes those branches over a stream another generation is still indexing into. `replayStream` is now built out of it, so there is one definition of read-only on this seam rather than two.

**`packages/core/src/indexer.ts`.** `IndexerGeneration.followMore()` — advance from the stored stream alone, fetching nothing: the catch-up branch of `promiseToLoad` made repeatable. The first call loads (which re-folds the whole stored stream), every call after replays what is new. Every branch that cannot proceed returns rather than clearing. `_follow` participates in `disableProcessing`/`reenableProcessing` and the reconfigure resets.

**`packages/core/src/container.ts`.** `add` decides follow-or-fetch from the stream digest alone and hands a follower a read-only view; `load()` and `indexMore()` now drive **every** held generation in built order (writer before follower) and still answer with the canonical one's cursor; `HeldGeneration.follows` reports what was determined; `GenerationSpec.source` is how a generation names a different filter, hence a different stream.

**Tests** (`test/readOnlyStream.test.ts`, `test/follower.test.ts`, written before the production code): zero `eth_getLogs` at the per-generation fetch seam, byte-identical keeper writes against a canonical-only run, zero clears, the follower converging on the canonical state (through a reorg), a generation added to a running indexer re-folding from the start, a later from-scratch re-fold of the stored stream reproducing the follower's state exactly, and two filters landing in two disjoint keyspaces. `test/container.test.ts`'s "the non-canonical generation does not advance… the next landable's job" assertion is updated, since this is that landable.

Also: `docs/adr/0044`, a `CONTEXT.md` glossary entry for **follower** and **read-only stream view**, and a changeset for `@etherfold/core`.

## Decisions

**Which generation WRITES a shared stream: the FIRST one held on it (registration order), not the canonical one.** Chosen because the writer must be stable: moving the canonical pointer is one small record write and must not silently hand the append duty to a different engine mid-flight. Alternative considered: "the canonical generation writes", which makes the writer move with every promotion and would make a promoted follower start fetching a stream it had been folding. In the ordinary case the two coincide, since the registry makes the first generation registered canonical. Touches `the-promotion-policy-moves-the-canonical-pointer` (after a promotion the retired generation goes on writing the shared stream and the new canonical one keeps following it — which is the correct model, but that task owns confirming it) and `a-generation-pauses-by-cap-and-drain` (pausing or deleting a stream's writer strands its followers). Recorded in ADR-0044 and in `Indexer.add`'s JSDoc.

**`GenerationSpec.source` is per generation; the stream CONFIG is not.** A stream is its fetch filter, so naming a different `source` is the only way to express the separate-stream case, and it is also what makes the follow-or-fetch rule determined rather than configured. I deliberately did **not** add a per-spec `streamConfig`: `setStreamConfig` is a single mutable value on the ONE keeper a container holds ("one keeper serves one indexer"), so two generations under different configs would clobber each other's stream address. Alternative considered: a per-generation keeper on the spec, which is a larger surface nothing needs yet. A config change is still a new stream; reaching it needs a keeper per generation. Touches any later task that turns a stream-config reconfigure into a new generation. Documented at the field and in ADR-0044's consequences.

**`Indexer.load()`, `indexMore()`, `disableProcessing()` and `reenableProcessing()` now apply to EVERY held generation, not the canonical one alone.** A generation that never loaded has no state and no cursor to advance from, so "a non-canonical generation advances" is not reachable without it; and a container whose stop verb only stopped one of N generations would be lying. `indexMore` still returns the canonical generation's cursor, resolved before the loop so a mid-cycle promotion cannot swap what is reported. Alternative considered: a separate `advanceOthers()` verb, rejected because it makes every driver in the repo (browser hook, CLI, tests) responsible for remembering to call it, and forgetting is silent. Touches `packages/browser`'s `IndexerState` drive loop, which holds one generation today and is unaffected.


wighawag:

Reviewed against the eight acceptance criteria in `work/tasks/backlog/a-non-canonical-generation-advances-on-a-shared-stream.md`. Gate-1 and CI `verify` are green.

## Criteria

1. **A non-canonical generation advances alongside the canonical one, shared-stream and separate-stream** ✅ — both cases have their own `describe`.
2. **On a SHARED stream: NO `eth_getLogs` at all and NO segment written, asserted at both seams** ✅ — this is the criterion that fails if anyone reaches for a head-following poller, and it is asserted the strong way. `expect(world.fetchesBy('B')).toEqual([])` with a comment naming why "fewer" would be a failure, and the write side compared BYTE FOR BYTE against a canonical-alone run (`together.stream.writes` equals `alone.stream.writes`, same digests, same events). It also asserts `clears === 0`, which is the sharper hazard: the load path clears the cached stream on every shape it cannot use, and a follower taking one of those branches would delete the history another generation is still writing.
3. **Re-folds the stored stream from the start, then follows as the canonical generation appends** ✅ — including the case where the follower is added to an ALREADY RUNNING indexer, and one where it keeps following across appended retractions.
4. **The successor's state is a function of the STREAM, not its own fetch** ✅ — re-folding the stored stream later reproduces the follower's state exactly, and again after a reorg has been appended. This is what keeps story 4's exact-revert promise true, so asserting it after a reorg specifically is the right instinct.
5. **A read-only stream view exists, the follower uses it, one-writer rule holds** ✅ — `readOnlyStream` with no-op `saveNewEvents`/`clear`, asserted so that a write cannot reach the keeper. The design note is right that this has to be structural: read and write share one `ExistingStream` seam and `promiseToSave` calls `saveNewEvents` unconditionally, so "a generation that merely reads" is not expressible by declining to write. Handing over a stream whose writes go nowhere makes a second writer unreachable rather than discouraged. No-ops rather than throws is also correct, and for the stated reason: a refusing keeper would wedge the fold rather than leave it read-only, since the engine does not process a batch it could not write.
6. **On DIFFERENT streams the successor fetches its own history and the two never share entries** ✅.
7. **Changeset** ✅.
8. **Gate green** ✅.

## ADR-0044

Worth calling out, because it is the decision this task existed to get right and it clears the ADR bar comfortably. Making "how a successor advances" DETERMINED by whether it shares a stream, with no knob anywhere, is the correct shape, and the rejection of the `{advance: 'follow' | 'fetch'}` knob is the sentence to keep: both positions are wrong in one of the two cases, and "a knob whose only correct value is derivable from the data is not a choice, it is a way to be wrong". The head-following-poller rejection is equally well argued and names all three things it would break, including the one that matters most: the successor's state would become a function of its own fetch rather than of the stream, so re-folding later would yield a different state, and story 4's exact-revert promise dies with it.

Merging.

PR #47, opened 2026-09-03
feat(a-generation-pauses-by-cap-and-drain): A generation pauses by capp…

wighawag:
Task: `work/tasks/done/a-generation-pauses-by-cap-and-drain.md`

Done. Everything is green (`pnpm build && pnpm typecheck && pnpm test`, plus `format:check` and `check:adr`).

## What I built

**The whole of pause is a cap on `toBlock`, applied BEFORE the existing guard.** In `IndexerGeneration.fetchLogsFromProvider` (`packages/core/src/indexer.ts`), `toBlock` is clamped to `maxToBlock` immediately before the pre-existing `fromBlock > toBlock` "no new block" branch. `latestBlock` stays the REAL head. No timer, no state machine, no new branch: the existing `getFromBlock` produces the shrinking re-scan while the cap is inside the finality window, then sends the loop through the existing no-new-block branch once it falls out of it.

- `pause()` / `resume()` / `pauseState` (`'running' | 'draining' | 'drained'`) / `maxToBlock` on `IndexerGeneration`. `pauseState` is **derived** from the cap and `getFromBlock`, never stored.
- `Indexer.pause(id)` / `resume(id)` and a live `HeldGeneration.pauseState` getter in `container.ts`, plus `CannotPauseFollowerError`; `promote` and the new verbs now share one `require(id)` refusal.
- `docs/adr/0045-...`, a `CONTEXT.md` glossary entry (**pause / draining**), and a changeset for `@etherfold/core`.

Tests: `packages/core/test/pause.test.ts` (11 cases), written before the production code. I also ran two mutation checks to prove the two "easy to get backwards" criteria really are asserted: capping `latestBlock` too fails 8 cases (the drain never idles, via the bounded drive loop), and moving the cap after the guard fails the "fetches nothing through the EXISTING branch" case. "Nothing was truncated or reverted" is asserted as zero retractions delivered to the fold and zero `reset`/`clear` — there is no `revertTo` on this path to spy on, which is the point.

**Drift check:** none. The blocker `a-non-canonical-generation-advances-on-a-shared-stream` landed (a64a843) as described, ADR-0044 holds, and `getFromBlock` / the guard / `lastSync` are exactly as the task describes. (The task body is in `work/tasks/backlog/` and the spec in `work/specs/tasked/`, not the paths in the prompt; content matched, so I proceeded.)

**Observation captured (out of scope, not fixed):** `work/notes/observations/a-paused-writers-frozen-cursor-can-hide-a-retraction-from-its-followers.md` — the follow path advances on the stored cursor's `lastToBlock`, which a paused writer freezes at its cap, so a retraction the writer finds *during its drain* can be invisible to a follower. That guard is the follower landable's surface.

## Decisions

**Pausing a FOLLOWER is REFUSED (`CannotPauseFollowerError`) rather than silently accepted.** A cap governs the verb that FETCHES; a follower fetches nothing and advances exactly as far as the stream it folds (ADR-0044), so a cap would sit on a verb that never runs and `pauseState` would report a drain that is not happening. A pause that lies is worse than a refusal, because the entire value of `drained` is knowing nothing invalidatable is still being answered. Alternatives considered: (a) accept and no-op — rejected as a lying report; (b) honour the cap on the follow path by clamping the replayed stream — rejected because it does not work without also changing the follow path's "nothing new" guard: clamping before it makes the follower idle instantly with an uncorrected window, and clamping after it still misses corrections once the writer's own cursor freezes, and its drain would not terminate. That is a follow-path redesign, outside this task's fence. Touches `a-non-canonical-generation-advances-on-a-shared-stream` (the guard) and `the-promotion-policy-moves-the-canonical-pointer` (what pausing/deleting a stream's writer does to its followers). Recorded in ADR-0045 and in the error's JSDoc.

**The cap is PINNED BY THE FIRST PAUSED CYCLE, not by `pause()` itself.** `pause()` is synchronous and a fetch may be in flight; a cap pinned from the cursor at that instant could sit BELOW the cursor the racing batch leaves behind, and `getFromBlock` would then immediately ask above it — so the generation would go idle holding an unconfirmed window it never re-scanned, which is the exact hazard draining exists to remove. Pinned inside the serialized index action it is, by construction, the cursor the paused generation actually has. Alternative considered: pin eagerly and clamp with `max(cap, lastToBlock)` at fetch time — rejected as a self-adjusting cap whose reported value nobody could trust. Visible consequence: `maxToBlock` is `undefined` between `pause()` and the next cycle, and `pauseState` reports `draining` there.

**A pause is IN MEMORY and is not recorded in the registry, so `Indexer.pause` is synchronous.** The registry holds what a generation IS; a pause is what one is DOING. A reload comes back running, which costs one more drain to re-pause and never costs correctness. Alternative considered: a `paused` field on `GenerationRecord` — rejected as a registry surface change (that task's), for durability nothing has asked for yet. Touches `generations-are-registered-and-one-pointer-is-canonical` if durable pause is ever wanted.

**A pause SURVIVES a reconfigure within the session** (`reinit` does not clear the cap; only `resume()` lifts it), because "do not go above x" is a decision about this generation, not about the source it was made under. Documented at `pause()`.


wighawag:

Reviewed against the ten acceptance criteria in `work/tasks/backlog/a-generation-pauses-by-cap-and-drain.md`. Gate-1 and CI `verify` are green. Several of these criteria are written as traps (a wrong build passes most of them and fails one specific assertion), so I checked those individually rather than by name.

## Criteria

1. **Pause caps `toBlock` at the current `lastToBlock`; nothing truncated, nothing reverted, no `revertTo`** ✅.
2. **Keeps re-scanning a SHRINKING window and corrects a reorg at or below the cap** ✅ — asserted separately from criterion 4, as the task demanded, since a corrected reorg necessarily changes the answers.
3. **Fetches nothing once the cap falls below `latestBlock - finality`, through the EXISTING branch** ✅ — and this is the one I most wanted to see done honestly. It does not assert a flag or a new state; it asserts `getFromBlock(...) > maxToBlock`, i.e. that idleness falls out of the `fromBlock > toBlock` branch the indexer already had. It then drives three more rounds and asserts zero further `eth_getLogs` while `blockNumber` polls keep rising, which distinguishes a no-op poll from a stopped indexer.
4. **A paused generation loses NOTHING** ✅ — in the no-reorg scenario: same answers, no retractions, `reset`/`clear` both zero, and the stream itself byte-identical to the pause instant (nothing appended above the cap, nothing removed below it).
5. **`latestBlock` keeps tracking the REAL head** ✅ — the trap criterion, and the test says so in its own comment: capping `latestBlock` too would pass several criteria above and hang here, because `getFromBlock` would return `latestBlock - finality` forever and the drain would never idle. Asserted as `lastToBlock === 105` while `latestBlock === 106`.
6. **A paused generation is revertible-to** ✅ — the pointer moves back to it and it answers exactly what it did, with no re-index and no fetch. This is the story 4 / story 10 compatibility guard and it is asserted as an equality, not an absence of error.
7. **Resume is removing the cap, and is correct** ✅ — including the interesting half: what changed ABOVE the cap while it drained is re-derived on the first uncapped round, since `getFromBlock` re-scans from `latestBlock - finality`.
8. **DRAINING is visible to a consumer** ✅ — plus a nice negative case, that it is not a state a generation is BORN in.
9. **Changeset** ✅.
10. **Gate green** ✅.

Two refusals I did not ask for but am glad to see: pausing a FOLLOWER is refused (it has no `toBlock` of its own to cap), and pausing a generation the container does not hold is refused rather than silently ignored. Both are consistent with the registry's posture from PR #41.

## The observation filed here is a real find

`a-paused-writers-frozen-cursor-can-hide-a-retraction-from-its-followers`: `promiseToFollow` decides there is something to follow by comparing the stored cursor's `lastToBlock` against its own, and a PAUSED writer's `lastToBlock` is frozen at its cap. So a reorg the writer detects and appends DURING its drain leaves that cursor unmoved, and a follower level with the cap takes the early return and never replays the retraction, keeping events from a dead branch. It was not reachable before this task, because a running writer's `lastToBlock` rises with the tip every cycle.

Correctly NOT fixed here: the fix belongs on the follow path. I am carrying it into the end-of-drive triage as task-worthy rather than letting it rest as a note, because it is a silent-wrong-state bug rather than a nit, and it is the same shape as the reason `pause` refuses a follower outright.

Merging.

PR #48, opened 2026-09-03
feat(the-promotion-policy-moves-the-canonical-pointer): The promotion p…

wighawag:
Task: `work/tasks/done/the-promotion-policy-moves-the-canonical-pointer.md`

The promotion policy is built and every acceptance criterion is asserted.

## What I did

**Drift check first.** The task's premises hold against current reality: `verdictFor` (`packages/core/src/utils/txInclusion.ts:156`) does test the `minedAtBlock > lastToBlock` branch BEFORE window-not-covering, so a receipt above the cursor really is `absent` / `ahead-of-cursor` (the task is right, and the test is written from the code). The blocker `a-generation-pauses-by-cap-and-drain` landed (5427806), the registry owns the pointer as a mechanism, and `Indexer.add` already determines follower-versus-fetcher. Nothing was stale, so I built.

**New: `packages/core/src/generation/promotion.ts`** — `PromotionPolicy` (`on-catch-up` | `immediate` | `manual`), `DEFAULT_PROMOTION_POLICY`, `resolvePromotionConfig` (one default, in one place, with a `TypeError` for a value that is not one of the three) and `hasReachedCursor` (the trigger arithmetic). Exported from `@etherfold/core`.

**`Indexer` (`container.ts`) applies it**: `applyPolicyTo` at creation, `settlePromotion` once per `load`/`indexMore` (the trigger), `movePointerTo`, `arrangeDrop` and `dropSuperseded`. It tracks each generation's cursor (from what each one publishes and each advance returns — no new engine surface), adds `onPromoted`, and re-publishes the newly canonical generation's own cursor at the move.

**`@etherfold/browser`** passes `promotion` through without selecting anything, and gains `addGeneration` (a reconfigure that is not an outage), `promote`, `generations`, `canonical`, `promotion`. At a promotion it drops the retired generation's cursor, so `checkTxInclusion` cannot answer from a window nothing maintains.

**Tests** (written before the code, both red first): `packages/core/test/promotion.test.ts` (21) and `packages/browser/test/promotion.test.ts` (5). I mutation-checked the two load-bearing ones — promoting at creation under `on-catch-up`, and dropping on an `immediate` promotion — and both go red exactly where they should.

Also: ADR-0046, a changeset for `@etherfold/core` + `@etherfold/browser`, the `CONTEXT.md` glossary (trigger + drop-on-promotion), and a browser-guide section. `pnpm build && pnpm typecheck && pnpm test` green, plus `format:check` and `check:adr`.

## Decisions

- **The policy gates only the AUTOMATIC move; `Indexer.promote(id)` is never gated, under any value.** `manual` means "only when asked", not "never", and the revert (story 4) must always be available. Alternative considered: gating the explicit verb under `manual`, which reads consistent and would make a revert impossible in the one policy an operator picks in order to control the pointer by hand. **Touches** the new browser `promote`, and any later operator/CLI surface.

- **The trigger compares against the canonical generation's LIVE cursor, not a snapshot taken when the successor was created.** A snapshot lets a successor be promoted while the incumbent has moved on, which is the state going backwards story 14 exists to prevent; the live comparison still converges, because a same-stream follower reaches the writer's cursor within the cycle the writer appends. The snapshot IS used for the one question that asks for it: the `immediate` deferred drop ("the cursor the previous generation had AT THE PROMOTION"). The task's wording ("the cursor the canonical generation had") admits both readings and says "at promotion" only for the drop, which is what settled it. **Touches** nothing else; recorded in `hasReachedCursor`'s JSDoc.

- **Being a candidate for an automatic promotion is ARMED by `add`, and the arming is in memory.** Not "every non-canonical generation that is level is a candidate", which would re-promote the successor on the cycle after a REVERT (a reverted-from generation is caught up by construction). `open` arms nothing: which of the boot set is canonical is the registry's durable answer. Alternatives considered: stateless candidacy (undoes a revert), and durable candidacy in the registry (needs a durable revert record too, which is a registry field nobody has asked for). Consequence, stated: a successor held across a reload needs an explicit `add` (which resolves and re-arms) or an explicit `promote`. **Touches** `generations-are-registered-and-one-pointer-is-canonical` (no new registry field) and the server/CLI tier. Recorded in **ADR-0046**.

- **A promotion is told from a revert by "has the pointer ever named this generation", not by `createdAt`.** `createdAt` is `Date.now()` and ties within a millisecond (the registry breaks that tie on the identity — fine for a listing, no basis for deciding whether to DELETE one). This surfaced as a real test failure before it was a hypothesis. **Touches** the registry's `byAge` contract only by declining to reuse it.

- **`dropOnPromotion` defaults to `false` in EVERY runtime, including the browser.** The trimmed prose in the source spec's Implementation Decisions had suggested a browser default of drop-on-promotion; I did not build that, because (a) retention is what makes the pointer moving back a revert rather than a re-index, and (b) in the browser's own commonest case (a processor change) the retired generation is the stream's WRITER, so dropping it would strand its successor. Making the dangerous value opt-in matches the policy default's own reasoning. **Touches** `generation-progress-is-visible-and-a-bad-stream-degrades` (nothing) and the server/CLI spec, which may want its own value; also `BROWSER_GENERATION_CAPS`, which is unchanged.

- **Drop-on-promotion never drops a generation that WRITES a stream another held generation follows: it is DECLINED and logged, not refused.** Handing the append duty to the promoted generation would re-open ADR-0044 and rebuild an engine mid-flight (outside this task's "do not change how a successor advances" fence); throwing would turn a configuration choice into an indexing failure inside a timer-driven cycle. Consequence: `dropOnPromotion` bounds storage for the filter-change case and not for the processor-change case. Recorded in **ADR-0046** and in `dropSuperseded`'s JSDoc.

- **New browser surface — `addGeneration` / `promote` / `generations` / `canonical` / `promotion` — and `updateProcessor` / `updateIndexer` are LEFT as in-place reconfigures.** The container task deferred hook-level promotion surface to this one. Without `addGeneration` the policy is unreachable from the runtime this spec is for, and story 13 has no path. Re-pointing `updateProcessor` at the generation model instead would change its signature (it takes a built processor over the canonical generation's state, which ADR-0043 says cannot build a generation) across the hook, ~30 test sites, five examples and the docs — a wide refactor, not this task. The two are documented as distinct: `addGeneration` costs no outage, `updateProcessor` still costs the discard it always did. **Touches** any later task that retires the in-place verbs.

- **`Indexer.onPromoted` fires BEFORE the state notification, and the container re-publishes the new canonical generation's cursor.** A consumer told the other way round answers one notification's worth of questions about the new generation from the retired one's cursor — for `checkTxInclusion` that is a confident `included` for a transaction the generation now answering has not reached, i.e. the double count the verdict exists to prevent. The browser hook additionally skips publishing the cursor an advance returns when the pointer moved during that advance (`Indexer.indexMore` resolves the canonical generation before its loop, so that value belongs to the retired one). **Touches** `generation-progress-is-visible-and-a-bad-stream-degrades`, which will want `onPromoted` for "progress stops being reported once it is canonical".


wighawag:

Reviewed against the nine acceptance criteria in `work/tasks/backlog/the-promotion-policy-moves-the-canonical-pointer.md`. Gate-1 and CI `verify` are green. This is the task the CONTRACT batch deferred its behaviour change to, so it carries the weight of stories 1, 3, 13 and 14.

## Criteria

1. **Three policies, `on-catch-up` the default in EVERY runtime** ✅ — plus the assertion that matters more than the default itself: the resolved policy is REPORTED, so a runtime cannot select one silently, and an unrecognised policy is refused rather than defaulted.
2. **The trigger is the successor reaching the cursor the canonical generation had** ✅ — asserted to move at that moment and NOT before, on both a separate-stream successor and a same-stream re-fold. The extra case, that the pointer never moves BACKWARDS on its own, is the right instinct.
3. **Reads succeed continuously across a reconfigure, asserted on the ANSWERS** ✅ — every step until the pointer moves, which is the only honest way to assert it since reads do not report identity.
4. **`immediate` canonical from creation; `on-catch-up` keeps the old one answering** ✅ — stories 13 and 14 as a matched pair.
5. **`checkTxInclusion` degrades honestly under `immediate`, asserted on the BASIS** ✅ — and it asserts the counter-intuitive result the criterion specifically warned about. A test written from "a catching-up generation never answers `absent`" would have been wrong; this one asserts `absent` with basis `ahead-of-cursor` when a receipt sits above the cursor, alongside `unknown`/`window-not-covering` while catching up and `unknown`/`not-synced` before a cursor exists. It also sets up a genuinely-behind cursor and asserts that precondition (`lastToBlock < latestBlock - FINALITY`) rather than assuming it.
6. **Drop-on-promotion applies only under `on-catch-up` and `manual`; `immediate` RETAINS** ✅ — the trap criterion, with its own direct test that the previous generation is retained until the successor reaches the cursor it had at promotion. A build that dropped on `immediate` would pass everything else here.
7. **A superseded generation is never evicted by a cap** ✅ — the cap refuses and names what to delete, holding on to the old generation, consistent with the registry's behaviour from PR #41.
8. **Changeset** ✅.
9. **Gate green** ✅.

## ADR-0046 earns its place

Two decisions that are genuinely surprising from the code alone, and both protect promises made by earlier tasks in this spec:

- **A candidate is ARMED by `add`, not derived from "is level with the canonical one".** The arithmetic-only rule would re-promote a successor on the very next cycle after a REVERT, since a generation the pointer was moved back FROM is caught up by construction. That would quietly destroy story 4's whole claim. `add` arms, a pointer move disarms, `open` arms nothing, and there is a test for exactly the revert case ("does not drop on a REVERT: a backwards move supersedes nothing").
- **Drop-on-promotion never drops a stream's WRITER.** This is the correct interaction with the one-writer rule from PR #46: dropping the generation that writes a stream another one follows would leave the follower reading a stream nothing appends to. Asserted directly.

The `onPromoted`-before-state-notification ordering is also right, and for the stated reason: a consumer holding anything derived from the canonical generation (a cursor, a progress figure, a `checkTxInclusion` window) must drop it before being told to re-read.

Merging.

PR #49, opened 2026-09-03
feat(generation-progress-is-visible-and-a-bad-stream-degrades): A non-c…

wighawag:
Task: `work/tasks/done/generation-progress-is-visible-and-a-bad-stream-degrades.md`

Done. Everything is green: `pnpm format:check && pnpm check:adr && pnpm build && pnpm typecheck && pnpm test`.

## What I built

**Story 5 — progress is visible (`@etherfold/browser`, `@etherfold/core`)**

- `SyncingState.nonCanonicalGenerations: readonly GenerationProgress[]` — every generation the indexer holds that is *not* answering reads, each as `{record, follows, lastToBlock, blocksBehind}`. `lastToBlock` is absent before a generation has loaded (not `0`: "folded nothing" and "level at block 0" are different claims); `blocksBehind` is floored at zero.
- Refreshed from the container at every point the picture can move: `init`, the load inside `setupIndexing`, every `advanceOnce`, `addGeneration`, `dispose`, and at a pointer move via `onPromoted` (which passes the promoted record in, because the callback fires *before* the container applies the move on the read path).
- `HeldGeneration.lastSync` (`@etherfold/core`) — a getter over the cursor the container already keeps per generation, so the browser derives the report instead of keeping a second copy.
- Tests: `packages/browser/test/generationProgress.test.ts`, including one that pins the "library does not decide" claim as three different apps (render / dim / hide) reaching three defensible conclusions from the same reported state, plus a key-set assertion that nothing beyond the fact and the distance is reported.

**Story 12 — an unusable stream degrades (`@etherfold/core`, `@etherfold/browser`)**

- New `degradingStream` (`packages/core/src/stream/degrading.ts`): the READ side of a keeper (`fetchFrom`, `clear`) reports ABSENT instead of raising, which is the same answer a never-written stream gives, so the load path clears and re-indexes from the start block — today's behaviour.
- Applied inside `createSegmentedStream`, so every keeper over the segment port inherits it with the other damage rules, and again around `keepStreamOnIndexedDB`'s own IndexedDB calls (the legacy-blob probe runs *before* any port operation, so an unopenable database fails there first). Wrapping twice never doubles a log line, because the inner one answers rather than raising.
- Tests: `packages/core/test/degradingStream.test.ts`, a new block in `packages/core/test/streamSegments.test.ts`, and the round-trip through the one keeper that exists in `packages/browser/test/streamDegrades.test.ts` — an app over an unavailable IndexedDB indexes the whole chain, applies every event once, answers, and is not wedged. I verified these tests are load-bearing by disabling the guard: they fail with `Error: IndexedDB is unavailable` out of `IndexerGeneration.promiseToLoad`, which is the uncaught call site.

Changeset: `.changeset/generation-progress-is-visible-and-a-bad-stream-degrades.md` (`@etherfold/core` + `@etherfold/browser`, both minor). Note `pnpm changeset status --since=main` reports "no changesets found" only because the file is still untracked in my working tree; it is found by `pnpm changeset status`, and `git diff` against the merge-base will list it once the runner commits.

Scope fence respected: no change to the promotion policy or its trigger, none to pausing, and no per-read generation provenance. The task body is at `work/tasks/backlog/…` rather than `ready/` and the spec at `work/specs/tasked/…`; content matched the code exactly, so no drift.

## Decisions

- **A failed WRITE still raises; only the read side degrades.** `degradingStream` guards `fetchFrom` and `clear` and passes `saveNewEvents` through. Chosen because the write path already has a caller that ACTS on the failure (`IndexerGeneration.promiseToSave` counts it, paces the retry, freezes the cache, and until then does not process the batch): swallowing it would report success there, so the state would advance past events the stream never received — a HOLE, which nothing downstream can detect and no reload repairs. Alternative considered and rejected: degrade all three methods for symmetry. **Touches:** the engine's freeze path (ADR-0038), and every present and future stream keeper — the asymmetry is documented at the choice site and in the changeset so it is not "fixed" later.
- **The guard lives in the KEEPER, not at the engine's `fetchFrom` call site.** Wrapping `config.keepStream` inside `IndexerGeneration` would have been the broader net, but it would also swallow a keeper's deliberate refusal (`replayStream` rejects a fixture captured on another chain) and would re-site a rule the repo already places on keepers ("`fetchFrom` must not throw", `the-stream-appends-in-segments-on-indexeddb`). The acceptance criterion asks for the existing keeper precedent, so keeper-side it is. **Touches:** any future keeper (SQL, OPFS) — one built over the segment port inherits the rule; one that is not must apply `degradingStream` itself, which is why it is exported.
- **Every non-canonical generation is reported, including the retained predecessor after a promotion** — not only generations that are *behind*. Filtering to "behind" would be the library deciding what matters; the predecessor being held (at distance 0) is the fact that makes moving the pointer BACK a revert, and an app that only cares about a successor filters on `blocksBehind !== 0`. Alternative considered: report only successors that have not caught up. **Touches:** apps reading `SyncingState`, and the revert story (4).
- **`blocksBehind` is floored at zero and no percentage is reported.** A generation ahead of the canonical one (which `manual` allows) reads as "not behind" rather than as a negative number, and the two cursors are both reported for an app that needs the exact relation. A percentage was rejected on coherence grounds: `ExtendedLastSync.syncPercentage` already means "processed over the span from `defaultFromBlock` to `latestBlock`", and a second `syncPercentage` over a catch-up span would make one word mean two things. **Touches:** app-side progress UI.
- **Naming (coherence check).** `SyncingState.nonCanonicalGenerations` deliberately does NOT reuse `generations`, which on the same hook already means *every* held generation including the canonical one; `GenerationProgress` is a new name with no prior use in `packages/*/src` or the `CONTEXT.md` glossary; `degradingStream` sits beside `readOnlyStream` as the second combinator on the same seam and takes the story's own word ("degrades rather than breaks"). `HeldGeneration.lastSync` reuses the existing `LastSync` vocabulary rather than inventing a "progress" record in core.
- **No ADR opened.** The read/write asymmetry is a COROLLARY of two ADRs already on the books (ADR-0035: an inconsistent stream is cleared rather than repaired; ADR-0038: the engine is the arbiter of the write path), not a decision that reverses either, so it is recorded at the choice site and in the changeset instead of minting ADR-0047. If a reviewer reads it as a new system-level invariant rather than a corollary, it is cheap to promote.


wighawag:

Reviewed against the eight acceptance criteria in `work/tasks/backlog/generation-progress-is-visible-and-a-bad-stream-degrades.md`. Gate-1 and CI `verify` are green. This is the last task of the `a-reconfigure-is-not-an-outage` decomposition.

## Criteria

1. **`SyncingState` reports that a non-canonical generation exists and how far it has caught up** ✅ — reported from the moment it is created, before it has folded anything, and the distance is in blocks.
2. **Reporting stops at the promotion** ✅ — and the complementary case is covered too: the generation the pointer moved OFF is reported, since it is retained and revertible. That pairing is what makes the report useful rather than merely present.
3. **The library reports; the app decides to render, dim or hide** ✅ — asserted as "carries enough for all three, and picks none of them", which is the right way to assert a negative design constraint.
4. **A generation whose stream is unavailable falls back to a FULL RE-INDEX rather than throwing** ✅ — asserted end to end: the app indexes from the chain and answers instead of becoming permanently unloadable. The core-level cases cover both a keeper that throws synchronously and one that rejects.
5. **The degradation matches the existing keeper precedent** ✅ — and the distinction it draws is the important one. READS degrade to absent, because absent is the same answer as far as the caller is concerned and it is what a full re-index is asked for by. WRITES raise through, deliberately, because the engine ACTS on a write failure (`StreamWriteOutcome`, and ADR-0038's freeze rule). Swallowing those would have been the easy symmetry and the wrong one. `setStreamConfig` is forwarded rather than swallowed, on the same reasoning task 7 used for the read-only view: it is the stream's identity, not a write to it.
6. **Round-trip through every keeper that actually exists** ✅, and better than asked. Rather than wrapping the one browser keeper, `degradingStream` is applied at the `createSegmentedStream` seam, so the damage rule has both halves in one place and any future keeper inherits them. The comment explains the split: damage a port can SHOW is inspected and cleared, damage it can only RAISE is what this wrapper answers for. No keeper was invented to satisfy the line.
7. **Changeset** ✅.
8. **Gate green** ✅.

With this the spec's fourteen stories are covered end to end: identity, registry, published verdict, container, follower, pause, promotion policy, and now visibility and graceful degradation.

Merging.

PR #50, opened 2026-09-03
tasking(one-command-runs-the-whole-pipeline): One command runs the whol…

wighawag:
Task: `work/tasks/done/one-command-runs-the-whole-pipeline.md`

Auto-tasked `one-command-runs-the-whole-pipeline` into 7 backlog tasks.

## Tasks

- **fetch-is-the-only-way-to-run-a-fetcher** — `etherfold fetch` is the only way to run a fetcher, and the standalone binary is retired
  - covers: US-3
  - blockedBy: one-configuration-path-for-every-command, run-follows-folds-and-answers-in-one-process
  - needsAnswers: true
- **index-receives-a-pushed-stream-and-owns-the-database** — `etherfold index` receives a pushed stream, owns the database, and makes the split a deployment choice
  - covers: US-3
  - blockedBy: the-node-server-starts-on-a-handle-it-was-given, fetch-is-the-only-way-to-run-a-fetcher
  - needsAnswers: true
- **one-configuration-path-for-every-command** — Every command resolves its processor, source and destination the same way: flags first, environment behind them
  - blockedBy: the-one-shot-is-build-and-serve-is-only-the-read-tier
  - needsAnswers: true
- **run-follows-folds-and-answers-in-one-process** — `etherfold run` follows a chain, folds into SQLite and answers HTTP, in one process
  - covers: US-1, US-2, US-6
  - blockedBy: one-configuration-path-for-every-command, the-node-server-starts-on-a-handle-it-was-given
  - needsAnswers: true
- **status-reports-the-cursor-through-an-injected-reporter** — `/status` reports the cursor, through a reporter injected beside the database
  - covers: US-6
  - needsAnswers: true
- **the-node-server-starts-on-a-handle-it-was-given** — The Node server starts on a database handle it was given, and carries an ingestion and a cursor reporter through
  - blockedBy: status-reports-the-cursor-through-an-injected-reporter
  - needsAnswers: true
- **the-one-shot-is-build-and-serve-is-only-the-read-tier** — The one-shot is `build`, `serve` is only the read tier, and no command is implicit
  - covers: US-4, US-5, US-7
  - needsAnswers: true

## Dependency graph

- Keystones: status-reports-the-cursor-through-an-injected-reporter, the-one-shot-is-build-and-serve-is-only-the-read-tier
- fetch-is-the-only-way-to-run-a-fetcher ← one-configuration-path-for-every-command
- fetch-is-the-only-way-to-run-a-fetcher ← run-follows-folds-and-answers-in-one-process
- index-receives-a-pushed-stream-and-owns-the-database ← the-node-server-starts-on-a-handle-it-was-given
- index-receives-a-pushed-stream-and-owns-the-database ← fetch-is-the-only-way-to-run-a-fetcher
- one-configuration-path-for-every-command ← the-one-shot-is-build-and-serve-is-only-the-read-tier
- run-follows-folds-and-answers-in-one-process ← one-configuration-path-for-every-command
- run-follows-folds-and-answers-in-one-process ← the-node-server-starts-on-a-handle-it-was-given
- the-node-server-starts-on-a-handle-it-was-given ← status-reports-the-cursor-through-an-injected-reporter

## Needs answers

- **fetch-is-the-only-way-to-run-a-fetcher**:
  - Three tasks require a test asserting that a processor-less server answers 501 on the ingestion routes, but that is not what the code does for an unauthenticated caller: getIngestAPI registers the token guard on the PATH (both /ingest and /ingest/*) BEFORE the getIngestion lookup, and authorized() fails CLOSED when INGEST_TOKEN is unset. So a serve/run/no-ingestion server answers 401, and 501 is reachable only by an AUTHENTICATED call. A test written to the literal criterion fails, and the tempting repair (move the capability check ahead of the guard) would let an unauthenticated caller probe whether a server hosts a processor, undoing the path-level guard the route comment says exists for exactly that reason. Fixed in the edits: each criterion now says configure the token, present a matching bearer header, assert 501 (the shape packages/server/test/ingest.test.ts:349 already uses), and states that the guard is not reordered. (packages/server/src/api/ingest.ts:106-118 (guard) vs 141/151 + notConfigured() at 199-204; affects run-follows (ingestion routes answer 501 on a run process), the-node-server (with none supplied they answer 501 exactly as they do today) and the-one-shot (a server started the way serve starts one answers 501))
- **index-receives-a-pushed-stream-and-owns-the-database**:
  - Three tasks require a test asserting that a processor-less server answers 501 on the ingestion routes, but that is not what the code does for an unauthenticated caller: getIngestAPI registers the token guard on the PATH (both /ingest and /ingest/*) BEFORE the getIngestion lookup, and authorized() fails CLOSED when INGEST_TOKEN is unset. So a serve/run/no-ingestion server answers 401, and 501 is reachable only by an AUTHENTICATED call. A test written to the literal criterion fails, and the tempting repair (move the capability check ahead of the guard) would let an unauthenticated caller probe whether a server hosts a processor, undoing the path-level guard the route comment says exists for exactly that reason. Fixed in the edits: each criterion now says configure the token, present a matching bearer header, assert 501 (the shape packages/server/test/ingest.test.ts:349 already uses), and states that the guard is not reordered. (packages/server/src/api/ingest.ts:106-118 (guard) vs 141/151 + notConfigured() at 199-204; affects run-follows (ingestion routes answer 501 on a run process), the-node-server (with none supplied they answer 501 exactly as they do today) and the-one-shot (a server started the way serve starts one answers 501))
- **one-configuration-path-for-every-command**:
  - Three tasks require a test asserting that a processor-less server answers 501 on the ingestion routes, but that is not what the code does for an unauthenticated caller: getIngestAPI registers the token guard on the PATH (both /ingest and /ingest/*) BEFORE the getIngestion lookup, and authorized() fails CLOSED when INGEST_TOKEN is unset. So a serve/run/no-ingestion server answers 401, and 501 is reachable only by an AUTHENTICATED call. A test written to the literal criterion fails, and the tempting repair (move the capability check ahead of the guard) would let an unauthenticated caller probe whether a server hosts a processor, undoing the path-level guard the route comment says exists for exactly that reason. Fixed in the edits: each criterion now says configure the token, present a matching bearer header, assert 501 (the shape packages/server/test/ingest.test.ts:349 already uses), and states that the guard is not reordered. (packages/server/src/api/ingest.ts:106-118 (guard) vs 141/151 + notConfigured() at 199-204; affects run-follows (ingestion routes answer 501 on a run process), the-node-server (with none supplied they answer 501 exactly as they do today) and the-one-shot (a server started the way serve starts one answers 501))
- **run-follows-folds-and-answers-in-one-process**:
  - Three tasks require a test asserting that a processor-less server answers 501 on the ingestion routes, but that is not what the code does for an unauthenticated caller: getIngestAPI registers the token guard on the PATH (both /ingest and /ingest/*) BEFORE the getIngestion lookup, and authorized() fails CLOSED when INGEST_TOKEN is unset. So a serve/run/no-ingestion server answers 401, and 501 is reachable only by an AUTHENTICATED call. A test written to the literal criterion fails, and the tempting repair (move the capability check ahead of the guard) would let an unauthenticated caller probe whether a server hosts a processor, undoing the path-level guard the route comment says exists for exactly that reason. Fixed in the edits: each criterion now says configure the token, present a matching bearer header, assert 501 (the shape packages/server/test/ingest.test.ts:349 already uses), and states that the guard is not reordered. (packages/server/src/api/ingest.ts:106-118 (guard) vs 141/151 + notConfigured() at 199-204; affects run-follows (ingestion routes answer 501 on a run process), the-node-server (with none supplied they answer 501 exactly as they do today) and the-one-shot (a server started the way serve starts one answers 501))
- **status-reports-the-cursor-through-an-injected-reporter**:
  - Three tasks require a test asserting that a processor-less server answers 501 on the ingestion routes, but that is not what the code does for an unauthenticated caller: getIngestAPI registers the token guard on the PATH (both /ingest and /ingest/*) BEFORE the getIngestion lookup, and authorized() fails CLOSED when INGEST_TOKEN is unset. So a serve/run/no-ingestion server answers 401, and 501 is reachable only by an AUTHENTICATED call. A test written to the literal criterion fails, and the tempting repair (move the capability check ahead of the guard) would let an unauthenticated caller probe whether a server hosts a processor, undoing the path-level guard the route comment says exists for exactly that reason. Fixed in the edits: each criterion now says configure the token, present a matching bearer header, assert 501 (the shape packages/server/test/ingest.test.ts:349 already uses), and states that the guard is not reordered. (packages/server/src/api/ingest.ts:106-118 (guard) vs 141/151 + notConfigured() at 199-204; affects run-follows (ingestion routes answer 501 on a run process), the-node-server (with none supplied they answer 501 exactly as they do today) and the-one-shot (a server started the way serve starts one answers 501))
- **the-node-server-starts-on-a-handle-it-was-given**:
  - Three tasks require a test asserting that a processor-less server answers 501 on the ingestion routes, but that is not what the code does for an unauthenticated caller: getIngestAPI registers the token guard on the PATH (both /ingest and /ingest/*) BEFORE the getIngestion lookup, and authorized() fails CLOSED when INGEST_TOKEN is unset. So a serve/run/no-ingestion server answers 401, and 501 is reachable only by an AUTHENTICATED call. A test written to the literal criterion fails, and the tempting repair (move the capability check ahead of the guard) would let an unauthenticated caller probe whether a server hosts a processor, undoing the path-level guard the route comment says exists for exactly that reason. Fixed in the edits: each criterion now says configure the token, present a matching bearer header, assert 501 (the shape packages/server/test/ingest.test.ts:349 already uses), and states that the guard is not reordered. (packages/server/src/api/ingest.ts:106-118 (guard) vs 141/151 + notConfigured() at 199-204; affects run-follows (ingestion routes answer 501 on a run process), the-node-server (with none supplied they answer 501 exactly as they do today) and the-one-shot (a server started the way serve starts one answers 501))
- **the-one-shot-is-build-and-serve-is-only-the-read-tier**:
  - Three tasks require a test asserting that a processor-less server answers 501 on the ingestion routes, but that is not what the code does for an unauthenticated caller: getIngestAPI registers the token guard on the PATH (both /ingest and /ingest/*) BEFORE the getIngestion lookup, and authorized() fails CLOSED when INGEST_TOKEN is unset. So a serve/run/no-ingestion server answers 401, and 501 is reachable only by an AUTHENTICATED call. A test written to the literal criterion fails, and the tempting repair (move the capability check ahead of the guard) would let an unauthenticated caller probe whether a server hosts a processor, undoing the path-level guard the route comment says exists for exactly that reason. Fixed in the edits: each criterion now says configure the token, present a matching bearer header, assert 501 (the shape packages/server/test/ingest.test.ts:349 already uses), and states that the guard is not reordered. (packages/server/src/api/ingest.ts:106-118 (guard) vs 141/151 + notConfigured() at 199-204; affects run-follows (ingestion routes answer 501 on a run process), the-node-server (with none supplied they answer 501 exactly as they do today) and the-one-shot (a server started the way serve starts one answers 501))


wighawag:
APPROVE ✅ — Gate-3 (conductor's review of the tasking output)

This is the task set for `one-command-runs-the-whole-pipeline`, emitted by `dorfl do spec:` against the spec as answered and re-scoped at `06d6659`. Reviewed as authored artifact rather than as a code diff: does the decomposition honour the spec's answers, is the graph sound, and are the criteria testable.

## The decomposition honours the re-scope

The four answers written into the spec are carried through faithfully. `status-reports-the-cursor-through-an-injected-reporter` is explicitly "the one piece of genuinely NEW surface in this spec" and carries the forward-note about the generation dimension. `the-one-shot-is-build-and-serve-is-only-the-read-tier` names the rename as "free today and none of them free after `publish-etherfold-and-deprecate-old-names` runs", which is the asymmetry the spec now records. The new `index` is correctly kept as the wire RECEIVER (`index-receives-a-pushed-stream-and-owns-the-database`), not dropped in the shrink. `fetch` refuses `--store`/`--db` rather than accepting them optionally.

## The graph is sound

Two free roots (`status-reports-the-cursor`, `the-one-shot-is-build`) that are deliberately file-orthogonal, fanning in through the Node adapter and the configuration path to `run`, then `fetch`, then `index`. Every `blockedBy` resolves to a task in this set, no cycle, and the build order reads as the dependency order. Historical records (changelogs, ADRs, written changesets, terminal-folder specs and tasks) are fenced off from the rename in the same words I used to uphold that boundary in PRs #44 and #45.

## The `needsAnswers` on all seven is a stamp, not a question

The tasker's improver loop hit its 3-pass cap and emitted the whole set as "uncertain", which puts `needsAnswers: true` and an identical `## Open questions` block on every task. Read the block: it is not a question. The reviewer found that three tasks required a `501` assertion on the ingestion routes that an unauthenticated caller can never see (the token guard is registered on the PATH, before the capability lookup, and fails CLOSED with no `INGEST_TOKEN`), noted that the tempting repair of reordering the guard would let an unauthenticated caller probe whether a server hosts a processor, and then FIXED the criteria in the bodies to configure a token and present a bearer header, and to state that the guard is not reordered. I verified both halves against `packages/server/src/api/ingest.ts` and against the edited criteria in `run-follows`, `the-node-server` and `the-one-shot`. So the finding is correct, the fix is already applied, and there is nothing left for a human to answer.

Merging as authored, then clearing the seven flags in one follow-up commit on `main` with this reasoning recorded, so the set is agent-buildable.

PR #51, opened 2026-09-03
feat(status-reports-the-cursor-through-an-injected-reporter): `/status`…

wighawag:
Task: `work/tasks/done/status-reports-the-cursor-through-an-injected-reporter.md`

Done. The task's premises checked out against the tree (no drift): `/status` reported no cursor, `getIngestion` is the live precedent, and the Workers host still constructs the app with a D1 binding and an environment only.

**What I built**

- `packages/server/src/cursor.ts` (new) — `CursorReport` (JSON-shaped, so a `bigint` does not compile), `StatusCursor` (the envelope), and `reportCursor`, which degrades every reporter failure to `{reported: false, reason}`: throw, reject, nothing-to-report, or unserialisable.
- `packages/server/src/types.ts` — `ServerOptions.getCursorReport?`, optional, injected exactly like `getIngestion`, with the doc stating what a reporter owes the server (a small, JSON-serialisable summary, never the raw serialized cursor, because the server reports verbatim and cannot bound it).
- `packages/server/src/api/status.ts` — the `cursor` field, computed only when a reporter is injected (absent otherwise, no `null`, no empty object), and unable to change `healthy` or the status code.
- Tests first, in `packages/server/test/server.test.ts` (in-process `app.request`, matching the neighbouring status tests): present, verbatim nested pass-through, async reporter, absent, throwing, rejecting, nothing-to-report, unserialisable, and an unhealthy server whose `healthy`/`503` is untouched. `platforms/cf-worker/test/status.test.ts` asserts the real store-less host reports no `cursor` key.
- Docs: `packages/server/README.md` (the option, the reporter's obligation, the field's two shapes, and that `/status` is the whole query surface for now with the richer layer deliberately out of this milestone), the Worker host's comment, `CONTEXT.md`'s query-layer bullet (which claimed the field did not exist yet), plus `docs/adr/0047-...md` and a changeset for `@etherfold/server` (minor).
- One out-of-scope signal captured in `work/notes/observations/the-worker-status-shape-test-asserts-nothing.md`.

Gate: `pnpm format:check`, `pnpm check:adr`, `pnpm build`, `pnpm typecheck`, `pnpm test` all green (server 56, cf-worker 12). `pnpm changeset status --since=main` only sees changesets through `git diff`, so it reports the new one once the runner commits; verified in a throwaway clone that it then passes with `@etherfold/server` bumped minor.

## Decisions

- **The reported field's SHAPE is an envelope: `cursor: {reported: true, value: <verbatim>}` / `{reported: false, reason}`, absent entirely when no reporter is injected.** Chosen over a bare scalar (`cursor: 4242` or the opaque string) and over a server-parsed shape (`{lastToBlock, latestBlock}`). A scalar could not grow the generation dimension without changing its type under readers, and the raw string is a serialized `LastSync` carrying an unconfirmed window of decoded events. A parsed shape would give `@etherfold/server` a cursor's meaning that ADR-0027 keeps behind the seam. **How the generation dimension grows inside it:** a later host that holds generations adds a key BESIDE `value` (`{reported: true, value: <the canonical generation's report>, generations: {...}}`), so `value` keeps meaning "the generation answering reads" and no existing reader breaks. Touches: the `/status` response body every client and dashboard reads, `the-server-and-cli-hold-generations-too`, and whichever CLI command (`run` / `build` / `index`) first builds a reporter. This is a published response body, so it met the ADR gate: `docs/adr/0047-the-status-cursor-is-a-reported-envelope-the-server-never-parses.md`.
- **Degraded means absent-WITH-A-REASON rather than an omitted field.** The criterion allowed either. A reason is what lets an operator tell "this deployment reports no cursor" (a Worker, a `serve` tier) from "this deployment's reporter is broken", which is the distinction the status page exists to make. Omission is reserved for the one honest absence: no reporter injected.
- **The server calls `JSON.stringify` on the report inside the degrade path.** An unserialisable report (a `bigint`, which is what every decoded `uint256` here is) would otherwise throw inside `c.json`, after the guarded code has returned, and answer `500` on the page an operator watches while something is wrong, which is exactly the failure the "a reporter must not take `/status` down" constraint forbids. Alternative considered: rely on the type alone, since `CursorReport` is JSON-shaped, and accept the runtime hole for a host that casts. The probe asks whether the value can be SENT, never what it means, so it is not parsing. Touches nothing outside this route.
- **The option is named `getCursorReport`, not `getCursor`, and it may be async.** `sync cursor` is a defined glossary term meaning the opaque string, and an option named `getCursor` whose documentation forbids returning the cursor would re-mean it; a reporter returns a REPORT about the cursor. Async because reading a cursor is a store read, unlike `getDB`/`getEnv`/`getIngestion`, which hand back handles a host already holds. The response FIELD is still named `cursor`, which is the name `CONTEXT.md` and the spec already promise.
- **`StatusCursor.value` is typed `unknown` while the reporter's return is the recursive JSON `CursorReport`.** Not a stylistic choice: the response type reaches the Hono RPC client type through `c.json`, and the recursive type made the compiler give up with `TS2589` (`hc<App>('')` in `packages/server/src/index.ts`). `unknown` is also what the server actually knows, and the serialisability obligation still binds where a host writes its reporter, which is the only place that can honour it. Touches any consumer of the typed client, which now sees `cursor.value` as `unknown` and must narrow it itself.
- **A failing reporter is logged but does NOT set `lastError`.** It follows the reorg counters' precedent in the same route (their read failure is swallowed), and the reason is already reported in the field itself; `lastError` stays "the error this process saw" about the server's own work rather than about a host's injected callback.


wighawag:

Reviewed against the eleven acceptance criteria in `work/tasks/backlog/status-reports-the-cursor-through-an-injected-reporter.md`. Gate-1 and CI `verify` are green. This is the one piece of genuinely new surface in the spec, so the two things that matter are that the server learned nothing about cursors and that the field can grow.

## Criteria

1. **Optional cursor reporter, injected exactly like the ingestion** ✅ — `getCursorReport?` beside `getIngestion?` in the options type, same optionality, same "absent is a missing capability, not an error" reasoning in its doc.
2. **With a reporter, `/status` reports the cursor as an OBJECT, verbatim** ✅ — `{reported: true, value}`; a nested report is asserted to pass through untouched, and an async reporter is awaited, "which is what reading a store is".
3. **With no reporter, `/status` is unchanged** ✅ — no field invented, existing assertions intact.
4. **No store dependency, no cursor table read, no cursor decoded** ✅ — I checked `package.json`: the only store-adjacent dep (`@etherfold/processor-sqlite`) was already there on `main` and is untouched. `cursor.ts` parses nothing.
5. **A failing reporter yields absent-with-a-reason and never fails the route** ✅ — throwing, rejecting, nothing-to-report, and a report that cannot be serialised each have a case, plus "does not flip `healthy` or the status code on an unhealthy server either". That last one is the assertion that matters for the page an operator refreshes while something is wrong.
6. **The Workers host still builds and serves with no cursor field** ✅ — exercised by a real host, and a new case in `platforms/cf-worker/test/status.test.ts` asserts the one key that matters.
7. **Documentation of what a reporter owes the server** ✅ — and it is enforced by the TYPE, not just prose: `CursorReport` is a recursive JSON type that excludes `bigint`, which is exactly the value a decoded `uint256` is throughout this project and which would otherwise compile and then fail `JSON.stringify` on the status page. Stating the obligation at the injection point, where a HOST writes its reporter, is the only place it can be honoured anyway.
8. **Documented where the query surface is documented, as the whole observability story for now** ✅.
9. **Tests: present, absent, throwing, verbatim** ✅ — nine cases.
10. **Changeset** ✅.
11. **Gate green** ✅.

## The forward-note was honoured with a mechanism, not a comment

The spec's answer 4 asked that the field be shaped to grow a generation dimension rather than hard-coding a scalar. The build did better than shape it: the envelope is an OBJECT precisely so that a host later reporting several generations adds a key BESIDE `value` rather than changing the type of a field clients already read, and that is recorded as ADR-0047. The envelope also carries the degraded case IN the field rather than by omission, so an operator can distinguish "this host reports no cursor" from "this host's reporter is broken". Both reasons outlive this milestone.

The `unknown`-here-vs-`CursorReport`-at-injection asymmetry is explained and is right: a recursive JSON type reaching the Hono RPC client type through `c.json` makes the compiler give up (`TS2589`), so the obligation lives at the one point that can honour it.

The off-path finding (the worker's "same status shape" test derives its expectation from the body it checks, so it cannot fail) was captured as an observation and NOT fixed here. Correct handling.

Merging.

PR #52, opened 2026-09-03
feat(the-node-server-starts-on-a-handle-it-was-given): The Node server…

wighawag:
Task: `work/tasks/done/the-node-server-starts-on-a-handle-it-was-given.md`

Gate is green (`format:check`, `check:adr`, `build`, `typecheck`, `test` across the workspace; `changeset status --since=main` only sees the changeset once it is committed, which the runner does).

## What I did

**Drift check first.** The blocking task `status-reports-the-cursor-through-an-injected-reporter` has landed (`ServerOptions.getCursorReport`, `CursorReport`/`StatusCursor`, ADR-0047), `getIngestion` has shipped and no host passes it, `d1-limits-reach-the-stores-batch-bounds` landed in `platforms/cf-worker/src/d1.ts` and does not touch this file, and the ingest token guard is on the path (`/ingest`, `/ingest/*`) ahead of the capability lookup exactly as the task body says. No drift; built as specified.

**Tests first** (`platforms/nodejs/test/serve.test.ts`, six new tests, red before the change): serving on a caller-built handle and handing that very handle back; shared-handle visibility against `:memory:` (a write via `recordReorg` outside the server is read back through `/status`, with an explicit assertion that a second `:memory:` connection does not even carry the schema, so it cannot pass by accident); auto-setup and `autoSetup: false` on the handle form; the reporter passthrough (present, and no `cursor` field when absent); the ingestion passthrough (`/ingest/expected-from-block` and `/ingest` both reach a supplied ingestion), and both directions of the absence: `501` for an authenticated caller with `INGEST_TOKEN` set through the adapter's `env` option and a matching bearer header, `401` for an unauthenticated one. The guard was not reordered.

**Production change** (`platforms/nodejs/src/index.ts`): `StartOptions.db` is now `string | RemoteSQL`; a handle is used as given, a string (or nothing) keeps today's defaulting (`DB`, else `file:./etherfold.db`) and `createNodeDB`. Auto-setup, the port binding, the address readback, the returned `db` and `close()` are untouched. `StartOptions.getIngestion` and `StartOptions.getCursorReport` are typed as `ServerOptions<NodeEnv>['getIngestion'|'getCursorReport']` and passed straight to `createServer`, so no new dependency was added: the adapter still names no store package, no engine package and no chain concept.

**Docs/changeset**: `platforms/nodejs/README.md` gains the handle form ("One database, two users"), the two capability options in the option table, the ownership rule for a supplied handle, and a corrected note where it previously claimed the host passes no `getIngestion`. `.changeset/the-node-server-starts-on-a-handle-it-was-given.md` records the three new options as a minor on `@etherfold/platform-nodejs`.

## Decisions

- **The two start forms are ONE option with a union type (`db?: string | RemoteSQL`), not two keys.** Considered a second key (`db` for the URL, `handle`/`database` for a `RemoteSQL`), which would make the discriminator a name instead of a `typeof`. Rejected because two keys create a third, meaningless state (both supplied) that then needs a refusal, and because the concept is one thing: a host has ONE database, and the only question is who opened it. A union keeps `db` meaning exactly what it means today and what `RunningServer.db` already returns, so a reader meets one word for one concept. Touches: any caller of `startServer` (today `packages/cli`'s `serve`, unchanged since a string is still accepted) and the CLI commands `run` / `index` that will pass a handle.
- **The ingestion and the reporter are accepted in the SERVER's own per-request shape (`(c) => ...`), not as plain values.** Considered accepting a `LogIngestion` value (and a plain report function) and wrapping it, which reads more naturally for a Node process that builds exactly one ingestion at startup. Rejected because the adapter's job is passthrough: re-shaping would make `platforms/nodejs` a second, subtly different injection surface that has to be kept in step with `ServerOptions`, and it would silently forbid a host that wants to resolve per request (which is the model the app is built on, because on Workers the database binding only exists per request). A caller with a single value writes `() => ingestion`. Consequence to note deliberately: `db` accepts a VALUE while the capabilities accept FUNCTIONS, and that asymmetry is intended — the database is a thing this adapter would otherwise have to build itself from a URL, the capabilities are things it can never build. Touches: `run` / `index` in `one-command-runs-the-whole-pipeline`, which will write `getIngestion: () => ingestion` and `getCursorReport: async () => report`.
- **Typed by indexing `ServerOptions<NodeEnv>` rather than importing `LogIngestion` / `CursorReporter` directly.** `CursorReporter` is exported by `@etherfold/server` and could have been imported by name, but `LogIngestion` lives in `@etherfold/core`, which this package does not depend on and must not start depending on ("it names no store package and no chain concept"). Indexing the options type gives both without a new dependency and makes the passthrough literal in the type: whatever the server accepts is what this accepts. The test spells its fake ingestion the same way, so the test adds no dependency either.
- **`close()` does NOT close the database handle, on either form.** That is today's behaviour for the URL form and it is now load-bearing rather than incidental: a server that closed a handle it did not build would take a store's connection down with it, which is precisely the failure the shared-handle form exists to avoid. Considered closing only handles the adapter itself built (it can tell), and rejected it: it would make shutdown mean two different things depending on how the server was started, and the URL form's handle is returned as `running.db` for exactly the sharing this task enables, so "the adapter built it" is not the same as "nobody else uses it". Stated in the doc comments on `StartOptions.db` and `RunningServer.close` and in the README. Touches: whichever CLI command owns process shutdown; it closes the handle it built, after stopping the server.


wighawag:

Reviewed against the ten acceptance criteria in `work/tasks/backlog/the-node-server-starts-on-a-handle-it-was-given.md`. Gate-1 and CI `verify` are green.

## Criteria

1. **A database HANDLE as an alternative to a URL, URL behaviour unchanged** ✅ — the handle form hands back the very handle it was given; the URL path is untouched.
2. **A shared handle is genuinely shared** ✅ — asserted the only way it can be: a write made OUTSIDE the server is visible to a read the server answers, against `:memory:`, where a second connection would see nothing. This is the criterion `run` depends on ("one handle, two users").
3. **Schema auto-setup on both forms, opt-out on both** ✅.
4. **Cursor reporter passed through unchanged** ✅ — verbatim on `/status`, and no field invented when none is supplied.
5. **Ingestion capability passed through** ✅ — a supplied ingestion receives logs over real HTTP, so a Node process can host a processor.
6. **Both directions driven through the ingest route's OWN contract** ✅ — this is the trap the tasker found and fixed, and the build honoured it exactly: `501` is asserted with `INGEST_TOKEN` configured and a matching bearer header, and a separate case asserts `401` for an unauthenticated caller "so the absence of a processor is not a probe". The guard was not reordered.
7. **The adapter names no store package and no chain concept** ✅ — grepped: no `state-store`, no `processor-*`, no `LogFetcher`/`StreamBuilder`/`chainId` in the adapter source. It decides database, environment, ingestion and reporter, and nothing else.
8. **Tests in the repo's style, real HTTP against a started server** ✅ — nine cases.
9. **Changeset** ✅.
10. **Gate green** ✅.

This is the adapter `run` will assemble on top of, and it leaves the seam exactly as narrow as the spec asked for: everything host-shaped arrives injected, nothing is derived.

Merging.

PR #53, opened 2026-09-03
feat(the-one-shot-is-build-and-serve-is-only-the-read-tier): The one-sh…

wighawag:


wighawag:

Reviewed against the eleven acceptance criteria in `work/tasks/backlog/the-one-shot-is-build-and-serve-is-only-the-read-tier.md`. Gate-1 ran green on the reconciled tip (`596829f`) and CI `verify` is green. This is the rename the spec's answer 2 says must land before anything is published; with this merged, `publish-etherfold-and-deprecate-old-names`'s new `blockedBy` edge is satisfied.

## Criteria

1. **`etherfold build` does everything today's `etherfold index` did** ✅ — same flags, `-n` still required with `ETHEREUM_NODE` as fallback, same refusals, same engine.
2. **`etherfold index` resolves to nothing** ✅ — asserted directly ("resolves to nothing, so the word is free for the wire receiver"). No error stub that could be mistaken for a command.
3. **A bare `etherfold` prints help; no default command** ✅ — both asserted, including "refuses the old default-command form rather than folding under no name", which is the case that catches a leftover `isDefault`.
4. **`serve` unchanged and documented as the read tier** ✅.
5. **The read tier refuses to write, `/status` still answers** ✅ — driven the way the tasker's finding required: `501` to an AUTHENTICATED caller, `401` to an unauthenticated one, guard not reordered. The comment block in the test records why both halves are needed.
6. **Live docs updated** ✅ — the one remaining `etherfold index` string in live docs is the CLI README sentence saying it "is an unknown command until the receiver lands", which is correct and forward-looking, not a stale reference.
7. **`CONTEXT.md` command-set bullet updated** ✅ — this is the bullet that conflicted with #51's cursor bullet next to it; resolved as the rename's bullet plus the cursor's bullet, each from the side that built it.
8. **The `501` test names its database explicitly** ✅ — `:memory:`, no stray `etherfold.db`.
9. **Historical records NOT edited** ✅ — I checked: no CHANGELOG, no ADR, no written changeset, no terminal-folder spec or task touched, other than this task's own done-move.
10. **Changeset says the break plainly** ✅ — `etherfold: major`, "BREAKING, and it is the whole point", with the before/after invocation.
11. **Tests + gate green** ✅.

## Integration note, for the record

This PR was built in parallel with #52 (the Node adapter), on the grounds that the two were file-orthogonal (CLI package vs platform adapter). They were not quite: both appended a `describe` block to `platforms/nodejs/test/serve.test.ts` and both edited `platforms/nodejs/README.md`. The test conflict was resolved by keeping BOTH blocks (all 14 cases pass) and dropping #53's duplicate `TOKEN` declaration; the README merged cleanly. The reconciled tip went back through `dorfl complete`, so the full gate ran on exactly what is being merged. The lesson is noted for the remaining tasks in this spec: they touch the CLI package in overlapping ways and will be built strictly serially.

Merging.

PR #54, opened 2026-09-03
feat(one-configuration-path-for-every-command): Every command resolves…

wighawag:
Task: `work/tasks/done/one-configuration-path-for-every-command.md`

## Drift check (first, as instructed)

The task's premises all held. Its blocker `the-one-shot-is-build-and-serve-is-only-the-read-tier` is in `work/tasks/done/`, and its Decision 3 explicitly hands the naming question here. The three disagreeing resolution paths are exactly as described: `packages/fetcher-host/src/config.ts` (environment-driven, refuses by name, takes overrides), `packages/cli/src/options.ts` (flag-driven, requiredness split between commander and the resolver), `platforms/nodejs/src/index.ts` (`DB`, `PORT`, with a `file:./etherfold.db` convenience default). Only file *locations* differ from the prompt's text (the spec is in `work/specs/tasked/`, this task body in `work/tasks/backlog/`), which is folder status, not a premise.

## What landed

**`packages/cli/src/config.ts` (new)** is the one resolution module. It holds the command table as code (`OWNERSHIP`, all five rows including the two asymmetries), the name of every input once (`INPUTS`: flag, variable, help text), and `resolveCommandConfig(command, options, env)` — pure, total, synchronous, opening and dialling nothing.

- `packages/cli/src/types.ts` now types all five rows (`RunConfig`, `BuildConfig`, `FetchConfig`, `IndexConfig`, `ServeConfig`). Every row was expressible. `Options` covers every flag any of the five takes and every field is optional.
- `packages/cli/src/program.ts` registers flags **from the table**. No `requiredOption`, no commander default.
- `packages/cli/src/serve.ts` (new) resolves the read tier's row and passes the database to `startServer` explicitly, so the adapter default is unreachable from a command.
- `packages/cli/src/index.ts`: `prepareIndexing('build', options, deps)`, and a `SourceOrigin` switch — `-d` → `INDEXING_SOURCE` → the module, the last being the only arm that can cost `eth_chainId`.
- `ETHEREUM_NODE` is gone from code and readme; `ETH_NODE_URI` is the only name.
- `docs/adr/0048-a-command-inputs-name-and-whether-it-may-default.md`, changeset, CLI readme section, one observation note.

Tests: `test/configuration.test.ts` (the resolver seam, 50 cases: flag-beats-env, env-behind-flag, every refusal by name, every refused cell, the chain-free source, what may default) and `test/fromTheEnvironment.test.ts` (one end-to-end pass per shipped command: `DB`/`ETH_NODE_URI`/`INDEXING_SOURCE` reach the pipeline, with `eth_chainId` asserted *not* called). CLI 94 tests; whole gate green (format, build, typecheck, 18 packages).

## Decisions

**1. The command table is ONE table, and it drives BOTH the parser and the resolver.** `OWNERSHIP` in `src/config.ts` is the only place the task's five-row table is written down; `registerInputs` reads it to register flags and `resolveCommandConfig` reads it to refuse. Alternative considered: register flags by hand per command and keep the table as documentation, which is what the CLI did before and is how the parser and the resolver came to disagree about requiredness in the first place. **Touches**: all three unbuilt commands — adding `run` / `fetch` / `index` is `program.command(...)` plus `registerInputs(cmd, name)` plus an assembly, with no new way to read a flag.

**2. A flag a command does not own is REGISTERED HIDDEN so it parses and reaches the resolver.** The criterion is that a not-accepted input is "REFUSED with a message saying why, never accepted and ignored", and commander's `unknown option` names neither a reason nor the command that does own it. So `etherfold serve -p ./processor.js` now says a read tier holds no processor and points at `index` / `run` / `build`, and `--help` still shows only what the command owns. Alternatives considered: leave them unregistered (the resolver refusal then only fires for programmatic callers, so the message is hollow at the terminal), or refuse them visibly in `--help` (which advertises flags that always fail). **Touches**: it changed an existing assertion — `commands.test.ts` asserted `serve -p` was `/unknown option/i` — and it is the mechanism all three future commands inherit.

**3. Which inputs get an environment variable at all: exactly the seven a deployable already publishes.** `INDEXING_SOURCE`, `ETH_NODE_URI`, `INGEST_ENDPOINT`, `INGEST_TOKEN`, `REQUESTS_PER_SECOND` (the fetcher host's) plus `DB`, `PORT` (the Node adapter's). `-p`, `--store`, `--retention`, `-d`, `--host`, `--no-auto-setup` are flags only, and their refusals say so rather than naming a variable that does not exist. The line: **the environment carries what varies between deployments of one image; a flag carries what the image IS.** Alternative considered: invent `PROCESSOR` / `STORE` so every refusal could name a variable literally — rejected as a second way to say what a container's `CMD` already says, and as six new names against the "one set" rule. Recorded in ADR-0048. **Touches**: every future command's refusal messages.

**4. Only the PORT may fall back to a default (2000). Nothing else does.** This is the one the prompt flagged. A port is not a claim about the deployment (a wrong one fails visibly, at once, writing nothing); a defaulted database or node URL fails silently, which is the whole reason `--db` was never defaulted. The asymmetry that settles it: adding a default later is free, removing one is breaking. Retention keeps its `unbounded` default because that is the store's own default and changes nothing about a store nobody configured; `autoSetup` keeps `true` because the task put the read tier's schema default explicitly out of scope. Recorded as **ADR-0048** (`docs/adr/0048-a-command-inputs-name-and-whether-it-may-default.md`). **Touches**: `run` and `index`, which serve, and any later argument for defaulting a database.

**5. `--port` lost its commander default, and that is a rule rather than a fix.** It was `'2000'` at the parser, so the flag was *always* present and `PORT` could never be reached — a commander default silently defeats "flags first, environment behind them". The rule is now stated where flags are registered: no `requiredOption` and no commander default, ever. **Touches**: `commands.test.ts` (which asserted `port: '2000'` reached the handler) and every future serving command.

**6. Refused = a FLAG a user typed; an ambient VARIABLE a command does not own is simply not read.** `ETH_NODE_URI` set on a host that runs `fetch` beside `index` is ordinary, and refusing `index` because the machine has a node URL would make the split deployment this system exists for unconfigurable. Alternative considered: refuse on the variable too, which is more literally "nothing is accepted and ignored" but breaks the one-host-several-commands case the wire exists to enable. **Touches**: `run`/`fetch`/`index` on a shared host; asserted by a test.

**7. `--ingest-token` is offered as a flag, with the hazard documented.** "Flags first, environment behind them" is a stated non-negotiable and the refusal has to name both, so an environment-only token would have been the one input off the path. But a secret on a command line is visible in `ps` to every process on the host, so the help text and the readme both say to prefer `INGEST_TOKEN`. The refusal never quotes a value, only the name — the same rule `FetcherConfigError` already follows (asserted by a test). Alternative considered: env-only, which is safer and inconsistent. **Touches**: `fetch` and `index`, which are the two commands that will consume it.

**8. The destination is ONE field with two arms (`Destination` = `StoreTarget | DatabaseTarget`), both carrying `db`.** The database is one input (`--db`, `DB`) whatever a command does with it; what differs is whether the command also owns a STORE — which is precisely the difference the task's table draws between "store + database" and "database". `StoreTarget` therefore gained a `kind` discriminant and kept its meaning. Alternative considered: `target?: StoreTarget` for writers plus a separate `database?: string` for `serve`, which puts two names on one input in the resolved shape — the exact muddle this task exists to remove. **Touches**: `run` and `index` (store arm), any later read-only command (database arm).

**9. `prepareIndexing` takes the command name first: `prepareIndexing('build', options, deps)`.** It is typed `'run' | 'build'` — `index` folds too but makes no chain call, so it builds no provider and no `LogFetcher` and will assemble differently while resolving through the same function. Alternative considered: have it take an already-resolved config, which is a cleaner seam but pushes resolution boilerplate into six test call sites and the example, for no assertion gain. It is a public API break; the changeset says so. **Touches**: `examples/event-processor-nfts/test/cli.test.ts` (updated) and `run`, which will pass `'run'`.

**10. `--rps` is parsed to a number in the resolver, and a value that is not a rate is refused.** This closes `work/notes/observations/cli-rps-is-typed-as-a-number-and-arrives-as-a-string.md`. I judged it inside rather than outside the fence: the resolver is the thing this task builds, `rps` is one of its outputs, and shipping the new shared path with a known-false type in it would be building the deliverable with a hole. It introduces one new refusal, which is why it is here rather than silent. **Touches**: nothing outside the CLI; `REQUESTS_PER_SECOND` is the fetcher host's existing name, reached here now instead of two layers down.

**11. `cli.ts` catches and prints the message rather than an unhandled rejection.** `serve` had no refusals before, so its first one arrived as a raw stack dump through commander's un-awaited async action. Configuration refusals are written to be read, so the entry point prints `err.message` and exits 1. `main`'s existing `console.error(err)` for `build` is untouched — a fatal indexing error is a different class and its stack is wanted. **Touches**: the presentation of every future command's refusals; a later task wanting uniform presentation would need an error class, which I did not introduce.


wighawag:

Reviewed against the ten acceptance criteria in `work/tasks/backlog/one-configuration-path-for-every-command.md`. Gate-1 and CI `verify` are green.

## Criteria

1. **One resolution module, both commands through it** ✅ — `packages/cli/src/config.ts` owns it; `build` and `serve` are asserted to resolve "off the same path", and the three not-yet-built rows (`run`, `fetch`, `index`) are already resolvable so the coming tasks add commands rather than configuration.
2. **Flag beats environment, environment behind the flag, neither is a refusal** ✅ — plus the case that catches a common bug: a BLANK variable reads as unset rather than as an empty answer.
3. **Every refusal names the flag AND the variable, and fires before the chain or a database is touched** ✅ — "refuses a missing node url before it imports the module or opens a database", and the token refusal never quotes the token itself.
4. **One set of variable names; `ETHEREUM_NODE` retired** ✅ — the only two surviving mentions are the README's retirement note and the resolver's comment saying it is retired, which is what a reader following the old name should find. `ETH_NODE_URI` is asserted as the one name behind `-n`.
5. **No command runs on an unnamed database** ✅ — the CLI resolves it and passes it explicitly, asserted as "hands the adapter the database it resolved, so the adapter default is never reached", and `serve` refuses rather than creating one nobody named.
6. **The two asymmetries are REFUSED, never accepted and ignored** ✅ — every cell of the 5x5 table is asserted by name with a reason, with the ordering rule that matters: a refused input is refused BEFORE a missing required one is asked for. And the necessary opposite: an AMBIENT variable a command does not own is ignored, not refused, so a shared `.env` does not break `serve`.
7. **A source resolves with no chain call, and the module route is refused where the caller cannot make one** ✅ — deployments folder first, `INDEXING_SOURCE` behind it, module route refused on `index` and on `fetch` for their two different reasons.
8. **Tests in the repo's style** ✅ — comprehensive; `fromTheEnvironment.test.ts` drives real `build` and `serve` invocations from the environment.
9. **Changeset, including the retired variable** ✅.
10. **Gate green** ✅.

Two things beyond the criteria, both right: `--rps` is now parsed to a number (closing the observation #53 filed) and falls back to the fetcher host's own `REQUESTS_PER_SECOND` rather than a second name; and `retention` stays in BLOCK NUMBERS per ADR-0019 with a duration refused by name. The `test:manual` script observation is correctly captured rather than fixed.

Merging.

PR #55, opened 2026-09-03
feat(run-follows-folds-and-answers-in-one-process): `etherfold run` fol…

wighawag:
Task: `work/tasks/done/run-follows-folds-and-answers-in-one-process.md`

Done. The whole gate is green (`format`, `build`, `typecheck`, `check:adr`, `test` across every package), and the tree holds only this task's changes.

## What I built

**`etherfold run`** — one process that follows a chain, folds into SQLite and answers HTTP.

- `packages/cli/src/run.ts` (new): `run()` assembles through the existing `prepareIndexing`, starts the Node server **on the handle the store folds into** with a cursor reporter, installs the signal handlers and starts the loop; `runMain()` is the process shape that resolves the exit code (`0` on a signal, non-zero on a fatal refusal or a failure to start).
- `packages/cli/src/cursorReport.ts` (new): `readCursorReport(store)` — `{lastFromBlock, lastToBlock, latestBlock, unconfirmedBlocks}`, never the stored cursor itself.
- `packages/cli/src/index.ts`: the one-shot's `indexToTip` became `driveCycles(command, …)`, whose single `stopAtTip = command === 'build'` **is** the difference between the two commands; `PreparedIndexing` gained `db` (the one handle) and `config` (the resolved row).
- `packages/cli/src/program.ts`: `run` registered, flags from the same `OWNERSHIP` table.
- `platforms/nodejs-fetcher`: `stopOnSignals(controller)` extracted from `startFetcher` and exported, so `run` reuses the adapter's signal answer instead of writing a second one.
- README (CLI: `run` documented first as the default thing to reach for; root; example) and a changeset covering both packages.

**Tests** (`packages/cli/test/run.test.ts`, 15 new cases + example + command-surface): the headline drives `run` with the provider, database, sleep and stop signal injected the way the one-shot's tests inject them, against a fixture chain over a real libSQL database and real HTTP — asserting state landed and `/status`'s cursor **advanced across cycles**; that it keeps cycling at the tip; that the store and the server are one handle (proved by reading the server's own `Meta` table through the store's handle); `501` authenticated / `401` anonymous on `/ingest` while `/status` answers; both exit codes; and refusals raised before any chain call, database or port. I verified each of these bites by mutating the production code (making `run` stop at the tip, dropping the reporter, passing a URL instead of the handle) and watching exactly the right tests go red.

**The two checks the spec asked for, both confirmed:** a reorg's retractions reach the processor in **one** `process` call carrying their replacements (asserted with a spy on the running processor), and a second `run` over the same database **resumes from the persisted cursor** (its first `eth_getLogs` starts above the start block).

**Story 2:** `examples/event-processor-nfts/src/entities.ts` is **zero lines different**; the example's CLI test now runs that same file under `run` and reads the same three facts back out.

No drift: `prepareIndexing`'s `'run' | 'build'` parameter, the resolved `run` row, `StartOptions.db` taking a handle and `getCursorReport` all landed as the task assumes. (Bookkeeping only: the body is in `work/tasks/backlog/` and the spec in `work/specs/tasked/`, not the `ready/` paths the prompt names; the runner owns those moves, so I left them alone.)

## Decisions

**1. The shared assembly is `prepareIndexing`, and the ONE difference is a single `stopAtTip` boolean inside one driver (`driveCycles`), not two drivers.** `indexToTip` was renamed to `driveCycles(command, host, deps)` and the abort now fires only when `command === 'build'`; `run` and `build` therefore go through the same assembly AND the same loop, and "`build` is `run` without the serving, stopping at the tip" is true of the code rather than of prose. Alternatives considered: (a) leave `indexToTip` for `build` and give `run` its own loop caller, which is a second `runFetcherLoop` call site whose fatal/abort/report handling would drift apart; (b) hoist the assembly into a new module and let each command own its driver, which is more files for the same one `if`. **Touches** every existing caller of `prepared.index()` (`build`, the CLI tests, the example test) — its meaning is now "drive it the way THIS command drives it", which is why the command name is already the first argument to `prepareIndexing`.

**2. `PreparedIndexing` gained `db` and `config`.** `db` because "one handle, two users" is a criterion and the handle was previously built and hidden inside `buildProcessor`; `config` because `run` needs the `serving` row it resolved and re-resolving it in `run.ts` would be a second call site for one answer (and a second place a refusal could be raised). `PreparedIndexing` grew a third, defaulted generic parameter (the command) so `config` narrows to `RunConfig` for `'run'`. Alternative: keep both private and have `run` call `resolveCommandConfig('run', …)` itself — cheap and pure, but it makes the resolution happen twice for one command and puts the "resolved once, before anything opens" property in two places. **Touches** anything constructing a `PreparedIndexing` value (nothing outside this package does).

**3. What the cursor reporter reports: four numbers, never the cursor.** `{lastFromBlock, lastToBlock, latestBlock, unconfirmedBlocks}` where `unconfirmedBlocks` is a **count**. The stored cursor is a serialized `LastSync` whose window carries whole blocks of decoded events with `bigint` args, and `/status` reports what a host hands it verbatim (ADR-0047), so a host that passed it through would put an unbounded blob on the operator's page and fail `JSON.stringify` on the first `uint256`. The four are chosen to answer the two questions a status page is asked: is it moving (`lastToBlock` across two reads) and how far behind is it (`latestBlock - lastToBlock`). Deliberately excluded: the `context` hashes (an identity, not progress) and anything needing more than one cursor read, since the reporter runs on every `/status`. Typed structurally in the CLI rather than importing `CursorReport`, so the CLI does not take a direct dependency on `@etherfold/server`; the obligation is still checked by the compiler at the injection point. No ADR: ADR-0047 already fixes the ENVELOPE, and the payload is per host and freely changeable. **Touches** `index`, which will want the same reporter (it is a separate exported module, `src/cursorReport.ts`, for exactly that reason), and any dashboard reading `cursor.value`.

**4. A retryable failure is NEVER bounded, for the follower or the one-shot.** `index-to-a-store-from-the-cli` left this decision to `run`; the answer is unbounded retry on the fetcher host's escalating, capped (60s default) backoff. A bound would turn a transient node outage into a stopped indexer that only a supervisor restarts — which is the loop we would then have written by hand — and what genuinely cannot be waited out is already a `fatal` that exits non-zero. Alternative considered: bound it for `build` only (a CI job would rather fail than hang), which is a real argument but is `build`'s to make and would break "one driver, one difference"; a `--max-retries` flag is the reversible way to add it later. **Touches** `build` (same driver) and any future `index` on the same loop; recorded in the `driveCycles` doc comment.

**5. `stopOnSignals` was extracted from `platforms/nodejs-fetcher`'s `startFetcher` and exported, rather than `run` growing its own handler or reusing `startFetcher` whole.** `startFetcher` builds its own host from configuration, and `run` must hand over the host `prepareIndexing` already built (with the direct ingestion in it), so reusing it whole would have meant a second way to say what to run. The signal half is the reusable third ("a process, its signals and an exit code"), and one answer to "which signals a container sends, and what happens to the cycle in flight" is worth the extracted export. Alternative: 12 lines in the CLI, which is a second answer that can drift. It also adds `@etherfold/platform-nodejs-fetcher` as a CLI dependency — directionally right, since `etherfold fetch` is a front door onto that same package. **Touches** `@etherfold/platform-nodejs-fetcher`'s published surface (changeset: minor) and the future `fetch` command.

**6. The server's `INGEST_TOKEN` still comes from the ambient environment on a `run` process, and is not passed or suppressed by the command.** `--ingest-token` is refused as a FLAG (there is no wire to configure), but the Node adapter reads `process.env` for the app it starts, which is what makes the authenticated-`501` criterion reachable at all: the guard is configured, and the capability is absent. This follows the landed rule that a command refuses a flag a user typed and simply does not read an ambient variable it does not own; suppressing it would have been a third behaviour (refusing an ambient variable) that the configuration task explicitly rejected. **Touches** `serve` (already behaves this way) and `index`, which will read the same variable as its own input.

**7. `run()` returns a handle and `runMain()` is the process wrapper; the subcommand's action calls `runMain`.** This mirrors `startServer` / `startFetcher` + `runFetcherProcess`, and it is what lets the headline test query a running process over real HTTP instead of inferring progress from reports. `main` (the one-shot's exit-code wrapper) is untouched, so the package now exports `main` for `build` and `runMain` for `run`; I judged renaming `main` to be churn outside this task's fence. **Touches** `program.ts`'s injectable handlers (a `run` handler was added beside `build` and `serve`) and any later command that wants a uniform `main`.


wighawag:

Reviewed against the twelve acceptance criteria in `work/tasks/backlog/run-follows-folds-and-answers-in-one-process.md`. Gate-1 and CI `verify` are green. This is the spec's headline: the command it exists to create.

## Criteria

1. **`etherfold run` follows, folds into libSQL, answers HTTP, and keeps running at the tip** ✅ — the headline test lands state and asserts the tip is reached and NOT stopped at ("stopping is a signal and not a report").
2. **One database handle, built once, shared by store and server** ✅ — asserted as the same object handed to both, on top of PR #52's handle form.
3. **`/status` cursor ADVANCES across cycles** ✅ — asserted across cycles, not once, through `readCursorReport` injected into the adapter from PR #51.
4. **No remote writer: 501 authenticated, 401 anonymous, `/status` still answers** ✅ — the shape the tasker's finding required.
5. **No component implemented twice; browser engine constructed nowhere** ✅ — this is the criterion I checked most carefully. `run.ts` imports `prepareIndexing` from the one-shot's module and the old `indexToTip` was refactored into `driveCycles` that BOTH commands drive, so it is genuinely shared rather than copied. The engine is `StreamBuilder` via `createDirectIngestion`; `IndexerGeneration` does not appear, and the CLI's existing guard test enforces that it never will.
6. **Terminates only on a signal (exit 0, cycle allowed to finish) or a fatal refusal (non-zero), never at the tip** ✅ — both exits asserted, "so a supervisor can tell a stop from a wedge".
7. **Every input through the shared configuration path; refusals before any chain call or database write** ✅ — plus a case I would not have thought to ask for: "leaves no signal handler on the process when it refused to start".
8. **The headline test is at the COMMAND** ✅ — it drives the same entry point the subcommand's action invokes, with provider, database builder, sleep and stop signal injected the way the one-shot's own test already does.
9. **Story 2 with the NFT example's unchanged declarations** ✅ — "follows the chain, folds the browser demo's processor and answers HTTP, unchanged", extending that example's existing CLI test as asked.
10. **Tests including the non-zero exit** ✅.
11. **README documents `run` as the default thing to reach for; changeset** ✅.
12. **Gate green** ✅.

Two assertions that were not required but are exactly the ones this spec's Implementation Decisions said to CHECK rather than assume while re-engining: a reorg's retractions reach the processor in ONE batch with their replacements through this path, and a restart resumes from the cursor in the store rather than the start block. Both are present.

With this merged, `etherfold run` exists and the combined pipeline is reachable from a terminal. Merging.

PR #56, opened 2026-09-03
feat(fetch-is-the-only-way-to-run-a-fetcher): `etherfold fetch` is the…

wighawag:
Task: `work/tasks/done/fetch-is-the-only-way-to-run-a-fetcher.md`

All green. Here's what I did.

## What I built

**`etherfold fetch` (new command, `packages/cli`)**
- `packages/cli/src/fetch.ts`: `prepareFetching` (resolve this command's row through the shared `resolveCommandConfig('fetch', …)`, then open the explicit source: deployments folder or `INDEXING_SOURCE`), `fetch` (starts the adapter's loop, returns its handle) and `fetchMain` (the process shape).
- `packages/cli/src/program.ts`: registered `fetch` with the flags `OWNERSHIP` says it owns (visible) and the ones it does not (hidden, so a copied command line is refused with a reason). Its default handler installs the console log sink and calls `fetchMain`.
- Exported from `packages/cli/src/index.ts`; `types.ts`/`program.ts` doc comments updated (four of five commands ship now).

**The standalone binary is retired (`platforms/nodejs-fetcher`)**
- `src/bin.ts` deleted, `bin` entry removed, and its two entry-point-only dependencies (`ldenv`, `named-logs-console`) with it. The library entry (`startFetcher`, `runFetcherProcess`, `stopOnSignals`, loop, signals, exit code) is untouched, which is the shape `platforms/nodejs` already had.
- Its "the adapter is scheduling and configuration, and nothing else" source scan now reads one file instead of two.

**Docs + changesets**: fetcher-platform README (no longer tells you to run a binary; points at the command, and says the sink hookup is the CLI's), CLI README (a `fetch` section, and "one name is still to come"), root README, `CONTEXT.md`'s command-set bullet, both package descriptions. Two changesets: `@etherfold/platform-nodejs-fetcher` **major** (published `bin` removed) and `etherfold` **minor** (the new command).

**Tests** (written at the seams first): `packages/cli/test/fetch.test.ts` (12 cases) drives the real loop against a fake chain into a real receiver the test supplies (`test/utils/receiver.ts`: the Hono app over a `StreamBuilder` over the same entity processor, reached through an in-process `fetch`) — logs land, ranges stay contiguous, it keeps cycling past the tip, a second run starts where the receiver says (no cursor anywhere), plus the resolution (flag beats variable, deployments folder without a processor), the refusals, and all three exit-code paths. `commands.test.ts` gained the word and the flag surface (including: no `--from-block`/state-file/lock flag exists).

Gate: `format:check`, `check:adr`, `build`, `typecheck`, `test` all green (workspace-wide, 2500+ tests). `changeset status --since=main` only sees changesets once they are committed (it diffs tracked files), which is the same as previous tasks here.

Not mine to do, per the task's integrator note: adding this slug to `publish-etherfold-and-deprecate-old-names`'s `blockedBy`. That task must not run before this lands.

## Decisions

- **The flag names were NOT invented here.** `--ingest-endpoint` / `--ingest-token` (with `INGEST_ENDPOINT` / `INGEST_TOKEN` behind them), `-n`, `-d`, `--rps` and the whole `fetch` row already existed in `packages/cli/src/config.ts`, landed by `one-configuration-path-for-every-command` and already asserted in `configuration.test.ts`. So this command CONSUMES that table rather than extending it: no new input name, no second way to read a flag. The only thing I added on that axis is passing `--rps` through as `requestsPerSecond`, since the table says `fetch` owns it. Alternative considered and rejected: a shorter `--endpoint` / `--token`, which would have been a second name for a variable that is already the published contract of a deployable.
- **`fetchMain` re-uses the adapter's `runFetcherProcess` WHOLESALE, so the adapter owns the exit code; `fetch()` uses only `startFetcher`.** The criterion is "the same exit code the retired binary did", and the cheapest way to be the same is to be the same function: `0` when asked to stop, `1` on a refusal no waiting fixes. The CLI adds only the half `runFetcherProcess` cannot see — a CONFIGURATION refusal raised before a loop exists — which it catches and exits `1` on, as every other refusal in this CLI does. The alternative (drive `startFetcher` and map `summary.stoppedBecause` in the CLI) would have put a second copy of that mapping one package away from the first, and it is the mapping most likely to drift silently. Cost, accepted: `runFetcherProcess` takes no `AbortSignal`, so the "exits 0" test stops it with a real `SIGTERM` (as `platforms/nodejs-fetcher/test/loop.test.ts` already does) rather than an injected signal. Touches: `platforms/nodejs-fetcher`'s exit-code contract is now depended on by `etherfold fetch`.
- **The console log sink is installed in `program.ts`'s default `fetch` handler, not globally in `cli.ts` and not at module scope.** It has to be at a PROCESS entry point (a library must not choose an embedder's sink), and the CLI is the process now. Module scope in `fetch.ts` was rejected because `etherfold` is also a library (`index.ts` re-exports `fetch`), so importing `named-logs-console` there would hook the console for anyone importing the package. `cli.ts` (global, for every command) was rejected as OUT OF SCOPE and user-visible: it would change what `run`, `build` and `serve` print. Consequence a later task may want to settle: `run` also drives a fetch loop and its `named-logs` lines still reach no sink, so the observability of the two commands is now asymmetric. Touches: `run`/`build`/`serve` output, and `ProgramDependencies.fetch` (a test substituting the handler installs no sink, which is why `commands.test.ts` stays silent).
- **The CLI's exported function is named `fetch`**, matching `run` / `serve` / `build` (named for the command). It shadows the global `fetch` inside `src/fetch.ts` (where the global is never called: the wire's `fetch` is passed through as a dependency) and for a consumer who imports it by that name. Alternative considered: `startFetching`, rejected because it reads as a near-duplicate of the adapter's `startFetcher` and breaks the one-word-per-command pattern.
- **Two changeset files rather than one covering both packages.** The criterion asks for "both changesets", and the two audiences differ: a `@etherfold/platform-nodejs-fetcher` reader needs "your binary is gone, here is what replaces it" at the top, an `etherfold` reader needs the new command. The repo has both patterns (`the-old-indexer-shape-is-deleted` bundles; most others do not).
- **`@etherfold/server` added as a `devDependency` of `etherfold`.** The loop test needs the REAL receiver on the other end of the wire (a mocked one would mock the thing under test: the cursor and the `409` live there), following `packages/fetcher-host/test/harness.ts`. Test-only, so it does not change what the CLI ships. `ldenv` and `named-logs-console` were removed from the fetcher platform's runtime dependencies, since only the deleted entry point used them.
- **`CONTEXT.md`'s command-set bullet was corrected, which also fixed drift I did not create.** It said three of the five commands "do not exist at all", naming `run` among them — already false before this task (`run-follows-folds-and-answers-in-one-process` did not update it), and my change makes it falser. Since that bullet is explicitly the thing a reader is told to land on, I corrected the whole sentence rather than only my half.


wighawag:

Reviewed against the eleven acceptance criteria in `work/tasks/backlog/fetch-is-the-only-way-to-run-a-fetcher.md`. Gate-1 and CI `verify` are green.

## Criteria

1. **`etherfold fetch` runs the loop and pushes contiguous ranges to a remote receiver, keeps running after the tip** ✅ — driven against a real fake chain and a real receiver (`test/utils/receiver.ts`), asserting every log lands.
2. **Stops on container signals, cycle in flight finishes, same exit codes as the retired binary** ✅ — 0 on a signal, non-zero on a refusal no retry fixes ("instead of staying up achieving nothing"), non-zero on a configuration refusal without starting a loop.
3. **`--store`/`--db` REFUSED on `fetch`, naming why; missing inputs refused by flag AND variable** ✅ — the spec's answer 3 correction, delivered exactly.
4. **Holds no cursor** ✅ — asserted the strong way: a second run starts where the RECEIVER says, not at the start block. No state file, no lock file, no from-block flag.
5. **Source without a processor module** ✅ — deployments folder with no module anywhere; and the module route is refused naming both explicit forms.
6. **The standalone binary is GONE** ✅ — `bin.ts` deleted, `bin` field removed, library entry kept.
7. **The console log sink moves to the CLI's fetch path** ✅.
8. **Fetcher platform README points at the command** ✅.
9. **Both changesets** ✅ — `etherfold: minor` for the command, and a separate `@etherfold/platform-nodejs-fetcher: major` that says BREAKING plainly for the removed `bin`. Splitting them is right: they are two packages with two different bump reasons.
10. **Tests: resolution reaching the host, refusals, one real drive of the loop** ✅.
11. **Gate green** ✅.

With this there is exactly one way to run a fetcher, which is what the spec asked for. Merging.

PR #57, opened 2026-09-03
feat(index-receives-a-pushed-stream-and-owns-the-database): `etherfold…

wighawag:
Task: `work/tasks/done/index-receives-a-pushed-stream-and-owns-the-database.md`

Done. `etherfold index` exists, the five-name set is complete, and the full gate is green.

## What I built

**`packages/cli/src/indexCommand.ts`** — the receiver. It resolves the `index` row of the existing configuration table, loads the processor module, builds the store on ONE libSQL handle, opens the explicitly-given source, constructs the same `StreamBuilder` `run` folds through, and starts the Node adapter on that handle with `getIngestion` (the stream-builder) and `getCursorReport` (the same `readCursorReport` `run` builds). No provider, no `LogFetcher`, no fetcher host anywhere in the path. `indexMain` is the process shape: `0` on a signal, `1` when it could not start. Registered in `src/program.ts` alongside the other four.

**`packages/cli/src/folding.ts`** — the folding assembly the three database-owning commands now share (`STREAM_CONFIG`, `buildProcessor`, `openExplicitSource`), extracted from `src/index.ts` so `index` reuses it rather than forking a second one. `fetch.ts` drops its private copy of the source opener.

**Tests.** `test/indexCommand.test.ts` (12 cases): the real `etherfold fetch` command driving the real `etherfold index` command over a real socket; the ingestion routes answering rather than `501`; `/status` cursor advancing; 404 on every plausible query route; a wrong secret refused `401` with nothing applied; no secret at all refusing to *start*; the resolved token beating an ambient one; a replayed batch refused `409` and the named resume accepted; one handle built once, proved by reading the server's own `Meta` through the store's handle; `-n` refused and a module-only source refused naming both explicit forms.

`test/equivalence.test.ts` (2 cases, the payoff): the same processor, the same declarations, one fixture chain including a reorg whose replacement branch carries **one** event where the dead branch carried two — run once through `run` and once through `fetch` + `index`, landing on identical state and an identical cursor; then `serve` over the receiver's database answering the same reads as `run` and reporting no cursor.

Docs: CLI README (a full `index` section, the read tier's cursorless `/status`, "the five names and the two compositions"), `CONTEXT.md`'s command-set bullet (now **BUILT**, caveat removed) and its query-layer bullet, root README, `platforms/nodejs` README, package description, and a changeset.

Gate: `format:check`, `build`, `typecheck`, `check:adr`, `test` all green; the two new files ran clean three times in a row.

## Decisions

- **The command's module is `src/indexCommand.ts`, not `src/index.ts`.** Every other command is `src/<command>.ts` (`run.ts`, `fetch.ts`, `serve.ts`), but `src/index.ts` is the package barrel (`"module": "dist/index.js"`) and also hosts the `build`/`run` assembly. Alternatives considered: renaming the barrel (touches the package entry point and every importer, for cosmetics), or putting the `index` assembly inside the barrel (mixes a command with the export surface and hides it from the file-per-command convention). Touches: nothing but this package's file layout; the exported symbol is still `index`.
- **The handle type is `RunningReceiver`, which now shadows a same-named type in `test/utils/receiver.ts`.** The src type is the command's public handle (`run`'s is `RunningIndexer`, so "Indexer" was taken and `indexer` means the named multi-tenancy unit in the glossary); the test util is fetch's hand-built receiver. They never meet in one file. Worth knowing: now that `index` ships, `test/utils/receiver.ts` is arguably redundant — `fetch.test.ts` could drive the real command — but rewiring another task's tests was out of scope. Touches: `packages/cli/test/utils/receiver.ts` (unchanged in meaning), `fetch-is-the-only-way-to-run-a-fetcher`.
- **How the source is given, and what is refused: reused verbatim, nothing invented.** `one-configuration-path-for-every-command` had already landed the `index` row (`nodeUrl`/`rps`/`ingestEndpoint` refused, `source` required through `requireExplicitSource` with the `NO_CHAIN_INDEX` reason, `ingestToken` required). I added no flag, no variable and no new refusal message; the command consumes that resolver, which is what that task said it should. The one consequence worth stating: **"no shared secret configured" is a STARTUP refusal, not a running endpoint that answers `401`.** The token is `required`, so `etherfold index` with none never binds a port. The acceptance criterion's "a push with no secret configured is refused" is therefore unreachable *through the command*; the server's fail-closed guard for that state is already pinned by `packages/server/test/ingest.test.ts`. I did not weaken the requirement to make the criterion literally testable — a receiver that comes up as an endpoint refusing everyone is strictly worse than one that refuses to come up.
- **The resolved secret is passed to the host through `StartOptions.env`, so a flag beats an ambient `INGEST_TOKEN`.** The Node adapter merges `{...process.env, ...options.env}`, so the token this command resolved wins. The alternative (letting the adapter read the ambient variable) would mean `--ingest-token` parsed, resolved, and then silently ignored by the guard — accepted-and-ignored, which this CLI's configuration doc forbids. Asserted directly. Touches: `platforms/nodejs` (no change to it; only how it is called).
- **Auto-setup stays ON by default, and `index` additionally migrates the STORE's own tables before binding a port.** The server's fixed-table schema default is right for a writer that owns its database (it is what `/status` reads, and what the ingest route counts reorgs into), and `--no-auto-setup` still takes it back. Separately I call `processor.load(source, streamConfig)` once at startup: it is idempotent, and everything `load` refuses — an illegal entity declaration, a retention window that does not cover what a reorg can reach — is a fact about the deployment, not about the batch that happened to arrive first. Discovered lazily it is a `500` to a sender on a process still reporting itself healthy; discovered here it is a refusal that never starts ("no command silently degrades"). Deliberately not gated on `--no-auto-setup`, which is about the server's fixed schema; the entity tables are the store's own and the first push would create them anyway. Asymmetry with `run`/`build` (which still migrate lazily on the first cycle) is intended: there a bad declaration becomes a `fatal` and a non-zero exit, which is already loud. Touches: nothing outside `index`, but it is a user-visible startup behaviour a reviewer should ratify.
- **The bounded read-tier assertion: `createQuerySurface` over the entity declarations, opened on a fresh handle to the database URL `serve` was pointed at.** The generated read surface is the criterion's own wording; I used the SQLite tier (`createQuerySurface`) rather than `createReadSurface` because the bounded tier cannot ask for a whole entity (a listing needs an id prefix), and a read tier over SQLite is exactly where that tier is meant to be used. Both tiers are exercised — `getCurrent` by id and `queryCurrent` for the whole table — and compared against the same surface over `run`'s database. No HTTP query route was added, and none of the databases is reachable over SQL.
- **The read-tier `/status` parity is asserted on the schema version, and on the reorg counters against the WRITER rather than against `run`.** `recordReorg` is called from exactly one place, the HTTP ingest route, so a combined process folds through the direct in-process wire and counts **no** reorgs: `run` reports `{absence: 0, contradiction: 0}` where `index` folding the identical chain reports the revert it made. Asserting equality with `run` would have pinned a number meaning "nobody counted". So the test asserts that the read tier reports exactly what the database its writer wrote holds (`{contradiction: 1, absence: 0}`), plus schema-version parity with `run`, and says why in the file. Closing the gap means deciding where an operational counter is written when there is no HTTP route (and whether the CLI takes `@etherfold/server` as a runtime dependency, and what `build` does with no `Meta` table) — a design change to two other commands, so it is captured in `work/notes/observations/a-run-process-counts-no-reorgs-on-status.md` rather than taken here. Touches: `run-follows-folds-and-answers-in-one-process`, `the-one-shot-is-build-and-serve-is-only-the-read-tier`.
- **Cursor parity between `index` and `run` is asserted on `lastToBlock` and `latestBlock`, not on `lastFromBlock`.** The last range's START depends on how many cycles each side ran while the chain sat at a tip, and a socket makes a cycle slower than a function call — it is timing, not a property of the deployment shape (observed: 1000083 vs 1000084 on an otherwise identical run). The two fields asserted are the ones the report exists to answer: is it moving, and how far behind is it. Never asserted against `serve`, which owns no store and is given no reporter; the test pins that its `/status` carries no `cursor` field at all.


wighawag:

Reviewed against the fourteen acceptance criteria in `work/tasks/backlog/index-receives-a-pushed-stream-and-owns-the-database.md`. Gate-1 and CI `verify` are green. This is the last task of `one-command-runs-the-whole-pipeline`, and it carries the two EQUIVALENCE criteria that are the spec's whole claim.

## Criteria

1. **`etherfold index` receives pushed batches, folds them into the named database, keeps running** ✅.
2. **NO chain call, asserted rather than inspected** ✅ — no node URL accepted (refused naming what this command is instead), the module route refused naming both explicit forms, and a module that keys its contracts per chain folded from the source it was GIVEN.
3. **Ingestion hosted through the Node adapter, not a second server host in the CLI** ✅ — the wire routes work rather than answering 501, on top of PR #52.
4. **No query API beyond `/status`** ✅ — asserted.
5. **Authenticates or refuses everyone** ✅ — and stronger than asked: it refuses to START with no secret configured rather than opening a write endpoint, it checks the secret THIS command resolved (a flag beats an ambient variable), and a wrong-secret sender lands nothing and exits on the refusal.
6. **One database handle** ✅.
7. **EQUIVALENCE: `run` equals `fetch` plus `index`, reorg included** ✅ — same state AND the same cursor from the same chain, with the transport as the only difference. This is the assertion the spec's engine choice (one `StreamBuilder` behind all four folding commands) was made to make meaningful, and it is exactly the one that would have been an equivalence between two implementations otherwise.
8. **EQUIVALENCE: `index` plus `serve` answer what `run` answers, bounded to existing surfaces** ✅.
9. **The cursor is deliberately NOT in read-tier parity** ✅ — "reports no cursor" is asserted on the read tier, as the task required.
10. **Idempotent by cursor** ✅ — a replayed batch is refused with a 409 carrying where to resume, applying nothing twice.
11. **Real sender against real receiver over real HTTP** ✅.
12. **README documents the split shape** ✅.
13. **`CONTEXT.md` no longer says DECIDED, NOT YET BUILT** ✅ — it now says BUILT, all five names exist.
14. **Gate green** ✅.

## The observation filed here is worth acting on

`a-run-process-counts-no-reorgs-on-status`: `recordReorg` is called from exactly one place, the HTTP ingestion route, so a COMBINED process folding through `createDirectIngestion` reverts state on a reorg and counts nothing. The absence-versus-contradiction rate is the signal that distinguishes "your logs are being truncated" from "the chain reorged" (ADR-0004), and it is silently unavailable on the deployment shape this milestone calls the default. Correctly not fixed here (it is a server/core seam, not the receiver's), and I am carrying it into the triage as task-worthy.

With this merged the spec is complete: `run`, `build`, `fetch`, `index`, `serve`, and the cursor on `/status`. Merging.

PR #58, opened 2026-09-03
feat(a-reconfigure-hashes-the-resolved-stream-config-everywhere): A rec…

wighawag:
Task: `work/tasks/done/a-reconfigure-hashes-the-resolved-stream-config-everywhere.md`

## What I built

**The defect, confirmed against current reality first.** I checked the task's premises before building: both sites are still exactly where it says (`packages/core/src/indexer.ts`, `reinit` hashing the resolved `this.config.stream` and `updateIndexer` hashing `update.streamConfig` as provided), `resolveStreamConfig` still fills `finality: 17`, `streamDigestOf` still takes the resolved config (ADR-0006/ADR-0034 as amended), and `sourceInvalidationOf` still invalidates **both** halves from block 0 on a config mismatch. No drift.

**The test first, and it was red.** `packages/core/test/updateIndexer.test.ts`, driving the engine over the core `streamCacheWorld` harness. Five behavioural cases went red against the unfixed `updateIndexer` (I verified by temporarily restoring the old expression) and are green now.

One finding worth stating, because it changed how the test asserts: the damage is **worse without a stream cache than with one**. With a cache, the discard re-folds from the stored stream (whose own context was written *resolved*, so it still matches) — the state is thrown away and the published verdict lies, but the node is not hit. With no cache there is nothing to rebuild from, so the same discard re-fetches from the start block. The suite therefore drives both worlds, and the ADR-0034 pair is asserted on all of: the verdict, `stateDiscarded`, the ranges asked of the node, the cache clears, and whether any event was delivered to the processor a second time.

**The fix.** `streamConfigHashOf(stream)` in `internal/engine/utils.ts`, beside `resolveStreamConfig`: resolve, then hash. Both verbs go through it, and so do `wireContextOf` and `captureStream`. `resolveStreamConfig` is idempotent, so every digest is byte-identical to what was persisted before — nothing is re-keyed. `simple_hash`, `canonical_form`, `resolveStreamConfig` and `streamDigestOf` are untouched, and the true positive is intact: `alwaysFetchTimestamps`, `alwaysFetchTransactions`, `parse.filters` and an explicitly different `finality` each still invalidate both halves from block 0, clear the cache and re-fetch from the start block, all asserted.

**Making a second divergence hard.** Three tests: no file under `packages/core/src` other than the defining one applies `simple_hash` to anything config-shaped; `indexer.ts` contains no `simple_hash` at all and reaches `streamConfigHashOf` exactly twice; and the digests themselves are pinned as literals (`h10lkzm2` for the resolved default, on both persisted cursors) so a re-key cannot pass silently.

Changeset: `.changeset/a-reconfigure-hashes-the-resolved-stream-config-everywhere.md`. `pnpm build && pnpm typecheck && pnpm test`, `format:check` and `check:adr` are green.

**Out-of-scope signal captured** (not fixed): `work/notes/observations/an-explicit-undefined-finality-resolves-to-no-finality-at-all.md` — `resolveStreamConfig({finality: undefined})` spreads the key back over the default and returns a config with no `finality` at all, so `finalityDepth` is `undefined` and the hash is that of `{}`. It is the resolver's own spread, which this task is forbidden to touch.

## Decisions

**Where the single resolve-then-hash step lives: `streamConfigHashOf` in `packages/core/src/internal/engine/utils.ts`, immediately beside `resolveStreamConfig`.** Chosen because that file already owns the other two questions this one is an instance of (`resolveStreamConfig`, `wireContextOf`), and because both halves of a split deployment already import their identity rules from there, so the sender and the receiver cannot reach two answers. Alternatives considered: `utils/hash.ts` (rejected — those bytes are the shared, persisted canonicalisation and a config-aware helper does not belong beside them), and a private method on `IndexerGeneration` (rejected — it would be unreachable from `captureStream` and `wireContextOf`, which is precisely how the split happened). The name is not a new concept: it names the existing `streamConfigHash` field / `ContextIdentifier.config` value, follows the repo's `...Of` convention, and adds nothing to the glossary.

**I exported `streamConfigHashOf` from `@etherfold/core` (a public-surface change, hence the changeset).** This is what keeps a *third* caller from reintroducing the split: a host or a test building a `ContextIdentifier` or `WireContext` of its own must reach the same digest the engine stored, and the structural test only guards `packages/core/src`. The rationale is `resolveStreamConfig`'s own, one step further along. Alternative: keep it internal (rejected — it leaves every cross-package caller re-deriving `simple_hash(resolveStreamConfig(x))` by hand, which is the exact expression that drifted). Touches: the published surface of `@etherfold/core`; nothing else changes call shape.

**I routed `wireContextOf` and `captureStream` through it too, not just the two named call sites.** `captureStream` was a genuine THIRD instance of the same defect — it hashed `options.streamConfig ?? {}` as provided, so a fixture captured with `finality` unset wrote a `config` digest no indexer running the resolved default could match, and the load path would read the replayed stream as belonging to another config and clear it. Making the acceptance criterion "no second site hashes a `ProvidedStreamConfig`" *true* required fixing it. Consequence to be aware of: a NEWLY captured fixture whose `streamConfig` omits `finality` now records the resolved digest instead of the unresolved one. No committed fixture moves — nothing is regenerated, and the one committed stream fixture (`stratagems-base.stream.json`) was captured with `{finality: 12}` spelled out, so its digest is identical either way. `wireContextOf` takes a `UsedStreamConfig` and so was already correct; it goes through the step for the structural rule rather than for a behaviour change, and idempotence is asserted so the wire keeps its bytes.

**The guard is a source scan, in the house style of `streamIdentity.test.ts`.** A comment does not satisfy the criterion, and a purely semantic test cannot see a second site that happens to agree today. The scan asserts the *set* of files that hash a config equals exactly one. Its known limit: it matches on `simple_hash(` textually, so an aliased import would slip past it — which is why the companion assertion is that `indexer.ts` contains no reference to `simple_hash` at all.


wighawag:
## APPROVE ✅ — Gate-3 (conductor's diff-vs-criteria review)

Reviewed the diff against every acceptance criterion, every "What this is NOT" fence, and the four hard constraints in the prompt. CI `verify` green.

- ✅ **Regression pinned FIRST and red before the fix.** Verified independently, not taken on trust: I reverted *only* the two hash expressions in `indexer.ts` to their pre-fix form in a throwaway worktree and ran `test/updateIndexer.test.ts` — **5/5 behavioural tests fail**, on `expectNothingMoved` finding events re-delivered to the processor (the fold was discarded and re-folded). Green on the branch as shipped. The test asserts the ADR-0034 pair (ranges fetched *and* state discarded) rather than restating the criterion.
- ✅ **Equivalent-but-unresolved config is a no-op.** `discards nothing and re-fetches nothing` asserts `stateDiscarded === false`, `sourceInvalidation === {state:{valid:true},stream:{valid:true}}`, no new clears, no batch re-delivery, and no range reaching back past `START_BLOCK`. Driven in *both* worlds, and the no-cache variant is the one that shows the actual outage (nothing to rebuild from → full re-fetch from the node).
- ✅ **Not "equal objects compare equal".** `is still a no-op when something else about the reconfigure DID change` moves the source (gains a view function, free per ADR-0034) and passes a config carrying a second field. What is compared is the resolved form.
- ✅ **A genuinely moved config still invalidates both halves from block 0.** Parameterised over `alwaysFetchTimestamps`, `alwaysFetchTransactions`, `parse.filters` and an explicit `finality: 5`, each asserting `{valid:false, invalidFromBlock:0, reason:'stream-config'}` on both halves; plus a test that the cached stream is cleared and the next range starts at `START_BLOCK`. The true positive is intact — this removes a false positive, not the rule.
- ✅ **`{finality: 17}` and unset are ONE config**, asserted in *both* directions (unset→spelled-out and spelled-out→unset).
- ✅ **The two call sites cannot diverge again — structurally, not by comment.** `streamConfigHashOf` is the single resolve-then-hash step, and the structural tests assert (a) exactly one file under `src/` hashes anything config-shaped, (b) `indexer.ts` contains no `simple_hash` at all and exactly two `streamConfigHashOf(` calls, comments stripped before matching so prose cannot satisfy or trip it.
- ✅ **No digest moves, nothing re-keyed.** `simple_hash`/`canonical_form` are not in the diff at all; `streamDigestOf` untouched. Idempotence is asserted (`streamConfigHashOf(provided) === simple_hash(resolveStreamConfig(provided))`), the persisted cursor bytes are pinned as literals on *both* the stream cursor and the saved state (`h10lkzm2`), and the raw digests are pinned (`h10lkzm2`/`h28y`/`ht6tzx8`). I also checked the one checked-in fixture (`stratagems-base.stream.json`): its cursor carries `h10lkvwl`, a resolved config, so routing capture through the new step strands nothing.
- ✅ **`ReconfigureOutcome.sourceInvalidation` publishes the corrected verdict** on the no-op case, asserted directly.
- ✅ **Changeset** present (`@etherfold/core`, patch) and it explains the blast radius honestly.
- ✅ `build && typecheck && test` green in the fresh-worktree gate and in CI.

### Scope fences honoured

- `resolveStreamConfig` **unchanged** — the new function sits beside it and calls it; the default `finality: 17` is untouched.
- `simple_hash` / `canonical_form` **byte-for-byte untouched** (not in the diff).
- **No stream re-keyed**; `streamDigestOf` not modified.
- One-writer/other-behaviour fences: n/a to this task.

### On the two sites beyond `indexer.ts`

`wireContextOf` and `captureStream` were also routed through `streamConfigHashOf`. This is **in scope, and required by the "no second site" criterion** rather than creep: `captureStream` hashed `options.streamConfig ?? {}` as *provided* — the same defect, which would write a fixture cursor no indexer running the default `finality` could match. `wireContextOf` already received a resolved config, so its bytes are unchanged by construction and the idempotence test pins exactly that.

### Off-path finding, correctly captured not fixed

The build filed `work/notes/observations/an-explicit-undefined-finality-resolves-to-no-finality-at-all.md`: `{finality: undefined}` spreads the key back over the default, yielding `finality: undefined` (reorg arithmetic on `NaN`) and a digest equal to `simple_hash({})`. That is a real, distinct defect on the same line of code, and it was right to record it rather than widen this task. Left for triage.

Merging.

PR #59, opened 2026-09-03
feat(a-follower-replays-a-retraction-a-paused-writer-appended): A follo…

wighawag:
Task: `work/tasks/done/a-follower-replays-a-retraction-a-paused-writer-appended.md`

## Drift check (first, as the task demanded)

The task's premise holds exactly as written, so I built it:

- `IndexerGeneration.promiseToFollow` (`packages/core/src/indexer.ts`) still had the guard verbatim: `if (lastSyncStored.lastToBlock <= current.lastToBlock) return current;`.
- Pause still caps `toBlock` before the `fromBlock > toBlock` guard and leaves `latestBlock` tracking the real head (ADR-0045), so a paused writer's `lastToBlock` really is frozen while it appends during its drain.
- ADR-0044 (a follower fetches nothing, holds a read-only view), ADR-0045 (pause, and its own last bullet already predicted this exact hole and pointed at the follow path) and ADR-0042 (a replay honours the verdicts it carries) all match. Both dependency tasks are in `work/tasks/done/`.

Two process notes for the human, no action taken by me: the body lives at `work/tasks/backlog/…` (not `ready/`), and `work/questions/task-a-follower-replays-a-retraction-a-paused-writer-appended.md` holds an unanswered "cancel this item?" question from a previous run that produced an empty diff. The defect was real and is now fixed, so that question should be answered "requeue", not "cancel".

## What I built

Test first, red before, green after (`packages/core/test/follower.test.ts`, new describe block plus small additions to the shared world: a `round` helper, a per-generation `reparse` counter, `batchesOf`, `heldOf`/`id`, and `cursorOf` on the keyed stream):

- **The pinned regression** (was red): writer and follower level at the tip of branch A, writer PAUSED, chain reorgs at 104 (below the cap), writer appends `R:0xa104` + `A:0xb104` while its stored `lastToBlock` stays 105. Before the fix the follower kept `0xa104`; now it replays both and equals the writer, and equals `refoldStoredStream` (a from-scratch fold of the stored stream, no node).
- Delivery order: retraction then replacement, as the writer emitted them.
- Zero `eth_getLogs`, zero clears, one stream digest, stream content is the writer's emissions only.
- **Idle costs nothing** (two tests: a running writer at a static tip, and a DRAINED paused writer): zero `reparse` calls and zero processor batches over repeated cycles. I verified these two go red if the early return is disabled, so they genuinely fence "the fix is not remove the early return".
- Running (unpaused) writer: unchanged behaviour, asserted explicitly.
- Resumed writer: follower keeps following the uncapped stream.

The fix in `packages/core/src/indexer.ts`: a follower remembers `followedEmissions`, the emissions it last folded over the range it resumes from (`emissionMarkOf` = block hash + log index + applied/retracted, the same identity `sameEvent` already uses), and the early return now fires only when the stored cursor does not reach past this fold **and** the stream holds emission-for-emission the list already folded. It is set only on the two paths that actually fold, and cleared in `forgetStoredStream` and on reconfigure/processor swap.

Also: `.changeset/a-follower-follows-the-stream-not-the-cursor.md` (patch, `@etherfold/core`), `docs/adr/0049-…`, and one clause in ADR-0045's last bullet now pointing at 0049 instead of the observation note.

`pnpm build && pnpm typecheck && pnpm test` green; `pnpm format:check` and `pnpm check:adr` green; working tree holds only these five files.

## Decisions

- **The question that replaced the cursor comparison: the stream's own EMISSIONS over the range the follower resumes from, not any cursor.** A follower now keeps `followedEmissions` (block hash + index in block + applied/retracted, in stream order, the identity `sameEvent` already defines) and returns early only while what `fetchFrom` hands back over that range is that exact list. **Why it cannot go stale the way `lastToBlock` did:** it is not a claim the writer makes *about* the stream that some other mechanism can freeze — it is the content of the stream over precisely the bytes the next replay would consume, so any append there differs whatever a cursor says. Its failure direction is the safe one: it is recorded only where this generation has just folded that exact slice, so "unknown" means fold, and an unnecessary replay de-duplicates against the unconfirmed window and delivers nothing. The stored cursor is still consulted for the one half it cannot be wrong about (a stream reaching *past* this fold is new by definition), which keeps the cheap integer test first. **Alternatives considered and rejected:** add `latestBlock`/`lastFromBlock` to the cursor comparison (they move for reasons unrelated to appends, so an idle follower would re-walk every cycle — a performance bug for a correctness one, and the same summary-can-lie mistake in a new place); delete the early return (correct but re-decodes and re-walks the window forever); have the writer publish an append counter or stream digest (a new field on a cursor contract ADR-0035 deliberately keeps to four properties, maintained by every keeper, and still a summary); make a paused writer move `lastToBlock` (that cursor is the drain's termination condition — explicitly out of bounds). **What it touches:** only the follow path (`promiseToFollow`); `indexMore`, pause/resume, the one-writer rule and the read-only stream view are untouched, and there is no new flag, config key or public API. Recorded as **ADR-0049** (`docs/adr/0049-a-follower-asks-the-stream-what-changed-never-its-cursor.md`), because a cursor comparison is the obvious thing a future optimiser will reintroduce here — which is exactly how this bug arose.
- **The memory is discarded on reconfigure and processor swap, and in `forgetStoredStream`.** A reconfigure can point the keeper at a different stream, so what was folded no longer describes what will be read. Forgetting costs one replay that de-duplicates to nothing; keeping it could skip a real append. Touches `updateIndexer` / `updateProcessor` / `reset` only in that they now also clear this field.
- **I amended one clause of ADR-0045** (its final bullet said this hole was open and recorded in `work/notes/observations/`) to point at ADR-0049 instead. Alternative was to leave 0045 untouched and let it contradict the code; a stale "this is still open" line in an accepted ADR is exactly the kind of drift the next reader trusts. No decision in 0045 was changed.


wighawag:
## APPROVE ✅ — Gate-3 (conductor's diff-vs-criteria review)

Reviewed the diff against every acceptance criterion, all three "What this is NOT" fences, and the three hard constraints. CI `verify` green.

### The two fences, verified by mutation rather than by reading

Both forbidden fixes are genuinely excluded by the shipped test suite. I checked this myself in a throwaway worktree:

| Mutation | Result |
| --- | --- |
| Guard reverted to the pre-fix `lastToBlock <= current.lastToBlock` | **2 failed** — exactly the pinned regression pair; the other 15 stay green |
| Early return removed outright (`if (false)`, the forbidden "just replay always") | **2 failed** — exactly the idle-cost pair |
| As shipped | **17 passed** |

Only the shipped fix satisfies both directions. That is the strongest thing this task could ask for: the regression is genuinely pinned *and* the performance fence has teeth, so neither can silently rot.

- ✅ **Regression pinned FIRST and red before the fix** (verified above). The scenario is built honestly: writer indexes branch A, is paused (`pauseState: 'draining'`), the reorg lands at block 104 **below** the cap, and the test asserts the stored cursor is still frozen at `lastToBlock: 105` — pinning the *premise* itself, not just the symptom — while the stream has grown by the retraction and its replacement.
- ✅ **Follower replays both, in emission order, landing where a from-scratch fold lands.** Order asserted as `[R:…, A:…]`; the terminal state is asserted equal to `refoldStoredStream(...)`, i.e. against an independent from-scratch fold rather than a hand-written expectation.
- ✅ **An unchanged stream still costs nothing.** Asserted by counting `reparse` calls (the first cost of an advance) and batches, over three further cycles, as **equal** and not "fewer" — and asserted twice, for an idle running writer and for a fully `drained` one.
- ✅ **Still zero fetch, zero writes, zero clears**: `fetchesBy('B') === []`, `clears === 0`, one stream digest, and the stream holds exactly the writer's emissions (`BRANCH_A.length + 2`).
- ✅ **A RUNNING writer behaves exactly as today**, asserted on the same reorg with no cap anywhere.
- ✅ **A RESUMED writer's follower is unaffected**, including a block appended above the old cap after `resume`.
- ✅ **Changeset** present (`@etherfold/core`, patch).
- ✅ `build && typecheck && test` green in the fresh-worktree gate and CI.

### Scope fences honoured

- **Pause is untouched.** No change to the cap, the drain, or `latestBlock` tracking. The only edit to ADR-0045 is a cross-reference on the very bullet that predicted this bug, rewritten to point at ADR-0049 and ending "The cap is unchanged." The drain's termination condition still rests on the frozen `lastToBlock`, and the ADR explicitly rejects moving it.
- **The early return is not deleted.** It is narrowed by conjunction — the cursor still contributes the half it cannot be wrong about (a stream reaching *past* this fold is new by definition), and the emission comparison settles the half where it could lie.
- **One-writer rule intact**: the follower still holds a read-only view, appends nothing, fetches nothing.
- **`CannotPauseFollowerError` untouched.**

### On the mechanism, and why it does not go stale

The replacement is content-derived: `emissionMarkOf` = `blockHash:logIndex:A|R`, the same identity `sameEvent` already uses. Deliberately **not** the block number (a reorg reuses it) and **not** decoded `args` (ADR-0034: those are what some ABI made of the bytes). It cannot go stale the way `lastToBlock` did because it is not a claim the writer makes *about* the stream — it is the stream's content over exactly the slice the next cycle would replay.

Two details I specifically checked, because they are where this shape usually goes wrong:

1. **No new read cost on the idle path.** `keepStream.fetchFrom(...)` already ran before the guard on `main`, so `hasAlreadyFolded(eventsStored)` compares a slice already in hand. The expensive work (`reparse`, `replay`, processor delivery) is still skipped. Confirmed against pre-fix `main`, not assumed.
2. **The memory's failure direction is the safe one.** It is set *only* after a successful fold (explicitly after `replay` returns, so a throwing replay does not record the slice as folded), and cleared on reconfigure, processor replacement, and stream reset. `undefined` therefore means FOLD, which costs a de-duplicated walk and never a skip.

ADR-0049 is warranted and well made: it enumerates and rejects the two forbidden fixes with reasons (a richer cursor comparison "repeats the original mistake in a new place"; unconditional replay "trades a correctness bug for a performance one"), records the honest cost (the shortcut fires slightly less often after the resume point moves — one extra de-duplicated walk, not a per-cycle cost), and leaves the standing warning that a cursor-derived shortcut is the thing to be suspicious of on this path. It also states the composition-defect framing the task asked for: neither ADR-0044's guard nor ADR-0045's pause is wrong alone; the bug lived only where they met.

Merging.

PR #60, opened 2026-09-05
feat(every-deployment-shape-counts-the-reorgs-it-concluded): Every depl…

wighawag:
Task: `work/tasks/done/every-deployment-shape-counts-the-reorgs-it-concluded.md`

Done. Everything is green: `pnpm format:check && pnpm check:adr && pnpm build && pnpm typecheck && pnpm test`.

## Drift check (done first)

The task's premises hold exactly as written: `recordReorg` was called from one place (`packages/server/src/api/ingest.ts`, `recordReorgSafely`), `run` folded through `createDirectIngestion` and reported `{absence: 0, contradiction: 0}` for ever, `build` had no `Meta` table, and `packages/cli/test/equivalence.test.ts` carried the exception with the comment pointing at `work/notes/observations/a-run-process-counts-no-reorgs-on-status.md`. ADR-0004 says what the task says it says. No drift; I built to the decided shape.

One imprecision, not load-bearing: the task cites ADR-0042 as "the shape ... caught once" for "one write in two places". ADR-0042 is *A replay honours the verdicts the stream carries*, whose rejection is "two contracts in one function", not a duplicated write. The principle is close enough that the decided design does not rest on the citation, so I proceeded and cited ADR-0004 and the new ADR-0050 instead.

## What I built

- **`@etherfold/core`** — new `src/reorgCounters.ts`: `ReorgRecorder`, `ReorgCounters`, `RecordedReorg`, and the durable key names `REORG_COUNTER_KEY` / `REORG_LAST_KEY`. `StreamBuilder` takes `recordReorg` in its options (`StreamBuilderOptions`, not hashed into the wire identity) and calls it **once**, in `noteReorg`, after `process()` succeeds, catching and logging any failure. The absence/contradiction log line moved here too, since a combined process has no route to log from.
- **`@etherfold/server`** — `recordReorg` deleted; `readReorgCounters` stays (a read tier owns no store and still has to answer) and now uses core's keys. The ingest route is a plain caller of `receive`. Dependency posture unchanged: still owns no store package.
- **`@etherfold/platform-nodejs`** — `ensureFixedSchema(db)` extracted from `startServer`'s auto-setup, so a process that binds no port can create the fixed tables.
- **`etherfold`** — new `src/reorgCounters.ts` holds the single writer; `buildProcessor` binds it to the handle the command folds into and returns it, so `run`, `build` and `index` all count into exactly the database they fold into. `build` now applies the fixed-table schema (`applyFixedSchema: command === 'build'`), and its `--no-auto-setup` refusal message says why it is refused rather than reusing the "answers no queries" one.

Tests (written before the production wiring at the equivalence and stream-builder seams):
- `packages/cli/test/equivalence.test.ts` compares `/status` counters between `run` and `fetch` plus `index` **directly**, exception and comment deleted: same counts, same classification, same block, and once each in both shapes (the split's `{contradiction: 1, absence: 0}` asserted unchanged). New: `run` under `--no-auto-setup` folds identically with no `Meta` table to count into, and a third `describe` asserting a database `build` emitted carries the schema *and* the counters.
- `packages/core/test/streamBuilder.test.ts` — six cases on the recorder: once per revert, the same detection the outcome carries, nothing on a batch that concluded nothing, a throwing and a rejecting recorder both leave the fold and the answer intact, identical fold with no recorder, and the recorder not hashed into `context`.
- `packages/core/test/oneReorgWriteSite.test.ts` — scans every `packages/*/src` and `platforms/*/src` and asserts exactly one write site.
- `packages/cli/test/reorgCounters.test.ts` — the writer itself, read back through the server's reader.
- Server and platform tests now supply the recorder / write the row the way a host does.

Docs: ADR-0050, root README, `packages/cli/README.md`, `packages/server/README.md`, the **reorg cause** glossary entry in `CONTEXT.md`, and a changeset for the four published packages whose surface changed.

## Decisions

**1. The single writer ended up in `etherfold` (the CLI), at `packages/cli/src/reorgCounters.ts`, bound by `buildProcessor`.** The task decided "persisted by whoever owns the store", and in all three folding shapes (`run`, `build`, `index`) that is the CLI: `buildProcessor` opens the libSQL handle and constructs the `VersionedStateStore`. Alternatives considered and rejected: `@etherfold/state-store-sqlite` (a store backend must not depend on `@etherfold/core`, which it would need for `ReorgCause` — ADR-0016, and the reason ADR-0027 keeps the sync cursor opaque); keeping it in `@etherfold/server` and importing it from the CLI (explicitly forbidden by the task, and it launders a server dependency into a write to a database the CLI opened); routing it through `@etherfold/platform-nodejs` (the same laundering, one level down). Touches: `run`, `build`, `index` all now go through `buildProcessor`'s new return value, and `@etherfold/server` lost an export.

**2. The durable key names moved to `@etherfold/core` (`REORG_COUNTER_KEY`, `REORG_LAST_KEY`), not to either end.** Consequence of (1): the writer and the reader are now in different packages *by design* (a read tier owns no store and still answers `/status`), so two packages have to name one row. Core is the only package both already depend on, and the one that defines `ReorgCause`. The alternative — the CLI spelling the strings and the server spelling them again — is exactly how two ends silently stop describing the same database. Coherence check: this adds no new vocabulary; `ReorgCounters` and `RecordedReorg` are the existing server types moved, and `ReorgRecorder` is named off the existing verb `recordReorg`. Touches: `@etherfold/server`'s `readReorgCounters` and its re-exported types.

**3. `IngestionOutcome` did NOT change shape; `reorg` became REPORTED-only.** I considered making it carry whether the count landed (`{detection, recorded}`), which would ripple into the HTTP response body, `IngestionResponse`, `createHttpIngestion`, `createDirectIngestion` and the fetcher host, for no assertion that is not already available. Instead the field's contract is now documented as reported-and-never-delegated, and the "no second site" scan test enforces that no caller counts from it. What moved is the *store seam* (`buildProcessor`'s return) and the *receiver's* construction options (`StreamBuilderOptions`), not the outcome. Touches: nothing on the wire; the wire response is byte-identical.

**4. `build` applies the fixed-table schema unconditionally, and `--no-auto-setup` stays refused there with a new reason.** This is a user-visible behaviour change: a `build` against an unmigrated database now creates `Meta`. It follows from the decided requirement that a database `build` emits is a publishable artifact carrying its provenance, and `build` binds no port so nothing else would ever create it. I kept the flag REFUSED (rather than making it optional on `build`) because the flag means "somebody else migrates this at *startup*" and there is no startup to decline at; the refusal message now says that instead of reusing `NOT_SERVING_BUILD`'s "answers no queries". `run` and `index` are untouched: `--no-auto-setup` still declines the tables there, and a counter that cannot then be written is a logged miscount. Touches: the `build` row of `OWNERSHIP`/`REFUSALS` in `packages/cli/src/config.ts`, and `packages/cli/README.md`.

**5. `@etherfold/core` added as a *devDependency* of `platforms/nodejs`.** Its `serve.test.ts` used to import `recordReorg` from `@etherfold/server` to set a scenario up. Rather than re-spell the key strings in a test, it now writes the row itself using core's exported keys. Dev-only; the adapter's runtime dependency posture is unchanged.

**6. Server tests supply a host recorder (`packages/server/test/utils/hostRecorder.ts`) instead of asserting the route counts.** `@etherfold/server` cannot depend on the CLI that owns the writer, so its end-to-end "a receiver counts its reverts and `/status` reports them" tests now play the HOST, which is what a deployment does. The alternative was deleting those assertions and leaving the counting covered only at the CLI seam; keeping them costs one small test helper and preserves the coverage.


wighawag:

Reviewed the diff against each acceptance criterion, the "What this is NOT" fences, and verified the regression pin is real rather than restated.

## Acceptance criteria

| # | criterion | verdict |
| --- | --- | --- |
| 1 | `run` reports the reorg with the same classification `index` reports | ✅ `equivalence.test.ts` compares `run`'s `/status` counters against the split shape's directly |
| 2 | Exception deleted from `equivalence.test.ts`, with its comment | ✅ the `recordReorg`-is-route-owned exception and the pointer to the observation are both gone, replaced by a direct three-way comparison |
| 3 | Written exactly once per shape; split shape's count asserted unchanged | ✅ `{contradiction: 1, absence: 0}` re-pinned explicitly for the split shape, plus `contradiction + absence === 1` asserted on *both* shapes — a second call site on either entrance makes one of them 2 |
| 4 | ONE path; a test asserts no second write site | ✅ `oneReorgWriteSite.test.ts` scans shipped sources and pins `writeSites === ['packages/cli/src/reorgCounters.ts']`, with a `> 50` guard so it cannot pass vacuously, and separate assertions that neither the route nor `directIngestion` writes |
| 5 | A failing counter never takes down the fold, asserted on the combined shape | ✅ `--no-auto-setup` removes the `Meta` table, then the fold is asserted to complete and land identical state |
| 6 | `build` writes the same `Meta` schema + counters, asserted on an emitted DB | ✅ asserted on the artifact itself via `readSchemaState` + `readReorgCounters`, not on a live process |
| 7 | Readme says which shapes carry counters and what the classifications mean | ✅ root, cli and server READMEs |
| 8 | Server's dependency posture unchanged | ✅ `packages/server/package.json` untouched; it now reads and writes nothing |
| 9 | Changeset | ✅ |

## Scope fences

- **NOT a correctness fix** — honoured. `streamBuilder.ts` is purely additive: the revert derivation is untouched, only a post-fold `noteReorg` call is added. Notably `recordReorg` is deliberately kept OUT of the wire hash, since hashing it would refuse batches from a sender wired to a different database.
- **NOT a richer query layer** — no endpoint added.
- **NOT a change to what the counters mean** — ADR-0004's absence/contradiction distinction preserved verbatim.
- **NOT a move of the store package into the server** — the server now writes nothing at all; the posture moved in the safe direction.
- **NOT a re-opening of the decided shape** — built to both decisions (core reports / store owner persists, and `build` gets its `Meta` table) rather than re-deriving them. ADR-0050 records the seam move and names the four rejected alternatives.

## Regression pin verified

The criteria pin the defect from `a-run-process-counts-no-reorgs-on-status`, so I confirmed it is a real pin rather than a restatement. In a throwaway clone (never the working checkout) I disabled the recorder call inside `StreamBuilder.receive`, reproducing the pre-fix "transport owns the write" behaviour. Result: **3 tests go red** with exactly the historical symptom —

```
AssertionError: expected { absence: +0, contradiction: +0 } to deeply equal { contradiction: 1, absence: +0 }
```

failing the combined comparison, the `build` artifact check and the read-tier check. Baseline on the branch is 4 passed. The pin holds.

Merging.

PR #61, opened 2026-09-05
feat(a-named-indexer-is-a-route-segment-and-a-registry-entry): A named…

wighawag:
Task: `work/tasks/done/a-named-indexer-is-a-route-segment-and-a-registry-entry.md`

All gates green. Here is the report.

## What I did

Made `@etherfold/server` host several **named indexers**, and moved both halves of the wire onto the namespaced routes.

**Drift check first.** The task's premises still hold: the code had `ServerOptions.getIngestion` resolving one `LogIngestion` and the unnamespaced `/ingest` + `/ingest/expected-from-block`; ADR-0004's envelope/refusal families and ADR-0036's named-indexer vocabulary are intact; `CONTEXT.md` explicitly recorded the route segment as NOT built. No re-scope needed.

**Test first, at the seam.** `packages/server/test/ingest.test.ts` was rewritten to deploy a host with N named indexers and to assert the new properties (per-name answers, isolation, unnamespaced pair gone, unknown name refused, 401 still ahead of the registry lookup). It went red (25 failures), then the production code landed.

**Server** (`packages/server`)
- New `src/registry.ts`: `IndexerRegistryEntry` (`{ingestion}`), `IndexerResolver`, and `indexerRegistry(record)`; all exported.
- `ServerOptions.getIngestion` → `getIndexer(c, name)`, resolving an entry.
- Routes are `POST /:indexer/ingest` and `POST /:indexer/ingest/expected-from-block`; the token guard moved with them (`/:indexer/ingest` and `/:indexer/ingest/*`). One `resolve()` helper produces either the entry or the refusal.

**Sending half** (`@etherfold/core`): `createHttpIngestion({endpoint, indexer, token})` posts to the namespaced paths, encodes the name into one segment, refuses a blank name at construction, and explains a `404` in the refusal it raises.

**Every caller moved**: `@etherfold/fetcher-host` (`indexer` config from `INDEXER_NAME`, demanded alongside `INGEST_ENDPOINT`/`INGEST_TOKEN`), `platforms/nodejs` (`StartOptions.getIndexer`), `platforms/cf-worker` (docs/comments), the CLI (`--indexer` on `fetch`/`index`, registry injection in `index`), and the tests in cli, core, fetcher-host, server, platforms/nodejs and platforms/nodejs-fetcher. Round trip: `fetcherRoundTrip.test.ts` drives a real `LogFetcher` through the real client into the real app under a name, plus a new case for a fetcher pointed at a name this host was not built with.

**Standing conventions**: two changesets (`@etherfold/server`, `@etherfold/core`, `@etherfold/fetcher-host`, `@etherfold/platform-nodejs`, `etherfold`; and `@etherfold/platform-nodejs-fetcher`), `CONTEXT.md` updated where it claimed the route segment was unbuilt, READMEs updated on server/cli/nodejs/cf-worker/nodejs-fetcher.

`pnpm format:check && pnpm check:adr && pnpm build && pnpm typecheck && pnpm test` all green (`changeset status --since=main` only sees changesets once they are tracked, which is the runner's commit).

## Decisions

- **The registry entry is `{ingestion: LogIngestion}` and carries NO `name`.** Chosen so the sibling widening (several live wire contexts per entry) is an added field rather than a change to the resolver's return type, which is exactly what the task asks for. The name is deliberately absent from the entry because the ROUTE SEGMENT is the single source of that value and a second copy is something a write path could key on while it disagrees with the URL; the next task (`the-emission-stream-table-...`) reads the name from `c.req.param('indexer')` in the same handler. Alternative considered: `{name, ingestion}` (self-describing, but two sources of one discriminator).
- **An unknown name is `404 unknown-indexer`; a host with NO registry at all keeps `501 ingestion-not-configured`.** A name is a route, so "no such tenant" is a routing refusal, not ADR-0004's payload `400` family (which stays exactly as it was) and not a resumable `409`. Keeping `501` for the no-registry case preserves what `run` and `serve` have always answered, and says the honest thing: one is a capability this host lacks, the other a tenant it was not built with. It is never a default to "the only indexer this host holds". Touches `createHttpIngestion` (404 joins the non-retryable 4xx family, with a hint naming the name) and the `run`/`serve` tests. Alternative considered: `400` for coherence with the wire families, rejected because it would re-mean the payload family.
- **The token guard stays AHEAD of the registry lookup**, so an unauthenticated caller gets `401` even for an unknown name and cannot enumerate the names a host was built with. Asserted.
- **New export `indexerRegistry(record)` in `@etherfold/server`.** Sugar over the resolver for hosts that know their names up front (used by the round-trip test and three harnesses); own-property lookup so `constructor`/`toString` resolve to nothing. The CLI's `index` deliberately does NOT use it and writes the two-line resolver inline, because that module imports `@etherfold/server` lazily on purpose (a folding process must not pull hono into its import graph).
- **`createHttpIngestion` throws on a blank name** (a plain `Error`, matching the existing "no global fetch in this runtime" refusal beside it). This is a new refusal, at construction rather than at the first push, on the ground that a missing name is a deployment that was never configured.
- **CLI: `--indexer <name>` / `INDEXER_NAME` is REQUIRED on `fetch` and `index`, and REFUSED on `run`, `build` and `serve`.** Required-and-never-defaulted follows the vocabulary ("supplied by the operator, refused when absent") and ADR-0048's rule that only `--port` may default. Refused on the other three because none of them routes a batch by name, and this repo refuses rather than accepts-and-ignores; refusing is also the reversible direction (accepting a flag later is free, refusing it later is breaking). What it touches: it is a new row in `INPUTS`/`OWNERSHIP`/`REFUSALS`, a new `INDEXER_NAME` variable in the CLI's published env contract, and a fence a later task must open if `run`/`build` ever need a name of their own (e.g. if the emission table keys `run`'s rows on one). Alternative considered: a constant default name in the CLI, rejected because a silent default is precisely the misdirected-batch hazard the discriminator exists to remove.
- **The fetcher-host's missing-variable message now formats three names as "A, B and C"** (two still reads "A and B", so the existing message is unchanged for the pre-existing case).


wighawag:

The largest task of this set, and the one most at risk of being mis-scoped in two specific ways: stopping at the server, and adding the namespaced routes while leaving the old ones live. Both were checked directly.

## Acceptance criteria

| # | criterion | verdict |
| --- | --- | --- |
| 1 | Namespaced routes answer for a registered name; an unregistered name is REFUSED, not defaulted | ✅ `404 unknown-indexer`, plus the batch is asserted to have gone nowhere — the name it was not sent to did not receive it either |
| 2 | **The unnamespaced pair no longer answers, asserted** | ✅ dedicated `describe('the unnamespaced routes')` asserting `404` on both paths, *and* that the cursor did not move — "nothing was applied by an old caller that never noticed" |
| 3 | `ServerOptions` resolves an entry OBJECT, so a later field is additive | ✅ `IndexerRegistryEntry {ingestion}` in the new `registry.ts`, with the widening for `the-server-and-cli-hold-generations-too` documented as the reason |
| 4 | `ingestClient` takes the name; full round trip | ✅ required, never defaulted (throws on empty), `encodeURIComponent`'d into the path; round trip covered in `fetcherRoundTrip.test.ts` |
| 5 | Every caller across core, fetcher-host, cli, server and platforms moved | ✅ 41 files across `core`, `fetcher-host` (incl. `src/config.ts`), `cli`, `server`, `platforms/nodejs`, `platforms/nodejs-fetcher`, `platforms/cf-worker` |
| 6 | Two named indexers on one server are isolated | ✅ a batch to `alpha` leaves `beta` with no owner, zero transfers and an unmoved cursor |
| 7 | Changesets | ✅ two |

## Scope fences

- **NOT a change to the wire envelope** — honoured, and verified structurally rather than by eye: no wire/envelope module is touched at all, and `WireBatch` carries no indexer field. The name lives beside the URL in `HttpIngestionOptions` and nowhere in the payload, with the rationale recorded (tenancy in the payload would turn a misdirected batch into a payload error rather than a routing one).
- **NOT the multi-generation registry** — exactly one live wire context per entry; the entry object is the seam that keeps the widening additive.
- **NOT authentication** — the guard and token are unchanged. The only auth-adjacent edits are a log-message rewording and the deliberate choice to answer `401` before `404`, so the registry cannot be probed for which names exist without the token. That is asserted, and is a tightening rather than a change to the scheme.

## Regression pin verified

Criterion 2 is the one a plausible-looking diff can miss while satisfying everything else, so I confirmed it is a real pin. In a throwaway clone I re-added an unnamespaced `POST /ingest` beside the new namespaced route — the shape the task explicitly warns about. Result:

```
FAIL  the unnamespaced routes > no longer answer, so the old surface is GONE rather than left live beside the new one
AssertionError: expected 200 to be 404
```

Baseline is 25 passed; the injected defect fails exactly that test and nothing else. The deletion is pinned, not merely performed.

Two details worth calling out as good: the registry lookup is an own-property read, so a name like `constructor` resolves to nothing rather than to something off `Object.prototype`; and the entry deliberately does NOT carry the name, since a second copy of a discriminator is something a write path can key on wrongly.

Merging.

PR #62, opened 2026-09-05
feat(the-emission-stream-table-is-created-with-every-column-it-needs):…

wighawag:
Task: `work/tasks/done/the-emission-stream-table-is-created-with-every-column-it-needs.md`

All acceptance criteria are met and the full gate (`format:check`, `check:adr`, `build`, `typecheck`, `test`) is green.

## What I did

**Checked the task against reality first.** The premises hold: `streamDigestOf` exists (`packages/core/src/stream/identity.ts`, from `a-stream-is-identified-by-the-digest-of-its-filter`), the indexer name is a route segment plus a registry entry (done, ADR-0036), ADR-0034's whole-source wire hash is still a 32-bit `simple_hash` per entry, and the fixed schema is still `packages/server/src/schema/sql/db.sql`. Nothing had drifted, so I built it.

**The table** (`packages/server/src/schema/sql/db.sql`, `SCHEMA_VERSION` 1 → 2): `EmissionStream`, append-only, PK `(indexer, stream, seq)`, with `removed`, `alive`, the block/transaction coordinates, `address`, `topic0..topic3`, `data`. Two indexes and only two: the partial `(indexer, stream, blockNumber, logIndex) WHERE alive = 1` for the canonical view, and `(indexer, stream, address, topic0, blockNumber)` for the later log API. No processor column, no generation column, no `chainId`.

**Core** (`@etherfold/core`): `LogIngestion` gains `streamDigest` (the wide digest, computed in `StreamBuilder`'s constructor from the *resolved* config), `IngestionOutcome` gains `emissions`, and `EmittedLog` is exported as the ABI-free raw-log shape a storing host needs.

**Server**: `packages/server/src/emissions.ts` (`appendEmissions`) and the call from `/{indexer}/ingest`, which now hands the route segment and `ingestion.streamDigest` in as the two discriminators.

**Tests**: `packages/server/test/emissionStream.test.ts`, 14 cases written before the production code, one per acceptance criterion. I mutation-checked them (breaking the `alive` flagging and the `indexer` binding fails 6 of 14), so they bite.

Also updated `CONTEXT.md`'s stream-builder entry and the server README, both of which claimed the emission stream was not stored; and dropped `work/notes/observations/an-entity-named-like-a-fixed-table-collides-silently.md` for an out-of-scope hazard I noticed (an entity declaration may name `Meta`/`EmissionStream`; `IF NOT EXISTS` makes the collision silent).

## Decisions

**Where the write lives: the ROUTE, not the fold.** ADR-0050 moved the reorg count the other way (into `StreamBuilder.receive`, injected by the store's owner) on the ground that a fact about the fold must not be a fact about the transport, and the emission stream is even more clearly a fact about the fold. I still put it on the route, because the constraint here is the KEY rather than the fact: half of it is the INDEXER NAME, and the route segment is the only place that value exists. `IndexerRegistryEntry` deliberately refuses to carry a name ("a second copy an entry could disagree with is a discriminator a write path might key on wrongly"), and `--indexer` is refused outright on `run`, `build` and `serve`, so there is no fold-side value to key on: injecting a writer into `StreamBuilder` would first require telling a receiver its own name, which is the duplication the registry rejects. Alternative considered and rejected: an `EmissionRecorder` seam mirroring `ReorgRecorder`. **What it touches:** a combined `etherfold run` / `build` folds through `createDirectIngestion`, reaches no route, and therefore stores no emission stream, so the two feed tasks that follow are servable only from `index`. That is not this module declining; that shape has no name to store one under. When `run` gets one, the write moves into `receive` and the route becomes a caller. Recorded in `emissions.ts`, `ingest.ts`, `CONTEXT.md` and the README rather than left to be re-derived.

**`seq` is allocated per `(indexer, stream)`, from the table, once per batch.** `SELECT COALESCE(MAX(seq), 0)` for the pair, then the batch is numbered from it and inserted in one `db.batch`. Per PAIR and not per stream alone (which is what the task's wording says literally) because two named indexers with byte-identical sources land on ONE digest and would otherwise punch holes in each other's cursor space; holes are legal by contract, but they should come from compaction, not from a neighbour. Read from the table rather than kept in a process counter, on the same argument that makes `StreamBuilder` re-read its cursor every call: the intended host is serverless and an in-memory sequence is one isolate's opinion of a value the database owns. It is not a lock; the PK `(indexer, stream, seq)` is, so a genuine concurrent append fails loudly instead of overwriting, and it cannot ordinarily happen because a second batch for one receiver is a `409` before it reaches here. Alternative considered: `MAX(seq)+1` as a SQL subquery per row, rejected as harder to read for no gain given the statements run in one batch anyway. **Touches** the next task's cursor, which addresses this `seq`.

**How `alive` is maintained on a retraction.** Two writes: the retraction row is APPENDED with `removed = 1` and `alive = 0` (a retraction is never something a canonical read may return), and an `UPDATE ... SET alive = 0` flags the emission it takes back, matched on `(indexer, stream, blockHash, logIndex) AND removed = 0 AND alive = 1`. Matched by block HASH rather than height, because a reorg retracts a named block and a height names whichever branch won; `alive = 1` is part of the match so a hash applied, retracted and applied again flags the row that is live at the time rather than the one already dead. Nothing is ever deleted and nothing but `alive` is ever updated. **Touches** the canonical-view task, whose read is exactly `WHERE alive AND blockNumber <= gate`.

**The composite index leads with the two discriminators**: `(indexer, stream, address, topic0, blockNumber)`, not the literal three columns `node-log-api` names. The spec decided the *shape* against D1's index footprint, and it did not have the tenancy discriminators in view; a range scan that could omit one would cross into another tenant's rows, which is the exact hazard the name column exists to prevent. Still ONE index, and `topic1..topic3` remain unindexed. **Touches** `node-log-api` when it is built: its query must bind both discriminators, which it has to anyway.

**Only the raw log is stored, never the decoded `args`/`eventName`.** `args` is what SOME ABI made of those bytes and is re-derived on replay against the source running now (`LogEventFetcher.reparse`), so persisting it would persist an opinion that a decode-only change invalidates, and it is the one thing that would break the "a decode-only change does not orphan the history" property at the row level rather than at the key. Consistent with `the-stream-stores-only-what-the-node-said`. **Touches** the feed task, which serves these rows.

**An append failure RAISES; it is not swallowed the way a reorg count is.** A lost count costs an operational number; a lost append costs the stream itself, and a stream silently missing a batch the state already applied is a HOLE, which this repo treats as the serious damage class precisely because it is invisible and self-consistent. So it becomes a `500` with `lastError` set. The sender's recovery is unaffected: the batch WAS applied, so its next attempt is the ordinary `409` and nothing is double-applied. **Touches** `etherfold index --no-auto-setup` against a database nobody migrated: ingest now fails loudly there instead of folding while storing nothing. That is consistent with `/status` already reporting such a database unhealthy.

**`streamDigest` and `emissions` are REQUIRED members, not optional.** An optional field on the fold's output is a hole with a polite name (the repo's own argument for making `getCodeFingerprint` required): a receiver that quietly omitted one would leave a host storing nothing under a key it could not form. The cost is that two hand-written fakes in tests had to grow, which is why `@etherfold/platform-nodejs` carries a patch changeset for a behaviour-neutral change.

**Naming.** Table `EmissionStream` and column `stream`, taken from the glossary's own terms rather than invented; the column holds the stream's identity, which the glossary already defines as the digest. `EMISSION_STREAM_TABLE` is spelled once so the coming reads cannot drift from this write.

**Left alone deliberately: ADR-0006's `status: accepted, not yet implemented`.** The storage half is now built and the two views and compaction are not, and `ADR-FORMAT.md` enumerates the legal status values with no "partly implemented" among them (it also cites ADR-0006 by name as the worked example of the pending case). Inventing a value would be a convention change outside this fence; the last of the six tasks should flip it.


wighawag:

This task had two fences that a helpful builder violates by accident, so both were checked against the DDL directly rather than against the prose.

## The two fences

- **NO generation column, and nothing carrying the processor.** Honoured. The DDL is `(indexer, stream, seq, removed, alive, blockNumber, blockHash, logIndex, transactionHash, transactionIndex, blockTimestamp, address, topic0..topic3, data)` — nothing processor-derived, no generation. The schema comment states the reason (a processor-only change is a new generation over the SAME stream; a processor column would fork the whole history, which is exactly the case the generation model promises is free). `chainId` is also correctly *absent*, since it is already inside `stream` via the block-0 skeleton entry.
- **The stream column is `streamDigestOf`, not the wire context's source hash.** Honoured, and pinned (below).

Also good and not required: decoded `args` are deliberately not persisted, because they are what some ABI made of the bytes and are re-derived on replay — persisting them would persist an opinion a decode-only change invalidates.

## Acceptance criteria

| # | criterion | verdict |
| --- | --- | --- |
| 1 | A batch on `/{indexer}/ingest` appends emissions, retractions included, both discriminators on every row | ✅ plus a case that a REFUSED batch writes nothing, so a cursor correction leaves no row behind |
| 2 | Superseded rows FLAGGED, never deleted, asserted on the rows | ✅ the retracted row is kept, its retraction appended, both flagged dead |
| 3 | Stream column holds `streamDigestOf`, asserted to differ from the wire context's hashes | ✅ asserted against `context.config` *and* every per-entry `hash`/`streamHash`/`legacyHash` |
| 4 | A decode-only source change leaves the digest UNCHANGED | ✅ and the test first asserts the wire identity **did** move, "which is what makes this a real case" — so it cannot pass vacuously. Seq continues `[1, 2]` over one stream rather than a second history starting |
| 5 | Two identically-sourced named indexers do not share rows | ✅ |
| 6 | `address` + `topic0..3` populated, composite index present, `topic1..3` unindexed | ✅ one index, with a dedicated test asserting there are not five |
| 7 | Table in the FIXED schema, both paths produce the same database | ✅ asserted, including that the statements come from `db.sql` and are not dynamic DDL |
| 8 | Changesets | ✅ two |
| 9 | Gate green | ✅ CI `verify` SUCCESS |

## One deliberate refinement, which I judge correct

The index landed as `(indexer, stream, address, topic0, blockNumber)` rather than the literal `(address, topic0, blockNumber)` the task names. This is honouring `node-log-api`'s decision in the presence of the two discriminators this task introduces, not re-deriving it: the column order, the single-index shape and the unindexed `topic1..3` are all as decided, and the discriminators must lead or a range scan would cross into another tenant's rows. The DDL comment says exactly that. Omitting them would have been the bug.

## Regression pin verified

Criterion 3 is the fence, so I confirmed it pins. In a throwaway clone I replaced `streamDigestOf(source, streamConfig)` with the wire context's `config` hash — the precise mistake the task warns about — and rebuilt core. Result: **5 tests red**, including

```
AssertionError: expected 'h1k67d73' to be 'b63b2d2373e0e3e74fb7c149dbfdb06e'
```

which shows the contrast the criterion exists for: the 32-bit wire change-detector versus the wide digest. Baseline is 14 passed. The decode-only and isolation tests fail with it, as they should.

## Observation filed, not fixed

The agent captured `work/notes/observations/an-entity-named-like-a-fixed-table-collides-silently.md`: a processor declaring an entity named `Meta` or `EmissionStream` gets `CREATE TABLE IF NOT EXISTS` against the server's fixed table, succeeding silently and surfacing later as a column error. Pre-existing with `Meta`; this task widens it by one name. Correctly left unfixed — it is a cross-package decision about where fixed names are declared and which layer refuses them, outside this fence. Carried to the drive report as task-worthy.

Merging.

PR #63, opened 2026-09-05
feat(a-consumer-follows-the-seq-ordered-stream-from-a-validated-opaque-…

wighawag:
Task: `work/tasks/done/a-consumer-follows-the-seq-ordered-stream-from-a-validated-opaque-cursor.md`

Full gate green (`pnpm build && pnpm typecheck && pnpm test`), plus `format:check` and `check:adr`. Working tree contains only this task's changes.

## What I built

**Drift check first.** The task's premises hold: ADR-0006 still decides the two views, `seq` holes and the cursor semantics; ADR-0027 still makes the sync cursor opaque for the reason cited; `the-emission-stream-table-is-created-with-every-column-it-needs` landed the table with `(indexer, stream, seq)` as its primary key; `a-named-indexer-is-a-route-segment-and-a-registry-entry` landed `/{indexer}/…` and `ServerOptions.getIndexer`. Nothing needed re-scoping.

New files:

- `packages/server/src/feed/cursor.ts` — the opaque cursor codec. Envelope `{view, indexer, stream, at}`; `at` is the view's own position (`{seq}` here) that the codec does not interpret, so the canonical view extends rather than re-mints. `openFeedCursor` returns one of four refusal kinds.
- `packages/server/src/feed/stream.ts` — the `seq`-ordered read: `seq > position ORDER BY seq LIMIT n+1` under both discriminators, returning the `seq` **actually served** as the next position. No `seq` and no `alive` on an entry.
- `packages/server/src/api/feed.ts` — `GET /{indexer}/feed`.
- `packages/server/src/api/resolve.ts` — the `501`/`404` name resolution, moved out of `ingest.ts` so both surfaces answer identically.
- `packages/server/test/feed.test.ts` — 23 tests, written before the production code: paging identity at every page size, retraction delivery, opacity (JSON, base64 and body-text probes plus the server decoder as the control), tamper refusal, foreign-indexer refusal with tenant isolation asserted, stale-stream refusal asserted on the body and then followed via `startCursor`, view-mismatch, holes (a hole at the head, a run of holes, a hole before the tail, with page size smaller than the widest hole), page-size rules, and the anonymous read beside the `401` ingest.

Also: `README.md` for the server (routes table plus a "The feed" section), a `CONTEXT.md` glossary entry for **the two views** and **feed cursor**, and a `minor` changeset for `@etherfold/server`.

## Decisions

**The cursor's encoding: a checksummed, per-payload-scrambled base64url framing of a small JSON envelope, and it is honestly labelled as obfuscation.** `fnv1a32` over the JSON payload, an xorshift32 keystream seeded from that checksum, the checksum prefixed in clear, the whole thing base64url with no padding. It is deterministic (the same position always yields the same cursor) and self-checking (an edited cursor fails the checksum and becomes `400 invalid-cursor` instead of decoding to some other position). Alternatives considered: **plain base64url of JSON**, rejected because a client can read it in one line and the acceptance criterion is precisely that it cannot, which is not pedantry (a readable number is what gets incremented, and holes make incrementing wrong); **an HMAC**, rejected because there is no key to hand (the only secret here is `INGEST_TOKEN`, which is the fetcher's write credential and must not become a read-path dependency), and because tamper-*proofing* is not the property needed; **server-stored cursors**, rejected as state about a consumer, which ADR-0005 says etherfold does not keep. The JSDoc says plainly that this is not encryption and not a signature: a determined client that reimplements the file can read one. What it stops is *accidental* dependence on the format, which is the failure that actually happens. **Touches** the canonical-view task, which shares this codec.

**The page-size rule: `limit` defaults to 100, is capped at 1000, and a larger one is REFUSED (`400 invalid-limit`, carrying `maxLimit`) rather than clamped.** A clamp is accept-and-ignore, and it is the dangerous kind: the caller's only check is the count it got back, and a clamped page is indistinguishable from "that is all there was". Non-integers, zero and negatives are refused for the same reason. Alternative considered: silent clamping (the common REST convention), rejected against the repo's standing "refuse rather than accept-and-ignore" stance. **Touches** the canonical view, which should use the same bound and the same refusal.

**The cursor carries a VIEW discriminator, which the task did not name.** The task said "name, stream, position". I added `view` because the two views count in different spaces (`seq` here, `(blockNumber, logIndex)` there) and the next task shares this codec, so without it a canonical cursor presented at `/feed` would either be read as a `seq` (the plausible-wrong-answer class this repo refuses everywhere) or collapse into a vague "invalid cursor". It is `400 view-mismatch`, and it makes the canonical view's cursor an *added field* rather than a re-mint. The word `view` is ADR-0006's and the spec's own ("two views over it"), not a new concept. **Touches** the canonical-view task directly.

**The stream-mismatch refusal answers `{stream, startCursor}`, and the field is named `startCursor` rather than `cursor`.** `stream` is the current stream digest (readable, comparable); `startCursor` is an opaque cursor at the position the current stream's feed begins at. Expressing the position as a *cursor* is forced: positions are otherwise not a vocabulary the consumer has. It is deliberately not called `cursor`, which in a success response means "present this next" and would invite a client to auto-follow the refusal; the task is explicit that re-subscribing is deliberate, and that this is **not** a rewind (asserted: the body's keys are exactly `error, message, startCursor, stream, success`, and it carries no fork block). **Touches** the canonical view, whose rewind response is a different shape for a different reason and should not reuse this name.

**All four cursor refusals are `400`, never `409`.** On the ingest side `409` is the ONE resumable refusal (re-send from `expectedFromBlock`, ADR-0004). None of these is resumable by re-presenting the same cursor, and re-meaning `409` on the read side would make a sender's and a consumer's contract say different things with one number. **Touches** the ingest contract by leaving it alone.

**`indexer-mismatch` names the indexer the caller ADDRESSED and never the one the cursor was minted at** (asserted: the refusal body contains no trace of the other tenant's name). Echoing it would both confirm that the encoding carries a name and hand one tenant's name to a caller poking at another.

**The feed is a PUBLIC read and is not behind `INGEST_TOKEN`.** That token is the fetcher's deployment secret guarding the routes that can WRITE; putting the feed behind it would mean issuing every consumer the credential that moves the cursor. A consumer is a third party built outside etherfold (ADR-0005). Alternative considered: a second read token, rejected as a new configuration surface and a new refusal for a milestone whose whole query surface is otherwise anonymous (`/status`). A deployment that needs a private feed puts it behind its own edge. **Touches** any later authorisation work, and `CONTEXT.md`'s "the QUERY LAYER is deferred" note.

**The feed requires `getIndexer`, so `etherfold serve` (the read tier) does not serve it today.** Validating a cursor's stream means knowing which stream is served, and only the registered receiver knows (`LogIngestion.streamDigest`). The table cannot answer it: one indexer's rows may span several streams over its life, nothing in them says which is current, and choosing by heuristic (for example "the only one present") is exactly the plausible wrong answer the discriminators exist to prevent. So the feed reuses the ingest resolver and a host with no registry gets `501`. I kept the existing error code `ingestion-not-configured` for that case rather than forking a second code for one condition, and generalised its message to "this server hosts no named indexer" (the code is pinned by tests in server, cli and platforms; the message is not). **Touches** `serve`, which refuses `--indexer` and therefore has no name to register: giving the read tier a feed is a separate task, and the honest options are a served-stream config or the canonical-pointer row the generation spec owns. Alternative considered: deriving the current stream from the table, rejected above.

**The response advertises `stream` on the success path too, not only on the refusal.** Symmetry, and a consumer that re-subscribed after a `stream-mismatch` can confirm where it landed. Noting the adjacent-concept hazard explicitly, in the code and the README: the stream identity says WHICH LOGS (`{source, config}`); it is **not** the generation identity (`{source, config, processor}`) that `every-feed-response-advertises-the-generation-it-answered-from` adds, which says WHICH FOLD. A consumer comparing `stream` across polls will not see a processor change. **Touches** that sibling task, which adds a second identity field beside this one.

**No `seq` and no `alive` on an entry, and no `seq` anywhere in the response** (asserted on the body text with the cursor removed, and on the exact key sets). Publishing the position is how a consumer ends up incrementing it, which holes make wrong; `alive` is the other view's derived flag and a retraction-aware consumer learns the same fact from the retraction arriving. **Touches** any later consumer-facing documentation.

**The cursor codec is NOT exported from the package index** (only the `FeedEntry` type is). Publishing a decoder would make the encoding a contract by the back door, which is what opacity exists to prevent; tests import the module path directly, as they already do for `schema.js` and `api/status.js`.

**`ADR-0006`'s `status: accepted, not yet implemented` is left alone**, on the previous task's recorded reasoning: `ADR-FORMAT.md` enumerates the legal statuses with no partly-implemented value, and the last of the six tasks should flip it.


wighawag:

The load-bearing task of this set: three siblings build on this cursor, so the single-codec property and the hole tolerance were checked hardest.

## Acceptance criteria

| # | criterion | verdict |
| --- | --- | --- |
| 1 | Ordered page, present the cursor, resume exactly, nothing repeated or skipped | ✅ asserted across pages *of every size*, plus a caught-up cursor answering an empty page rather than a refusal |
| 2 | Retractions DELIVERED in `seq` order beside what they retract | ✅ apply, retraction and replacement all arrive |
| 3 | Cursor is opaque; a client cannot read a position out of it; nothing documents the encoding | ✅ and it cannot pass vacuously: asserted against the raw string, JSON and base64, *and* the server's own decoder is shown to read `{indexer, stream, at: {seq: 3}}` out of it. The response-shape test pins the exact key set and asserts `seq` appears nowhere outside the cursor |
| 4 | A cursor from indexer A presented at B is REFUSED, not re-interpreted | ✅ plus the other indexer is served none of A's stream |
| 5 | A stale-stream cursor is REFUSED, carrying the current stream identity and its feed's start position | ✅ `400 stream-mismatch` with `stream` + `startCursor`, and a follow-up test actually *re-subscribes* from it and gets the whole new stream — so the refusal is proven actionable, not just well-shaped |
| 6 | Holes in `seq` tolerated, nothing skipped, no stall | ✅ pinned (below) |
| 7 | Changeset | ✅ |

## Scope fences

- **NOT the canonical gated view** — honoured and asserted: a dedicated test that this view *does not* filter on `alive`, "which is the OTHER view's rule", and `stream.ts` consults neither `alive` nor a block gate.
- **NOT a rewind response** — asserted structurally: the refusal body's key set is pinned and the text is checked to contain neither `rewind` nor `forkBlock`, on the stated ground that a filter change produced logs that were never on the old stream.
- **NOT a generation check** — no generation is consulted; that stays task 6's.

## ONE cursor codec, which the siblings depend on

`feed/cursor.ts` is a single encoder whose `at` is a view-opaque position the codec never interprets, with an explicit `view` discriminator (`STREAM_FEED_VIEW = 'stream'`). The header states the rule outright: "the canonical view adds its block hash by putting one in there rather than by minting a second encoding. Two encoders would be two refusal paths that drift." It already refuses a cursor minted for the `canonical` view with `view-mismatch`, so task 5 inherits the seam rather than needing to invent one. This is the property `the-canonical-view-is-gated-and-rewinds-on-a-reorg` will be checked against.

Worth noting the codec is honest about what it is: a checksummed, scrambled base64url framing, explicitly NOT encryption and NOT a signature. It stops accidental format dependence (a client that base64-decodes, finds a number and starts incrementing), and the checksum makes an edited cursor a refusal rather than a plausible answer. That framing is correct and is documented rather than implied.

## Regression pin verified

Criterion 6 is what makes compaction safe to enable later, so I confirmed it pins. In a throwaway clone I replaced the next position (`the seq of the last row actually served`) with `after + page.length` — the exact "derive by incrementing" bug the task warns about. Result: **2 tests red**, with the pathology visible:

```
AssertionError: expected [ 102, 102, 106, 106, 106, 106, 108 ] to deeply equal [ 102, 106, 108 ]
```

entries repeating as the cursor limps across holes. Baseline is 23 passed. The holes test is deliberately shaped to catch this: a hole at the start, a run in the middle, one immediately before the end, and a page size smaller than the widest hole.

Merging.

PR #64, opened 2026-09-05
feat(the-canonical-view-is-gated-and-rewinds-on-a-reorg): The canonical…

wighawag:
Task: `work/tasks/done/the-canonical-view-is-gated-and-rewinds-on-a-reorg.md`

All green: `pnpm build && pnpm typecheck && pnpm test` (server: 127 tests, whole monorepo clean), `pnpm format:check` and `pnpm check:adr` pass.

## Drift check (done first)

No drift. ADR-0006 still specifies exactly this view (`WHERE alive AND blockNumber <= gate` ordered by `(blockNumber, logIndex)`, cursor carrying the block hash, "rewind to fork block F", and the contiguous-suffix proof). ADR-0015 still names this feed's cursor validation as the read-side twin of its own rule. The blocking task landed as described: `/{indexer}/feed`, one opaque codec carrying `{view, indexer, stream, at}`, and the `indexer-mismatch` / `view-mismatch` / `stream-mismatch` / `invalid-cursor` refusal family. `EmissionStream` already has `alive` and the partial index `EmissionStreamCanonical` the view needs. Nothing to route to needs-attention.

## What I built

- **`packages/server/src/feed/canonical.ts`** — the view: `readCanonicalView` (`alive = 1 AND blockNumber <= gate`, strictly after `(blockNumber, logIndex)`, `limit + 1` for an honest `hasMore`, riding the existing partial index), `isStillCanonical` (one lookup), `forkBlockSince` (the fork block).
- **`packages/server/src/api/feed.ts`** — `GET /{indexer}/canonical` beside the seq feed, sharing the limit parser, the cursor codec and one refusal mapper; plus the `409 rewind-required` answer.
- **`packages/server/src/feed/entries.ts`** — the row shape and mapper both views share; `CanonicalEntry` (no verdict) and `FeedEntry = CanonicalEntry & {removed}`.
- **`packages/server/src/feed/cursor.ts`** — `CANONICAL_FEED_VIEW` only; the codec itself is untouched.
- **`packages/server/src/emissions.ts`** — `readStreamHighWaterMark` extracted so the append and the view read the mark through one expression.
- **`packages/server/test/canonical.test.ts`** (25 tests) + **`test/utils/feedHarness.ts`** (extracted from `feed.test.ts`, which now imports it). Test-first at the seam; I verified the three rewind tests are load-bearing by breaking `isStillCanonical` and `forkBlockSince` in turn (3 failures each time) before restoring.
- README, `CONTEXT.md` glossary (canonical view now BUILT, rewind vs stream-mismatch distinguished), and a changeset.
- One observation captured, out of scope: `work/notes/observations/a-reorg-drops-new-logs-below-the-lowest-block-we-held.md`.

## Decisions

**The route is `GET /{indexer}/canonical`, a sibling path rather than `?view=canonical` on `/feed`.** The two views take different parameters (a `gate`, here and nowhere else), answer different refusals (a REWIND here and never there) and hand back different entry shapes, so one path per contract keeps each surface's parameters and refusals true of the whole path. Alternatives: `?view=` on `/feed`, rejected because half the query string would be meaningful only for one value of it, and the `501`/`404`/`limit` answers would have to be documented conditionally; `/{indexer}/feed/canonical`, rejected because nesting implies a sub-resource of the seq stream, and it is a sibling view over the same table, not a slice of that one. The glossary's own words are "the retraction-aware feed" and "the canonical view", so the segment matches the concept. **Touches** `every-feed-response-advertises-the-generation-it-answered-from`, which must add its field to two routes.

**The rewind is `409 rewind-required`, carrying `{stream, forkBlock, rewindCursor, message}`, and `409` is deliberate.** ADR-0004 already makes `409` the ONE resumable refusal in this system ("your position is not where mine is, carry on from here"), and a rewind is that same sentence spoken to a consumer, so using the same number for the same meaning is what keeps a sender's and a consumer's contract legible together. The previous task decided all four *cursor mismatch* refusals are `400` precisely because none of them is resumable; this one is, so it is the exception and not a contradiction — those four are unchanged. Alternatives: `400` beside the mismatches, rejected because it files a correction the caller is MEANT to follow among ones it must not; `200` with an empty page and a rewind field, rejected outright because a consumer that ignores an unknown field reads it as "caught up", which is the silent skip in a new costume. `rewindCursor` is named to say "present this next", deliberately unlike the stream mismatch's `startCursor`, which the previous task named to *discourage* auto-following; here auto-following is the correct behaviour, because this view's whole promise is that the consumer implements no reorg handling. **Touches** `createHttpIngestion`'s reading of status codes only by staying on the read side (that client never calls these routes), and any future consumer SDK.

**The fork block is the LOWEST block the stream has retracted anything at SINCE the cursor was minted — and to make that answerable the canonical cursor carries a `since` mark beside its position and hash.** The mark is the stream's `seq` high-water mark when the cursor was handed out. Why a mark is needed at all: only `seq` records TIME in that table, and the fork is exactly "what changed after you last looked"; the cursor's own block cannot answer it, because a reorg can fork BELOW the cursor and answering the cursor's block is the silent skip this validation exists to prevent. Taking the MINIMUM over several retractions is correct rather than approximate — a second, deeper reorg must move the answer DOWN — and there is a test for it. Alternatives: **look the mark up from the cursor's own row** instead of carrying it, rejected because pair-compaction (the sibling ready task) reclaims a dead row together with its retraction, which is exactly the row that lookup depends on; **walking back over the window**, unnecessary by ADR-0006's contiguous-suffix proof. **The coherence risk, named**: ADR-0006 says "a synthetic sequence is wrong for the canonical view", and it is — as an ORDER. `since` is not a position, never advances the read and never appears outside the opaque envelope; the JSDoc at the field says so in those words. **Touches** `pair-compaction-is-off-by-default`, whose "answer-preserving for the canonical view" property this design deliberately survives (`isStillCanonical` asks "is it alive", so a compacted pair and a flagged-dead row give the same answer).

**`gate` is REQUIRED, and absent or malformed is a new `400 invalid-gate`.** "Do not default it" means the only honest treatment of an absent gate is a refusal: serving unbounded would be defaulting to the tip, which is the specific thing forbidden. Every candidate default is wrong for somebody and none of them says so. ONE error code covers absent and malformed, with the message distinguishing them, following `invalid-limit`'s precedent of one code per parameter. **Touches** any consumer-facing documentation and ADR-0007's two lanes (a `safe` consumer passes a low gate, a `fast` one a high one).

**A canonical entry carries no `removed` field at all** (new exported `CanonicalEntry`, beside `FeedEntry`). A flag that is `false` on every entry a view can ever serve invites `if (entry.removed)` handling that can never fire, which is exactly the reorg handling this view exists to remove from a consumer. Alternative: `removed: false` on every entry, for one entry type across both views; rejected on that ground. **Touches** the package's public types.

**Three refactors inside the fence, each to avoid a second copy of something the acceptance criteria require to be single.** (1) The row shape and mapper moved to `feed/entries.ts`, so `FeedEntry` is now exported from there rather than `feed/stream.js` (internal path only; the package export name is unchanged). (2) `refuse` is parameterised by the served view and its start position instead of hard-coding the seq view — one refusal mapper for both, which is the same argument as one codec. (3) The test fixture (`deploy` / `post` / `batchOf` / `transfer`) moved to `test/utils/feedHarness.ts`, so both suites drive reorgs through the same real ingest path rather than each defining its own idea of one. `feed.test.ts`'s 23 tests are unchanged and still pass.

**`ADR-0006`'s `status: accepted, not yet implemented` is left alone**, on the previous task's recorded reasoning: `ADR-FORMAT.md` enumerates no partly-implemented value, and the last of the six tasks should flip it. Two remain.


wighawag:

## Acceptance criteria

| # | criterion | verdict |
| --- | --- | --- |
| 1 | Only `alive` entries, ordered by `(blockNumber, logIndex)`, never a retraction | ✅ and the ordering test explicitly asserts it is not merely the stored row order |
| 2 | The caller's gate bounds it; raising it serves the entry | ✅ plus the gate is REQUIRED, a non-whole gate is refused, and a gate below every entry answers an empty page rather than a refusal |
| 3 | One advancing position, landing where a from-scratch read of the same gate lands | ✅ asserted at every page size, and a caught-up cursor keeps being answered and re-issued |
| 4 | **A non-canonical cursor answers rewind-to-fork-block F, via a REAL reorg through ingest** | ✅ driven through the actual `/{indexer}/ingest` route, never a row edit. `409 rewind-required`, `forkBlock: 103` — below the cursor at 105 — with the key set pinned so a rewind cannot be missed by ignoring an unknown field, and no page returned alongside it |
| 5 | Events not yet received are delivered after the rewind | ✅ the replacement branch from the fork inclusive, including block 106 "which a naive resume from (105, 0) would have been the ONLY thing to deliver"; nothing below the fork is re-delivered |
| 6 | ONE cursor codec, asserted; name/stream refusals still behave | ✅ pinned structurally (below), and the name/stream/page-size refusals are re-asserted on this view |
| 7 | Changeset | ✅ |

Two cases beyond the criteria that I would have asked for: a consumer whose own block *survived* the reorg is correctly NOT rewound (and raising its gate afterwards serves the new branch, never the old), and a later reorg reaching further back moves the fork DOWN.

## Scope fences

- **NOT a second cursor codec** — honoured, and this is the criterion I checked hardest since it is the defect the single-encoder rule exists to prevent. The test scans the package's sources and pins `encode*Cursor`, `decode*Cursor` **and** the base64 alphabet (what a hand-rolled rival would need) to `src/feed/cursor.ts` alone, with a `files.length > 4` guard against passing on an empty scan. Functionally confirmed too: the canonical cursor is opened by the seq feed's decoder, and each view refuses the other's cursor with `view-mismatch`.
- **NOT retraction delivery** — asserted: the table is shown to hold the retractions and dead originals while this view serves none of them.
- **NOT a server-chosen gate** — the gate is required rather than defaulted to tip or finality; the server does not pick the consumer's risk appetite.

## Regression pin verified

The fork block is the whole guarantee, so I confirmed it pins. In a throwaway clone I changed `MIN(blockNumber)` to `MAX(blockNumber)` over the retracted rows — reporting the cursor's own block instead of the lowest changed one, which is exactly the silent skip. Result: **3 tests red**, and the consequence is visible rather than abstract:

```
AssertionError: expected 105 to be 103
AssertionError: expected [ [ 106, '0xb106' ] ] to deeply equal [ [ 103, '0xb103' ], …(2) ]
```

The consumer receives only block 106 and never the new 103 and 104. Baseline is 25 passed (48 across both feed suites).

I also checked that the `feed.test.ts` reduction (−163 lines) is a move into `test/utils/feedHarness.ts` and not lost coverage: the seq feed still runs 23 tests, exactly as before.

## Observation filed, not fixed

`work/notes/observations/a-reorg-drops-new-logs-below-the-lowest-block-we-held.md`: `generateStreamToAppend` sets `startingBlockForNewEvent = reorgBlock.number` from `unconfirmedBlocks`, which holds only event-bearing blocks — so if the chain forked below the lowest block we held logs for, a log the new branch carries in that gap is dropped by the filter and never re-fetched. Correctly left unfixed: it is an engine-side fetch/derive property, and this view's rewind already answers the best the stored stream can support. Carried to the drive report as task-worthy, since it is a potential correctness gap rather than a nit.

Merging.

PR #65, opened 2026-09-05
feat(every-feed-response-advertises-the-generation-it-answered-from): E…

wighawag:
Task: `work/tasks/done/every-feed-response-advertises-the-generation-it-answered-from.md`

All gates green. Here is what I did.

## What I built

`generation` is now on every response both feed views give, page and refusal alike.

**`@etherfold/core`**
- `packages/core/src/generation/identity.ts` (new): `generationDigestOf(GenerationId)` renders the existing `{stream, processor}` identity as one 128-bit lowercase-hex digest, under a versioned rule tag, using the same `sha256`-truncated rendering as `streamDigestOf`. Exported from the package index.
- `LogIngestion` grows `readonly generation: GenerationId`, implemented on `StreamBuilder` as a **getter** so the processor half is read live (`getVersionHash()` covers a processor's config, which `configure()` can move after construction). Its `stream` half is `streamDigest` itself, so the two cannot disagree.

**`@etherfold/server`**
- `GET /{indexer}/feed` and `GET /{indexer}/canonical` both compute `generationDigestOf(entry.ingestion.generation)` next to the stream digest and put `generation` on: the 200 page, all four cursor refusals (via the shared `refuse` mapper), `invalid-limit`, `invalid-gate`, and the `409 rewind-required`.
- README updated (both response samples plus the advertise-don't-dictate paragraph); `CONTEXT.md`'s **generation** glossary entry now records "two fields inside, one opaque value on the way out".

**Tests** (failing first, at the seam): `packages/server/test/generation.test.ts` (17) covers both views advertising, one value across both, refusals and the rewind carrying it, stability across polls/pages/empty pages/new logs, the same-fold-two-names property, movement on a stream change, the paired processor-change assertion (value moves *while* cursors stay valid and delivered logs stay identical), opacity (source scan that nothing splits/slices it, nothing reads it back in, the cursor codec never mentions it, and it leaks neither part), and no generation/processor column on `EmissionStream` plus byte-identical rows across a processor change. `packages/core/test/generationIdentity.test.ts` (7) covers the digest itself. `streamBuilder.test.ts` gains two tests for the live getter.

Test-harness tidy-up inside the fence, to avoid third and fourth copies of things the harness exists to define once: `readFeed`/`readCanonical` and the `indexedThroughBlock105`/`reorgAt103` reorg fixture moved into `test/utils/feedHarness.ts` (the two existing suites now import them; their 127 tests are unchanged and pass), and `deploy` takes an optional `processorVersion` so a redeploy over the same database is a new fold over an already-stored stream.

**Drift check (done first):** premises hold. ADR-0006 still keys the stream on `{source, config}` and the state on `{source, config, processor}`; the code narrows the stream key to the wide digest per ADR-0034/0035; both views exist as the two done tasks built them. No stop needed. `pnpm build && pnpm typecheck && pnpm test` green, `format:check` and `check:adr` clean, working tree only the intended files.

## Decisions

**The advertised value is composed from `GenerationId` (`{streamDigest, processorVersionHash}`), not from ADR-0006's `{source, config, processor}` wire context.** The task's snapshot named the wire context and said the composition "can later be replaced by the stream-digest-plus-processor-hash identity"; that replacement already exists and is the repo's word for this concept (`GenerationId`, `CONTEXT.md`'s **generation**: "identified by its STREAM plus the processor's `version` hash"). Building the older composition and swapping it later would have meant keying an identity on the wire context's 32-bit `simple_hash` source halves, which ADR-0034 rules out as a key. Alternative considered: build `{source, config, processor}` literally as written and swap later, rejected as paying twice for a value the criteria declare opaque. **Touches** nothing outside this task by construction (opacity), but it settles which concept the word `generation` names on the wire.

**It is rendered as an OPAQUE digest, not as a structured `{stream, processor}` object nor a delimited string.** A consumer handed two named fields will read one of them, and a consumer handed `a-b` will split it; either makes the composition a contract by the back door, which is the exact failure `feed/cursor.ts` argues against for the cursor and the thing "replaceable later" needs to survive. Alternatives: the structured object (more diagnosable, rejected because it makes criterion 4 a promise the shape invites breaking) and a concatenation (same objection, no upside over the object). Cost, accepted: an operator cannot read the processor version out of the value; the response still carries `stream` beside it, and the server knows the preimage. **Touches** any future runtime that advertises a generation (browser/CLI) — the rendering lives in core so there is one, and the glossary now says so.

**Every response *that resolved a named indexer* carries it; `501 ingestion-not-configured` and `404 unknown-indexer` do not.** Those two come from the shared `resolveIndexer` before any entry exists, so there is no generation that answered — a host with no registry, or a name it was not built with. Alternative: put a placeholder on them, rejected as inventing an identity for an answer no fold gave. The refusals that *do* carry it (four cursor mismatches, `invalid-limit`, `invalid-gate`, `409 rewind-required`) do so because a refusal is still an answer this fold gave. **Touches** `a-named-indexer-is-a-route-segment-and-a-registry-entry`'s shared resolver (unchanged by me) and any consumer error handling.

**`LogIngestion` gains `generation: GenerationId` rather than a bare `processorVersionHash`, and it is derived per read.** The composition rule then lives once in core instead of the server assembling an identity from two fields; the `stream` half is derived from `streamDigest` so it is not a second copy that can disagree. Derived rather than snapshotted because `getVersionHash()` covers a processor's configuration and `configure()` can move it after construction, so a captured field could advertise a fold that is no longer running. This is a required member on a published interface: `StreamBuilder` supplies it, and the two in-repo test fakes (`core/test/directIngestion.test.ts`, `platforms/nodejs/test/serve.test.ts`) were updated. **Touches** every implementer of `LogIngestion` and `the-server-and-cli-hold-generations-too`, where one entry holds several generations — this reads as "which generation this receiver is", which stays true when an entry holds several receivers.

**`streamIdentity.test.ts`'s "is the ONLY digest of its kind" now allows `generation/identity.ts`, rather than extracting a shared hashing helper.** That test guards against a *second implementation of the stream digest* (two that must agree byte-for-byte would address one stream twice); a digest of a different value, built on top of the stream digest, is not that. Alternative considered and rejected: factor the `sha256(stringToHex(canonical_form(x))).slice(2, 34)` rendering into `utils/hash.ts` so there is literally one hashing site — it would have touched the derivation of a persisted key for cosmetic gain, and would have weakened the deliberate in-file assertions on `stream/identity.ts`. The new file is asserted to keep the same discipline (sync, no `crypto.subtle`, no `simple_hash`) in `generationIdentity.test.ts`. **Touches** `a-stream-is-identified-by-the-digest-of-its-filter`'s invariant, which is narrowed in wording, not in force.

**No ADR, and ADR-0006's `status: accepted, not yet implemented` is left alone.** The composition choice fails the ADR gate's "hard to reverse" leg on purpose — opacity plus a test that nothing depends on the composition is what makes it reversible — so the JSDoc, the changeset and this block carry the rationale. The status line stays for the reason the previous task recorded: `ADR-FORMAT.md` enumerates no partly-implemented value and the last of the six tasks should flip it; `pair-compaction-is-off-by-default` remains.


wighawag:

## Acceptance criteria

| # | criterion | verdict |
| --- | --- | --- |
| 1 | Every response from BOTH views carries the generation as a plain readable field | ✅ and beyond the letter of it: also on the refusals and on a rewind, "so a consumer logging one has the whole context" |
| 2 | Stable across polls while nothing changes | ✅ across polls, pages and an empty page at the tip; explicitly does not move when new logs arrive, since "more data is not a different fold" |
| 3 | A PROCESSOR change moves it while cursors stay valid and logs stay identical — asserted TOGETHER | ✅ in one test, as the criterion demands: the value moves, the stream digest and entries are unchanged, and the cursors minted by the old generation are honoured and caught up rather than refused or replaying. A second test delivers the not-yet-reached logs across the change |
| 4 | Opaque: nothing parses it or depends on its composition | ✅ source-scanned for decomposition (`split/slice/substring/startsWith/indexOf/match/replace`) and for it ever being read back IN, with an empty-scan guard; plus asserted not to be, or contain, the stream digest or the processor version, and not to be JSON |
| 5 | No generation column on the log table, asserted | ✅ via `PRAGMA table_info`, matching `/generation\|processor\|version\|fold/i` against a column list guarded as non-trivial, *and* rows asserted byte-identical across a processor change |
| 6 | Changeset | ✅ |
| 7 | Gate green | ✅ CI `verify` SUCCESS |

Two further cases worth crediting: one stream and one fold under two different tenant names advertises ONE value ("says WHICH FOLD and not which tenant"), and the value moves when the stream moves, so a re-subscribing consumer is not told nothing changed.

## Scope fences

- **NOT a generation column** — honoured and asserted twice over (schema introspection, and byte-identical rows). This is the fence that keeps a processor change free, and the test says so.
- **NOT cursor validation** — asserted directly: the generation appears nowhere in the decoded cursor, and the source scan pins that `src/feed/cursor.ts` never mentions it. A generation change therefore cannot invalidate a cursor, which is the required behaviour.
- **NOT the multi-generation model** — one generation per named indexer; the field's opacity is what lets the composition be swapped for the stream-digest-plus-processor identity later without touching a consumer.
- **ADVERTISES, does not DICTATE** — no rule about consumer behaviour was added.

## Regression pin verified

Criterion 3 is the substance, so I confirmed it pins. In a throwaway clone I dropped the processor half from the digest preimage, leaving the value a function of the stream alone — the composition mistake that would make the field useless for the one case it exists for (same logs, different fold). Result: **2 tests red**:

```
AssertionError: expected '0d867bdd0323121773f2abce02896a01' not to be '0d867bdd0323121773f2abce02896a01'
```

the advertised value failing to move across the processor change. Baseline is 17 passed. The digest's own docstring already anticipates this: the processor hash alone would make two indexers folding different streams look identical, and the stream alone misses the fold change — both halves are required, and the test enforces it.

Merging.

PR #66, opened 2026-09-05
feat(pair-compaction-is-off-by-default): Pair-compaction reclaims a ret…

wighawag:
Task: `work/tasks/done/pair-compaction-is-off-by-default.md`

Done. Everything green: `pnpm format:check && pnpm check:adr && pnpm build && pnpm typecheck && pnpm test` (server: 168 tests, 24 of them new).

The task's premises still hold: ADR-0006 (stored emission stream, two views, `seq` holes legal, pair-compaction a config that is off by default), ADR-0019 (block numbers only, finality as the floor, refused not clamped), ADR-0022 (a host-scheduled call deleting an explicit bounded set), and the three landed tasks it builds on (`the-emission-stream-table-is-created-with-every-column-it-needs`, `a-consumer-follows-the-seq-ordered-stream-from-a-validated-opaque-cursor`, `the-canonical-view-is-gated-and-rewinds-on-a-reorg`). The code already anticipated this task in three places (`emissions.ts`, `feed/stream.ts`, `feed/canonical.ts` on `isStillCanonical`), and nothing contradicted it. No STOP.

## What I built

- **`packages/server/src/compaction.ts`** (new): `resolvePairCompaction(setting, {finality})` and `compactEmissionPairs(db, {indexer, stream, compaction?, finality, latestBlock, maxPairs?})`, plus `COMPACTION_OFF`, `DEFAULT_MAX_PAIRS` and the five types. A pair is one dead application (`removed = 0, alive = 0`) and one retraction (`removed = 1`) of the same `(blockNumber, blockHash, logIndex)`; both `seq` values go into one `DELETE ... seq IN (...)` inside one `batch()`. Live rows are never candidates, `seq` is never renumbered, unmatched rows are left alone.
- **`packages/server/src/index.ts`**: exports the verb and the resolver, wired to no route and no timer.
- **`packages/server/test/compaction.test.ts`** (new, written before the module): the depth's unit and floor, off-by-default, pairs-go-together-and-only-below-the-depth, no half pairs (an orphan retraction inserted directly is left alone), no live row dropped, bounded work and the budgeted loop, the canonical view **byte-identical** before/after at seven gates, the rewind path still naming the right fork block after a compaction, and a `seq` consumer following the real holes at a page size smaller than the hole. I mutated the floor and the pairing to confirm the tests bite (6 failures).
- **`packages/server/test/utils/feedHarness.ts`**: `followFeed` moved here from `feed.test.ts`, so the two suites that assert "a consumer reaches the end" share one loop.
- **`packages/server/src/feed/canonical.ts`**: doc-only note on `forkBlockSince`, saying where compaction is observable and to whom.
- **`CONTEXT.md`** glossary entry for **pair-compaction** (and how it differs from `prune`), **`packages/server/README.md`** section, and a changeset for `@etherfold/server` (minor).

## Decisions

**Where the configuration lives: an argument to the verb, resolved by an exported `resolvePairCompaction`, and NOT a field on `IndexerRegistryEntry` or `ServerOptions`.** A field on the registry entry would be config this package reads nowhere, since nothing in the server calls the verb (nothing may: the cadence is the host's, ADR-0022). So the deployment holds the setting where it holds its other deployment config, validates it at startup with `resolvePairCompaction` (so a bad depth is a boot failure, not a 3am surprise from a scheduler), and passes it to `compactEmissionPairs`. The verb resolves again internally, so the finality floor cannot be bypassed by handing it a window nobody checked. Alternatives considered: a `compaction` field on `IndexerRegistryEntry` (rejected: dead config, and it would need `finality` on `LogIngestion` to enforce the floor, widening a core interface for a feature that is off); an admin route (rejected: a new user-visible surface and a second place off-by-default could be got wrong). **Touches:** `IndexerRegistryEntry` stays as it was; no CLI flag was added either (see below).

**No `--compaction` CLI flag.** `--retention` exists on the CLI because retention bounds READS from the moment it is set even if nothing prunes; a compaction depth changes nothing at all until a host calls the verb, so a flag would be a knob that does nothing on every command the CLI has today. Alternative: add the flag now and leave it unwired (rejected: an operator setting it would reasonably believe compaction was happening). **Touches:** `packages/cli` config, if a scheduled maintenance verb is ever added there.

**The bound on one call's work: `maxPairs`, defaulting to 490, with the scan bounded at `maxPairs * 2` rows.** One call reads at most `maxPairs * 2` candidate rows and deletes at most `maxPairs` pairs, named row by row (never `DELETE ... WHERE`, per ADR-0022, since a hosted backend caps rows written per request and `remote-sql` reports no affected-row count). `PAIRS_PER_STATEMENT = 49` makes a delete statement carry exactly 100 bound parameters (2 discriminators + 2×49 `seq`s), which is D1's cap, and 490 is ten such statements: one SELECT plus ten DELETEs fits the FREE D1 plan's 50 queries per invocation with room to spare. `complete` reports whether the scan reached the end, so an amortised policy and a whole sweep are both expressible without the store inventing a cadence. Alternative considered: prune's shape (unset budget = loop until done), rejected because a compaction call that can grow unboundedly is exactly what a serverless host cannot schedule. **Touches:** a future `d1CompactionBudget`-style helper in `platforms/cf-worker` if a Worker ever schedules this.

**A depth exactly AT the finality depth is legal, and compaction is strictly BELOW `latestBlock - depth`.** The criterion says a depth that would compact *at or above* `latestBlock - finality` is refused. Making the deletion strictly below the floor means `depth === finality` touches nothing at or above `latestBlock - finality`, so the floor is a legal value rather than the first illegal one, which is exactly how `resolveRetention` treats `blocks < finalityDepth`. One rule, spelled the same way in both places. Alternative: refuse `depth <= finality`, rejected because it would make retention and compaction disagree about what "the floor" means by one block.

**The candidate scan is ordered by `(blockNumber, logIndex, blockHash, seq)` and not by `seq`, and no index is added for it.** Block order puts the two rows of a pair adjacent, so pairing is local to the page and a budget can never cut between an entry and its retraction; `seq` order (the primary key, the cheaper scan) would leave a small budget reading a page of entries whose retractions all sit just past the end, making no progress. Neither order rides an index anyway, because the only relevant index is PARTIAL on `alive = 1`, so a scan over dead rows is a table scan either way. I deliberately did not add a third index to `EmissionStream`: it would be paid for by every deployment, on a table already weighed against D1's 10GB ceiling, to serve a feature that is off by default. **Touches:** the schema (unchanged, still version 2) and `node-log-api`'s index budget, if compaction is ever scheduled hot enough to want one.

**`latestBlock` is supplied by the host, not read from the table.** The highest block the table carries a log at is the highest *event-bearing* block (median 429 blocks behind on the real measured stream), so measuring the depth from it would silently mean something other than what an operator configured. A non-integer or negative tip is refused rather than compacted against.

**Recorded, not decided: pair-compaction is observable on the canonical view's fork search too.** `forkBlockSince` finds a fork by looking for retraction rows above a cursor's mark, so a reorg whose pair has been reclaimed can no longer be named. That needs a consumer holding a cursor minted before a reorg and presenting it only after that reorg fell below the configured depth, i.e. lagging further behind than finality, which is the same observer ADR-0006 already accepts and is why the finality depth is the floor. Documented at both sites and asserted the other way round (a consumer inside the window still gets its rewind after a compaction). No ADR: ADR-0006, ADR-0019 and ADR-0022 already carry the substance, and these are follow-through rather than new decisions.


wighawag:

## Acceptance criteria

| # | criterion | verdict |
| --- | --- | --- |
| 1 | Entry and retraction removed TOGETHER, only below the depth, asserted on the rows | ✅ both rows of the deep pairs go, neither row of the shallow one |
| 2 | OFF by default, asserted | ✅ three ways: configuring nothing, saying `off`, and a deployment that configured nothing compacting nothing |
| 3 | A depth at or above `latestBlock - finality` REFUSED, naming the floor | ✅ refused rather than clamped, floor named, and a depth exactly AT the floor is accepted |
| 4 | A unit other than block numbers refused | ✅ including a duration refused for the stated reason (it prunes on the wrong clock) |
| 5 | Appending never compacts | ✅ every row survives four batches and two reorgs |
| 6 | One call performs BOUNDED work | ✅ stops at its budget and reports `complete: false`; repeated bounded calls reach the same end as one unbudgeted call |
| 7 | **Canonical view byte-identical before and after** | ✅ pinned (below), and stronger than asked: byte-identical *at every gate*, including gates inside the compacted range, cursor included, with `rowsDeleted > 0` guarding against a vacuous pass |
| 8 | A `seq` consumer follows the holes, nothing skipped, no stall | ✅ at a page size smaller than the hole, and without repeating an entry on resume |
| 9 | Changeset | ✅ |

Beyond the criteria: a consumer inside the window still gets its rewind after a compaction, with the fork correctly naming the reorg it actually missed rather than the reclaimed pairs.

## Scope fences

- **NOT half a pair** — `pairsIn` pairs the k-th dead entry with the k-th retraction within a `(blockNumber, logIndex, blockHash)` group and leaves any remainder alone, with a test that a retraction with no entry of its own is untouched.
- **NOT on by default / NOT automatic** — no cadence invented; a host-scheduled verb, per ADR-0022.
- **NOT inside the finality window** — the floor is enforced at the API.
- **NOT a renumbering of `seq`** — asserted: surviving rows keep the numbers they had, leaving the holes task 4's cursor already tolerates.

## Regression pin verified

Criterion 7 is the property the whole design rests on, and the task explicitly says to prove it in a test rather than assert it in a comment. So I checked it two ways in a throwaway clone.

First I removed the `alive = 0` filter from candidate selection. The canonical tests stayed green — because the pairing logic *independently* prevents a live row from being reclaimed (a live row has no matching retraction in its group, so no pair forms). That is defense in depth rather than a gap, and it is worth noting: the safety property has two independent guards.

So I injected a defect that genuinely reaches a live row: deleting each pair by `(blockNumber, logIndex)` coordinates instead of by `seq`, which also takes the live replacement row sitting at those coordinates. Result: **9 tests red**, including both of the ones that matter here —

```
× serves a byte-identical response over the same gate
× serves a byte-identical response at every gate, including gates inside the compacted range
```

along with `never drops a LIVE row, however far below the depth it is` and `never renumbers seq`. Baseline is 24 passed. The answer-preserving property is genuinely pinned.

Merging. This completes the `indexer-server-feed` task set.

PR #67, opened 2026-09-05
feat(a-refetched-block-is-new-unless-the-window-already-holds-it): A re…

wighawag:
Task: `work/tasks/done/a-refetched-block-is-new-unless-the-window-already-holds-it.md`

Done. Summary of the work:

**Drift check first.** The task's premises all still hold: `generateStreamToAppend` still computed `startingBlockForNewEvent` exactly as described, ADR-0004's absence/contradiction split and ADR-0050's count-in-`receive` placement are intact, and `work/tasks/done/` holds nothing that supersedes them. Nothing routed to needs-attention.

**Test first, at the agreed seam.** Added `describe('a re-fetched block is new unless the window already holds it')` to `packages/core/test/utils.test.ts`: the task's reproduction verbatim (fork at 195, window holding only 200, gap log at 196), plus a no-reorg gap case, the once-only cases, a two-cycle case proving the gap block is delivered once and never again, the retracted-then-re-offered case, and the window-ordering case. Four failed on the unmodified engine; the two once-only ones passed, pinning the behaviour the fix had to keep.

**The fix** (`packages/core/src/internal/engine/utils.ts`): the scalar height is gone. The retained window is `lastUnconfirmedBlocks.slice(0, reorgedBlockIndex)` (what survived this cycle's retraction), and an incoming block is delivered unless that set holds it by `(number, hash)` via `windowKeyOf`. The carry-forward loop reads the same retained prefix. Reorg detection, retraction generation, the finality prune and the re-fetched range are untouched. The rebuilt `unconfirmedBlocks` is now sorted ascending, because membership removes the ordering that a height threshold gave for free. Doc comments updated where they described the removed mechanism (`generateStreamFromReplay`'s cross-reference, `assertAscendingByBlock`'s rationale).

**End to end**: `packages/server/test/reorgBelowTheWindow.test.ts` drives the same fork through `/{indexer}/ingest` (so the stream-builder derives the reorg, the route writes rows and the entity fold runs), and asserts the gap log lands in the stored `EmissionStream`, on the retraction-aware feed, and on the canonical view — and that re-offering both blocks on the next cycle appends nothing.

**Standing conventions**: changeset for `@etherfold/core` + `@etherfold/server`, ADR-0051, one-clause `CONTEXT.md` glossary update on `reorg / removed / finality`. `pnpm format:check`, `check:adr`, `build`, `typecheck` and the full `test` run are green, with no existing test edited (including the CLI equivalence suite and the stratagems conformance workload). Working tree carries only the intended changes.

## Decisions

- **The no-reorg branches of the old scalar were replaced too, and one of them CHANGED BEHAVIOUR deliberately.** The scalar served three cases. (1) Empty window (`logEventsGroupedPerBlock[0].number`, or `0`): unchanged — an empty window holds nothing, so everything incoming is new, exactly as before. (2) Non-empty window, no reorg (`top + 1`): **changed** — a block inside the re-fetched range that the window does not hold is now delivered even when it sits below the window's top and nothing reorged. It is the same silent loss wearing different clothes (by the finality invariant we can never have applied it, or it would be in the window), so admitting it is the point rather than a side effect. Covered by `delivers a block the window never held even when NOTHING reorged`. Alternatives considered: keep the scalar for the no-reorg branches and only change the reorg branch (rejected: two rules for one question, and it leaves the same loss reachable without a fork), or lower the threshold to `newLastFromBlock` (rejected: it re-delivers the whole window every cycle, since the range always reaches back over it). Touches: every caller of `generateStreamToAppend` — `IndexerGeneration.feed`/`indexMore` and `StreamBuilder.receive` — and therefore the stored emission stream and both feed views.
- **The rebuilt `unconfirmedBlocks` is now SORTED ascending.** Under a height threshold nothing below a retained block could be delivered, so the window came out ordered for free; membership removes that, and the next cycle's reorg walk (which stops at the first contradiction) and `cursorSyncedThrough` (which cuts the window as a prefix) both read it in block order. `generateStreamFromReplay` already sorts, with the same comment. Alternative considered: insert delivered blocks in order instead of sorting (rejected: more code for a list bounded by the finality window, and the sort is what the replay path already does, so the two paths read alike). Touches: anything reading `LastSync.unconfirmedBlocks` — `checkTxInclusion`, `cursorSyncedThrough`, the replay walk.
- **A delivered gap block can sit BELOW a block the processor already applied, and that is safe rather than tolerated.** Any block above it that contributed state carried events, so it is in the window, so it was retracted in the same stream, so `forkPoint` reverts below it before the gap block is applied; a block with no events changed no state. This is why the fix needs no change to `applyEventStream` or to any store. Recorded because a reader meeting an apply at 103 after a revert to 104 would reasonably suspect an ordering bug.
- **Recorded as ADR-0051** (`docs/adr/0051-a-refetched-block-is-new-unless-the-window-already-holds-it.md`): hard to reverse (it decides what reaches the persisted emission stream), surprising without context (a fetch delivering below the fork point), and a real trade-off (widening the range to hunt the true fork point was the rejected alternative, explicitly out of scope here). ADR-0042 names `startingBlockForNewEvent`, which no longer exists; ADR-0051 states that it is what that rule became rather than editing the older record.


wighawag:

The data-loss fix. Both the delivery property and the once-only property it replaces were pinned by injecting defects, and one second-order consequence the task did not anticipate was found and handled by the build.

## Acceptance criteria

| # | criterion | verdict |
| --- | --- | --- |
| 1 | The reproduction passes: a fork below the lowest held block delivers the gap logs | ✅ `delivers the new branch logs BELOW the lowest block we held logs for` |
| 2 | Asserted END TO END through a receiver, into the stored stream and both views | ✅ `packages/server/test/reorgBelowTheWindow.test.ts`: the gap log reaches the emission stream beside its retraction, the `seq` feed in stream order, and the canonical view in block order |
| 3 | Nothing delivered TWICE | ✅ three ways: a block the window already holds is applied once however often re-offered; the gap block is not delivered a second time once the window holds it; and end to end, applied once across repeated re-offers |
| 4 | Reorg behaviour unchanged, existing suites green WITHOUT being edited to fit | ✅ and this is verifiable at a glance: `utils.test.ts` is **+230 −0**. No existing assertion was touched |
| 5 | Reorg counters still count once per shape (ADR-0050) | ✅ untouched; equivalence suite green |
| 6 | Deliberate no-reorg change named and tested | ✅ `delivers a block the window never held even when NOTHING reorged`, called out in the report rather than slipped in |
| 7 | Changeset | ✅ |

## Scope fences

All honoured: reorg detection and the absence/contradiction split untouched, retraction generation untouched, finality pruning untouched, and the re-fetched range was **not** widened. The fix delivers logs that were already in the range, which was the constraint.

## Two things the build got right that I would have missed

- **`retainedBlocks`, not the window we arrived with.** Membership is tested against `lastUnconfirmedBlocks.slice(0, reorgedBlockIndex)`, the window that *survived* this cycle's retraction. This is load-bearing in the opposite direction from the main fix: a reorg concluded at the first window block retracts every later one, and a re-fetch still carrying one of them must re-apply it under the same hash. A retracted block has left the window, so its re-offer is new again.
- **The window's ascending order stopped being free.** Under the height threshold, nothing below a retained block could ever be delivered, so `newUnconfirmedBlocks` came out sorted by construction. Membership removes that, since a block the window never held can now be delivered from below the lowest retained one. The build spotted this, added an explicit stable sort, and documented why. The next cycle's reorg walk and `cursorSyncedThrough`'s prefix cut both depend on that order, so this would have been a real second bug.

It also corrected the `assertAscending` docstring honestly: under the old rule an out-of-order group was silently DROPPED, and under membership it is merely misordered, so the refusal is now about order of delivery alone. The comment says so rather than leaving a stale justification in place.

## Regression pins verified

**Restoring the height threshold** (the pre-fix rule): 4 core tests red *and all 4 end-to-end tests red*, so the bug is pinned at both the unit seam and through the receiver into the feeds.

**Membership against the whole window** instead of the retained one: exactly one test red, `still re-applies a block that was RETRACTED and then re-offered under the same hash` — the subtle case above, caught by precisely the test written for it.

Baselines: 31 core, 4 end to end.

ADR-0051 records the decision, why membership is complete (the `getFromBlock` finality invariant), and why the retained window is the right one to read.

Merging. This closes the first of the two promoted observations, and the more serious one.

PR #68, opened 2026-09-05
feat(every-fixed-table-lives-in-the-reserved-underscore-namespace): Eve…

wighawag:
Task: `work/tasks/done/every-fixed-table-lives-in-the-reserved-underscore-namespace.md`

Working tree holds only this task's changes.

## What I did

**Drift check first.** Every premise still holds: `packages/state-store-sqlite/src/ddl.ts` still emits `CREATE TABLE IF NOT EXISTS "<entity.name>"`; `isReserved` in `packages/state-store/src/entities.ts` still reserves `_` and `_blocks`/`_cursor` still live there; `buildProcessor` (`packages/cli/src/folding.ts`) still hands ONE handle to both the store and the server; `db.sql` still created `Meta` and `EmissionStream` outside the namespace. ADR-0016's dependency posture (store imports nothing but `remote-sql`/`named-logs`) and ADR-0006's stored stream are unchanged and unviolated. No re-scope needed.

**The rename** (`packages/server/src/schema/sql/db.sql`): `Meta` → `_meta`, `EmissionStream` → `_emissions`, `EmissionStreamCanonical` → `_emissions_canonical`, `EmissionStreamByAddressTopic` → `_emissions_by_address_topic`. Columns, keys, indexes and the version row are byte-for-byte the same otherwise. Call sites followed: `schema.ts`, `reorgs.ts`, `emissions.ts` (`EMISSION_STREAM_TABLE`), `feed/canonical.ts`, `packages/cli/src/reorgCounters.ts`, plus comments in `core/src/reorgCounters.ts`, `cli/src/folding.ts`, `platforms/nodejs/src/index.ts`, the server README and `CONTEXT.md`. No migration, no compatibility read.

**The guarantee** (test written before the rename compiled clean): `packages/server/test/reservedNamespace.test.ts` scans the `.sql` FILE, strips comments, and asserts every `CREATE TABLE`/`CREATE INDEX` name begins with `_`. Its guard is three-part: a floor on how many CREATEs it found, at least one of each kind, and an equality between what it parsed and the number of `CREATE` keywords in the stripped DDL (so a statement the regex cannot read fails the scan rather than being skipped). A negative control proves the matcher can fail, and I verified it live by appending an unprefixed `Sessions` table: the suite went red naming it, then I restored the file.

**The refusal, in the shape the collision lived in**: `packages/cli/test/fixedTableNamespace.test.ts` drives the real `buildProcessor` with the server's fixed schema on the same handle, and asserts `_meta` and `_emissions` are refused with the existing reserved-identifier error, before any DDL is issued (the database's table list is unchanged after the rejection). A third case shows the fixed tables, the store's `_blocks`/`_cursor` and the processor's `nft`/`counter` all in one file, with everything that is not a declared entity inside the namespace.

**`isReserved`'s docstring** now says the prefix means "not a user entity", names both packages that place tables there, and records why parameterising the reserved set was refused.

Changeset at `.changeset/every-fixed-table-lives-in-the-reserved-namespace.md` (`@etherfold/server` minor, `etherfold` minor, `@etherfold/state-store` patch).

`pnpm build && pnpm typecheck && pnpm test` green (server 12 files/176 tests, cli 15/156, cf-worker's D1-migration suite included), `pnpm format:check` and `pnpm check:adr` green.

## Decisions

- **The new names are `_meta` and `_emissions`, exactly as the task decided; the INDEX names are mine.** `_emissions_canonical` and `_emissions_by_address_topic`. The task named the tables but not the two indexes, and SQLite puts tables and indexes in ONE namespace, so an unprefixed index is the same collision. I followed the store's existing convention (`_blocks_timestamp`, and `_<entity>_<suffix>` for derived entity indexes) rather than transliterating the camel-case names, so all fixed names in the database read alike. Alternative considered: `_emissionsCanonical`, keeping the old spelling with a prefix; rejected because it would be the only camel-case name in the reserved namespace. Checked for overlap: an entity named `emissions` derives `_emissions_open|history|lower|upper`, none of which collide with these two. Touches: `packages/server/README.md` and the `feed/canonical.ts` comment, which name the partial index.
- **The schema-file scan parses the DDL with a regex over the comment-stripped file, and proves it parsed everything.** It reads `packages/server/src/schema/sql/db.sql` from disk (not the generated `db.sql.ts`, not a live database), strips `--` comments the way `schemaStatements` does, then matches `CREATE [UNIQUE] (TABLE|INDEX) [IF NOT EXISTS] [quote]<name>`. Comments must go first because one of them says "CREATE TABLE" while describing what is deliberately NOT in the file. The honest weakness of a regex scan is that it passes when it matches nothing, so the guard asserts the parsed count EQUALS the number of `CREATE` keywords in the stripped DDL: a future statement spelled in a way the matcher cannot read fails the test instead of silently escaping the rule. Alternative considered: scanning `schemaStatements` or `sqlite_master` after `applySchema`; rejected because the file is what BOTH application paths consume (wrangler runs the file and calls nothing of ours), and the criterion asks for a file scan. Touches: nothing outside the server package.
- **`SCHEMA_VERSION` stays at 2 rather than bumping to 3.** The version row lives in the table that was renamed, so a database migrated by an older build has no `_meta` at all and `readSchemaState` reports `applied: false` — a stronger and more accurate signal than a number mismatch, and the one the task calls correct. No database can hold a `_meta` row this build did not write, so "2 in `_meta`" is unambiguous. Alternative considered: bumping to 3 on a literal reading of the constant's docstring ("bumped whenever db.sql changes in a way an existing database has to be told about"); rejected because no existing database can be told anything through a row it cannot find, and a bump would imply a version-2 `_meta` exists somewhere. I recorded the reasoning at the constant. Touches: `packages/server/test/server.test.ts`, which asserts the SQL row and the constant agree (still does), and `/status`, which reports both.
- **The reserved-identifier ERROR MESSAGE changed with the docstring.** It said `Names starting with "_" belong to the store.`; it now says `... are not user entities.` The criterion only asked for the docstring, but leaving the message asserting the thing the docstring now corrects would re-mean the same term one line apart. No test or doc asserts the old string (checked). It is user-visible text on a refusal, hence recorded rather than silent. Touches: `@etherfold/state-store` consumers who match on the message; none in this repo.


wighawag:

Option A as decided: bring the server's fixed tables into the namespace the store already reserves, then pin the convention with a schema scan so it becomes a guarantee rather than a habit.

## Acceptance criteria

| # | criterion | verdict |
| --- | --- | --- |
| 1 | An entity named like a server fixed table is REFUSED, in the combined shape | ✅ `packages/cli/test/fixedTableNamespace.test.ts`, for both names, through the shape that shares one database handle — and refused at DECLARATION time, so nothing was created before it failed |
| 2 | `Meta` and `EmissionStream` renamed; no reference to the old names in shipped code or SQL | ✅ now `_meta`, `_emissions`, `_emissions_canonical`, `_emissions_by_address_topic`; a grep over `packages/*/src` and `platforms/*/src` returns nothing, and a test asserts the schema file itself carries no trace |
| 3 | Every `db.sql` table and index begins with `_`, with a guard against a vacuous scan | ✅ pinned (below), with a *stronger* guard than asked for |
| 4 | Counters, schema-version report and both feed views behave as before; existing suites green | ✅ CI green across server, feed, canonical, compaction, cli and equivalence |
| 5 | D1 migration path and Node schema path still agree | ✅ the previous task's assertion still holds |
| 6 | `isReserved` docstring says "not a user entity" | ✅ docstring and the user-facing error message both, and it names both packages' tables |
| 7 | Changeset | ✅ |

## Scope fences

- **NOT a change to what the tables contain** — a pure rename. `_emissions` keeps every column and both indexes, and still carries no generation or processor column.
- **NOT a new dependency** — and this is the point of choosing A. `packages/state-store/src/entities.ts` is **+21 −2, all of it docstring and one error-message rewording**. `isReserved` is byte-for-byte the same logic. The store learned nothing about the server.
- **NOT a widening of entity legality** — `IDENTIFIER`, the case-folding rule and the keyword posture are untouched.
- **NOT a migration** — none written, no compatibility read of the old names.

The docstring also records why the parameterised reserved set was refused, so the option is not silently re-proposed later.

## The guard is better than the criterion asked for

I asked for a scan with a guard against passing on an empty read. The build wrote three:

- at least 4 `CREATE`s found;
- at least one TABLE and at least one INDEX among them;
- **the parsed count equals the raw `CREATE` occurrence count in the file**, so a statement the parser fails to understand fails the test instead of being silently skipped.

Plus a meta-test, `would notice a fixed table that forgot the prefix`, which runs the scanner over synthetic DDL containing `Sessions`/`SessionsById` and asserts it flags them. That validates the detector itself, which is the thing a scan-based guarantee actually rests on.

## Regression pins verified

**A future fixed table forgetting the prefix** (appended `CREATE TABLE IF NOT EXISTS Sessions` to `db.sql`):

```
× creates every table and index inside the reserved `_` namespace
AssertionError: expected [ { kind: 'TABLE', name: 'Sessions' } ] to deeply equal []
```

So the convention really is enforced at the gate now, which was the whole second half of the task.

**The namespace guard itself disabled** (`isReserved` returning `false`, reintroducing the original collision): 3 of the 4 cli tests red, including both names and the declaration-time refusal. The one that stays green is the premise test that the shared database holds fixed tables and entity tables together, which is correctly independent of the guard.

Baselines: 4 and 4.

Merging. This closes the second promoted observation.

PR #69, opened 2026-09-05
feat(every-deployment-shape-stores-the-stream-it-folded): Every deploym…

wighawag:
Task: `work/tasks/done/every-deployment-shape-stores-the-stream-it-folded.md`

All gates green (`pnpm build && pnpm typecheck && pnpm test`, plus `format:check` and `check:adr`), working tree clean.

## What I did

**Checked the task against reality first.** Every premise held: `appendEmissions` had exactly one call site (the ingest route), `run`/`build` folded through `createDirectIngestion` into databases with an empty `_emissions` table, ADR-0050 is as described (and its `build`-indistinguishability consequence was contradicted), ADR-0036's never-defaulted rule is a routing rule, ADR-0048 allowed exactly one default (the port), and `CONTEXT.md`'s **hole** entry already prescribes "write the stream BEFORE processing and do not process a batch that was not saved". No drift, so I built it.

**The seam (test first).** `packages/core/test/streamBuilder.test.ts` gained a `describe` pinning the port and its ordering before any production code existed (5 of 7 red). Then:

- **`@etherfold/core`**: new `EmissionAppender` / `EmissionWrite` (`src/emissionStream.ts`) and `StreamBuilderOptions.appendEmissions`. `receive` appends **before** `processor.process`, with no `try/catch`; empty batches write nothing; the port is not hashed into the wire identity.
- **`@etherfold/server`**: `emissionAppenderFor(db, indexer)` beside `appendEmissions`; the ingest route now performs **no durable write at all** and is a caller of `receive`.
- **`etherfold`**: `buildProcessor` builds the appender beside the reorg recorder against the same handle and now requires an `indexer` name; `run`, `build` and `index` inject it. `--indexer` is optional-with-default on `run`/`build`, still required/never-defaulted on `fetch`/`index`, still refused on `serve`.
- **Tests**: `packages/server/test/oneEmissionAppendSite.test.ts` (workspace scan + vacuous-scan guard); `equivalence.test.ts` compares the two shapes' stored streams row-for-row and column-for-column, asserts the split shape's rows are unchanged and not doubled, asserts the `build` artifact carries the same seven rows under the default name, drives a refused append (`--no-auto-setup` on an unmigrated database) proving no advance and no hole and then catch-up after `applySchema`, and serves both feed views over a `run`-produced database.
- **Docs**: ADR-0052, `CONTEXT.md` (stream-builder, command set, indexer glossary), `packages/server/README.md`, `packages/cli/README.md`, `db.sql` comments, the cf-worker host example, and a changeset for `etherfold` / `@etherfold/core` / `@etherfold/server`.

## Decisions

**Where the append sits relative to `processor.process`, and how the failure path guarantees no hole.** The append is the first durable act of `receive`, immediately after the delta is derived and strictly before `processor.process`, and it is *not* wrapped: a failure propagates out of `receive`, so the processor is never called, the store's transaction never opens, the cursor never moves, and the next cycle re-derives the identical delta. Alternative considered and rejected: append after the fold (that is the order the route had, and it is exactly what produces a hole). The accepted cost in the other direction, documented at the call site and in the ADR: if the append lands and the fold then fails, the stream is one batch ahead and the re-derived delta is appended again — legal by ADR-0038 ("ahead or level, never behind"), whereas a hole is not. Touches: `StreamBuilder.noteReorg` stays catch-and-log, so the two ports next to each other now have deliberately different failure semantics.

**A refused append stops the fold, which changes what `--no-auto-setup` does.** Because `_emissions` is a fixed table, a folding command pointed with `--no-auto-setup` at a database nobody has migrated now retries its cycle and reports no progress instead of indexing on. The error carries no `retryable: false`, so the fetcher host classifies it as `retry` and the deployment catches up by itself once the schema arrives. This is user-visible and it invalidated an existing equivalence case ("folds identically when it cannot write the counter at all"), which I replaced with the no-hole/catch-up case; the counter's best-effort guarantee is still pinned at the unit level in `packages/core/test/streamBuilder.test.ts`. Touches: `--no-auto-setup` on `run` and `index`, and the README paragraphs about it.

**The combined-shape default name is the constant `default` (`DEFAULT_INDEXER_NAME`, `packages/cli/src/config.ts`).** Alternatives: derive it from the database file name (rejected — a copied or moved file would silently fork the stored stream), or require the flag (rejected — it breaks `etherfold run`, the readme's headline one-liner, for a discriminator that only matters once two answer sets share a database). `run` and `build` share the one constant so an artifact one built is one the other continues. Touches: ADR-0048 ("only the port may default") is amended by ADR-0052; ADR-0036's never-defaulted rule is *narrowed to routing*, not re-opened — `fetch`/`index` are unchanged and a test covers both halves. It also touches `--help` (the flag is now visible on `run`/`build`) and `RunConfig`/`BuildConfig`, where I put `indexer` as a plain field and deliberately *not* inside a `Wire`.

**The SQL write stays in `@etherfold/server`, so `etherfold` takes it as a real dependency (dev → prod).** This is the one place I diverged from ADR-0050's placement, which put the reorg SQL in the CLI. Reason: `@etherfold/server` owns the emission table's DDL, both of ADR-0006's views, and the `seq` high-water read that the append and the canonical view's cursor must agree about; moving the write out would split one table's contract across two packages and duplicate the row shape. The CLI already starts that server in three of its five commands, and the import is dynamic like the rest of `buildProcessor`. Alternative considered: put the SQL in the CLI and have the server test keep a copy (as `test/utils/hostRecorder.ts` does for counters) — rejected as duplication of the table's own semantics.

**The one-append-site test lives in `packages/server/test/`, not `packages/core/test/`.** Its sibling is in core because the durable *key names* it scans for are in core; the durable name here is `EMISSION_STREAM_TABLE`, which the server owns, so importing it there is what keeps the test and the table from drifting. It still scans the whole workspace and keeps the vacuous-scan guard.

**The port is OPTIONAL on `StreamBuilder`, like `ReorgRecorder`.** A host that supplies none stores no stream and folds identically (asserted). Rejected making it required: it would be a constructor break for every embedder, and the honest statement is that storing a stream is the store owner's capability, not the engine's. Every shipped shape supplies one; the consequence (such a host has no feed to serve) is documented on the type.


wighawag:

The prerequisite for `the-server-and-cli-hold-generations-too`, and a real defect in its own right: `appendEmissions` had exactly one call site, the HTTP ingest route, so `run` and `build` produced databases with an empty `_emissions` table.

## Acceptance criteria

| # | criterion | verdict |
| --- | --- | --- |
| 1 | All three shapes append; rows compared as an EQUIVALENCE | ✅ full row-for-row `expect(storedInOneProcess).toEqual(storedViaWire)` over `SELECT *`, beside the ADR-0050 counter comparison |
| 2 | Asserted on a database `build` EMITTED | ✅ rows read out of the artifact, retractions included and superseded rows flagged |
| 3 | ONE path; a test asserts no second append site | ✅ `oneEmissionAppendSite.test.ts`, same shape as `oneReorgWriteSite.test.ts` including the empty-scan guard, plus explicit assertions that neither the route nor `directIngestion` writes |
| 4 | No double-append; split shape's rows unchanged | ✅ split rows re-pinned as "the same six rows it stored when the route wrote them", and `seq` asserted `[1..7]` with a single indexer value |
| 5 | **A failed append does not advance the state, and leaves no hole** | ✅ pinned (below); asserted at the unit seam and end to end |
| 6 | Retractions, `alive`, monotonic `seq` unchanged on every shape | ✅ |
| 7 | `--indexer` optional on `run`/`build`, required on `fetch`/`index`, both halves tested | ✅ |
| 8 | Both feed views answer over a database a COMBINED process folded | ✅ the feed is no longer a split-shape-only surface |
| 9 | Changeset | ✅ |

## Scope fences

- **NOT a schema change** — honoured, and I checked this specifically because `db.sql` appears in the diff. It is **comments only**: no column, index or key touched. The edit corrects the now-false claim that `indexer` "arrives as the route segment", which had to change.
- **NOT a change to what `_emissions` means, nor to the feed views** — their suites pass untouched.
- **NOT best-effort** — the fence I was most concerned about, and it is honoured emphatically rather than narrowly (below).
- **NOT the generation model** — nothing of it built.
- **NOT a re-opening of ADR-0036's never-defaulted rule for the wire** — the default is scoped to the combined shapes, which have no wire, and the rule is left intact where it does work.

## The part that matters most

The task's one deliberate difference from ADR-0050 was that this write must **not** copy the counter's catch-and-log. The build did more than comply: it made the asymmetry explicit in the class docstring ("They are ORDERED DIFFERENTLY on purpose, and that is the only asymmetry between them"), put the append before `processor.process` with no `try`/`catch`, and — the part I would have accepted its absence of — **named the residual trade honestly**:

> if the APPEND lands and the FOLD then fails, the stream is one batch AHEAD of the state and the re-derived delta is appended again. The stream may be ahead of the state or level with it, never behind (ADR-0038), so that is the survivable side of the trade and a hole is not.

That is the correct reading of `CONTEXT.md`'s hole-versus-gap entry. It also correctly excludes both ports from the wire hash, since hashing them would refuse batches from an identically-configured sender wired to a different database.

## Regression pins verified

**Append made best-effort** (wrapped in `try`/`catch`, the counter's shape): exactly the two no-hole tests go red —
`does NOT process a batch whose append failed` and `re-derives the SAME delta on the next cycle`. Nothing else moves, so the ordering is pinned precisely and narrowly.

**Fold stores nothing** (the pre-fix behaviour): 4 of 5 equivalence tests red, including the `build` artifact test and `serves the retraction-aware feed and the canonical view from a combined fold`.

Baselines: 36 core, 5 equivalence.

ADR-0052 records the seam, why this write is not best-effort unlike the count beside it, and the `--indexer` default.

Merging. This unblocks story 8 of `the-server-and-cli-hold-generations-too`, which rebuilds from the locally stored stream on exactly the shapes that previously stored none.

PR #71, opened 2026-09-05
feat(the-generation-registry-is-durable-on-sql): The generation registr…

wighawag:
Task: `work/tasks/done/the-generation-registry-is-durable-on-sql.md`

## What landed

**The durable substrate** (`packages/server/src/generations.ts`): `generationRegistryPortOnSQL(db, indexer, {dropState?})` supplies the five-operation `GenerationRegistryPort` over `RemoteSQL`, and `openGenerationRegistryOnSQL(db, indexer, {caps, dropState?})` opens `openGenerationRegistry` on it. It supplies rows and re-decides nothing: registration/resolution, the caps that refuse, deletion, reaping, the pointer and the sweep are all inherited from `@etherfold/core`.

**Two fixed tables in the static schema file** (`packages/server/src/schema/sql/db.sql`), both in the reserved `_` namespace: `_generations` (indexer, stream, processor, createdAt; PK on the three) and `_generation_pointer` (one small row per named indexer: the canonical identity, plus the commit guard). `SCHEMA_VERSION` and the version row both move to 3 together.

**`commit` is atomic over `prepare` + `batch`** by a guarded batch carrying a revision token (ADR-0054, below). `listStreamDigests` / `dropStreamSubtree` answer over `_emissions`, scoped to one indexer name, with the row count and the delete in one batch since `remote-sql` reports no affected-row count.

**The writer rule** is now code: `writerOf(records, stream)` plus `GenerationRegistry.writerOf(stream)` in `@etherfold/core` (the rules module, so all three substrates inherit one rule), amended into ADR-0044 and into `CONTEXT.md`'s `follower` entry.

**Tests** (`packages/server/test/generationRegistry.test.ts`, 19 cases; plus 4 new core cases): rows and pointer, indexer scoping, cap refusals naming candidates with nothing evicted, canonical-delete refusal, reap, a second handle on a file database seeing the same generations and the same canonical one, writer succession on the rows in one writing batch, the sweep hitting exactly the unclaimed subtree, the port having exactly five operations, and no caps anywhere in the schema. The concurrency case is asserted by interleaving a rival's commit immediately before the first *writing* batch reaches the database, never by timing; I verified it discriminates by disabling the guard (both concurrency cases go red, everything else stays green) and then restoring.

Gate: `pnpm build`, `pnpm typecheck`, `pnpm test` (all packages, including the cf-worker suite that runs this DDL through D1 migrations), `pnpm check:adr` and `pnpm format:check` are all green. Working tree holds only the intended changes.

## Decisions

**1. How `commit` stays atomic: a GUARDED BATCH carrying a REVISION TOKEN (ADR-0054, written).** A commit reads the records, the pointer and a `revision` token in one batch; every statement it then writes is conditional on that token (`COALESCE((SELECT revision ...), '') = ?`); the last write swaps the token for a fresh unique one; and a final `SELECT` *inside the same batch* reports the token the row now holds. Ours means we won and every guarded statement applied; anything else means our whole batch applied to nothing, and we re-read, re-run the decision function and retry. Alternatives weighed: pushing the decision into SQL (rejected: the decision is not one predicate, it resolves duplicates, applies two caps and refuses naming every deletable generation and stream, so SQL would fork the model this task consumes); a monotonic integer revision (rejected: a loser reads `expected + 1`, which is the winner's value *and* the number the loser meant to write, so a win is indistinguishable from a loss, and `remote-sql` reports no affected-row count); a lock row (rejected: a crashed holder wedges the registry and needs a second timeout mechanism); in-process serialisation (rejected: the target host is serverless, so two isolates are two writers). Touches: any future substrate over this seam, and the port contract, since `plan` must remain a re-runnable pure decision (it already is).

**2. The revision guard lives ON the canonical-pointer row, not a table of its own.** Every commit already writes that row, so the guarantee costs no extra row and no extra statement, and "absent canonical means leave it where it is" falls out of binding `NULL` and `COALESCE`ing. The cost is one column on `_generation_pointer` that is not about the pointer; said out loud in the schema, the module and the ADR. Table names: `_generations` and `_generation_pointer`.

**3. Writer succession: the rule is DERIVED, and only its durable half is built here.** `writerOf` is the oldest surviving registered generation on a stream, computed from the records, so succession is atomic with the delete *because it is stored nowhere*: no writer column, no second write for a crash to land between. That is the whole of what this task's seam can honestly enforce. The ENGINE half named in the task body (swapping a successor's `readOnlyStream` for the real writable `ExistingStream`, and moving the wire context's receiver) is **not built here and cannot be**: on the server tier an `IndexerRegistryEntry` holds exactly one live wire context and there is no generation container above `StreamBuilder`, which are `one-registry-entry-holds-several-live-wire-contexts` and `a-changed-context-creates-a-successor-instead-of-clearing`. I did not fake it with a test that cannot fail (today's receiver is generation-blind, so "a batch still appends after a delete" passes for the wrong reason). The ADR-0044 amendment states the rule, states that succession is atomic because it is stored nowhere, and names those two tasks as the place the engine half lands, so they inherit it rather than deciding one of their own. Touches those two tasks and `the-rebuild-replays-the-local-stream-in-bounded-chunks`, which already cites this rule.

**4. ADR-0046 is deliberately NOT reopened.** Drop-on-promotion still DECLINES to drop a writer another held generation follows: an automatic drop taken inside an advance cycle is a different act from an operator deleting a generation on purpose, and I changed no container behaviour in `@etherfold/core`. The amendment says so explicitly, because "the writer succeeds on a delete" and "the automatic drop declines" otherwise read as two rules in tension.

**5. `writerOf` went into `@etherfold/core`'s registry (the rules), not into the SQL substrate.** The rule is substrate-neutral and is now asserted over memory and SQL; putting it in the server package would fork it across three substrates, which is the duplication the rules-over-a-port shape exists to prevent. It adds one method to the `GenerationRegistry` type (nothing else implements that type).

**6. A new refusal: `GenerationCommitContentionError`, after `MAX_COMMIT_ATTEMPTS = 8` lost guards.** Nothing is written and the caller may retry; the bound exists so sustained contention surfaces as a refusal naming it rather than as an unbounded loop that is indistinguishable from a hang. It is exported from `@etherfold/server`.

**7. `openGenerationRegistryOnSQL` DEFAULTS no caps: `caps` is required.** The browser has `BROWSER_GENERATION_CAPS` (two of each) because a tab keeps the previous generation only until promotion; a server or a CLI should be far more generous and the number is a deployment's statement, so a substrate module picking one would be an invisible user-visible default. Touches `a-changed-context-creates-a-successor-instead-of-clearing`, which the task names as the owner of the cap VALUES on this runtime.

**8. Consequence worth flagging to the next tasks: the first registry open on a PRE-GENERATION database sweeps its stored emission stream.** `listStreamDigests` answers from `_emissions`, so a stream written by today's `run` / `index` (which register no generation) is claimed by nobody and the sweep on open drops it. That is the registry's designed behaviour rather than a new one, and its documented cost is a re-fetch rather than a hole, but it is real for `a-changed-context-creates-a-successor-instead-of-clearing` (which opens the registry before it can create anything) and for `the-rebuild-replays-the-local-stream-in-bounded-chunks`. Documented on `listStreamDigests`; no migration written, per the repo's "a one-time migration between two unreleased formats has no beneficiary" convention.

**9. `CONTEXT.md`'s `follower` entry was updated** so the glossary's writer sentence is the restated rule rather than the superseded half of it.


wighawag:

1. **`GenerationRegistryPort` over `RemoteSQL`, scoped to one named indexer, passing what memory and IndexedDB pass** — met. `generationRegistryPortOnSQL` supplies exactly `read`/`commit`/`listStreamDigests`/`dropStreamSubtree`/`dropState`; register, resolve-instead-of-duplicate, cap refusal naming candidates, delete and reap are each asserted over SQL, and the rules themselves are untouched in `@etherfold/core`.
2. **Deleting the writer hands the duty to the oldest survivor, in the SAME commit** — met at the seam this task owns, and the residue is honestly declared rather than faked (see below).
3. **Dropping the last generation on a stream reaps the subtree** — met (`deletion.reaped === STREAM_B`, emission rows for that stream gone, the other stream and the other indexer's identically-digested rows untouched).
4. **Dropping the canonical generation is refused** — met (`GenerationIsCanonicalError`, row count unchanged).
5. **Caps refuse loudly, evict nothing, name the deletable generations** — met, on both `maxGenerations` and `maxStreams`, asserting `cap` and `candidates` and that no row was evicted.
6. **The pointer survives a fresh handle** — met, and correctly on a FILE-backed database opened twice rather than a shared in-memory handle, which is what "a restarted process" actually means.
7. **A concurrent cap decision cannot be beaten, asserted against the mechanism and never by timing** — met, and this is the strongest part of the diff. `interleaved()` commits a rival's write in the exact window between the decision and the first WRITING batch, deterministically.
8. **Nothing about the caps is persisted; the port still has five operations** — met, and asserted structurally: a regex over the comment-stripped `db.sql`, and `Object.keys(port).sort()`.
9. **The sweep finds an unclaimed subtree and drops exactly it** — met, with the negative half (another indexer's rows under the same digest survive) asserted alongside.
10. **Static schema, `SCHEMA_VERSION` and the version row agree, reserved-namespace test still passes** — met: `SCHEMA_VERSION` 2→3 and `INSERT INTO _meta ... '3'` move together, both tables are `_`-prefixed, and CI is green on the existing assertions.
11. **Tests in the repo's style** — met (vitest, per-package `test/`, 19 new server cases plus 4 in core).

ADR-0054 claims "removing the guard turns that test red". I checked it rather than taking it: in a throwaway clone I neutralised the revision guard (`COALESCE((SELECT revision …), '')` → a tautology, and the pointer upsert's `WHERE … revision = ?5` with it), which is precisely the read-then-write the ADR refuses. Result: **exactly the two concurrency tests go red, and the other 17 stay green.** So the pin is specific and load-bearing, not incidental.

### Scope fences honoured

- **No caps table, no caps column, no sixth port operation** — held, and `openGenerationRegistryOnSQL` defaults no caps at all rather than inventing a server-side number.
- **`dropState` is injected, not implemented** — held; defaults to a no-op, so `a-generation-folds-into-its-own-tables` still owns the table namespace and the two tasks stay in different packages.
- **The rules are not rewritten** — held. The one addition to `@etherfold/core` is `writerOf` plus the registry method that reads it, which is a substrate-neutral rule and belongs there rather than being forked across three substrates.
- **No runtime-computed names in the static schema** — held: `db.sql` gains two fixed `_` tables and nothing derived.

### The one place the criterion over-reaches, and why the honest partial is right

Criterion 2 also asks that "the wire context resolves to it, so an incoming batch is still received and still appends". That half is not buildable at this point in the graph: an `IndexerRegistryEntry` holds exactly ONE live wire context today, and there is no generation container above `StreamBuilder`. Those are `one-registry-entry-holds-several-live-wire-contexts` and `a-changed-context-creates-a-successor-instead-of-clearing`, both DOWNSTREAM of this task. The build declined to fake it with a test that would pass for the wrong reason (today's receiver is generation-blind, so "a batch still appends after a delete" is vacuously true), and instead built the durable half, derived the writer so succession is atomic by construction, and wrote the ADR-0044 amendment naming those two tasks as where the engine half lands. That is the correct call, and because it is recorded in a committed ADR rather than a PR comment, the two successor tasks inherit it. I am tracking it and will check it at their Gate-3.

### Notes carried forward

Decision 8 in the done record is worth flagging loudly: the first registry open on a PRE-generation database sweeps its stored emission stream, because `listStreamDigests` answers from `_emissions` and today's `run`/`index` register no generation. It is documented on `listStreamDigests` and its cost is a re-fetch rather than a hole, but `a-changed-context-creates-a-successor-instead-of-clearing` opens the registry before it can create anything, so it must not trip over this.

Merging.

PR #72, opened 2026-09-05
feat(a-generation-folds-into-its-own-tables): A generation folds into i…

wighawag:
Task: `work/tasks/done/a-generation-folds-into-its-own-tables.md`

All checks green: `pnpm build`, `pnpm typecheck` (0 errors), `pnpm test` (18 packages, all passing, 508 in `state-store-sqlite`), `pnpm format:check`, `pnpm check:adr`.

## What I did

**Drift check first.** ADR-0053 is present and unchanged, `packages/server/src/generations.ts` landed with `dropState` injected and a comment explicitly deferring the per-generation naming to this task, and `buildProcessor` is untouched as instructed. No stale premise found, so I built.

**`@etherfold/state-store-sqlite` gained a table-name namespace.**

- `src/identifiers.ts`: the namespace shape rule (`[A-Za-z0-9]+`), `assertStorableTableNamespace` (refusing an underscore, and `sqlite`), and `inTableNamespace(ns, name)`, which puts the namespace *inside* the reserved `_` prefix: `token` → `genA_token`, `_blocks` → `_genA_blocks`.
- `src/ddl.ts`: `TableNames` + `tableNames(namespace?)` (one value, resolved once, the only place a table or index name is spelled), `FIXED_SCHEMA_DDL` → `fixedSchemaDDL(names)`, and `dropSchemaStatements` (the `DROP` that retiring a generation is).
- `src/statements.ts`: every builder now takes the resolved `TableNames` as a required argument; none of them spells a table name.
- `src/store.ts`: `VersionedStateStoreOptions.tableNamespace` (validated at construction, beside the `sqlite_` entity rule), every read/write routed through the resolved names, and `VersionedStateStore.drop()`.
- README section "Several generations in one database", plus a changeset.

**Tests** (written before the code): `test/namespace.test.ts` covers two stores over one handle (isolated rows, independent cursors under the same `lastSync` key, separate block tables), a revert and a prune in one leaving the other untouched, a drop removing exactly its tables and indexes with a *read* afterwards on the survivor, the byte-for-byte unnamespaced names, and the construction-time refusals. `test/conformance.test.ts` gained a fourth run: the suite inside a namespace, beside another generation already migrated into the same handle, so a statement that forgot the namespace fails loudly. `packages/cli/test/generationNamespaceBesideFixedTables.test.ts` covers the boundary from the server's side: `_meta`, `_emissions`, `_generations` and `_generation_pointer` keep their names and are still found by `readSchemaState`, `appendEmissions` and the registry, with two namespaced generations in the same file, and survive one of them dropping its state.

## Decisions

**The option is `tableNamespace`, not `namespace`.** "Namespace" already means the reserved `_` prefix everywhere in this repo (`CONTEXT.md`'s entity-declaration entry, `every-fixed-table-lives-in-the-reserved-underscore-namespace`), so a bare `namespace` option in the same package would make one word mean two things. `tableNamespace` is ADR-0053's own noun ("a table-name namespace"). Rejected: `generationNamespace`, because the store must not learn what a generation is; the caller holds that identity. Touches: whatever the follow-on wiring task passes at construction.

**The namespace goes INSIDE the reserved `_` prefix (`_<ns>_blocks`), not in front of it (`<ns>__blocks`).** Uniform prefixing is one rule instead of two, but it makes the store's fixed tables stop starting with `_`, which breaks the invariant `every-fixed-table-lives-in-the-reserved-underscore-namespace` established and `packages/cli/test/fixedTableNamespace.test.ts` asserts on the one database a combined deployment shares between this store and `@etherfold/server`. Touches: that test's rule, and anything that identifies a fixed table by its prefix.

**A namespace is `[A-Za-z0-9]+`, and an underscore in one is a NEW refusal at construction.** The underscore is the join separator, so allowing one inside a namespace makes `a_b` + `c` and `a` + `b_c` one table: two generations silently sharing rows, the exact failure this exists to prevent. Banning it makes the join injective by construction rather than by a uniqueness argument a single store cannot check. Leading digits and mixed case ARE admitted, unlike a declaration at the seam, because every name this package emits is either quoted or begins with `_`, and a rendered `generationDigestOf` (32 lowercase hex, possibly digit-leading) must be usable as it comes. Alternatives considered: allowing `_` and demanding uniqueness from the caller (rejected: unverifiable and silent when wrong); demanding a letter-leading identifier (rejected: forces the wiring task to decorate a digest for no engine reason). Touches: the follow-on task that names namespaces from generation identities, which must not use `_`.

**`sqlite` is refused as a namespace.** Its entity tables would be `sqlite_<entity>`, which the engine refuses however quoted. Same rule and same placement as the existing `sqlite_` entity-name refusal, so it fails where it was configured rather than at `migrate()`.

**Two residual hazards left open and documented rather than closed.** (1) SQLite folds identifier case, so `genA` and `GENA` are one namespace here; a store sees only its own name and cannot refuse the collision, and a caller deriving the name from a digest never meets it. (2) An unnamespaced store declaring an entity literally named `genA_token` lands on namespace `genA`'s `token` table; closing that would mean refusing an entity name for what some *other* store might be called, which is the "legality depends on the neighbours" failure the derived index names were prefixed to avoid, and it needs a database mixing namespaced and unnamespaced stores, which the model does not produce.

**`drop()` is a verb on this backend, not on the `StateStore` seam.** The acceptance criterion needs a production path for "dropping one generation's state", and `@etherfold/server`'s registry already expects an injected `dropState` that owns no naming convention. Putting it on the seam would hand a handler a verb that erases the store, and a backend whose whole storage is a keyspace expresses the disposal differently. Touches: `SQLGenerationRegistryOptions.dropState`, which a later task will wire to this.

**`TableNames` is a REQUIRED parameter on every exported DDL/statement function** (breaking, changeset written; nothing outside this package calls them). A default would make forgetting it compile, and forgetting it does not throw: it quietly reads and writes the unnamespaced table beside the generation it belongs to.

**No new ADR.** ADR-0053 already owns this decision and its rejected alternatives; everything above is implementation detail inside it, and nothing is on disk anywhere yet (nothing constructs a namespaced store), so none of it is hard to reverse. The rationale lives in the module notes at each choice site.


wighawag:

1. **Two stores over ONE handle under different namespaces: writes invisible across, cursors independent** — met, and asserted THROUGH THE SEAM rather than by comparing table listings, which is the right instrument: a listing cannot tell a shared row from two rows that look alike. The cursor case is the sharp one and it is done properly, writing under the literal `lastSync` both folds would use, so only the namespace can be what separates them.
2. **A revert in one does not touch the other's blocks or rows** — met, and extended with a `prune` case, which is the other whole-table sweep that would have crossed over.
3. **Dropping one generation's state removes exactly its tables and indexes, and leaves the other READABLE** — met, with the read-after-drop the criterion explicitly demands (`getCurrent`, `getAsOf`, `getBlock`, `readCursor`), not just a listing diff. Also idempotent, and re-migratable after a drop.
4. **The server-owned fixed tables are NOT namespaced and are still found by their owners** — met, and asserted where it actually matters: `packages/cli/test/generationNamespaceBesideFixedTables.test.ts` builds the combined shape (server schema + two namespaced stores on one libSQL handle) and shows `readSchemaState` over `_meta`, `appendEmissions` over `_emissions`, and the registry over `_generations`/`_generation_pointer` all still working, plus a negative assertion that no generation grew a copy of one of them.
5. **With no namespace configured, names are exactly today's** — met, asserted byte for byte against the literal eight-name list.
6. **Conformance still passes, including with a namespace configured** — met, and the fourth run is better than the criterion asked for: it migrates a DECOY generation into the same handle first, so a statement that spelled an unnamespaced name for itself addresses a table that does not exist and fails loudly instead of passing on a store that was quietly alone.

I injected the exact defect this task's body calls out — leaving the store's own fixed tables shared (`inTableNamespace` namespacing only entity tables, so `_blocks` and `_cursor` stay common) — in a throwaway clone. Result: **7 isolation cases go red** (cross-generation reads, cursor independence, separate blocks, revert isolation, prune isolation, the drop, and the name-shape assertion), while the no-namespace and refusal cases stay green. So the isolation is pinned specifically and not incidentally.

### Scope fences honoured

- **The boundary is exact**: the namespace covers only what the STORE owns. `_meta`, `_emissions` and the generation registry are untouched and shared on purpose, and the reason (a processor-only change re-folds the same stored stream, which is what makes it free) is stated at the assertion site.
- **No leak into the static schema file** — `db.sql` is not touched by this diff at all, which is exactly right: the store's DDL is runtime-generated and the server's is static, and ADR-0053 is explicit that they are principled opposites.
- **Wiring left alone** — `buildProcessor` is untouched; nothing constructs a namespaced store yet, so this is purely additive substrate as the task asked.
- **Absent namespace = today's names, byte for byte** — held and pinned.
- **Identifier discipline stays this backend's** — the seam's declaration rules are unchanged; the new refusals live in `identifiers.ts` and fire at CONSTRUCTION, matching the placement of the existing `sqlite_` rule.

### Judgement on the two knowingly-open hazards

The done record leaves two residuals open with reasons, and I agree with both. (1) SQLite folds identifier case, so `genA` and `GENA` collide; a store sees only its own name and cannot refuse it, and the intended caller derives the namespace from a 32-hex digest, which never meets it. (2) An UNNAMESPACED store declaring an entity literally named `genA_token` would land on namespace `genA`'s `token`; closing that would make a declaration's legality depend on what some other store is called, which is precisely the neighbours-decide-legality failure the derived index prefixing was introduced to avoid, and it needs a database mixing namespaced and unnamespaced stores, which the model does not produce. Documenting rather than half-closing is the right call, and both are written down at the choice site rather than only in the PR.

### Notes carried forward

`drop()` is a verb on this backend and deliberately not on the `StateStore` seam, sized to fit `SQLGenerationRegistryOptions.dropState` from #71. The task that wires them together must pass a namespace that is `[A-Za-z0-9]+` (no underscore) — a rendered `generationDigestOf` qualifies as it comes.

Merging.

PR #73, opened 2026-09-06
feat(the-stored-emission-stream-is-a-stream-a-successor-can-refold): Th…

wighawag:
Task: `work/tasks/done/the-stored-emission-stream-is-a-stream-a-successor-can-refold.md`

All checks green: `pnpm build && pnpm typecheck && pnpm test && pnpm format:check && pnpm check:adr`.

## What I did

**Drift check first.** `the-generation-registry-is-durable-on-sql` has landed (`SCHEMA_VERSION` 3, `_generations` + `_generation_pointer`), so the schema-race the task warned about is resolved rather than live; every ADR the task names (0006, 0035-as-amended, 0038, 0042, 0044, 0052) is current, and `readOnlyStream` / `replayStream` / `IndexerGeneration.replay` are exactly as described. Nothing was stale, so I built it.

**The reader** — `packages/server/src/streamReader.ts`, `storedEmissionStream(db, indexer)`: an `ExistingStream` over `_emissions` for one `(indexer, stream)`, yielding the stored stream from a given block in `seq` order with retractions INCLUDED (`alive` never consulted), holes tolerated, nothing renumbered. The stream half of the key is derived per call from `streamDigestOf(source, streamConfig)`; the name is closed over exactly as `emissionAppenderFor` closes over it. It goes out through `readOnlyStream` (one-writer rule, structural) and `degradingStream` (a DB error must not make a generation unloadable). It deliberately does not ride `createSegmentedStream`. It never clears: every shape it cannot serve is reported ABSENT, because it owns none of those rows.

**Question 1, answer (a)** — a coverage row. The rows alone cannot report coverage (a quiet range moves the cursor without adding one), so `_stream_coverage` holds one claim per `(indexer, stream)` — the fetch-filter identity plus the three block numbers — upserted inside the existing `appendEmissions` `batch()`. `SCHEMA_VERSION` → 4, `db.sql` and the version row agree, reserved-namespace test still green. Question 2: the ADR-0035 cursor-contract properties are asserted (one claim per stream; claim and rows in one `batch()`, and a refused batch leaves neither; an empty save costs one statement), and no segment-port conformance material, since this does not ride the segment helper.

**Tests** — `packages/server/test/storedStreamRefold.test.ts` (23 cases) drives a real `StreamBuilder` over a fixture containing a reorg *and* a quiet range, then re-folds from `_emissions` alone: same state, zero `eth_getLogs`, table byte-for-byte unchanged, `saveNewEvents`/`clear` no-ops, no cross-indexer or cross-stream leakage. I verified the suite is not vacuous by mutating the production code two ways (dropping the `indexer` discriminator; restoring the empty-batch early return) — 11 cases go red.

**Also**: ADR-0055, two changesets (`@etherfold/core` minor, `@etherfold/server` minor), and a `coverage claim` glossary entry in `CONTEXT.md`.

## Decisions

**1. Widened the core `EmissionWrite` port with `coverage`.** The task's scope fence says the coverage row is "written inside the EXISTING `appendEmissions` batch", and there is a place for it — but the payload the appender receives (`{stream, emissions}`) carries no cursor at all, so the data for it does not exist server-side. `EmissionWrite` now carries `StreamCoverage` (`{source, config, latestBlock, lastFromBlock, lastToBlock}`), ABI-free like `EmittedLog` beside it. Alternative considered and rejected: have the host wire a second write — that is the "second appender" the fence forbids, and it could not be atomic with the rows. **Touches** `@etherfold/core` (`emissionStream.ts`, `StreamBuilder.storeStream`) and every host supplying an appender (only `emissionAppenderFor` exists). No emission write moved; ADR-0052's ordering is untouched.

**2. The appender is now called on EVERY batch, including one that emitted nothing.** This reverses a deliberate, pinned behaviour (`StreamBuilder.storeEmissions` early-returned; `packages/core/test/streamBuilder.test.ts` asserted "an empty cycle costs no write", which I rewrote). It is forced by answer (a) being honest: a claim only written on log-bearing batches under-claims over quiet ranges, which is precisely the `MAX(blockNumber)` failure question 1 rules out — a successor permanently behind, and a promotion whose `expectedFromBlock` is too far back, whose re-sent batches ADR-0052 appends twice. It is the shape ADR-0035's contract already requires (`writeCursorOnly`: an empty save writes the cursor and no segment). Visible edge, recorded rather than hidden: the append is not best-effort, so a store that cannot be written now refuses a **quiet** batch too. **Touches** `@etherfold/core` and the sibling task `the-rebuild-replays-the-local-stream-in-bounded-chunks`, which shares question 1.

**3. Named it `_stream_coverage` / "coverage claim", not `_stream_cursor`.** Coherence check: `_cursor` already exists in the *same database*, meaning the state store's opaque **sync cursor** (ADR-0027) — how far a *processor* got. A second `_..._cursor` one table away holding a different thing is the one-word-two-meanings hazard the reserved `_` namespace exists to prevent. "Coverage" is already `CONTEXT.md`'s word for what a cursor asserts. Recorded in the glossary so later tasks inherit the term rather than forking it.

**4. The claim stores only the STREAM half of the context (`source` + `config`), and the reader returns `processor: ''`.** The IndexedDB cursor record stores the whole `ContextIdentifier`; here that would file a generation's fact under a shared key, which is the rule `_emissions` already states about its own columns ("nothing about the PROCESSOR, and no GENERATION"). Verified that `sourceInvalidationOf`'s stream verdict reads `source` and `config` alone, and that no consumer of `fetchFrom`'s `lastSync` reads `context.processor`. Alternative: store the whole context — simpler, but stores a value nothing reads and a later reader could wrongly trust.

**5. A `startBlock` column, written once, and a stream that does not reach back to the requested block is reported ABSENT.** Mirrors the segment keeper's identical check (ADR-0035), so a partial history is not replayed as though it were whole. It is an *absence*, not a new error class, and — unlike the segment keeper — nothing is deleted in response, because this view owns none of those rows. **Touches** the schema (a column) and the migration story: a version-3 database gains the table empty, so a stream reads ABSENT until its writer's next batch, at which point `startBlock` records *that* batch's first block rather than the true history's — which the reader then honestly refuses to serve from below. Called out in the ADR and the changeset.

**6. Exported only the wrapped `ExistingStream`, not the bare `StreamReader`.** You cannot obtain a writable handle on `_emissions` through this module, so the one-writer rule is unreachable-by-construction rather than a convention a caller could skip.

**7. Note for the human, since I must not edit the task body:** the task says whoever answers question 1 with (a) must add `the-generation-registry-is-durable-on-sql` to `blockedBy`. That task is already **done**, so the serialisation it protected is satisfied and `blockedBy: []` is now correct rather than stale; I took its `SCHEMA_VERSION` 3 and bumped to 4 without racing it.


wighawag:

1. **A reader over `_emissions`, scoped to one (indexer, stream), from a given block in `seq` order, retractions included, holes tolerated** — met. Both discriminators are bound on every query, the order is `seq` and never `(blockNumber, logIndex)`, `alive` is deliberately never consulted (that is the canonical VIEW's rule and applying it here would hide the very reorgs a re-fold must replay), and the hole case is asserted directly.
2. **Wrapped through `readOnlyStream`, `saveNewEvents` and `clear` are NO-OPS** — met, and asserted the strong way: a byte-for-byte table snapshot before and after a re-fold, after a `saveNewEvents` pushed straight at the view, and after a `clear`. The `clear` case earns its place — a view that passed it through would have deleted the stream out from under its live writer.
3. **Re-folding reproduces the state the original fold produced, on a fixture containing a REORG** — met, and the reorg is real: the dead 104 and its retraction and its replacement are all in the stored rows, and the rebuild lands on the live branch starting from an EMPTY unconfirmed window, which is only possible by honouring the stored verdicts rather than re-deriving them.
4. **It reads only its own (indexer, stream)** — met, including the sharp case: another named indexer on a BYTE-IDENTICAL stream digest.
5. **Question 1's answer (a) implemented and asserted; question 2 answered accordingly** — met. The coverage claim is one row per `(indexer, stream)`, written in the SAME `batch()` as the rows it covers (asserted by counting statements in the recorded batch, not by inspection), and the cursor-contract properties are asserted as question 2 requires.
6. **Static schema and `SCHEMA_VERSION` agree; the reserved-namespace test still passes** — met, 3 → 4, table `_stream_coverage`, CI green.
7. **Tests in the repo's style** — met, 23 server cases plus the core changes.

The load-bearing claim here is that coverage must NOT be derivable from the rows, because a range that carried no logs moves the stream's reach without adding one. I injected exactly that defect in a throwaway clone — restoring the old `emissions.length === 0` early return, so quiet batches are skipped and coverage effectively becomes `MAX(blockNumber)`. Result: **11 of 23 server cases go red plus the core case**, including `resumes at the coverage the stream claims, which is BEYOND its last log`, `moves on a batch that emitted NOTHING`, and `is PRESENT with no rows at all`. The pin is real and it is the right pin.

### Scope fences honoured

- **The appender did not move.** `storeStream` is called from exactly where `storeEmissions` was, before `processor.process`, and it still catches nothing. ADR-0052's ordering and its not-best-effort rule are untouched, and the "leaves NEITHER behind when the store refuses, and the fold does not advance past it" case pins that.
- **The ADR-0052 asymmetry is preserved, not "fixed".** The schema comment states out loud that `lastToBlock` may sit ONE BATCH AHEAD of every generation's state, and why that is the allowed direction (ADR-0038). Nothing tries to make the append transactional with the fold.
- **No `createSegmentedStream` over SQL** — held, with the reasoning recorded: what is inherited from ADR-0035 is the CONTRACT (one authoritative claim per stream, written with the rows, and presence IS the claim), not the segment layout.
- **Exactly one writer change** — the coverage row, inside the existing `appendEmissions` batch. No second appender, no emission write moved.
- **No runtime-computed names in `db.sql`** — held; `_stream_coverage` is a fixed `_`-prefixed table with static columns.

### The one behaviour reversal, and why it is in scope

Decision 2 reverses a pinned behaviour: the appender is now called on EVERY batch, including one that emitted nothing, and the test that asserted "an empty cycle costs no write" was rewritten. I checked this rather than waving it through, and it is FORCED by the answer the task itself records. Answer (a) rules out (b) precisely because `MAX(blockNumber)` under-claims over quiet ranges; a claim that only moves on log-bearing batches has that identical defect, so honouring (a) requires the quiet batch to reach the appender. The write stays empty and cheap — one upsert, nothing proportional to the history, asserted — which is the shape ADR-0035's empty save already has on the segment keeper. The rewritten test asserts the new behaviour specifically (one write, zero emissions, coverage moved) rather than being loosened, and the reversal is explained at the assertion site instead of only in the PR. The real user-visible consequence — a store that cannot be written now refuses a QUIET batch too, where before it advanced — is a direct restatement of ADR-0052's own rule that a fold which cannot record what it folded must not advance past it. In scope, correctly reasoned, correctly recorded.

### Naming and identity calls I agree with

`_stream_coverage` rather than `_stream_cursor` is right: `_cursor` already exists in the same database meaning the state store's opaque SYNC cursor (ADR-0027), and a second `_..._cursor` one table away holding a different thing is the one-word-two-meanings hazard the reserved namespace exists to prevent. Storing only the STREAM half of the context and returning `processor: ''` is also right, and was verified rather than assumed — `sourceInvalidationOf`'s stream verdict reads `source` and `config` alone.

### Notes carried forward

- **A version-3 database gains `_stream_coverage` EMPTY**, so a stream reads ABSENT until its writer's next batch, at which point `startBlock` records THAT batch's first block rather than the true history's — and the reader then honestly refuses to serve from below it. Documented in the ADR, the schema and the changeset. The rebuild task inherits this.
- **Decision 7 confirms the ordering call**: the task body required `the-generation-registry-is-durable-on-sql` to be serialised ahead of this one under answer (a). I built them in that order, so the `SCHEMA_VERSION` bump was taken cleanly (3 → 4) with no race, and `blockedBy: []` is now correct rather than stale.

Merging.

PR #74, opened 2026-09-06
feat(a-changed-context-creates-a-successor-instead-of-clearing): A chan…

wighawag:
Task: `work/tasks/done/a-changed-context-creates-a-successor-instead-of-clearing.md`

All checks green. Here's what I did.

## What changed

**`packages/core/src/receivingContainer.ts` (new)** — `ReceivingIndexer` / `openReceivingIndexer`, the generation container above `StreamBuilder`: the chain-free sibling of `Indexer`. For one named indexer it opens the durable registry (running its sweep), computes the stream digest, builds the fold `createState` → `createProcessor(state)` (ADR-0043), registers it from the processor's own `getVersionHash()`, and constructs the receiver wired to itself. It exports `SERVER_GENERATION_CAPS` (`{maxGenerations: 4, maxStreams: 2}`) as the documented default, and `ReceivedGenerationSpec` is a `Pick` of `GenerationSpec` so the build order and state-keying rule are inherited, not restated.

**`packages/core/src/streamBuilder.ts`** — new `GenerationContainer` port and `StreamBuilderOptions.container`. `currentLastSync` resolves-or-creates the generation *before* reading the cursor (so a cap refuses before anything is folded), and the mismatch branch calls `processor.clear()` only when there is no container. Absent a container, behaviour is byte-for-byte today's.

**`packages/processor-entities`** — `entityProcessorVersionHash(processor, config)` exported; `getVersionHash()` now *is* that call. It is what makes ADR-0053's "the namespace is derivable before the processor exists" reachable by a host.

**Tests** — `packages/core/test/receivingContainer.test.ts` (the contract at the `processor.clear()` call site, resolve-not-duplicate, restart, caps, no-container-unchanged) and `packages/cli/test/aChangedContextCreatesASuccessor.test.ts` (the durable half over a real libSQL database: rows in `_generations`, per-generation table namespaces, incumbent's state identical before and after, cap refusal leaving no orphan record and no orphan tables, no double-append). Both were checked to FAIL when the guards are removed.

**Also** — two changesets, a `CONTEXT.md` glossary entry recording that there are now two containers and what distinguishes them, and one observation note (`work/notes/observations/a-re-folding-successor-would-re-count-reverts-into-the-shared-reorg-counters.md`).

## Decisions

**The container lives in `@etherfold/core`, as `ReceivingIndexer` in `receivingContainer.ts`.** In core because everything it composes is already there (the registry, the promotion/identity model, `StreamBuilder`) and because both the server and the CLI must reach it — putting it in `@etherfold/server` would make `@etherfold/core`'s own receiver depend upwards, and putting it in the CLI would leave the Worker host unable to hold generations. Named `ReceivingIndexer` rather than a new noun: `CONTEXT.md` already defines *indexer* as the named unit holding generations, and both containers ARE that; the adjective says which half of the wire (`StreamBuilder` is already "the receiving half"). Rejected: `StreamIndexer` (re-means *stream*, which is a filter+config identity, not a runtime), `GenerationContainer` as the class name (kept for the narrow PORT the receiver holds, which is what "container" means from below).

**Rules LIFTED versus rules deliberately NOT taken from the existing container.** Lifted by reuse rather than by copying: the two factories (a `Pick` of `GenerationSpec`, so the order and the keying rule have one home), and the ONE-WRITER rule via the already-shared `writerOf` — the emission appender is handed only to the writer of the stream, so a successor over a shared stream cannot double-append (ADR-0052/ADR-0044); `writesStream` reports it and nothing can set it. Deliberately NOT taken: `CanonicalGenerationNotHeldError`. On this runtime a generation's state is a table namespace and the read tier resolves the pointer to name it (ADR-0053), so the canonical generation answers with no engine at all — which is exactly the ordinary upgrade, where the host holds only the new processor. Refusing there would turn every upgrade back into the outage. Touches: `two-named-indexers-never-touch-each-others-data` and `the-cli-and-the-server-hold-generations-the-same-way`, which will build hosts on this container. Also not taken: `promotion` — creating is this task, moving the pointer is not.

**The caps are an option on the container (`caps`), and the container takes the registry PORT rather than an opened registry.** `openGenerationRegistry` takes caps as an argument, so a host handing over an opened registry would already have decided the bound inside whichever module built the port — which is how a bound nobody can see gets chosen by a substrate. Taking the port also puts the sweep at container open. Default `SERVER_GENERATION_CAPS = {maxGenerations: 4, maxStreams: 2}`: four is the incumbent + the successor + the predecessor kept so a revert stays free + one spare so a routine upgrade never meets the bound as a surprise; two streams (not four) because a generation re-folds a stream that is already stored while a STREAM is the expensive thing to re-fetch — the browser's own reasoning at the same number for the same resource. Alternative considered: no default at all, as `openGenerationRegistryOnSQL` does. Rejected because a host that must state a number before it can hold one generation makes the additive path harder than the destructive one. Touches: `ReceivingIndexerOptions.caps`, and the server options / CLI config that task 10 threads into it.

**The generation is registered at container OPEN, not on the first batch.** So a cap refuses at start-up, in an operator's boot log, naming what to delete — rather than as a `500` on somebody's ingest — and the record exists with the widest possible margin before anything writes the stream (the registry's own ordering rule). The receiver still resolves-or-creates on every cursor read (memoised on the container), which covers a version hash that moves after open and keeps `StreamBuilder` honest with a container it did not build. Alternative considered: register lazily at the `clear()` branch only; rejected because that branch is unreachable once state is per generation, so the generation would never be registered on the happy path.

**`entityProcessorVersionHash` is a new export on `@etherfold/processor-entities` (a second package).** ADR-0053 asserts the namespace is derivable before the processor exists; nothing exposed the value it is derivable from, so a host would have had to re-spell `${version}-${simple_hash({entities, config})}`. This is not the thing ADR-0043 rejected (a caller *declaring* a hash that can silently disagree): `getVersionHash()` now returns exactly this call, so the two cannot diverge. Touches: any host naming a generation's table namespace, which is task 10's job.

**Scope held at the container; the commands are NOT rewired.** `prepareIndexing`, `indexCommand` and `folding.ts` are untouched, because `the-cli-and-the-server-hold-generations-the-same-way` owns that move (and is blocked on two tasks that also rework `folding.ts`). The host assembly this container needs is exercised in `packages/cli/test/aChangedContextCreatesASuccessor.test.ts` — real SQL registry port, real namespaced stores, one handle — so the container is proved over the real substrate and task 10 has the shape it will move into `folding.ts`.

**No new ADR.** Every choice above sits inside a decision already recorded (ADR-0043, ADR-0044, ADR-0052, ADR-0053) or is cheap to reverse (the caps default, the container's name, where registration happens). The one that came closest to the ADR gate — not lifting `CanonicalGenerationNotHeldError` — is already implied by ADR-0053's "a read tier must resolve the canonical pointer before it can name a state table", and is a refusal that could be added later without moving any bytes.


wighawag:

1. **A batch whose context names an unseen fold CREATES a generation instead of calling `processor.clear()`, asserted AT the call site** — met. `currentLastSync`'s discard branch is now conditional on the container, and the spy (`world.cleared`) is asserted empty. See the note below on which of the two cases actually carries the assertion.
2. **The canonical generation's state is untouched by the creation** — met, read before and after and compared, plus the same thing over the real SQL substrate in the CLI test.
3. **Registered durably, state in its own namespace, nothing written into the canonical generation's tables or cursor** — met, asserted on real tables in one libSQL handle: the two namespaces share no table, so a successor cannot reach the cursor the canonical generation resumes from.
4. **Re-presenting the same context RESOLVES rather than creating a second record** — met, asserted at both seams and by counting rows in `_generations`.
5. **At the cap, creation is REFUSED naming what to delete, with no partial generation left behind** — met, and the "nothing partial" half is done properly: no orphan record AND no orphan tables, by comparing the full table listing before and after, with the observation that a refused generation's store creates nothing until its first write so a refusal costs no DDL.
6. **The caps are configurable by the host, with a documented default, and a different bound refuses at that bound** — met. `SERVER_GENERATION_CAPS = {maxGenerations: 4, maxStreams: 2}` with the reasoning stated (incumbent + successor + the predecessor kept so a revert stays free + one spare so a routine upgrade never meets the bound as a surprise), it is reported through `caps`, and both the lower-bound refusal and the raised-bound success are asserted.
7. **A restart comes back holding the same generations with the same canonical one** — met, at both seams.
8. **A `StreamBuilder` built WITHOUT a container behaves exactly as today, including the discard** — met, and asserted POSITIVELY: the control case checks `cleared === ['v2']` and the rows gone, rather than just checking the new path is off.
9. **A changeset accompanies the public API change** — met, two of them.
10. **Tests in the repo's style** — met.

### Regression pins verified for real

I injected two defects in a throwaway clone:

- **Dropping the one-writer rule** (handing `appendEmissions` to every generation, not just the stream's writer): `does NOT append the stream a second time` goes red. This matters more than a normal nit — under ADR-0052 a duplicate append is indistinguishable from a real second emission, for ever.
- **Restoring the outage** (forcing the `clear()` branch even with a container): `does not clear even when the caller handed both folds ONE store` goes red.

Everything else stayed green in both runs, so the pins are specific.

One honest observation on the first criterion. The headline test's `expect(world.cleared).toEqual([])` is belt-and-braces rather than the real pin: with separate namespaces the successor's `load()` finds no cursor, so the discard branch is never reached and the assertion passes vacuously. The case that genuinely exercises the call site is `does not clear even when the caller handed both folds ONE store, which is the pre-generation database`, and that one does go red when the defect is injected. So the criterion IS pinned; it is just pinned by the second test. Not a block, and arguably the better arrangement, since the pre-generation database is the real-world shape of the branch.

### Scope fences honoured

- **Processor-only change only** — held; the filter/config case and the wire widening are left to `one-registry-entry-holds-several-live-wire-contexts`.
- **Out of scope stayed out** — no rebuild driver, no pointer move (the test explicitly asserts the pointer did NOT move), no `/status` surface, no multi-name hosting.
- **Behaviour with no container is exactly today's** — held and pinned, so the Worker host and existing tests are untouched.
- **Reuse the model, do not restate it** — held, and this is the part I checked hardest. The two factories are a `Pick` of `GenerationSpec` rather than a copy, and the one-writer rule comes from the ALREADY-SHARED `writerOf` rather than a second implementation. No second copy of the promotion trigger, because promotion is not in this task at all.
- **Commands not rewired** — `prepareIndexing`, `indexCommand` and `folding.ts` untouched, correctly left to task 10, while the host assembly is still proved over the real substrate in the CLI test so task 10 inherits a working shape rather than a sketch.

### Carry-forward from #71, partially delivered here

ADR-0044's amendment named this task as one of the two places the ENGINE half of writer succession lands. This PR delivers the part it owns: `writesStream` is derived from `writerOf`, and the appender is structurally withheld from a non-writer rather than a caller being trusted not to append. What remains is moving the wire context's receiver on a delete, which needs several live wire contexts per entry and is `one-registry-entry-holds-several-live-wire-contexts`. Still tracked.

### The deliberate NON-lift, which I agree with

`CanonicalGenerationNotHeldError` was deliberately not lifted from the chain-facing container. On this runtime a generation's state is a table namespace and the read tier resolves the pointer to name it (ADR-0053), so the canonical generation answers with no engine at all — which is exactly the ordinary upgrade, where the host holds only the new processor. Refusing there would turn every upgrade back into the outage this task exists to remove. Right call, and recorded rather than silently skipped.

### Observation filed by the build (not fixed in place, correctly)

`work/notes/observations/a-re-folding-successor-would-re-count-reverts-into-the-shared-reorg-counters.md`: the emission append now has a one-writer rule but `recordReorg` does not, and its counters are per NAMED INDEXER in `_meta` (ADR-0050), so a successor re-folding a stream with retractions would count those reverts a second time. Not reachable today because nothing yet drives a successor through a re-fold; it belongs to `the-rebuild-replays-the-local-stream-in-bounded-chunks`, which is the next task but one. Good capture, right destination.

Merging.

PR #75, opened 2026-09-06
feat(one-registry-entry-holds-several-live-wire-contexts): One registry…

wighawag:
Task: `work/tasks/done/one-registry-entry-holds-several-live-wire-contexts.md`

All checks green (`format:check`, `check:adr`, `build`, `typecheck`, full `test` across every package). Working tree holds only this task's changes.

## What I did

**The registry entry now holds several live wire contexts** (`packages/server/src/registry.ts`). `IndexerRegistryEntry` was `{ingestion}`; it is now two questions — `liveIngestions(): Promise<readonly LogIngestion[]>` (one receiver per live context) and `canonicalGeneration(): Promise<GenerationId>`. `singleContextEntry(ingestion)` is the one-context entry, and `indexerRegistry` builds a registry of those, so every existing host behaves exactly as before.

**The ingest route selects the indexer by segment and the receiver by the batch's own `{source, config}`** (`api/ingest.ts`), using core's own `sameWireContext` (now exported, so the selection rule and the refusal rule cannot drift). `POST /{indexer}/ingest/expected-from-block` answers `{success, contexts: [{context, expectedFromBlock}, …]}`, one per live context, asked in turn (answering can write). A context no live receiver holds is the ordinary non-resumable `400 context-mismatch`, whose `expected` now names *every* live context.

**The container holds several folds** (`packages/core/src/receivingContainer.ts`): `ReceivingIndexer.add(spec)` builds a fold beside the ones held — its own state, processor and receiver — registering it before the receiver exists (so a cap refuses with nothing partial). `liveIngestions()` derives liveness from `registry.list()`; `canonicalGeneration()` reads both halves in one registry read. A `ReceivingIndexer` structurally *is* an entry, so a host registers it directly.

**Both feed views answer from the canonical generation**, stream and fold read together once per request (`api/feed.ts`).

**Sender kept green**: `IngestionTarget.expectedFromBlock(context)` now names the asker; `LogFetcher` passes its own context, `createHttpIngestion` finds its entry in the list, `createDirectIngestion` ignores it (one receiver).

Tests: new `packages/server/test/severalLiveWireContexts.test.ts` (9 cases at the HTTP seam over a real `ReceivingIndexer` + SQL registry: two streams fed independently, cursors isolated, the `409`, the `400` naming both, `404`, deletion/reaping ending a context's life, the pointer moving *not* ending one, and both feed views following the canonical generation including the `stream-mismatch` on the stale cursor), plus a `one container, SEVERAL live wire contexts` block in `packages/core/test/receivingContainer.test.ts` for the selection seam beneath. Changeset added; `CONTEXT.md`'s *indexer* entry and `packages/server/README.md` updated; a dated addendum on ADR-0044 records where the engine half of writer succession now stands.

## Decisions

**The entry became TWO ASYNC QUESTIONS (`liveIngestions()` / `canonicalGeneration()`), not a list field.** Liveness and the canonical pointer are facts only the durable generation registry holds, and both move without this host being told (a generation deleted by an operator making room at a cap; a pointer moved by another process), so a record captured at boot could express neither. Alternatives considered: a plain `ingestions: LogIngestion[]` field (cheap, but then "live" could only mean "what this process last thought", which is exactly the hard-coded retirement rule the task forbids), and keeping `ingestion` beside a new plural (two ways to say one thing, and every route would have to choose). Touches every host's resolver — `packages/cli/src/indexCommand.ts`, `platforms/nodejs` (type only), and the `getIndexer` docs — and it is what lets `ReceivingIndexer` be registered as the entry with no adapter. The resolver itself stays SYNCHRONOUS: resolving a NAME is a lookup in what the host was built with (so `404`/`501` still cost no database read), while resolving what the name currently HOLDS is the async part.

**ONE RECEIVER PER STREAM, and `add` REFUSES a second fold on a stream already held.** A batch carries `{source, config}` and nothing that could tell two folds over one stream apart, so a second receiver at that address would be reachable only by iteration order. Such a fold is a PROCESSOR-change successor, and ADR-0044 already says how it advances (re-fold the stream its writer stores), so the refusal points there. It is a new REFUSAL, deliberately a plain `Error` at host-assembly time rather than a typed one (precedent: `createHttpIngestion`'s "needs the NAME") because no sender or fetcher host classifies it. Consequence I did NOT build, and which ADR-0044's amendment already reserved: deleting a stream's writer removes the only receiver that stream had, so its context stops being live rather than the wire moving to the next-oldest generation — the records succeed instantly, the engine half does not. Recorded as a dated addendum to ADR-0044 rather than a new ADR, since it is that ADR's own rule applied rather than a new decision. Touches `the-rebuild-replays-the-local-stream-in-bounded-chunks` and `the-cli-and-the-server-hold-generations-the-same-way`, which build the hosts that would hold such a fold.

**`expected-from-block` answers a `contexts` LIST with no top-level pair, and the ASK NAMES THE ASKER.** The task requires the list; what it does not settle is how a sender picks its entry, and `createHttpIngestion` cannot know its own `{source, config}` (only `LogFetcher` does). So `IngestionTarget.expectedFromBlock(context)` takes a REQUIRED argument. Alternatives: keep a top-level pair beside the list (a single pair can only ever name one of several — silently wrong, and this repo does not do accept-and-ignore), or have the client take the FIRST entry when it cannot match (the plausible-wrong-answer this design refuses). Consequence, and the one user-visible behaviour change on the sending side: over HTTP a fetcher pointed at a receiver that does not fold its context now fails at the ASK with `IngestionRefusedError` code `context-mismatch` instead of `WireContextMismatchError`. Both are fatal, non-retryable and raised before a single log is fetched, and `@etherfold/fetcher-host` classifies both as `fatal`; the direct (in-process) transport still raises `WireContextMismatchError` from `LogFetcher`'s own check, unchanged. Touches `@etherfold/core`'s public `IngestionTarget` and every hand-written target (one test wrapper in `fetcherRoundTrip.test.ts` updated).

**`400 context-mismatch` now carries `expected` as an ARRAY, always — even for one live context.** Naming every live context is information; picking one to call "expected" is a policy with no basis, which is the same call `GenerationCapReachedError` makes when it lists every deletable generation. A conditional shape (scalar for one, array for several) was rejected: a sender would have to branch on it. The receiver's own `WireContextMismatchError` (should the route and the receiver ever disagree) is mapped into the same array shape rather than kept as a second body.

**Liveness compares the fold's REGISTRATION RECORD, not its current version hash.** A processor whose `getVersionHash()` moves after open (via `configure()`) must still be able to receive — `StreamBuilder` resolves-or-creates the new generation on the next cursor read — so filtering on the live hash would drop the receiver just before the moment it is supposed to create its successor.

**`ReceivedGenerationSpec` gained `source` and a per-fold `stream` config.** `source` is the only way to say "a different stream", which is what a second live wire context IS. The stream CONFIG is per fold here although the chain-facing container refuses it per generation (one keeper serves one indexer there, ADR-0044's last consequence); on the receiving side each fold has its own `StreamBuilder`, which holds its own config, so the reason for that limit does not exist. Both default to the container's, so a single-fold host states neither twice.


wighawag:

1. **An entry holds several live wire contexts; the route selects the indexer, the batch's `{source, config}` selects the stream-builder** — met.
2. **A successor on a NEW stream receives its own batches while the incumbent keeps receiving its own, neither advancing the other's cursor** — met, asserted on both cursors rather than only on delivery.
3. **`expected-from-block` answers one `{context, expectedFromBlock}` per live context; a single-context sender still resumes, including through a `409`** — met, and the `409` case is kept as its own test rather than folded into the list case.
4. **The refusal families are unchanged** — met: `404` unknown name, `501` no registry, `400` for a context in no entry (not resumable), `409` carrying the expected value. The `404` case is asserted specifically as "does not reach one of these contexts", which is the failure a widening like this would plausibly introduce.
5. **A context stops being live when its generation is deleted or its stream reaped, and a batch for it is then the ordinary `400`** — met.
6. **The live set is DERIVED from the registry, not from a hard-coded promotion-retires-the-previous rule** — met, and this is the criterion I pinned hardest (below). The complementary test — `does NOT stop being live because the pointer moved: a retained generation is still fed` — is exactly the assertion the criterion is asking for.
7. **Both feed views answer from the CANONICAL generation's stream and advertise its generation, read once per request** — met. `canonicalGeneration()` is one call and both halves come from it, with the comment stating why: read separately, a response could pair one generation's stream with another's fold.
8. **A changeset records the response-shape change** — met.
9. **Tests in the repo's style** — met, 9 new server cases plus core and platform updates.

I injected the exact thing criterion 6 forbids — a hard-coded "a promotion retires the previous context", by filtering `liveIngestions()` to the canonical generation. Result: **8 of the 9 cases in `severalLiveWireContexts.test.ts` go red**, including the retained-generation case and the `400`-naming-every-live-context case. The derivation is pinned, not merely written.

### Scope fences honoured

- **Receiver plus keeping the existing sender green** — held: a fetcher pushing one context finds its own entry and behaves as before, asserted in `fetcherRoundTrip` and `logFetcher`.
- **A fetcher host running one loop per context is NOT built** — held; it is enabled and left.
- **The retained-generation policy is NOT baked in** — held, and this is the fence that mattered most. The question stays a policy input to the registry rather than something a future task would have to unpick from the routing.

- **The entry became two ASYNC questions rather than a list field.** Right, and the reasoning is the load-bearing part: liveness and the pointer are facts only the durable registry holds and both move without this host being told, so a record captured at boot could express neither — which is precisely the hard-coded rule the task forbids. Keeping the NAME resolution synchronous is a good split: `404`/`501` still cost no database read.
- **`expected-from-block` returns a `contexts` list with no top-level pair, and the ask names the asker.** A single pair can only ever name one of several, and taking the first entry when it cannot match is the plausible-wrong-answer this repo consistently refuses. The consequence is real and is stated: over HTTP a fetcher pointed at a receiver that does not fold its context now fails at the ASK with `IngestionRefusedError` code `context-mismatch` rather than `WireContextMismatchError`. Both are fatal, non-retryable, raised before a single log is fetched, and both classified `fatal` by `@etherfold/fetcher-host`, so no caller's handling changes.
- **`400 context-mismatch` carries `expected` as an ARRAY always, even for one.** Consistent with `GenerationCapReachedError` naming every deletable generation, and a conditional shape would make a sender branch.
- **Liveness compares the REGISTRATION RECORD, not the current version hash.** Subtle and correct: a processor whose hash moves via `configure()` must still be able to receive, because `StreamBuilder` resolves-or-creates the new generation on the next cursor read. Filtering on the live hash would drop the receiver at exactly the moment it is supposed to create its successor.

### The residue I am tracking, and why it does not block

`add` REFUSES a second fold on a stream already held — one receiver per stream — because a batch carries `{source, config}` and nothing that could tell two folds over one stream apart. That is right, and it follows ADR-0044's own rule that a processor-change successor re-folds the stream its writer stores rather than being fed.

What follows from it is the residue: deleting a stream's WRITER removes the only receiver that stream had, so its context stops being live rather than the wire moving to the next-oldest generation. The records succeed instantly (that is #71's rule, working); the engine half does not. ADR-0044's amendment originally named this task and `a-changed-context-creates-a-successor-instead-of-clearing` as where the engine half would land, and neither has fully delivered it — but rather than let that quietly lapse, this PR writes a dated addendum recording precisely where the rule now stands and why, and names the rebuild driver and the CLI/server assembly as what would have to arrange it. That is the honest disposition, and the criteria for THIS task never included writer succession.

I am carrying it forward: I will check whether `the-rebuild-replays-the-local-stream-in-bounded-chunks` or `the-cli-and-the-server-hold-generations-the-same-way` picks it up, and if neither does it goes in the drive report as an item needing a task rather than being lost between three ADR amendments.

Merging.

PR #76, opened 2026-09-06
feat(the-rebuild-replays-the-local-stream-in-bounded-chunks): The rebui…

wighawag:
Task: `work/tasks/done/the-rebuild-replays-the-local-stream-in-bounded-chunks.md`

## What I built

**The rebuild driver**, so a processor upgrade costs a local scan instead of a re-index.

**`packages/core/src/generation/rebuild.ts` (new)** — `GenerationRebuild`, the platform-neutral, chain-free, bounded catch-up: `more()` reads the durable checkpoint, reads one bounded slice of the stored stream through a `ReplaySource` port, decodes it against the source running now, replays it through `generateStreamFromReplay` (honouring the verdicts the stream carries, ADR-0042), and returns a `RebuildReport` — the `prune` / `compactEmissionPairs` shape (ADR-0022). It holds no position between calls, so a fresh object graph resumes from what the store committed. Its decoder is built over a provider that throws on every call, so a reach for the chain is loud rather than a quiet re-index.

**`packages/server/src/streamReader.ts`** — `storedEmissionReplaySource`, the same `_emissions` rows read in bounded slices over the coverage claim (ADR-0055), read-only by construction (the port has no write on it).

**`packages/core/src/receivingContainer.ts`** — a fold on a stream the container already holds is now a **follower** (determined by the stream, never configured, ADR-0044): no receiver, a `GenerationRebuild` instead. `rebuildMore()` is the host-scheduled call; it advances every follower by a chunk and then settles the pointer. Promotion policy is applied (not re-decided), with `promote(id)` ungated. The retired generation is retained, keeps its receiver, keeps folding, and keeps the append duty.

**Lifted rather than copied** (the task's explicit instruction): `readyForPromotion` and `promotionOnAdd` are new in `generation/promotion.ts` and both containers now go through them; `batchStreamForDelivery` is a new internal shared by `IndexerGeneration.promiseToFeed` and the rebuild — a replayed stream can carry `+X`, `-X` and `+Y` at one block, and one `process()` call for all three is a primary-key collision, not a fold.

**Tests**: `packages/core/test/rebuild.test.ts` (contract, report, resumability through fresh containers, kill between chunks, pointer once, retention, policy), `packages/server/test/rebuildInBoundedChunks.test.ts` (the SQL cut over real rows), `packages/cli/test/theRebuildReplaysTheLocalStream.test.ts` (end to end on one libSQL handle with real entity processors, real namespaces, reads resolved through the pointer). Plus a changeset and `docs/adr/0056`.

`pnpm format:check && pnpm check:adr && pnpm build && pnpm typecheck && pnpm test` all pass. (`changeset status --since=main` reports no changesets only because mine is still untracked; it compares via `git diff`, so it passes once the runner commits — the same shape as the previous task in this spec.)

## Decisions

**Where the checkpoint lives: it IS the successor's own sync cursor, and there is no second durable value.** `EventProcessor.process` persists the `LastSync` describing each block in the same transaction as that block (ADR-0027 puts the cursor behind the storage seam precisely because only the store holds that transaction), so "the state and the checkpoint commit together" is a guarantee the seam already makes rather than one this driver arranges. Alternative considered: a `_rebuild` checkpoint table keyed per generation. Rejected because it cannot be atomic with the state write, so a crash between the two leaves a checkpoint that re-applies or skips a chunk — the exact failure the chunking exists to survive. Touches: any future host that wants to inspect rebuild progress reads the generation's own cursor, not a rebuild row. Recorded in ADR-0056.

**How a chunk is sized: a budget in EMISSIONS, cut on a BLOCK boundary, and always ending ABOVE the fold's own position.** Emissions because that is what the work is proportional to; a block boundary because the stream is `seq`-ordered and a reorg puts an application, its retraction and its replacement at one block at arbitrarily separated `seq` values, so a row-aligned cut would leave rows below its own resume point and skip them for ever. The third clause is the non-obvious one and it fixes a real stall I hit while building: `getFromBlock` REACHES BACK over the reorg window, so a fold level with the tip re-reads blocks it has already folded, and a budget spent inside them cuts the chunk where the fold already is — the identical chunk is then asked for for ever, with no error anywhere. So `ReplayChunkQuery` carries `foldedThrough` and the reader guarantees a cut above it; the extra work is bounded by the reorg window, which is what one live cycle already pays. `DEFAULT_MAX_EMISSIONS_PER_CHUNK = 2000`. Touches: any second `ReplaySource` implementation (an IndexedDB one) owes the same rule. Recorded in ADR-0056 with a regression test at both the reader and the driver level.

**"Caught up" is measured against the stream's COVERAGE CLAIM, not the emission `seq` high-water — a deliberate deviation from this task's answered open question 1.** The launched answer says catch-up is decided on `readStreamHighWaterMark`. I could not implement that honestly: evaluating it needs a persisted consumed-`seq`, which cannot be written in the store's block transaction and would therefore be the second durable checkpoint the decision above rejects; and the naive in-memory form is simply wrong in the steady state, because the resume point walks past old rows so a level fold reads none and its highest observed `seq` sits permanently below the high-water. The coverage claim (ADR-0055, landed by the sibling task since this one was written) removes the under-claim the original answer was guarding against — it moves on every batch including quiet ones — and it is the same space the promotion trigger already compares in (`hasReachedCursor` over `lastToBlock`, ADR-0046), so no second notion of "level" enters the system. The high-water is still read and REPORTED on every chunk (`RebuildReport.highWater`) as the honest measure of what is being folded. Alternative considered and rejected: `complete = the read hit its budget`, which is the same stall wearing a different hat. Touches: `a-rebuild-in-progress-is-never-an-empty-answer` (the `/status` dimension will report from this) and `the-canonical-pointer-moves-back-without-re-ingesting`. Recorded in ADR-0056.

**`immediate` together with `dropOnPromotion` is REFUSED on the receiving container.** ADR-0046 requires that combination to DEFER the drop until the successor reaches the cursor the previous generation had at the promotion, and that interlock is per-advance bookkeeping the chain-facing container keeps and this one does not. `on-catch-up` and `manual` drops ARE implemented, including the never-drop-a-followed-writer decline (which is what fires on the ordinary processor upgrade). Alternative considered: accept the flag and ignore it — rejected as accept-and-ignore on the one setting whose cost is unrecoverable (a complete state discarded for one that has proved nothing). Touches: `ReceivingIndexerOptions.promotion`, and whatever `the-cli-and-the-server-hold-generations-the-same-way` threads into it.

**`ReceivingIndexer.add` no longer refuses a second fold on a held stream; it makes it a FOLLOWER.** That refusal was a placeholder whose own message named this task as the fix, and ADR-0044 says the choice is determined by the stream and is never a knob — so a separate `follow()` verb would have made it one. Consequence: `HeldFold.ingestion` is now optional and `HeldFold.follows` is reported beside `writesStream`. A container given no `replay` source still refuses, naming the missing port, rather than registering a generation that could never advance. Touches: `receivingContainer.test.ts` (rewritten assertion), `severalLiveWireContexts.test.ts` (one optional-chain), and the server's `IndexerRegistryEntry`, whose `liveIngestions()` contract is unchanged.

**Named the container verb `rebuildMore` rather than `followMore`.** Coherence check: `followMore` already means "advance from the stored stream alone", UNBOUNDED, on `IndexerGeneration` — reusing it for the bounded version would give one name two bound-semantics. `rebuild` already means exactly this in ADR-0008 and `CONTEXT.md` (blue-green rebuild), so it re-means nothing. The JSDoc and the `CONTEXT.md` **follower** entry say the two are one mechanism at two bounds.

**No new index for the block-ordered probe**, matching `compactEmissionPairs`' own precedent: the canonical index is partial on `alive = 1`, so a scan over a stream including retractions is a table scan whatever the order, and a third index on a table weighed against D1's 10GB ceiling is paid for by every deployment. Nothing here has been measured against a stream large enough to say it earns its keep; recorded as a consequence in ADR-0056 so the day it does is a decision and not a discovery.


wighawag:

This is the largest task of the spec (stories 2, 8, 9, 10) and the diff carries its weight: a platform-neutral driver, a bounded reader over `_emissions`, the promotion policy applied, and three test files at three seams.

1. **A successor on a SHARED stream reaches the incumbent's position by replaying `_emissions` alone, ZERO chain calls and ZERO writes to the stream** — met, and asserted the strong way in both directions: the successor is handed a provider that REFUSES, so a reach for the chain is loud rather than a quiet re-index, and the table is compared byte for byte after the rebuild.
2. **Bounded chunks, each call bounded and REPORTING whether it finished** — met, in the `prune`/`compactEmissionPairs` shape the task named. A budget of zero is refused rather than read as "do nothing", which is the right call for a footgun that would otherwise look like completion.
3. **Resumability asserted through a FRESH container between chunks, including a kill** — met at all three seams, with "applying nothing twice and skipping nothing" asserted rather than just "it finished".
4. **Reads answer from the canonical generation throughout, unchanged until the pointer moves** — met.
5. **The pointer moves ONCE, atomically, at level; rollback before the move is a no-op; after it the retired generation is still registered, still holds its own state and can still answer** — met, including `does not move the pointer again on the chunk after a REVERT`, which is the case that would break `the-canonical-pointer-moves-back-without-re-ingesting` if it were wrong.
6. **Question 2 implemented and asserted** — met, precisely as answered: the retired generation KEEPS FOLDING, the append duty does NOT move with the pointer, `writerOf` still names v1 after promotion to v2, its context stays live, and the promoted successor keeps following. The test drives a real later batch through the retired writer and checks both generations' rows.
7. **A rebuild over a stream containing a REORG lands on the same state the original fold produced, compared as STATE** — met at all three seams, state-compared and not row-counted.
8. **Tests in the repo's style** — met.

ADR-0056's least obvious rule is that a chunk must end strictly ABOVE the fold's own position, because `getFromBlock` reaches back over the reorg window, so a budget spent inside that window cuts the chunk where the fold already is and the identical chunk is asked for for ever — a rebuild that never completes, with no error anywhere. I removed that guarantee (`floor = max(foldedThrough + 1, fromBlock)` → `fromBlock`) in a throwaway clone: `never cuts AT OR BELOW what the fold already covers` and `finishes over a stream whose whole tail sits inside the reorg window` both go red, 13 others stay green. Note they FAIL rather than hang, so the driver has a bound rather than relying on the reader's good behaviour.

This is the best kind of finding in the drive so far: a real stall discovered while building, fixed, explained in an ADR, and pinned at both the reader and the driver level.

### Scope fences honoured

- **It replays, it does not fetch** — held and asserted with a refusing provider.
- **The verdicts the stream carries are honoured** (ADR-0042), and the unconfirmed window is rebuilt by WALKING the stream rather than filtering `removed` out.
- **Platform-neutral** — the driver holds no position between calls at all, so a Node cron, a CLI loop and a browser can each drive it. The Worker's scheduling stays the deferred follow-on the spec named.
- **The promotion policy is applied here and re-decided nowhere** — held; `on-catch-up` default, `manual`, and the never-drop-a-followed-writer decline all come from the shared model.

### The deliberate deviation, which I examined closely and accept

**Catch-up is measured against the stream's COVERAGE CLAIM, not the emission `seq` high-water — which overturns this task's own answered open question 1.** I did not wave this through, because overturning an answered question recorded jointly with a sibling task is exactly the kind of thing that should either be a STOP or be very well argued. It is very well argued:

- The launched answer cannot be implemented honestly. Evaluating it needs a persisted consumed-`seq`, which cannot be written inside the store's block transaction, so it would be the SECOND durable checkpoint that the same ADR rejects on correctness grounds — a crash between the two writes leaves a checkpoint that re-applies or skips a chunk, which is the exact failure the chunking exists to survive.
- The naive in-memory form is not merely weaker, it is wrong in the steady state: the resume point walks past old rows, so a level fold reads none and its highest observed `seq` sits permanently below the high-water. It would never report caught up.
- The reason `seq` was chosen over blocks in the first place was the under-claim of `MAX(blockNumber)` over quiet ranges. ADR-0055's coverage claim — which landed in the sibling task AFTER this task's answer was written — removes exactly that under-claim, because it moves on every batch including the quiet ones. So the hazard the original answer was guarding against no longer exists.
- It is the same space the promotion trigger already compares in (`hasReachedCursor` over `lastToBlock`, ADR-0046), so no second notion of "level" enters the system, and because the append is ordered before the fold (ADR-0052), reaching the stream's claim implies reaching the incumbent's cursor. The predicate can be conservative, never premature.
- The high-water is still read and REPORTED on every chunk as the honest measure of what is being folded. It is demoted from predicate to telemetry rather than discarded.

All of that is in ADR-0056 with the rejected options, so a later reader meets a decision rather than a discrepancy. I am flagging it in the drive report anyway, because it supersedes prose in a done task body that a human answered.

### Other calls I checked and agree with

- **The checkpoint IS the successor's own sync cursor, with no second durable value.** Right, and for the reason that matters: ADR-0027 puts the cursor behind the storage seam precisely because only the store holds the transaction, so "state and checkpoint commit together" is a guarantee the seam already makes rather than one this driver arranges.
- **`add` no longer refuses a second fold on a held stream; it makes it a FOLLOWER.** This correctly retires the placeholder from #75 whose own message named this task as the fix, and it avoids the trap a separate `follow()` verb would have set: ADR-0044 says the choice is determined by the stream and is never a knob. A container given no `replay` source still refuses, naming the missing port, rather than registering a generation that could never advance.
- **`immediate` + `dropOnPromotion` is REFUSED rather than accepted and ignored.** The right answer on the one setting whose cost is unrecoverable.
- **`rebuildMore`, not `followMore`** — `followMore` already means the UNBOUNDED version on `IndexerGeneration`, and one name with two bound-semantics is the coherence hazard this repo keeps catching. Good.
- **No new index**, with the reasoning recorded as an ADR consequence so the day it earns its keep is a decision and not a discovery.

### Residue still open after this PR

The writer-succession ENGINE half (moving the wire to the next-oldest generation when a stream's WRITER is deleted) is still not built. This PR changes the picture — a stream can now hold several folds, one receiver plus followers — so it is now *buildable* where #75 recorded that it was not, but nothing here does it, and the ADR-0044 addendum's named owners are this task and `the-cli-and-the-server-hold-generations-the-same-way`. I will check task 10 and, if it does not land there, raise it in the drive report as needing its own task rather than leaving it distributed across three ADR amendments.

Merging.

PR #77, opened 2026-09-06
feat(two-named-indexers-never-touch-each-others-data): Two named indexe…

wighawag:
Task: `work/tasks/done/two-named-indexers-never-touch-each-others-data.md`

Done. Everything the task asks for is built and the gate is green.

## What I did

**Checked the task against reality first.** ADR-0053, ADR-0036 and ADR-0052 all still say what the task quotes; the blocking tasks landed as described (`IndexerRegistryEntry` is the two-question object from `one-registry-entry-holds-several-live-wire-contexts`, and the per-generation table namespace exists in `@etherfold/state-store-sqlite`). No drift, so I built it.

**The change (`@etherfold/server`).** `IndexerRegistryEntry` grew a **required** `db: RemoteSQL`: a named indexer IS a database (ADR-0053), so the handle is part of what the name resolves to. `getDB` is unchanged and now documented as the HOST-LEVEL handle (`/status`, `/admin/setup`). Both feed views (`/{indexer}/feed`, `/{indexer}/canonical`) read through `entry.db`; the ingest routes hold no handle at all and never did (their durable writes go through the host-supplied appender/recorder/store bound to that name's database), and the JSDoc now says so. `singleContextEntry(db, ingestion)` takes the handle first, `indexerRegistry` takes entries, and `indexerEntryOn(db, holds)` is new for a host holding a `ReceivingIndexer` (core knows no database, so it cannot carry one). Call sites updated: `etherfold index`, the Node/Worker/fetcher hosts and every harness. `CONTEXT.md`, both READMEs and the core container docstring follow.

**The guard (`packages/server/test/twoNamedIndexers.test.ts`, written before the production change).** Two named indexers with identical sources, contracts, stream config and processor (so the stream digest and the wire context cannot tell them apart), on a host whose own `getDB` is a **third** database holding neither one's rows, so any forgotten discriminator answers with nothing rather than something plausible. It asserts end to end: the routes (both views, both ingest routes), the store, the stored stream, the cursor, the reorg counters and the generation registry; a cap reached in one refusing in that one only; and a `DROP` of one database with no `WHERE` clause anywhere, after which the other is still complete and READABLE (asserted by reads). Verified the guard bites by temporarily reverting the feed to the host handle (alpha's feed went empty).

**Gate:** `format:check`, `check:adr`, `build`, `typecheck`, `test` all green (254 server tests, whole workspace green). `changeset status --since=main` reports red only locally because the new changeset file is still untracked (changesets uses `git diff`, which excludes untracked files); verified in a scratch clone outside the repo that once committed it reports exactly `@etherfold/server` minor plus the five patch bumps.

**Observation captured (not fixed):** `work/notes/observations/the-sqlite-processor-convenience-cannot-take-a-table-namespace.md` — `VersionedStateProcessorOptions` is a `Pick` that omits `tableNamespace`, so the SQLite convenience class cannot put two generations in one database; the shipped entity-path assembly can, so nothing shipped is wrong.

## Decisions

- **`IndexerRegistryEntry.db` is REQUIRED rather than optional-with-fallback-to-`getDB`.** Chosen because it makes the isolation structural: a host adding a second named indexer cannot omit the handle and silently inherit the first one's rows, because there is nothing to omit, and that is the "fails loudly if a discriminator is forgotten" property the task asks for. The alternative (optional `db`, defaulting to the host handle) keeps every existing host compiling untouched, but it is a default where a tenant-keyed handle is required, which is the same hazard class ADR-0036 forbids for the NAME. Third alternative considered and rejected: widening `getDB(c, name?)`, which the task explicitly rules out and which would give one function two meanings (host-level and tenant-level). *Touches:* every host that passes `getIndexer` (`etherfold index`, `@etherfold/platform-nodejs`, the Worker docs, six test harnesses), and it is a breaking type change covered by the changeset.
- **`indexerRegistry` now takes ENTRIES, and `singleContextEntry` takes the db first; `indexerEntryOn` is the new second builder.** A registry keyed by name must carry each name's handle, and I preferred `Record<string, IndexerRegistryEntry>` over `indexerRegistry(db, Record<string, LogIngestion>)` (which would have made colocation the easy path and one-database-per-name the awkward one, inverting ADR-0053) and over `Record<string, {db, ingestion}>` (which could not register a generation container). `indexerEntryOn` exists because `ReceivingIndexer` can no longer be registered as the entry itself: it answers the two questions but `@etherfold/core` knows no database by design. *Touches:* every `getIndexer` call site above, and the `@etherfold/server` public API.
- **Colocation (two names, one handle) stays expressible and is now explicit.** ADR-0053 keeps `_emissions.indexer` and the registry's name column for exactly that future, and two existing suites (`emissionStream`, the feed harness) assert what those columns do. Rather than forbid it, a host passes the same handle twice and the harnesses say why in a comment. The alternative (refusing a duplicate handle) would have deleted a documented future capability to enforce a rule the type cannot check across requests anyway.
- **The guard test gives a SUCCESSOR generation its own state handle instead of a table namespace inside the tenant database.** ADR-0053's namespace form is asserted over a real shared handle in `packages/cli/test/aChangedContextCreatesASuccessor.test.ts`, and it is not reachable from `@etherfold/server`'s dependency set (see the observation above); adding `@etherfold/processor-entities` + `@etherfold/state-store-sqlite` as devDeps purely for test convenience, or widening another package's options type, both grow past this task's fence. The *opening* fold still folds into the named indexer's own database, so the delete assertion remains a real claim: everything that name holds, state included, goes with the one `DROP`. *Touches:* nothing shipped; it bounds what this file proves, and the comment says where the other half is proved.


wighawag:

1. **A host can register several named indexers, each resolving to its own database** — met, via `IndexerRegistryEntry.db` and the two builders (`singleContextEntry`, `indexerEntryOn`).
2. **Two named indexers identical in everything but their name are fed independently — state, cursor, stored stream, reorg counters and generation registry** — met, and all five are asserted separately rather than as one blanket claim. The setup is the sharp one the criterion asks for: the two index ONE stream identity, so nothing but the name and its database tells them apart.
3. **Deleting everything in one leaves the other complete and READABLE, asserted by reads** — met, and thoroughly: the fold, both feed views, the cursor, the counters, the registry and the canonical pointer are all read back after the drop and compared to what they were before, plus the negative half (the deleted name no longer answers `200`).
4. **A cap reached in one refuses in that one only** — met.
5. **Routes acting on one name use that name's handle; host-level surfaces keep working** — met, and this is the crux of the diff: `/status` and `/admin/setup` stay on `getDB` (facts about the DEPLOYMENT), while every tenant-scoped read goes through `entry.db`. The distinction is now documented on `ServerOptions.getDB` itself, which is where someone would otherwise get it wrong.
6. **Unknown name still `404`, no registry still `501`** — met, asserted on ingest and on both views.
7. **Changeset for the registry/options shape change** — met.
8. **Tests in the repo's style** — met.

I made the feed routes read through the HOST handle instead of the tenant's (`options.getDB(c)` in place of `entry.db`) — the exact cross-tenant read this task exists to make impossible. `are fed independently: the state, the cursor and the stored stream stay in their own database` goes red; the rest stay green. The isolation is pinned by a test that actually exercises a cross-tenant read rather than by a table listing.

One honest note on the shape of that pin: the delete case did NOT go red under the same defect, because with both reads redirected to the host handle the before/after comparison passes vacuously. It is still a correct and valuable test of the criterion it is written for; it just is not a second pin for this particular defect. Not a block.

### The design call I checked hardest, and agree with

**`IndexerRegistryEntry.db` is REQUIRED rather than optional-with-fallback.** This is a breaking type change touching every host and six test harnesses, and the cheaper option (optional `db` defaulting to the host handle) was available and would have left everything compiling. Taking the breaking one is right, and for the reason the task itself frames the whole story around: with a fallback, a host adding a second named indexer can OMIT the handle and silently inherit the first one's rows. That is a default where a tenant-keyed value is required, which is the same hazard class ADR-0036 forbids for the NAME. With `db` required there is nothing to omit, so the isolation is structural rather than a discipline. Widening `getDB(c, name?)` was correctly rejected too — the task rules it out, and it would give one function a host-level and a tenant-level meaning.

**Colocation stays expressible.** Two names on one handle is still allowed and now explicit, rather than being forbidden. Right call: ADR-0053 keeps `_emissions.indexer` and the registry's name column precisely for that future, two existing suites assert what those columns do, and refusing a duplicate handle would delete a documented capability to enforce a rule the type cannot check across requests anyway.

### Scope fence, honestly bounded

The guard test gives a successor generation its own state HANDLE rather than a table namespace inside the tenant database. That narrows what this one file proves, and the done record says so plainly and says where the other half is proved — ADR-0053's namespace form is asserted over a real shared handle in `packages/cli/test/aChangedContextCreatesASuccessor.test.ts` (landed in #74). The reason is a genuine dependency boundary rather than convenience: the namespace form is not reachable from `@etherfold/server`'s dependency set, and pulling `@etherfold/processor-entities` and `@etherfold/state-store-sqlite` in as devDeps purely for test convenience would grow past this task's fence. The opening fold still folds into the named indexer's own database, so the delete assertion remains a real claim. Correctly bounded and correctly declared.

### Observation filed by the build (not fixed in place, correctly)

`work/notes/observations/the-sqlite-processor-convenience-cannot-take-a-table-namespace.md`: `VersionedStateProcessorOptions` is a `Pick` of `'retention' | 'finalityDepth'`, so `tableNamespace` cannot be passed through the SQLite convenience class, and two generations built that way over one handle would land on the same tables. Nothing shipped is wrong — the entity-path assembly the CLI actually uses takes it — and the narrow `Pick` simply predates the namespace. Good catch, right destination, and worth a follow-up task rather than a silent widening here.

Merging.

PR #78, opened 2026-09-06
feat(the-canonical-pointer-moves-back-without-re-ingesting): The canoni…

wighawag:
Task: `work/tasks/done/the-canonical-pointer-moves-back-without-re-ingesting.md`

The task is implemented and the repo's gate (`pnpm build && pnpm typecheck && pnpm test`, plus `format:check` and `check:adr`) is green.

## What I built

**Drift check first.** All three blockers landed as described (`a-changed-context-creates-a-successor-instead-of-clearing`, `one-registry-entry-holds-several-live-wire-contexts`, `the-rebuild-replays-the-local-stream-in-bounded-chunks`), the shared promotion trigger and arming are in `generation/promotion.ts` consumed by both containers, the feed resolves the canonical generation once per request, and ADR-0056's answer to open question 2 (the retired generation keeps folding) holds. No stale premise; I built on what the rebuild task left.

**Core (`@etherfold/core`, `receivingContainer.ts`).** Two gaps stood between "promotion works" and "the way back is real":

- `ReceivingIndexer.promote` required this container to hold a FOLD for the target, which makes the ORDINARY revert impossible: a host redeployed with the new processor holds only the new fold, and the generation an operator wants back is in the durable registry. It now moves the pointer to any REGISTERED generation, with the registry's own `UnknownGenerationError` as the refusal (reads resolve the pointer to a table namespace, ADR-0053, so the reverted-to generation answers with no engine).
- A backwards move could DROP the generation it moved away from under `dropOnPromotion`. The container now tracks whether the pointer has EVER named a held fold (the chain-facing container's `everCanonical`, as a set) and a revert drops nothing, per ADR-0046.

**Server (`@etherfold/server`).** The operator affordance answered by the task's open question 1: `GET`/`POST /{indexer}/admin/canonical-generation`, guarded by a NEW `ADMIN_TOKEN` that fails closed, sharing ONE constant-time comparison with the ingest guard (`api/auth.ts`) at a separate credential. `IndexerRegistryEntry` gained two optional questions (`generations()`, `promote(id)`) that `ReceivingIndexer` already answers and `indexerEntryOn` forwards; a host with no registry answers `501 generations-not-held`. `POST /admin/setup` is untouched.

**Tests.** `packages/core/test/theCanonicalPointerMovesBack.test.ts` (7) at the container seam and `packages/server/test/theCanonicalPointerMovesBack.test.ts` (11) over HTTP, written before the production code. They assert the old answers come back byte for byte from the generation's own state, the stored stream and every fold are untouched across the revert, the successor is not re-promoted (including across a new batch), the reverted-from generation is still registered and promotable, the once-per-request pointer resolution on both feed views, the fail-closed guard (including refusing the INGEST token), and the refusals. Plus ADR-0057, a changeset, the server README route table, and a `CONTEXT.md` line on the **canonical pointer**.

## Decisions

**The revert is `POST /{indexer}/admin/canonical-generation`, ONE path with a `GET` beside it, and there is no `revert` verb.** The task's open question 1 fixed the shape (an authenticated admin HTTP route with a new `ADMIN_TOKEN`); what I chose is the NAME and the granularity. `/{indexer}/admin/...` because a pointer belongs to one named indexer while `/admin/setup` is host-level, and `canonical-generation` rather than `canonical` because `/{indexer}/canonical` already MEANS the canonical VIEW of the emission stream (coherence check: reusing the word would have made it mean two things one path apart). A single verb in both directions because moving the pointer IS promotion and moving it back IS revert (`CONTEXT.md`), so two verbs would be two names for one record write; the direction is DERIVED (has the pointer named this generation before), not declared. The `GET` exists because a feed advertises its generation as an opaque digest, so without a listing an operator has nothing to match it against and cannot name a target. Alternatives considered: a CLI flag or a sixth verb (cannot serve a Worker; the command set is pinned at five), and accepting the opaque digest as the target (rejected as the sole form: the registry keys on `{stream, processor}`, so the `GET` reports both forms per generation and the `POST` takes the pair). Touches: `IndexerRegistryEntry` (every host's resolver), `Env`, and the CLI task below. Recorded in ADR-0057.

**`ReceivingIndexer.promote` no longer refuses a generation this container holds no FOLD for; the refusal becomes the registry's `UnknownGenerationError`.** This reverses a refusal the rebuild task introduced. Requiring a held fold would mean a revert could only be performed by a process first rebuilt with the OLD processor, which is the re-index the design exists to remove, and it contradicts the receiving container's own rule 1 (it does not refuse a canonical generation it holds no engine for) applied to the write side. Alternative considered: a second method for the unheld case, rejected as a second pointer-move path that would drift from this one. Touches: any caller matching on the old "holds no fold" error text (there is none in the tree), and the admin route's `400 unknown-generation`. Recorded in ADR-0057.

**A move to a target this container holds no fold for counts as a REVERT for drop purposes.** With no fold there is no in-memory record of whether the pointer ever named it, so the direction is unknowable; declining to drop is the safe side (retaining is the default, and the cost is bytes rather than an unrecoverable delete). Touches: `dropOnPromotion` deployments, which now keep a generation in one more case than before.

**The CLI is deliberately NOT touched, and no `--admin-token` flag exists.** `run`/`index` host the same app, so the route is already present in their process; it answers `501 generations-not-held` because their registry entry holds one `StreamBuilder` and no container, which is `the-cli-and-the-server-hold-generations-the-same-way`'s to change (it owns the CLI's generation container, and there is nothing to revert TO on the CLI until it lands). `ADMIN_TOKEN` reaches those processes from the environment through the Node host's `{...process.env, ...options.env}` merge, so no new command INPUT is introduced and ADR-0048's axis is untouched. Alternative considered: adding the flag now, rejected as a user-visible input for a capability that command cannot yet answer.

**Two refactors of existing files, both to avoid a second copy of one rule.** (1) The timing-safe secret comparison moved out of `api/ingest.ts` into `api/auth.ts`, parameterised by WHICH credential, so "is a token accepted" has one answer for two surfaces; `platformAgnostic.test.ts`'s source-reading guard was updated to assert the shape where it now lives and that BOTH surfaces reach it at their own credential. (2) `rebuild.test.ts`'s ~330-line world (stored stream + coverage claim, registry substrate, a store per generation namespace) moved to `packages/core/test/utils/receivingWorld.ts` and is now shared with the revert suite, on the precedent of `streamCacheWorld.ts` — a second copy would be a second definition of what a reorg and a quiet range mean in a test.


wighawag:

1. **Moving the pointer back makes the previous generation answer reads again, from its own state, with nothing the successor wrote in it** — met at both seams, and asserted as open question 2 requires: the honest property (the old processor's answers over the CURRENT chain), not the unassertable "the exact answers from the promotion instant".
2. **The revert makes ZERO chain calls and appends NOTHING to the stored stream** — met, asserted as "one row of bookkeeping".
3. **The automatic promotion policy does not move the pointer forward again on the next advance, over the SHARED trigger with no second one added** — met, and checked across further advances rather than one. No second trigger appears in the diff.
4. **A read resolving through the pointer holds one generation for its whole unit of work** — met, asserted directly as "a read never straddles a move".
5. **The generation reverted FROM is still registered, still has its state, and can be promoted again** — met, plus the case that would quietly ruin it: it is NOT dropped by a backwards move even under `dropOnPromotion`.
6. **Question 1 implemented, refusing clearly when asked to point at a generation this host does not hold** — met, and the refusal surface is better than the criterion asked for: it names every generation the host DOES hold, refuses a body naming no generation, and distinguishes "this host holds no generations" from "this route does not exist".

I made `authorizedWith` fail OPEN when no credential is configured — the classic inversion of exactly the property this task promises. Both `refuses EVERY caller when no ADMIN_TOKEN is configured` AND the pre-existing ingest case `refuses everyone when no token is configured, rather than letting everyone in` go red, with 45 others green. That both surfaces are caught by one defect is the refactor paying for itself: there is now one answer to "is a token accepted", so it cannot rot on one surface while staying right on the other.

### The security shape, which I looked at carefully because it is a new credential

- **A separate `ADMIN_TOKEN`, not the ingest token.** Right, and for the reason stated: the ingest credential is handed to a log shipper and guards the WRITE path; letting it also change which generation answers reads would hand a fetcher control-plane authority.
- **Fails closed when unset**, matching the ingest guard's existing rule, and the message names the variable so an operator is not staring at a 401 they configured themselves.
- **An unauthenticated caller cannot enumerate the names this host holds.** This is a test I did not expect and was glad to see — the refusal that names every generation an operator could target is genuinely useful and would have been an information leak if it were reachable before auth.
- **`/admin/setup` is left UNAUTHENTICATED and that is flagged rather than quietly retrofitted.** Correct scope discipline: the task introduces the first admin credential, it does not re-litigate an existing bootstrap choice. Worth a follow-up decision, and it is now written down where someone will find it.

- **One path, `POST /{indexer}/admin/canonical-generation`, with no `revert` verb.** Moving the pointer IS promotion and moving it back IS revert, so two verbs would be two names for one record write, and the direction is DERIVED rather than declared. The name avoids a real collision: `/{indexer}/canonical` already means the canonical VIEW of the emission stream, and reusing `canonical` would have made one word mean two things one path apart.
- **The `GET` beside it earns its place.** A feed advertises its generation as an opaque digest, so without a listing an operator has nothing to match it against and literally cannot name a target.
- **Reversing #76's "promote refuses a generation this container holds no fold for".** This is the call I scrutinised most, since it undoes something one PR old. It is right: requiring a held fold would mean a revert could only be performed by a process first rebuilt with the OLD processor, which is precisely the re-index this whole spec exists to remove, and it contradicts the receiving container's own rule 1 (it does not refuse a canonical generation it holds no engine for) applied to the write side. The refusal becomes the registry's `UnknownGenerationError`, so there is still exactly one refusal, and the test `reverts from a host restarted with the NEW processor alone` pins the case that motivated it.
- **A move to a target with no held fold counts as a REVERT for drop purposes.** With no fold there is no in-memory record of whether the pointer ever named it, so the direction is unknowable and declining to drop is the safe side: the cost is bytes, the alternative is an unrecoverable delete.
- **The CLI is deliberately not touched and no `--admin-token` flag is added.** `ADMIN_TOKEN` already reaches those processes through the Node host's env merge, so no new command INPUT appears and ADR-0048's axis is untouched; the route answers `501 generations-not-held` there until `the-cli-and-the-server-hold-generations-the-same-way` gives the CLI a container. Adding a user-visible flag for a capability the command cannot yet answer would have been the worse choice.

### On the two refactors

Both are justified by "avoid a second copy of one rule" and both check out. I specifically verified the larger one is a pure extraction: `rebuild.test.ts` drops ~314 lines to a shared `receivingWorld.ts`, and its list of `it()` names is BYTE-IDENTICAL before and after (18 cases either side), so no coverage was lost in the move. The `platformAgnostic` source-reading guard was updated to assert the shape where the helper now lives and that both surfaces reach it at their own credential, which keeps that guard meaningful rather than stale.

Merging.

PR #79, opened 2026-09-06
feat(a-rebuild-in-progress-is-never-an-empty-answer): A rebuild in prog…

wighawag:
Task: `work/tasks/done/a-rebuild-in-progress-is-never-an-empty-answer.md`

Done. Full gate green (`build` + `typecheck` + `test` + `format:check` + `check:adr`), clean tree.

## What I built

**A rebuild in progress is visible on `/status`** — inside the field ADR-0047 reserved for it, no endpoint added.

- `packages/server/src/cursor.ts`: the reporter now hands over the envelope's two slots explicitly (`StatusReport` = `{value?, generations?}`), and `StatusCursor` carries `generations` beside `value` on **both** branches. `reported` keeps its exact meaning (is there a *cursor*), which is what lets a **first build** — generations to report, no cursor yet — report both halves instead of losing one. An entry is `{generation, canonical, follows, value?}`: the digest the feed already advertises, which one answers reads, whether a **rebuild** is what advances it (`follows`, ADR-0044), and how far it has got (its own sync cursor, which is the rebuild's only checkpoint, ADR-0056). The server fixes every field but `value`, so the size bound is structural rather than a plea in a doc comment.
- `packages/cli/src/cursorReport.ts`: `readStatusReport({folds, canonical})`, one cursor read per generation, the canonical one's answer reused as the top-level `value`. `run` and `index` now report their one fold as one generation. A fold whose cursor cannot be read yet is still an *entry*, with no `value` — never zeroed, never dropped.

**A read that would otherwise lie refuses** — `503 no-canonical-generation` (new **ADR-0058**), through the *same* canonical-generation resolution both feed views already do, now lifted below the routes into `resolveCanonicalGeneration` (`api/resolve.ts`). `IndexerRegistryEntry.canonicalGeneration()` may answer `undefined`; the admin route *reports* that state rather than refusing it.

**Tests**: `packages/server/test/aRebuildInProgressIsNeverEmpty.test.ts` (the envelope + both views' refusal, over a read-tier entry resolving the durable pointer) and `packages/cli/test/aRebuildInProgressIsVisible.test.ts` (a real container, real folds, real stored stream: progress advancing across chunks with two reads either side, the canonical flag moving exactly once, and the envelope staying small). Written before the production code at both seams.

Also: changeset, `CONTEXT.md` glossary updated (query layer + the two views), and one out-of-scope signal captured in `work/notes/observations/`.

## Decisions

- **The reporter's return shape is two NAMED slots (`{value?, generations?}`), a breaking change to `getCursorReport`, rather than sniffing the reported value for a `generations` key.** ADR-0047 forbids the server parsing what it reports verbatim, and a host may legitimately put a key called `generations` *inside* its cursor summary — `packages/server/test/server.test.ts` has asserted exactly that shape since the seam landed. Sniffing would make a legal report change how the envelope is built. Alternatives considered: (a) sniffing (rejected, above); (b) a SECOND injected option `getGenerationReport` (rejected: ADR-0047 says the dimension grows *inside* the field a single reporter fills, and two reporters would be two degrade paths for one page). It touches every host that injects a reporter — `run`, `index`, `platforms/nodejs` tests — and is documented in the changeset with the one-line migration. Nothing is published (CONTEXT `## Conventions`), so it costs a changeset.
- **`reported` keeps its exact meaning and `generations` sits beside it on the `reported: false` branch too.** A first build is the load-bearing case: it has generations and no cursor. The alternative — `reported: true` with an absent `value` — would silently re-mean the flag every existing reader branches on.
- **The per-generation flag is `follows`, the established word (ADR-0044 / CONTEXT "follower"), not a new `rebuilding`.** Coherence check: "follower" already means "advanced by re-folding the stored stream", which is exactly the condition under which the entry's `value` is rebuild progress. A second near-synonym would fork the vocabulary.
- **The refusal lives BELOW the routes** (`resolveCanonicalGeneration` in `api/resolve.ts`), which the task invited me to record. Both feed views resolve through it, so there is one place a read decides which generation answers, and a surface added later inherits it. It touches `one-registry-entry-holds-several-live-wire-contexts`'s once-per-request resolution (I replaced it in place rather than adding a second) and will be inherited by `two-named-indexers-never-touch-each-others-data`'s handles unchanged.
- **`503 no-canonical-generation`, and the refusal is scoped to WHICH generation answers, never to how far it has got.** A canonical generation mid-backfill still serves a short page and a cursor: refusing on progress would mean a feed could not be followed until its backfill finished. Alternatives: `409` (rejected — ADR-0004 pins it as the one *resumable* refusal, and there is no position to resume from), `404`/`501`/`400` (rejected — the name resolved, the host can serve feeds, and nothing about the request is wrong). Recorded as **ADR-0058** because it is a published refusal on a public read surface with four plausible codes.
- **The new wire field is `building` (opaque digests).** New name on a new refusal; `generations` was rejected because that key already carries objects on the admin route and on `/status`, and a third shape under one name is the muddle worth avoiding.
- **`IndexerRegistryEntry.canonicalGeneration(): Promise<GenerationId | undefined>`**, and the ADMIN route *reports* the absent case (`canonical` absent, generations still listed) rather than refusing it — the opposite of what a READ does with the same answer. A read served from nothing is a wrong answer; "nothing answers reads yet, here is what is registered" is what an operator opened that route to see.
- **`serve` is NOT wired to a registry here, so the criterion's parenthetical "(and by a `serve` process over such a database)" is satisfied structurally rather than by a running `serve`.** `serve` registers no named indexer today (it refuses `--indexer` outright, so it has no name to register under) and its feed answers `501`; making it resolve the canonical pointer is `the-cli-and-the-server-hold-generations-the-same-way`'s own acceptance criterion. My test drives exactly the read-tier entry that task will build (`generationRegistryPortOnSQL(db, name).read()`), so the refusal is already the one it will inherit — nothing about it will need rewriting there.


wighawag:

1. **`/status` reports one entry per generation: which is canonical, and how far a rebuild has got** — met.
2. **Inside the existing cursor envelope: no new endpoint, no new top-level field beyond what ADR-0047 reserves, and a host with no reporter still has no `cursor` field** — met, all three asserted, including the negative "invents no `generations` key on a host whose reporter names none".
3. **The envelope stays small and JSON-serialisable, no unconfirmed window, no raw serialized cursor** — met, and asserted concretely rather than as a vibe: "four numbers a generation, no window and no serialized cursor".
4. **The rebuild's progress ADVANCES across chunks, asserted either side of a chunk** — met, plus the complementary case that the canonical flag moves exactly once, when the rebuild finishes.
5. **A read against an indexer with no canonical generation is REFUSED, naming what has not caught up, never answered empty, through the SAME resolution the feed already does** — met. The resolution is now one function below both routes rather than duplicated, and it REPLACED #75's once-per-request read rather than sitting beside it, so there is still exactly one place a read decides which generation answers.
6. **A reporter that throws or returns nothing still yields absent-with-a-reason, never fails the request or changes `healthy`** — met, and extended to the case a generation entry cannot be serialised, which degrades rather than taking the page down.

I made `resolveCanonicalGeneration` fall back to a placeholder generation instead of refusing — literally "serve the empty page of a build in progress", which is the one thing ADR-0058 exists to forbid. Three cases go red: both feed views' refusals and the naming case, with 7 green. The refusal is pinned on both surfaces rather than on one and inherited by assumption.

### The shape decisions I checked and agree with

- **Two NAMED slots (`{value?, generations?}`) rather than sniffing the reported value for a `generations` key.** This is the right call and the reasoning is exact: ADR-0047 forbids the server parsing what it reports verbatim, and a host may legitimately put a key called `generations` INSIDE its cursor summary — `server.test.ts` has asserted that very shape since the seam landed, so sniffing would let a legal report change how the envelope is built. Rejecting a second injected `getGenerationReport` was also right: ADR-0047 says the dimension grows inside the field one reporter fills, and two reporters would be two degrade paths for one page.
- **`reported` keeps its exact meaning, and `generations` sits beside it on the `reported: false` branch too.** The load-bearing case is a FIRST BUILD: generations, no cursor. Re-meaning the flag every existing reader branches on would have been the silent-breakage option.
- **`follows`, not a new `rebuilding`.** "Follower" already means "advanced by re-folding the stored stream", which is exactly when the entry's value is rebuild progress. No near-synonym forked into the vocabulary.
- **`503 no-canonical-generation`, scoped to WHICH generation answers and never to how far it has got.** The scoping is the important half: a canonical generation mid-backfill still serves a short page and a cursor, so refusing on progress would mean a feed could not be followed until its backfill finished. The code choice is argued against all four plausible candidates and `409` is correctly ruled out as ADR-0004's one RESUMABLE refusal, with no position to resume from here.
- **The admin route REPORTS the absent case where a read REFUSES it.** Opposite treatments of the same answer, and both right: a read served from nothing is a wrong answer, while "nothing answers reads yet, here is what is registered" is exactly what an operator opened that route to see.

### Two scope boundaries, both declared rather than glossed

1. **`serve` is not wired to a registry here**, so the criterion's parenthetical "(and by a `serve` process over such a database)" is satisfied structurally. This is a real sequencing boundary and not a dodge: `serve` refuses `--indexer` outright today, so it has no name to register under, and making it resolve the canonical pointer is `the-cli-and-the-server-hold-generations-the-same-way`'s own acceptance criterion. The test drives exactly the read-tier entry that task will build, so the refusal it inherits is already this one. **I am checking this at task 10's Gate-3.**
2. **`ReceivingIndexer.canonicalGeneration()` still falls back to its opening fold**, so a container-backed host cannot reach the new refusal — captured as an observation rather than changed, because changing it would change what every container-backed host ANSWERS, which is outside a task fenced to the read path.

### Observation filed by the build (a sharp one)

`work/notes/observations/the-receiving-container-answers-its-opening-fold-where-the-registry-names-no-canonical.md`. The reachable case is narrow but genuinely real: `openGenerationRegistry.canonical()` resolves the pointer against the RECORDS and answers `undefined` when the pointer names a generation whose record has gone (another process deleting it, a half-written substrate). A container-backed host then serves reads from its opening fold, which may not be the generation the pointer named, where a read tier over the same rows would refuse. The note is honest that the existing JSDoc's justification for the fallback predates the answer being expressible at all, and it lays out two candidate fixes without picking one. Exactly the right disposition, and it goes in my drive report as task-worthy.

Merging.

PR #80, opened 2026-09-06
feat(the-cli-and-the-server-hold-generations-the-same-way): The CLI hol…

wighawag:
Task: `work/tasks/done/the-cli-and-the-server-hold-generations-the-same-way.md`

All checks are green. Here is my report.

## What I built

The five CLI commands now hold generations the way the server does, asserted at the commands over one fixture chain.

**`packages/cli/src/folding.ts`** — the shared folding assembly now opens a `ReceivingIndexer` (the chain-free generation container) instead of a bare `StreamBuilder` over one un-namespaced store. It supplies the three things the container cannot have because `@etherfold/core` knows no database: the registry substrate (`generationRegistryPortOnSQL`, with `dropState` dropping the generation's namespace), the per-generation **table namespace** the state folds into (ADR-0053, named from `{stream digest, processor version hash}` before the processor exists), and the stored stream's two ends (`emissionAppenderFor` + `storedEmissionReplaySource`). It states `SERVER_GENERATION_CAPS`. Split into `openFoldingDatabase` (handle + fixed tables) and `openFolding` (container), because a generation's tables are named from the source, so the store can no longer be built before the source is resolved — the database still is, so a database this command cannot open is still refused before `eth_chainId`.

**`run` / `build` (`src/index.ts`, `src/run.ts`)** — both fold through that container. `run` exposes it (`RunningIndexer.container`), reports one `/status` entry per generation held (`foldingStatusReport`), and advances any FOLLOWER by one bounded chunk in the gap its fetch loop already waits between cycles, so a successor added in-process catches up and the pointer moves on its own. `build` schedules none: it opens with one fold and exits, so it never adds a second and never promotes, and a second `build` over the same inputs resolves the same generation. `run` still registers no named indexer, so its namespaced routes still answer `501`.

**`index` (`src/indexCommand.ts`)** — same container; the name it registers now resolves to it, so the entry answers all four questions and `GET`/`POST /{indexer}/admin/canonical-generation` (ADR-0057) answer here instead of `501 generations-not-held`.

**`serve` (`src/serve.ts`) + `src/readTier.ts`** — the read tier resolves the canonical pointer over the database it was pointed at, with no name (it learns it from the rows), no processor and no registry open (opening one sweeps, which is a write), and says which generation answers beside the URL. `canonicalGenerationIn` / `canonicalStateNamespaceIn` / `heldGenerationsIn` are exported, thin over a new read-only `readHeldGenerations(db)` in `@etherfold/server`, which owns those tables.

**Tests** — `packages/cli/test/equivalence.test.ts` extended (not duplicated): the reader now resolves the pointer; a new `build`-versus-`run` block compares the registered generation, the canonical pointer, the state namespace, the stored stream and the reorg counters over one fixture including a reorg; a new block adds a successor to a live `run` and watches it promote while the incumbent answers throughout; the read-tier block asserts what `serve` resolved. Plus the admin-surface test on `index`. Docs: `packages/cli/README.md`, an ADR-0057 consequence update, and the three test headers that pointed at this task as future work.

Also: a changeset, and one out-of-scope observation captured (`work/notes/observations/server-readme-says-the-server-holds-no-generations.md`).

Gate: `format:check`, `check:adr`, `build`, `typecheck`, full `test` (all packages/platforms/examples) green. `changeset status --since=main` cannot see the changeset until it is committed (it diffs tracked files) — the file is there, untracked, per the no-git rule.

## Decisions

- **`--no-auto-setup` against an unmigrated database is now a refusal to START, not a retry loop.** A fold registers its generation before it reads or writes anything, and the registry and pointer are rows, so there is nowhere to register it. Alternatives considered: (a) apply the schema anyway — rejected, it re-means the flag, which says somebody else owns those migrations; (b) open the container lazily on the first batch — rejected, it loses "a cap refuses at start-up where an operator reads it" and turns a configuration fault into a `500` on somebody's ingest. Touches: `run` and `index` (the two commands that take the flag), and the equivalence test that used an unmigrated database to exercise ADR-0052's no-hole rule — that test now drops `_emissions` under a running, migrated process instead, and a new test pins the refusal. `build` is unaffected (it always migrates).
- **`run` schedules the rebuild chunk in the gap its fetch loop already waits between cycles.** ADR-0022 says the host schedules and invents no cadence; `run` is the host, and the loop's own backoff is its clock. Alternatives: a separate timer/pump (invents a cadence and a second writer thread against one handle), or fire-and-forget inside `onReport` (concurrent writes on one libSQL handle). Consequence, documented at the site: a process still catching up flat out (`CATCH_UP_DELAY_MS=0`) advances followers only once it reaches the tip. Touches nothing outside `run`; `build` schedules none, which is where "holds exactly one generation" becomes structural.
- **`run` still registers no named indexer, so the pointer's HTTP affordance lands on `index` only.** Registering the container on `run` would open its ingest routes to a remote sender — the second writer that command exists without. So `run`'s add-and-promote is in-process (the container it exposes, plus the automatic on-catch-up move) while the operator's `admin/canonical-generation` route answers on `index`. Touches ADR-0057 (amended in its consequences) and `packages/cli/README.md`.
- **`serve` resolves the pointer and REPORTS it in its startup lines; it injects no `/status` reporter.** The read tier's only HTTP surface is `/status`, and its cursor field is absent by design ("a read tier owns no store and is given none", pinned by the equivalence test). A reporter here could only claim `generations` with a `follows` flag it cannot know, so the resolution is said out loud instead and exported as `canonicalGenerationIn` / `canonicalStateNamespaceIn` for the readers that actually open the state. Alternative considered: report generation entries on `/status` with `follows: false` — rejected as a claim the read tier cannot make. Touches: what a reader of an etherfold database must do (two steps), which is now documented and exported.
- **A database holding SEVERAL named indexers refuses `canonicalGenerationIn` without a name rather than guessing.** Each name has its own pointer; picking one would be choosing a tenant for the caller. On every shape this repo builds a database is one named indexer, so the refusal is unreachable today. Touches: nothing else — `serve` reports every name it finds and refuses nothing.
- **The CLI states `SERVER_GENERATION_CAPS` (4 generations, 2 streams) and adds no flag for it.** The command set is pinned at five verbs with a fixed input table; a new flag would be a new user-visible input for a number that constant already argues for. Touches the command table (deliberately unchanged); an operator who hits the bound deletes a generation, and the refusal names what to delete.
- **A database written by the pre-generation CLI is not resumed.** Its state sits in tables nothing points at, and its stored stream is claimed by no registered generation, so the registry's own sweep collects it on first open (documented behaviour of `listStreamDigests`). Nothing is published and no disk holds state this project must preserve, so that costs a re-index rather than a migration.


wighawag:
APPROVE ✅ — Gate-3 (conductor's diff-vs-criteria review). CI `verify` COMPLETED/SUCCESS, mergeStateStatus CLEAN. Tenth and last task of `the-server-and-cli-hold-generations-too`.

1. **`build` creates exactly one generation, never adds a second, never promotes; twice over the same inputs resolves the SAME generation** — met, and structurally rather than by discipline: `build` schedules no rebuild chunk at all, which is where "holds exactly one" stops being a rule and becomes a property of the command.
2. **A `build`-produced database is indistinguishable from a `run`-produced one on the generation axis, through ONE fixture** — met, and this is the best test in the PR. One fixture, both shapes fed the same two chain states INCLUDING a reorg, then compared on every axis the criterion names and then some: the same registered generation, the same canonical pointer, the same state namespace, the same tables under it, the same reads over them, the same stored stream, and the same reorg counters. "Indistinguishable" is asserted as a claim about all of them at once rather than one axis at a time.
3. **`run` can add a successor and promote it in-process, and `index` folds through the same container** — met, driven end to end: the incumbent answers until the rebuild catches up, then the successor does.
4. **`serve` answers over a database written elsewhere by resolving the canonical pointer, holding no processor and registering nothing; still refuses `--indexer`** — met. See the note below on how this interacts with #79.
5. **The name still defaults on `run` and `build` and is still required on `fetch` and `index`** — met, ADR-0052's axis untouched.
6. **No sixth command, no new default command, bare `etherfold` still prints help** — met, and the caps are stated as a constant rather than becoming a flag, which keeps the five-verb input table exactly as pinned.

I removed the `--no-auto-setup` refusal so an unmigrated database degrades instead of refusing to start. `refuses to START against a database it may not migrate, naming what is missing` goes red, 7 others green. The new behaviour is pinned, not merely described.

### The behaviour change I looked at hardest

**`--no-auto-setup` against an unmigrated database now refuses to START rather than retrying.** This is a real, user-visible change to a flag's behaviour, so it deserved scrutiny. It is forced rather than chosen: a fold registers its generation before it reads or writes anything, and the registry and pointer are ROWS, so against an unmigrated database there is literally nowhere to register it. Both alternatives are worse and are argued rather than waved at — applying the schema anyway re-means a flag whose entire content is "somebody else owns these migrations", and opening the container lazily on the first batch loses "a cap refuses at start-up where an operator reads it" and converts a configuration fault into a `500` on somebody's ingest.

The knock-on was handled honestly too: the equivalence test that used an unmigrated database to exercise ADR-0052's no-hole rule now drops `_emissions` under a running, migrated process instead, so that rule is still pinned rather than quietly losing its test. `build` is unaffected because it always migrates.

### `serve`, and how it squares with #79

#79 left criterion 5's parenthetical — "(and by a `serve` process over such a database)" — to be satisfied here, so I checked what actually landed. `serve` resolves the pointer, reports which generation answers in its startup lines, and exports `canonicalGenerationIn` / `canonicalStateNamespaceIn` as the two-step resolution any reader over an etherfold database must perform. It still registers nothing, which is what THIS task's criterion 4 requires, so its feed views continue to answer `501` rather than ADR-0058's `503`.

That is consistent rather than a gap, and the property both tasks are protecting holds: a read tier over a database whose first build has not finished never returns an empty page. It says so on startup, and `canonicalGenerationIn` returns `undefined` as a real answer meaning "nothing answers reads here yet". The `501` is the pre-existing, correct answer for a host built with no named indexers, not a degradation. Worth recording plainly in the drive report, because a reader of #79's criterion alone might expect the `503` on `serve` specifically.

- **`readTier.ts` deliberately does NOT go through `openGenerationRegistryOnSQL(...).canonical()`**, because OPENING the registry sweeps every stored stream no registered generation claims — a WRITE. A read tier must not delete a stream on its way to asking which generation answers. This is exactly the hazard #71's Decision 8 flagged forward, caught and avoided at the one place it would have bitten. Good.
- **`run` schedules the rebuild chunk in the gap its fetch loop already waits.** ADR-0022 says the host schedules and invents no cadence; the loop's own backoff IS the clock, so no timer and no second writer thread against one handle. The consequence (a process catching up flat out advances followers only once it reaches the tip) is documented at the site rather than discovered later.
- **`run` still registers no named indexer**, so the pointer's HTTP affordance lands on `index` only. Registering the container on `run` would open its ingest routes to a remote sender, which is the second writer that command exists in order not to have. ADR-0057 is amended in its consequences rather than left to disagree.
- **A colocated database refuses `canonicalGenerationIn` without a name rather than guessing** — picking one would be choosing a tenant for the caller, which is the failure ADR-0036's discriminator exists to prevent. Unreachable on every shape built today, and correct for the shape the columns keep possible.
- **A pre-generation database is not resumed**, and the reasoning is the repo's standing one: nothing is published, no disk holds state this project must preserve, so it costs a re-index rather than a migration.

### Observation filed by the build

`work/notes/observations/server-readme-says-the-server-holds-no-generations.md`: `packages/server/README.md` still describes the generation dimension of the `/status` envelope as future work, but `StatusReport.generations` is built and both `run` and `index` now fill it. Stale prose from before #79 landed, captured rather than fixed in a task fenced elsewhere. It is a one-paragraph docs fix and goes in my drive report as task-worthy.

Merging — this completes all ten tasks of the spec.

PR #81, opened 2026-09-06
feat(decoding-preselects-the-event-by-topic0-instead-of-re-searching-th…

wighawag:
Task: `work/tasks/done/decoding-preselects-the-event-by-topic0-instead-of-re-searching-the-abi.md`

Done. Everything is green (`format:check`, `check:adr`, `build`, `typecheck`, `test`), and `changeset status --since=main` was verified to pass once the files are committed (it only fails locally because the new changeset file is still untracked; confirmed in a throwaway clone with the change committed, exit 0).

## What I did

**Drift check first.** Every premise held: `decodeOnto` did hand `decodeEventLog` the whole per-address event list once per log; ADR-0031's collapse-or-refuse on `topic0` is intact and is what makes the map unambiguous; ADR-0034's unconditional `reparse` is intact. One sub-premise did not hold (anonymous events, see Decisions) but the pure-optimisation contract resolves it unambiguously, so I proceeded.

**Step 1, the pinning test, written before any production change** (`packages/core/test/decodePreselection.test.ts`, 13 tests). Two independent pins over a multi-event, multi-address fixture (2 declared addresses + 1 undeclared, 5 events including an anonymous one and one whose declared order differs from its decoded arg order):
- a **golden table** of `eventName`/`args`/`decodeError`, canonicalised to a string so argument **key order** is pinned, not just values;
- an **oracle**: the whole-ABI algorithm transcribed into the test and run over the same logs, so the check is against the rule rather than a snapshot.

Cases covered: a hit, a real **miss** (a topic0 declared at the *other* address), a hit whose **data fails to decode**, an **anonymous** event, an **undeclared address**, the stale decoded half being dropped, and fetch-path/replay-path agreement. Run against unmodified production code: **10 passed, 3 failed** (only the three that read the not-yet-existing map). After the change: **13 passed**.

**Step 2/3, the map and the fallback** (`packages/core/src/internal/decoding/LogEventFetcher.ts`). `abiEventPerAddressAndTopic` is built once in the constructor from the same de-duplicated per-address lists `decodeOnto` decodes against, anonymous members excluded. `decodeOnto` looks up and falls back: a hit passes `[member]`, a miss passes exactly what it passed before. `parseAllEventsIrrespectiveOfAddresses` is untouched and grows no map (asserted, both directions).

**Step 4, re-measured through the built fetcher.** First run was garbage (load average 11, a `chrome-headless` at 309% CPU); I waited for the machine to quiet and re-ran. Production-to-production, 31,330 real Base logs, all four variants asserted to decode identically:

| | median | per event |
| --- | --- | --- |
| production `reparse`, pre-change whole-ABI search | 1,824 ms | 58.2 µs |
| production `reparse`, **shipped, preselected** | **576 ms** | **18.4 µs** |
| map build | 0.24 ms | once per fetcher |

**3.2x through the shipped path.** The finding is updated with these numbers; the spike memo's §5 carries a pointer note rather than being rewritten.

Also: `.changeset/decoding-preselects-the-event-by-topic0.md`, and one observation note (`work/notes/observations/a-decode-failures-real-error-is-overwritten-by-a-generic-one.md`) for a pre-existing bug I spotted but deliberately did not fix.

## Decisions

**1. The anonymous-event criterion's premise is false in viem 2.52.0; I pinned output IDENTITY instead of "still decodes correctly".** The task and the finding both assume an anonymous event decodes today and would "stop decoding" under a map-only implementation. It does not decode today: `decodeEventLog` selects a member by matching `topics[0]` against a computed event selector *even for a one-member ABI* (there is no `abi.length === 1` shortcut in 2.52.0, verified in `packages/core/node_modules/viem/_esm/utils/abi/decodeEventLog.js`), and an anonymous event's `topics[0]` is an indexed argument, so no member is found and the log records a `decodeError` on either route. I did not treat this as STOP-worthy drift because the task's actual contract ("same inputs, same outputs, byte for byte") settles it with no judgement left over: the anonymous event must keep taking the whole-ABI route and produce the identical answer, which is what the test asserts directly, against both the golden and the oracle. Alternative considered: STOP and route to needs-attention. Rejected because the sub-premise is not load-bearing for the design (the fallback is required either way, and is what keeps the two routes agreeing whichever answer viem gives in future). Touches: the finding, which I corrected in place rather than leaving to mislead the next reader.

**2. Shipped a changeset although no published surface moved.** The acceptance criterion says "ship a changeset if any published surface changes (expected: none)". None did: the map and the lookup are private, `parse`/`reparse` return the same. But the repo's `verify` gate runs `pnpm changeset status --since=main`, which fails for *any* changed package regardless of whether the API moved, so "no changeset" is a red gate at land time. I wrote a `patch` changeset for `@etherfold/core` that says explicitly that no published surface moves. Alternative considered: `changeset add --empty`, rejected because the release notes should carry a 3.2x decode win. Touches: the release the publish task eventually cuts.

**3. Deleted the dead private field `abiEventPerTopic`.** It was assigned in the constructor and read by nothing, and its comment claimed "kept because the map below decodes by it", which was not true. Leaving it beside a new `abiEventPerAddressAndTopic` would be two nearly-identically-named maps where only one means anything, which is exactly the coherence trap the new concept had to avoid. The local of the same name is untouched, so the duplicate-`topic0` refusal it guards is unchanged (and the comment now describes what actually decodes by it). This is a small scope expansion beyond the task's letter; recorded rather than silent.

**4. Measured the "before" number through production's address-agnostic route rather than a hand-rolled transcription** (`docs/spikes/replay-decode-cache/decode-breakdown.ts` gained a `productionWholeAbi` variant). `parseAllEventsIrrespectiveOfAddresses` deliberately does not preselect, and each fetcher in that harness holds exactly one contract, so it decodes against the same member list and differs only in the search. That makes the published delta production-to-production instead of production-against-transcription, which is what the task asked for; the pre-change production number it produces (1,824 ms) re-confirms the 1,791 ms recorded at `2efd858` on the same machine. It also gave a free real-data check of the pure-optimisation claim: all 31,330 logs decode identically preselected and unpreselected, both through the shipped code. Touches: the spike's result JSON shape (`correctness.allThreeAgree` → `allVariantsAgree`, a new `productionWholeAbi` entry, and `speedup` now production-to-production).

**5. Left the spike README as the pinned memo it is, with a note rather than a rewrite.** It states it is measured at `2efd858` and its §5 table is the argument that was made; rewriting its numbers would rewrite the record of a decision. I added a short block above that table pointing at the finding for the post-landing numbers, so a reader re-running the script does not think the memo is wrong.


wighawag:

1. **The pinning test exists and passes before and after: byte-identical `eventName`/`args` across a multi-event, multi-address ABI, including a failing decode and an anonymous event** — met, and asserted two ways: against a golden, and against an oracle that decodes the whole-ABI way.
2. **An anonymous event still decodes correctly via the fallback** — the criterion's PREMISE is false, and the substitution is right. See below.
3. **A `topic0` in the map and one NOT in it both decode correctly, so the fallback is exercised** — met.
4. **`parseAllEventsIrrespectiveOfAddresses` unchanged, asserted** — met, three cases.
5. **A `topic0` collision is still REFUSED at construction (ADR-0031), asserted rather than assumed** — met, and it earns its place: the one-member and whole-ABI routes are only interchangeable because no two members of one ABI can answer to one `topic0`.
6. **The map is built ONCE per fetcher, asserted structurally rather than by timing** — met, and done properly: it patches `Map.set` to count writes across two full reparse passes and asserts zero, plus identity of the map instance. No timing anywhere.
7. **The delta re-measured through the BUILT fetcher, finding updated** — met, and better than asked: a `productionWholeAbi` variant makes the published number production-to-production rather than production-against-transcription.
8. **Changeset if any published surface changes** — shipped, saying explicitly that none does.
9. **Build, typecheck, test green** — CI confirms.

### The overturned premise, which I verified rather than accepted

The task and the finding both assert that an anonymous event decodes today and that a map-only implementation would "silently stop decoding it". The build claims that is false in viem 2.52.0. **I checked the source rather than taking it**: `decodeEventLog` does `abi.find((x) => x.type === 'event' && signature === toEventSelector(formatAbiItem(x)))` with **no `abi.length === 1` shortcut**, and throws `AbiEventSignatureNotFoundError` when nothing matches. An anonymous event's `topics[0]` is an indexed argument rather than a selector, so no member matches on either route and the log records a `decodeError` either way. The claim is correct.

Not stopping on it was the right call. The task's actual contract is "same inputs, same outputs, byte for byte", and that settles the case with no judgement left over: the anonymous event must keep taking the whole-ABI route and produce the identical answer, which is what the test asserts. The sub-premise was never load-bearing — the fallback is required regardless, and is what keeps the two routes agreeing whichever way viem answers in future. The finding was corrected in place rather than left to mislead the next reader.

### Regression pin verified, and one honest note on its shape

I made the lookup map-only (`abi: preselected ? [preselected] : []`), removing the whole-ABI fallback. **Three tests go red** — all three `parseAllEventsIrrespectiveOfAddresses` cases — so the fallback is genuinely pinned.

Worth recording precisely which tests do NOT catch it, because it is not the ones you would expect: the anonymous case and the `topic0`-not-in-the-map case both stay green under that defect. Both assert identity with the whole-ABI route, and both inputs FAIL to decode on either route, so the outputs match either way. They match all the more exactly because of the `decodeError` overwriting this PR filed as an observation — the real error is replaced by a constant string, so two different failures are indistinguishable to a test comparing outputs.

That is not a defect in this PR and it does not weaken the change: the fallback is pinned by the flag tests, and Decision 4 adds the strongest evidence of all — all 31,330 real Base logs decode identically preselected and unpreselected, through the shipped code. But it does mean those two tests document intent more than they enforce it, and it is a second, quieter argument for fixing the observation below.

### Scope calls I checked and agree with

- **Deleting the dead private `abiEventPerTopic`** — it was assigned in the constructor, read by nothing, and carried a comment claiming "kept because the map below decodes by it", which was untrue. Leaving it beside a new `abiEventPerAddressAndTopic` would have been two near-identically-named maps where one means nothing, which is exactly the coherence trap the new concept had to avoid. A small expansion beyond the task's letter, recorded rather than silent, and the local of the same name (which guards the duplicate-`topic0` refusal) is untouched.
- **Leaving the spike README as a pinned memo** with a pointer to the finding, rather than rewriting its numbers. Right: it states it was measured at a specific commit, and rewriting it would rewrite the record of a decision rather than update a document.
- **A `patch` changeset despite no published surface moving** — the gate runs `changeset status --since=main`, which fails for any changed package regardless, and the release notes should carry a 3.2x decode win rather than an empty entry.

### Observation filed (good catch, correctly not fixed here)

`a-decode-failures-real-error-is-overwritten-by-a-generic-one`: in `decodeOnto` the catch sets `decodeError` to `decoding error: ${err}` and `parsed = null`, and the next block overwrites it with the constant `'parsing did not return any results'` because `parsed` is falsy. So the informative branch is unreachable in the output and every failure records the same uninformative string. Correctly left alone — this task's contract was byte-identical output and `decodeError` is stored on a `LogEventWithParsingFailure`, so changing it is a behaviour change with its own blast radius.

Merging. This unblocks tasking `the-stream-stores-only-what-the-node-said`, which decided this lands first and separately.

PR #82, opened 2026-09-06
tasking(the-stream-stores-only-what-the-node-said): The stream stores o…

wighawag:
Task: `work/tasks/done/the-stream-stores-only-what-the-node-said.md`

Auto-tasked `the-stream-stores-only-what-the-node-said` into 5 backlog tasks.

## Tasks

- **a-fixture-reader-is-not-a-stream-keeper** — A fixture reader is not a stream keeper
  - needsAnswers: true
- **no-configuration-can-strip-the-raw-log** — No configuration can strip the raw log
  - covers: US-5
  - needsAnswers: true
- **the-stored-event-type-refuses-a-decoded-event** — The stored-event type refuses a decoded event
  - covers: US-2
  - blockedBy: no-configuration-can-strip-the-raw-log
  - needsAnswers: true
- **the-stream-is-saved-without-its-decoded-half** — The stream is saved without its decoded half
  - covers: US-1, US-3
  - blockedBy: the-stored-event-type-refuses-a-decoded-event
  - needsAnswers: true
- **the-stream-seam-takes-only-the-stored-event** — The stream seam takes only the stored event
  - covers: US-2, US-4
  - blockedBy: the-stream-is-saved-without-its-decoded-half, a-fixture-reader-is-not-a-stream-keeper
  - needsAnswers: true

## Dependency graph

- Keystones: a-fixture-reader-is-not-a-stream-keeper, no-configuration-can-strip-the-raw-log
- the-stored-event-type-refuses-a-decoded-event ← no-configuration-can-strip-the-raw-log
- the-stream-is-saved-without-its-decoded-half ← the-stored-event-type-refuses-a-decoded-event
- the-stream-seam-takes-only-the-stored-event ← the-stream-is-saved-without-its-decoded-half
- the-stream-seam-takes-only-the-stored-event ← a-fixture-reader-is-not-a-stream-keeper

## Needs answers

- **a-fixture-reader-is-not-a-stream-keeper**:
  - The decomposition mints StoredLogEvent without ever naming EmittedLog, which already exists in the SAME module and already claims almost the same meaning. Lens 4c fork risk, and a concrete hole in the-stream-seam-takes-only-the-stored-event: its criterion forbids re-typing an implementation to the base event type, but EmittedLog is a SECOND supertype with the identical hole (a decoded event satisfies it), and it sits right beside the server keeper this task narrows. A keeper annotated EmittedLog would compile and enforce nothing, which is exactly the failure the spec minted a new name to prevent. Fixed by the edits: task 3 must state the relation in the new type docstring and leave EmittedLog untouched; task 5 forbids both supertypes. (packages/core/src/types.ts:49 exports EmittedLog = NumberifiedLog, docstring: one entry of the emission stream as a host that STORES it sees it, deliberately does NOT promise the decoded half. Used by streamBuilder.ts, emissionStream.ts, server/src/emissions.ts.)
- **no-configuration-can-strip-the-raw-log**:
  - The decomposition mints StoredLogEvent without ever naming EmittedLog, which already exists in the SAME module and already claims almost the same meaning. Lens 4c fork risk, and a concrete hole in the-stream-seam-takes-only-the-stored-event: its criterion forbids re-typing an implementation to the base event type, but EmittedLog is a SECOND supertype with the identical hole (a decoded event satisfies it), and it sits right beside the server keeper this task narrows. A keeper annotated EmittedLog would compile and enforce nothing, which is exactly the failure the spec minted a new name to prevent. Fixed by the edits: task 3 must state the relation in the new type docstring and leave EmittedLog untouched; task 5 forbids both supertypes. (packages/core/src/types.ts:49 exports EmittedLog = NumberifiedLog, docstring: one entry of the emission stream as a host that STORES it sees it, deliberately does NOT promise the decoded half. Used by streamBuilder.ts, emissionStream.ts, server/src/emissions.ts.)
- **the-stored-event-type-refuses-a-decoded-event**:
  - The decomposition mints StoredLogEvent without ever naming EmittedLog, which already exists in the SAME module and already claims almost the same meaning. Lens 4c fork risk, and a concrete hole in the-stream-seam-takes-only-the-stored-event: its criterion forbids re-typing an implementation to the base event type, but EmittedLog is a SECOND supertype with the identical hole (a decoded event satisfies it), and it sits right beside the server keeper this task narrows. A keeper annotated EmittedLog would compile and enforce nothing, which is exactly the failure the spec minted a new name to prevent. Fixed by the edits: task 3 must state the relation in the new type docstring and leave EmittedLog untouched; task 5 forbids both supertypes. (packages/core/src/types.ts:49 exports EmittedLog = NumberifiedLog, docstring: one entry of the emission stream as a host that STORES it sees it, deliberately does NOT promise the decoded half. Used by streamBuilder.ts, emissionStream.ts, server/src/emissions.ts.)
- **the-stream-is-saved-without-its-decoded-half**:
  - The decomposition mints StoredLogEvent without ever naming EmittedLog, which already exists in the SAME module and already claims almost the same meaning. Lens 4c fork risk, and a concrete hole in the-stream-seam-takes-only-the-stored-event: its criterion forbids re-typing an implementation to the base event type, but EmittedLog is a SECOND supertype with the identical hole (a decoded event satisfies it), and it sits right beside the server keeper this task narrows. A keeper annotated EmittedLog would compile and enforce nothing, which is exactly the failure the spec minted a new name to prevent. Fixed by the edits: task 3 must state the relation in the new type docstring and leave EmittedLog untouched; task 5 forbids both supertypes. (packages/core/src/types.ts:49 exports EmittedLog = NumberifiedLog, docstring: one entry of the emission stream as a host that STORES it sees it, deliberately does NOT promise the decoded half. Used by streamBuilder.ts, emissionStream.ts, server/src/emissions.ts.)
- **the-stream-seam-takes-only-the-stored-event**:
  - The decomposition mints StoredLogEvent without ever naming EmittedLog, which already exists in the SAME module and already claims almost the same meaning. Lens 4c fork risk, and a concrete hole in the-stream-seam-takes-only-the-stored-event: its criterion forbids re-typing an implementation to the base event type, but EmittedLog is a SECOND supertype with the identical hole (a decoded event satisfies it), and it sits right beside the server keeper this task narrows. A keeper annotated EmittedLog would compile and enforce nothing, which is exactly the failure the spec minted a new name to prevent. Fixed by the edits: task 3 must state the relation in the new type docstring and leave EmittedLog untouched; task 5 forbids both supertypes. (packages/core/src/types.ts:49 exports EmittedLog = NumberifiedLog, docstring: one entry of the emission stream as a host that STORES it sees it, deliberately does NOT promise the decoded half. Used by streamBuilder.ts, emissionStream.ts, server/src/emissions.ts.)


wighawag:
APPROVE ✅ — Gate-3 on the TASKING. CI `verify` COMPLETED/SUCCESS, mergeStateStatus CLEAN.

Merging admits nothing to the agent pool: all five tasks land in `work/tasks/backlog/` (staging) with `needsAnswers: true`, so neither CI `advance` nor a `run` daemon can claim one. The single unresolved question goes to the human.

### The decomposition

Five tasks, and the graph is sound:

- `no-configuration-can-strip-the-raw-log` (covers 5) and `a-fixture-reader-is-not-a-stream-keeper` are the two roots;
- `the-stored-event-type-refuses-a-decoded-event` (2) ← the first;
- `the-stream-is-saved-without-its-decoded-half` (1, 3) ← that;
- `the-stream-seam-takes-only-the-stored-event` (2, 4) ← that plus the fixture root.

Stories 1–5 are all covered, with no task claiming coverage it does not deliver.

### The spec's load-bearing decisions each landed somewhere

I checked these individually rather than trusting the decomposition:

- **`logValues` DELETED, not relocated** — with the full sweep scoped (the field, `LogValuesFlags`, the `OptionsFlags` helper, the projection branch, the indexer TODO, the root `TODO.md` line, the `CONTEXT.md` entry), and the stream-config digest explicitly NOT narrowed.
- **The detect-and-clear guard SURVIVES**, and the task is honest that nothing today pins it — the sibling's unparseable-SEGMENT tests are a different guard on a different layer, so this task WRITES that test rather than preserving one. It also tells the builder to grep first and correct the premise if wrong.
- **The strip produces NEW OBJECTS** — with the reason: a new array over the same references strips nothing, and those references are the processor's own event objects.
- **`lastSync` stripped, building a NEW `LastSync`** — with the trap named: the same object goes to the state keeper on the same tick, so a mutating strip silently empties the live reorg window.
- **Expand-then-contract handled explicitly.** The task that strips must typecheck on its own before the seam narrows, and it is told exactly how: ONE assertion at the single call site where the stripped batch meets `saveNewEvents`, commented with why it is temporary and which task removes it — and explicitly NOT by widening the seam, narrowing it early, or declaring the strip's output to be decoded events. That is the right shape and the kind of thing that goes wrong silently when it is left implicit.
- **`replayStream` stops being an `ExistingStream`**, with the `fixture.ts` docstring rewrite owed in the same change.
- **A changeset in every task**, recording the breaking `@etherfold/core` change.

### One drift finding, in the tasker's favour

The spec instructs that its tasks must carry `blockedBy` onto `a-reconfigure-is-not-an-outage`'s landable-1 tasks, to serialise against a merge collision in `@etherfold/core` and both keepers. **Those tasks are all in `work/tasks/done/`**, so the collision is moot and the emitted `blockedBy` edges are correctly internal to this spec only. The spec's prose still reads as though that risk is live; it is not.

### The one question, and why the loop did not converge

The review loop ran its three passes, applied 11 edits, and stopped with one blocking issue outstanding — so it gated all five tasks with the same question. That is the right failure mode, and the question is a good one:

**The spec mints `StoredLogEvent` without ever naming `EmittedLog`, which already exists in the same module and claims almost the same meaning.** I verified this rather than accepting it: `EmittedLog` is exported from `packages/core/src/types.ts:49`, its docstring says it is "one entry of the emission stream as a host that STORES it sees it" and "deliberately does NOT promise the decoded half", and it is used by `streamBuilder.ts`, `emissionStream.ts` and `server/src/emissions.ts`. The spec mentions it **zero times** — and `EmittedLog` landed on 2026-09-05 in #62, *after* this spec was written. So this is real drift, not a misreading.

The concrete hole is sharper than the naming overlap. `EmittedLog = NumberifiedLog` is a plain alias, and `BaseLogEvent = NumberifiedLog & {…}`, so a decoded `LogEvent` is structurally assignable to `EmittedLog`. It is a SECOND supertype with the identical hole that made the spec refuse `BaseLogEvent` — a keeper annotated `EmittedLog` would compile and enforce nothing, which is exactly the failure the new name exists to prevent, and it sits right beside the server keeper task 5 narrows.

The tasker's own edits already address it (task 3 states the relation in the new type's docstring and leaves `EmittedLog` untouched; task 5 forbids annotating an implementation with EITHER supertype). What it could not do is confirm convergence, so it gated rather than guessed. Correct call.

Merging so the staged tasks are on `main` where the human can act on them; the question goes to them as a batch of one.

PR #83, opened 2026-09-06
feat(no-configuration-can-strip-the-raw-log): No configuration can stri…

wighawag:
Task: `work/tasks/done/no-configuration-can-strip-the-raw-log.md`

## What I did

**Drift check first.** Every premise in the task held against the code: the projection branch in `LogEventFetcher.parse`, the `reparse` docstring naming the knob, the `OptionsFlags`-derived `LogValuesFlags` and the `logValues` field on `LogParseConfig`, the indexer TODO, the two load-path clear branches (only the first named the knob; the third `reparse` site is the follower path, which returns rather than clears and names nothing), `docs/reviews/todo-triage.md`, the root `TODO.md`, and the `CONTEXT.md` entry. The task's warning was also correct: **no existing test drove `reparse` to "cannot re-read"** (the neighbours cover an unreadable stored *segment*: `streamSegments.test.ts`, `degradingStream.test.ts`), so this task wrote that test. The only stale detail was a path: the source spec has moved `work/specs/proposed/` → `work/specs/tasked/` since the snapshot.

**Test first** (`packages/core/test/rawLogIsNeverStripped.test.ts`, new): written and run before the deletion. It was RED on `pnpm typecheck` (two `Unused '@ts-expect-error' directive` errors, because `logValues` still compiled) and green on the guard half, which is what pins the branch across the deletion. I also mutation-checked it: removing the raw-half check in `reparse` kills 5 of its 16 cases (file restored byte-identically afterwards).

**The deletion** (`@etherfold/core`): `logValues` off `LogParseConfig`, the `LogValuesFlags` type and the now-unused private `OptionsFlags` helper, and the projection branch in `parse` (which now always pushes the whole raw log, with a comment saying why). No dead helper left; `parseConfig` still carries `filters` and `parseAllEventsIrrespectiveOfAddresses`.

**The guard, re-explained rather than removed:** `reparse`'s docstring and the indexer's clear-branch comment now name the real remaining cause (a stream already on disk, written by an older version whose parse could strip `topics`/`data`) and cite ADR-0034 plus the pinning test. ADR-0034, the `.changeset/` entry and CHANGELOGs are historical and untouched. A workspace grep for `logValues` / `OptionsFlags` leaves only: history (ADR, changeset, done-task records, specs) and the new test's own prose/`@ts-expect-error` lines.

**Sweep:** indexer TODO gone, root `TODO.md` gone, `CONTEXT.md`'s stream-identity entry now states the deletion as done and explicitly says the digest is NOT narrowed (`parse` stays in on `filters`). No digest code was touched.

**Changeset:** `.changeset/no-configuration-can-strip-the-raw-log.md`, `@etherfold/core: major`.

**Gate:** `pnpm build`, `pnpm typecheck`, `pnpm test` (full workspace), `pnpm format:check`, `pnpm changeset status` and `pnpm check:adr` all pass. `git status` shows only this task's changes.

## Decisions

- **Deleted the root `TODO.md` file outright, not just its line.** It held exactly one item, the one this task closes by deciding not to build it, so removing the line would have left a zero-byte file in the repo root. Alternative considered: keep an empty `TODO.md`, or keep it with a "nothing outstanding" placeholder; both preserve a file whose whole content is gone, and this repo tracks work in `work/`, not in a root TODO. Touches: two docs that referenced "the root `TODO.md`" (see next two entries). Reversible in one commit if a human wants the file back.
- **Annotated `docs/reviews/todo-triage.md` as closed-by-deletion rather than leaving it silent** (the task offered either). Both of its `logValues` entries (the root-TODO item and the "type `logValues`" standalone cleanup) now carry a note pointing at `work/specs/tasked/the-stream-stores-only-what-the-node-said.md`, and the dated triage text itself is left verbatim underneath, because it is a record of what was true when it was written. It is also where the "flag STUB" evidence lives, so a silent leave would have had it still recommending a feature this task decided against.
- **Touched `docs/design/historical-state-database.md` (one line), which is outside the task's enumerated sweep.** That line said raw `data`/`topics` retention "is mutually exclusive with the root `TODO.md` idea of trimming log fields via stream config" — a live design doc pointing at a file I was deleting, so it would have dangled. Rewritten to say the conflict is settled by the deletion. Alternative considered: leave it and accept the dangling reference; rejected for exactly the reason this task exists (a live doc explaining today's design by something that no longer exists).
- **The type test's refusal literal spells out EVERY raw-log flag, rather than a partial `{topics: true}`.** `LogValuesFlags` was a mapped type with all-required properties, so a partial literal was already a type error today: the `@ts-expect-error` would have been satisfied before and after the deletion and proved nothing. The full literal was VALID before (hence RED, verified) and fails only on the unknown property after. Touches nothing else; noted because the obvious shorter form is a silently useless assertion.
- **The guard's load-path test drives the REAL `reparse`.** The shared stream-cache world (`test/utils/streamCacheWorld.ts`) replaces the log fetcher wholesale and its `reparse` is a pass-through, so the clear branch is unreachable through it as-is. Rather than change that shared fake (it is used by five other suites, whose assertions are about cache arithmetic), the new test overlays a real `LogEventFetcher.reparse` onto the fake fetcher for its two cases. Alternative considered: give the shared world a real `reparse`; rejected as a wider blast radius than this task needs.
- **No ADR written.** The durable "why" for deleting rather than relocating the knob already lives in the source spec and is now restated in the changeset; nothing here is hard to reverse or surprising without that context, so it does not meet the ADR gate.


wighawag:

1. **`LogParseConfig` has no `logValues`, `LogValuesFlags` is gone, asserted at the TYPE level, no dead private helper left behind** — met. Both `LogValuesFlags` AND the `OptionsFlags` helper it was built from are deleted, so nothing dead is left; the refusal is a `@ts-expect-error` evaluated by `pnpm typecheck`.
2. **`parse` has no projection branch** — met; every parsed event carries the full raw log with no configuration able to change that.
3. **The detect-and-clear guard SURVIVES and is pinned by a test of its own, BOTH halves** — met, and this is the best part of the diff. See below.
4. **The indexer TODO and the root `TODO.md` line are gone, `CONTEXT.md` no longer speaks of the deletion as future work, and the stream-config digest is NOT narrowed** — met. The glossary now says the knob is gone and states explicitly that `parse` stays in the digest "on the strength of `filters` alone".
5. **No comment or docstring in `packages/*/src` still explains the guard by naming the deleted knob** — met; zero residual mentions, and both sites now name a stream written by an older version instead.
6. **A changeset records the breaking `@etherfold/core` change** — met.

I removed the guard itself (`if (!stored.topics || !stored.data || !stored.address) return undefined`) so a stripped stream would be replayed on trust. **Five tests go red**: all four "cannot be re-read when the projection dropped …" cases and the load-path clear. The guard is pinned, and it is pinned at both levels the criterion asked for — the rule (`reparse` answering "cannot re-read") and the consequence (the load path clearing rather than replaying).

The task was honest that nothing pinned this before and that the sibling's unparseable-SEGMENT tests are a different guard on a different layer; it wrote the missing test rather than claiming an existing one. It also asserts the negative — a stream whose raw half is intact replays untouched with the node asked for nothing — so the guard cannot pass by being permanently on.

### The premise that had drifted, and the deletion I checked

The task expects "the root `TODO.md` line that stubs it". My own pre-flight grep for `logValues` in `TODO.md` found nothing, so I checked what was actually deleted before accepting a whole-file removal. The file held exactly one line:

`- [] NumberifiedLog / LogEvent could have fields removed, configuration on config.stream`

That IS the stub — it describes the feature (removing raw fields via stream config) without using the word, which is why my grep missed it and why the build was right where I was wrong. Deleting the file rather than leaving it zero-byte is the correct call, it is reversible in one commit, and the two docs that referenced "the root `TODO.md`" were both updated rather than left dangling.

### Two judgement calls worth calling out, both right

- **The type test spells out EVERY raw-log flag rather than a partial literal.** This is subtle and it matters: `LogValuesFlags` was a mapped type with all-required properties, so `{topics: true}` was already a type error *before* the deletion — the `@ts-expect-error` would have been satisfied both before and after and proved nothing. The full literal was valid before (verified red) and fails only on the unknown property now. That is the difference between a real assertion and a decorative one.
- **The guard's load-path test overlays a REAL `reparse` onto the shared fake.** The stream-cache world replaces the log fetcher wholesale and its `reparse` is a pass-through, so the branch under test is unreachable through it — the test would have passed while exercising nothing. Overlaying the real rule for two cases, rather than changing a shared fake five other suites depend on for cache arithmetic, is the right blast radius.

### One scope expansion, declared

`docs/design/historical-state-database.md` gained a one-line edit outside the enumerated sweep, because it explained a live design decision by reference to the `TODO.md` being deleted and would have dangled. Correct to fix, and correctly declared rather than slipped in. `docs/reviews/todo-triage.md` was annotated as closed-by-deletion with the dated text left verbatim underneath — the right treatment for a record rather than live documentation, and it matters here because that file is also where the "flag STUB" evidence lives, so a silent leave would have had it still recommending a feature this task decided against.

Merging.

PR #84, opened 2026-09-06
feat(a-fixture-reader-is-not-a-stream-keeper): A fixture reader is not…

wighawag:
Task: `work/tasks/done/a-fixture-reader-is-not-a-stream-keeper.md`

Task complete. Working tree holds only this task's changes.

## What I did

**Drift check first.** The task's premises still hold: `logValues` deletion (`no-configuration-can-strip-the-raw-log`) landed, `StoredLogEvent` does not exist yet, `replayStream` still declared `ExistingStream` and was still built out of `readOnlyStream`, and both named tests were exactly as described. No stale premise, so I built it.

**Test first**, at the agreed seam: `packages/core/test/streamFixture.test.ts` grew a `replayStream` block asserting the new reader (serves from a block, refuses another chain, serves the same events on repeated fetches, reports an empty unconfirmed window per ADR-0035, hands back DECODED events) plus a compile-time refusal block (`@ts-expect-error` on `saveNewEvents`, `clear`, and assignment to `ExistingStream`) and a runtime check that no write half exists. It went red on the write-half assertion before the production change.

**Production change** (`packages/core/src/stream/fixture.ts`): new exported `StreamFixtureReader<ABI>` (`fetchFrom` only, decoded `LogEvent`s, always answers), `replayStream` returns it and no longer imports `readOnlyStream`. Format, format number, provenance, `blocksOf`, `replayFixtureInto`, `captureStream` untouched.

**`readOnlyStream` is unchanged in behaviour** and keeps its keeper-seam callers (follower + server's `storedEmissionStream`); its own tests keep covering that case. Its `replayStream is that same view` block moved out of `readOnlyStream.test.ts` and its guarantees now sit on the fixture's own type.

**Prose corrected:** the replay helper's docstring, `readOnly.ts`'s "one view, two callers" section and its `StreamReader` note, `ExistingStream.setStreamConfig`'s "a replayed fixture" example in `types.ts`, `segments.ts`'s "shipped third implementation of this seam", the core README bullet, `CONTEXT.md`'s **read-only stream view** and **seeding** entries, ADR-0044's rejected-alternatives bullet (marked stale on its example only), ADR-0035's "third implementation" clause, and a `processor-sqlite` test comment that said `replayStream` "builds an `ExistingStream`" (found by grepping, not on the task's list). The package CHANGELOG was left alone.

**New:** `docs/adr/0059-a-fixture-reader-is-not-a-stream-keeper.md`, and `.changeset/a-fixture-reader-is-not-a-stream-keeper.md` (`@etherfold/core` minor, naming what a caller has to change).

**Gates:** `pnpm build`, `pnpm typecheck`, `pnpm test`, `pnpm format:check`, `pnpm changeset status`, `pnpm check:adr` all green (core: 770 passed, 1 skipped).

## Decisions

- **The fixture reader has NO write half at all, rather than a no-op `saveNewEvents`/`clear` of its own.** Chosen because it makes "writing through a fixture does not change what it serves" unexpressible instead of merely swallowed, and because a second no-op write beside `readOnlyStream`'s is exactly the parallel read-only shape ADR-0044 refuses. Alternative considered: keep a no-op write half so the fixture still *looks* like the seam (rejected: preserves the conflation the change removes). Touches: the read-only-stream test's "swallows a write" assertion about the fixture is now a compile-time claim, pinned with `@ts-expect-error` under `pnpm typecheck` in `streamFixture.test.ts` rather than dropped. Recorded durably as **ADR-0059**.
- **`StreamFixtureReader.fetchFrom` returns a value, not `| undefined`.** A fixture is a stream that is present by construction; ABSENT is a keeper's answer, and a wrong chain is still a THROW, not an absence. Alternative: mirror `StreamFetcher`'s optional result for familiarity (rejected: it would invite `?.` on a thing that never answers nothing, and copying the seam's shape is what got the fixture onto the seam in the first place). Touches: any caller written against the old `?.` (only core's own tests today) and the changeset says so.
- **Named `StreamFixtureReader`, and it lives in `stream/fixture.ts`.** Coherence check: `StreamReader` already means "the READ half of the KEEPER seam" in `stream/readOnly.ts`, so I did not reuse or re-mean it; the new name is qualified by its subject and sits with the fixture it reads, so the seam narrowing (`the-stream-seam-takes-only-the-stored-event`) can move `StreamReader` onto the stored-event type without touching the fixture. Touches that sibling task, which already expects the fixture to be off the seam.
- **ADR-0044 is amended in place with a stale-on-its-example note pointing at ADR-0059, not rewritten**, and ADR-0035's "third implementation of this seam" clause got the same one-clause correction (its evidence about the empty window is unaffected). Alternative: leave both ADRs alone as historical records (rejected: ADR-0044's rejected-alternatives bullet is cited as current rule, and a reader arriving with it in hand would find the code contradicting it).


wighawag:

1. **The replay helper returns its own reader type (decoded events), not `ExistingStream`, and does not go through the read-only wrapper** — met. `replayStream` returns `StreamFixtureReader` with exactly one operation, `fetchFrom`, and no write half at all.
2. **The read-only wrapper is unchanged in BEHAVIOUR and still serves the follower/one-writer case; prose corrected; the tests that asserted the fixture through it keep asserting the same guarantees against the fixture's own type** — met on both halves, and I checked the second one specifically rather than trusting a line count. See below.
3. **The fixture serialization format is byte-for-byte unchanged** — met: an existing captured fixture still parses and replays the same events from the same block, a fixture for another chain is still refused, the BigInt round-trip and the real-BigInt-versus-look-alike-string case survive, and a format-1 fixture is still refused rather than half-decoded.
4. **The replay helper's docstring, the core README line, BOTH `CONTEXT.md` entries, and the segment keeper's comment no longer describe the fixture as the keeper seam or as a caller of the wrapper** — met, all five sites.
5. **The stream-fixture test still pins what a fixture guarantees, adjusted rather than deleted** — met, and strengthened.
6. **A changeset records the `@etherfold/core` API change** — met.

### The 51 deleted test lines, checked rather than assumed

`readOnlyStream.test.ts` lost two cases and `streamFixture.test.ts` gained a set, which is exactly the shape a silent loss of coverage would also have. It is not one: `still serves a fixture and still writes nothing` and `still refuses a fixture captured on another chain` moved into the fixture's own suite as `serves the captured events from a block, and refuses another chain` and `offers NO write at all…`, and the guarantee got STRONGER on the way — what was a runtime observation that a no-op write changed nothing is now a type fact plus a runtime absence. The five cases that remain in `readOnlyStream.test.ts` are all about the wrapper itself, which is right, since the fixture is no longer one of its callers.

I gave the fixture its write half back (`saveNewEvents`/`clear` as borrowed no-ops), which is precisely the state this task exists to leave behind. `offers NO write at all, so "writing through a fixture changes nothing" is a type fact` goes red; the other 11 stay green. The separation is pinned rather than merely performed.

### The ADR earns its place

ADR-0059 is written for the right reason and says so in its own second paragraph: **ADR-0044 deliberately built the fixture out of `readOnlyStream`** so there would be one definition of read-only on that seam, so a reader arriving with that ADR in hand would find the opposite and be right to ask. Recording the reversal is what stops that being a mystery.

The argument is also correct. A keeper stores what the node said; a fixture's events are decoded ONCE at capture, deliberately, so a replay does not re-run the decoder — and that decoded half is the whole point of it as a test input. They shared an interface only for as long as nothing made them differ, and the seam narrowing to a stored event that structurally refuses a decoded one is what made them differ. Nothing ever wired a fixture as `keepStream`, so they never met at runtime; the fixture merely DECLARED the seam.

And the ADR-0044 question is answered rather than dodged: that rule is about the KEEPER seam and is untouched, `readOnlyStream` is unchanged in behaviour (I verified — the diff to that file is docstring-only), and what changed is a caller LEAVING rather than a second implementation ARRIVING. `readOnlyStream`'s callers are now all keepers — the container's follower and the server's `storedEmissionStream` — which is a sharper statement of ADR-0044's rule than the fixture ever was.

Merging. Both roots are down; the chain unblocks from here.

PR #85, opened 2026-09-06
feat(the-stored-event-type-refuses-a-decoded-event): The stored-event t…

wighawag:
Task: `work/tasks/done/the-stored-event-type-refuses-a-decoded-event.md`

## What I did

**Checked the task against reality first.** Its blocker `no-configuration-can-strip-the-raw-log` has landed (`work/tasks/done/`, `logValues` is gone and `packages/core/test/rawLogIsNeverStripped.test.ts` pins both halves of the ADR-0034 guard). `BaseLogEvent` / `ParsedLogEvent` / `LogEventWithParsingFailure` / `LogEvent` and `EmittedLog` are where the task says, `reparse` is still `readonly LogEvent<ABI>[]`, the keeper seam still declares decoded events, and the follower path's raw-only readers are there as described (the next task's). The one transcription fix the task itself flagged applies: `BaseLogEvent` declares `Extra extends JSONObject | undefined = undefined`, so the constraint is carried over verbatim. The named type-test exemplar `packages/browser/test/processorKinds.test.ts` no longer exists (deleted with ADR-0037); the prompt's own pointer, `packages/browser/test/callShape.test.ts`, is the live one and is what I followed. No drift worth stopping for.

**`packages/core/src/types.ts`** — added and exported `StoredLogEvent<Extra extends JSONObject | undefined = undefined> = BaseLogEvent<Extra> & {args?: never; eventName?: never; decodeError?: never}`, immediately after `LogEvent` and above `EmittedLog`. Its docstring says what it governs (WRITES from here on; reads tolerate and ignore a decoded half, no migration), why it is not `BaseLogEvent`, names the widening hole (a value already typed as the supertype still assigns; the guard is at the seam), and states the relation to `EmittedLog`: different seams, `EmittedLog` permissive and ABI-free for the server's emission row, this one carrying `extra`/`removedStreamID` and refusing a decoded event. `EmittedLog` itself is untouched.

**`packages/core/src/internal/decoding/LogEventFetcher.ts`** — `reparse` widened to `readonly (LogEvent<ABI> | StoredLogEvent)[]`, still returning `LogEvent<ABI>[] | undefined`. Runtime body unchanged apart from the internal destructure's assertion target moving from `LogEvent<ABI> & {...}` to `BaseLogEvent & {...}`, which is what makes the union destructure without an assertion that lies about the input. Docstring extended with why it takes both shapes.

**`packages/core/test/storedLogEvent.test.ts`** (new) — the type test in the repo's `@ts-expect-error` style, body deliberately never called: a parsed event with ABI inputs, a parsed event whose ABI event declares none, a parsing failure, the `LogEvent` union, and an array of the union are each REJECTED where the stored type is expected; a fresh raw literal and an array of them are ACCEPTED. Below it, three ordinary vitest cases: `reparse` takes a `StoredLogEvent[]` with no cast and returns decoded events, produces byte-identical decoded halves from the stored and the decoded routes, and still answers "cannot re-read" for a stored event whose raw half is missing.

**`.changeset/the-stored-event-refuses-a-decoded-one.md`** — records the added `@etherfold/core` type and that nothing narrowed.

Gate: `pnpm build`, `pnpm typecheck`, `pnpm test` (core 46 files / 775 tests, whole workspace green), `pnpm format:check`, `pnpm changeset status`, `pnpm check:adr` all pass. `pnpm typecheck` is load-bearing here: an unused `@ts-expect-error` is itself an error, so every refusal above is proven to fire. Working tree holds only the four intended files.

## Decisions

- **`reparse` takes an explicit union (`readonly (LogEvent<ABI> | StoredLogEvent)[]`) rather than the supertype (`readonly BaseLogEvent[]`).** Chosen because the union NAMES the two things a read may be handed, while annotating the parameter with the supertype would accept literally any event including ones nothing decoded, which is the exact "enforces nothing" property the spec refuses `BaseLogEvent` for elsewhere; a reader of the signature would then have no way to tell that a stored stream is what this is for. Alternatives considered: the supertype (rejected, above), and making `reparse` generic over `Extra` (rejected as scope creep — every caller runs at `Extra = undefined` and no consumer asked for it). What it touches: `the-stream-seam-takes-only-the-stored-event`, which narrows `StreamFetcher` to hand back stored events. That task needs no further change here, since `readonly StoredLogEvent[]` already satisfies this parameter; if it ever decides reads should be stored-only, dropping the `LogEvent<ABI>` arm is a one-line contraction, but I have NOT done it, because reads still legitimately carry decoded events off segments written before this change.
- **The internal destructure now asserts `BaseLogEvent & {args?: unknown; eventName?: unknown; decodeError?: unknown}` instead of `LogEvent<ABI> & {...}`.** A union input cannot be asserted to one member without claiming something false about the other. The supertype is the honest common shape here, and it is an internal assertion inside the one function that immediately drops those three keys, not a seam annotation — the "never annotate a seam with the supertype" rule is untouched. Alternative considered: destructuring without any assertion (it does typecheck, both arms being assignable), rejected because the explicit optional-`unknown` shape is what documents that the three keys are being deliberately discarded.
- **The known widening hole is asserted as a COMPILING line in the type test, not only described in prose.** `storesOne(widened)` (a `BaseLogEvent`) sits in the refusal function with no `@ts-expect-error`, commented as the hole. This makes the limit executable knowledge rather than a paragraph nobody re-reads. The cost, stated at the site: if a later change ever closes the hole, that line goes red and the comment must be rewritten. Alternative considered: documenting it in the docstring only (rejected — a hole that is only prose is exactly how the supertype gets reached for again).
- **No new ADR.** The decision this task carries (a distinct exclusion-form type rather than reusing the supertype, and `EmittedLog` surviving beside it) was made by the human in the task body and by the spec, and its durable WHY now lives in the exported type's docstring plus the changeset. Adding an ADR would be a third copy of a decision already recorded in two durable places, and the constraint it operates under (ADR-0034) already exists.


wighawag:

### Acceptance criteria

1. **A stored-event type exported from `@etherfold/core` whose shape refuses a decoded event, with a docstring saying what it governs (writes), naming the widening hole, and stating its relation to `EmittedLog`** — met, all four parts, and the docstring is genuinely load-bearing rather than decorative.
2. **`EmittedLog` unchanged: not re-pointed, not re-meant, not deleted** — met, and I diffed it rather than eyeballing: byte-identical.
3. **The re-decode widens to accept the stored type** — met, and correctly: `reparse` takes `LogEvent<ABI> | StoredLogEvent`, still returns `LogEvent<ABI>[]`, because a READ produces decoded events. The internal destructure moved to `BaseLogEvent` with the three optional-unknown escapes, which is the honest way to strip a half the static type says is not there.
4. **A changeset** — met.
5. **Tests in the repo's style** — met.

### Your `EmittedLog` decision, delivered as decided

The docstring states the relation exactly as agreed, and states it in terms of KIND and AUDIENCE rather than name: `EmittedLog` is the SERVER's emission-row shape (ADR-0006), free of an ABI type parameter, PERMISSIVE — a plain `NumberifiedLog` alias that does not promise the decoded half but does not refuse it — and it can express neither `extra` nor `removedStreamID`. `StoredLogEvent` is what a `keepStream` keeper persists, carries both, and REFUSES. It closes with "neither substitutes for the other, and `EmittedLog` is not to be re-pointed at this", which is the instruction task 5 has to honour.

The build did not merge them and did not touch `EmittedLog`, which is what the task's STOP clause required if it had concluded otherwise.

### The hole is stated, not hidden

The docstring says plainly that an event whose STATIC type has already been widened to `BaseLogEvent` still assigns, because nothing is left for the `?: never` clauses to catch — so the guard is at the SEAM and not through a widening. That is the spec's own known limitation, carried forward rather than quietly dropped, and it is exactly why task 5's criterion forbids re-typing an implementation to a supertype.

It is also honest about scope: the type governs WRITES from here on. Segments written before it existed still hold `args` and `eventName` forever, no migration rewrites them, and a READ tolerates and ignores that half because the re-decode drops and re-derives it anyway.

I collapsed the exclusion to the plain supertype — `StoredLogEvent = BaseLogEvent<Extra>`, which is precisely the shape the spec refused and precisely the hole `EmittedLog` has. **Five typecheck errors**, one per `@ts-expect-error` that stopped catching: a parsed event, a parsed event whose declaration has no inputs, a parsing failure, the union through its members, and an ARRAY of them — that last being the shape a keeper is actually handed, which is the one that matters most.

Worth noting this is a TYPE-level pin: `vitest` cannot see it, and `pnpm typecheck` is what enforces it. That is the only way to assert a configuration is not accepted, and the test file says so at the top rather than leaving a reader to wonder why the body is never called.

Merging.

PR #86, opened 2026-09-06
feat(the-stream-is-saved-without-its-decoded-half): The stream is saved…

wighawag:
Task: `work/tasks/done/the-stream-is-saved-without-its-decoded-half.md`

## What I built

**The strip, in core, at the one save call site.**

`packages/core/src/internal/stream/strip.ts` (new) holds three functions and two types:

- `storedEventOf` — one event rebuilt without `args` / `eventName` / `decodeError`, by the same destructuring `LogEventFetcher.reparse` uses on the way back, so both halves of "the decoded half is a cache" are spelled the same way. Returns `StoredLogEvent`.
- `storedStreamOf` — the batch, every element rebuilt (a new array over the same references would strip nothing, and those references are the processor's own objects).
- `storedLastSyncOf` — a NEW `LastSync` with new blocks and new events inside them, because the same object is handed to the state keeper on the same tick. Cursor numbers and `context` carried over unchanged.

`packages/core/src/indexer.ts`, `promiseToSave`: `toWrite` is stripped, `lastSync` is stripped, and the pair is handed to `saveNewEvents` under ONE assertion over the whole argument, commented as temporary and naming `the-stream-seam-takes-only-the-stored-event` as its remover. Nothing else moved: the seam is untouched, no keeper changed, the read path's re-decode is untouched, and `streamWrittenNotProcessed` still holds the decoded batch the engine compares against (it compares raw fields only).

The assertion has to go through `unknown`: `tsc` refuses the direct conversion as insufficiently overlapping (`StoredLogEvent` vs `LogEventWithParsingFailure`), which is exactly the refusal the stored type was minted for. I verified that by trying the single `as` and reading the error.

**Tests** — `packages/core/test/savedStreamIsRawOnly.test.ts` (7 cases), written at the keeper seam in the `streamCacheWorld` style, over a chain that serves DECODED logs:

- events handed to `saveNewEvents` carry none of the three fields; also for reorg retractions, and for the retractions `streamRemainderOf` builds from a batch the fold refused (the two other producers of events at the save path);
- the events are NEW objects (what `process` was handed on that same tick is still decoded, and the served objects are untouched);
- membership, order and raw halves are what was fetched;
- the handed `lastSync` carries a stripped window and unchanged cursor numbers;
- the strip does not mutate: the state keeper's persisted window is still decoded and the live window is still populated on the next cycle.

I checked both traps are actually caught: with the strip removed 4 of the 7 go red; with a *mutating* strip substituted, the two non-mutation cases go red.

**Changeset** — `.changeset/the-stream-is-saved-without-its-decoded-half.md` (`@etherfold/core: minor`): a behaviour change to what core persists, no type change, no migration, reads still tolerate a decoded half.

## Drift check

No drift. `StoredLogEvent` landed as the task assumes (blocker `the-stored-event-type-refuses-a-decoded-event`, commit 241e684) and `reparse` is already widened; the seam still declares decoded events on both halves; the load path reads a stored `lastSync` for its three block numbers and `context` only (`promiseToLoad` both branches, `promiseToFollow`, and `replay`'s own docstring); ADR-0034 and ADR-0035 (as amended) are consistent with stripping the window rather than storing one.

## Verification

`pnpm build`, `pnpm typecheck`, `pnpm test` (all packages, browser included: 128 pass unchanged, invalidation tests among them), `pnpm format:check` and `pnpm check:adr` all pass. `pnpm changeset status --since=main` reports "no changesets found" **only because the new changeset file is still untracked** (changesets reads `git diff`, which cannot see it); I verified in a throwaway clone under `/tmp` that once these four files are committed it reports `@etherfold/core` at minor and passes. Working tree holds only the four intended paths.

## Decisions

- **The stripped-cursor type lives INTERNAL and unexported, not in `types.ts`.** `storedLastSyncOf` needs a return type whose window says STORED, so I minted `StoredLastSync` / `StoredEventBlock` in `packages/core/src/internal/stream/strip.ts` rather than beside `StoredLogEvent` in the public types. Alternative considered: exporting it from `@etherfold/core` now. I rejected that because the *next* task (`the-stream-seam-takes-only-the-stored-event`) explicitly owns the choice between "leave the seam's window typed decoded" and "give the seam a stored-window variant", and publishing the name here would pre-empt it and grow the public API for a shape nothing outside core can use yet. Touches: that task, which may promote or delete these two types. They carry no ABI type parameter deliberately — the ABI is what the decoded half was made under.
- **The strip module is a new file under `internal/stream/`, not a function in `internal/engine/utils.ts` and not in `src/stream/`.** `src/stream/*` is entirely re-exported from the package index, so putting it there would have published it; `internal/engine/utils.ts` is the fold's grab bag and this is a storage-boundary rule. Touches: only the import in `indexer.ts`.
- **`context` is carried over by reference into the stripped cursor.** Nothing is stripped from it, so copying would buy an allocation and no guarantee, and handing the whole `lastSync` over already shared it. Alternative considered: a defensive clone, rejected as scope creep with no failure it prevents (the load path's `context.source` mutation predates and is unaffected by this change).
- **The strip runs AFTER `streamRemainderOf`, not before.** So the retractions that helper synthesises from `streamWrittenNotProcessed` are stripped too, and its `sameEvent` comparison keeps reading the decoded in-memory list. Order is not observable today (the comparison reads raw fields only), so it is pinned by a test rather than by a comment.


wighawag:

### Acceptance criteria

1. **Keepers receive raw-only events: no `args`, no `eventName`, no `decodeError`** — met, including the two cases easiest to miss: events that are RETRACTIONS the chain reorged away, and retractions of a batch the processor never accepted.
2. **The stripped events are NEW objects; what the processor was handed on the same tick still carries its decoded half** — met, and asserted from inside `process`, which runs after the save on that same tick, so a strip over shared references would show up there. It also checks the objects the chain served are untouched.
3. **The `lastSync` a keeper is handed carries a stripped window, and the strip does NOT mutate** — met, and asserted on the sharp edge: the STATE keeper's persisted window is read from the very object the stream save was handed one line earlier, and the LIVE window is confirmed still populated afterwards.
4. **`fetchFrom` answers the same membership, order and raw halves** — met.
5. **A changeset** — met.
6. **Tests in the repo's style** — met.

### The expand-then-contract boundary, which is what I said I would check

This is the task that cannot typecheck on its own, and it is handled exactly as instructed.

**One temporary assertion, at the one boundary.** It sits on the single `saveNewEvents` call, covers the whole argument the saver takes rather than the batch and the window separately, and carries a comment saying it is temporary, why it exists, and — by name — which task removes it (`the-stream-seam-takes-only-the-stored-event`). The strip's own locals and return types say STORED throughout, so the lie is confined to the seam and does not leak into the new module.

**It did not take any of the three forbidden escapes.** The seam is not widened, it is not narrowed early, and the strip's output is not declared to be decoded events.

The comment also records something worth having: a DIRECT conversion is refused by the compiler as insufficiently overlapping, which is why it goes through `unknown`. That refusal is `StoredLogEvent` working — the type minted one task ago demonstrably will not accept a decoded event, and the assertion has to say so out loud to get past it.

Two other assertions exist, both inside `storedEventOf` and neither at the seam: widening the input union to destructure the three optional keys off it, and asserting the rest-spread result. Both are the standard shape for "rest of a union minus keys", which TypeScript cannot express, and both mirror the identical pattern `reparse` has carried for some time. Not scope creep.

**I will verify at task 5's Gate-3 that the seam assertion is actually deleted** rather than quietly becoming permanent.

I made the strip mutate in place — `delete` the three keys off the same references instead of building new objects — which is trap (1) and trap (2) at once, and is the failure the task warns has "already been paid for once in this codebase". Both cases go red: `are NEW objects…` and `does not MUTATE…`, with 5 others green.

That second failure is the one that matters. An in-place strip empties the LIVE reorg window, because the state keeper is handed the same object on the same tick — which takes `checkTxInclusion` and the next cycle's retraction derivation down with it, silently. The test catches it by driving another `indexMore` and checking the live window is still populated, rather than by inspecting the strip.

### One design call worth recording

`context` is carried over BY REFERENCE in the new `LastSync`, deliberately: nothing is stripped from it, so copying it would buy an allocation and no guarantee. That is the right line to draw — the copy exists to prevent mutation of what is stripped, not to deep-clone for its own sake.

Merging. One task left, and it is the one that removes the assertion above.

PR #87, opened 2026-09-06
feat(the-stream-seam-takes-only-the-stored-event): The stream seam take…

wighawag:
Task: `work/tasks/done/the-stream-seam-takes-only-the-stored-event.md`

All gates are green. Summary of what I did.

**Reality check first (no drift found).** Both blockers had landed: `StoredLogEvent` exists and refuses a decoded event, `reparse` is already widened, the indexer already strips both halves at the save call site behind one temporary cast, and `replayStream` has already left the seam (ADR-0059). The previous task had even minted `StoredEventBlock` / `StoredLastSync` internally in `internal/stream/strip.ts` with a docstring explicitly deferring the seam's window decision to this task. Nothing in ADR-0034/0035/0044 contradicted the task.

**The seam** (`packages/core/src/types.ts`): `StreamFetcher` and `StreamSaver` now speak `StoredLogEvent[]` and `StoredLastSync`; `StoredEventBlock` / `StoredLastSync` moved from the internal strip module to the public types (used by these two function types and nowhere else). `LastSync` is untouched for the processor, load and wire callers. `EmittedLog` is untouched.

**Implementations moved with it:** the segmented keeper (`StreamSegment` is now `{events: StoredLogEvent[]}`, non-generic; `carryForward` / `cursorRecord` take `StoredLastSync`; the readback assertion at `isSegment` is documented as where reads tolerate a decoded half), the server's `storedEmissionStream` (`storedLogOf` now returns `StoredLogEvent`; the bounded `ReplayChunk` port — a different seam — gets one documented widening on its own path), and the browser keeper + degrading/read-only wrappers, which are generic and needed no change.

**Consumers widened, not asserted:** `emissionMarkOf` and `hasAlreadyFolded` take either shape. The three `fetchFrom` call sites build the engine's own cursor via a new `cursorFromStream` (which also replaces the in-place mutation of the object a keeper returned). The temporary cast at the save call site is **gone entirely** — no residual conversion at that boundary.

**Tests:** `storedLogEvent.test.ts` extended with four `@ts-expect-error` refusals *at the seam* (decoded fetcher, decoded window, decoded saver, decoded `ExistingStream`) plus the accepted stored forms; new `packages/core/test/anOldSegmentStillReplays.test.ts` writes segments the previous version's way (decoded halves and all) and proves they are served, not rewritten, replayed end to end out of the cache with no re-fetch, the stale `args` dropped, and a later batch appended in the new shape beside them. Every other stream test moved by type annotation only.

**Also:** `CONTEXT.md`'s **stream / keepStream** entry now names `StoredLogEvent` / `StoredLastSync`, `docs/adr/0060-…` records the durable why, and a `major` `@etherfold/core` changeset names what a third-party keeper implementor has to change.

`pnpm build`, `pnpm typecheck`, `pnpm test` (48 core files, all packages green), `pnpm format:check`, `pnpm changeset status` and `pnpm check:adr` all pass. Working tree holds only the intended changes.

## Decisions

**The `lastSync` window: option (b), a stored-window variant on the seam (`StoredLastSync` + `StoredEventBlock`), rather than (a) leaving it typed `LastSync<ABI>` with a permanent documented cast.** Chosen because under (a) the seam's own type would *lie about the value*: core strips the window at runtime, so an implementor reading the declared, non-optional `unconfirmedBlocks[].events[].args` would find it always `undefined` — the same class of defect this spec exists to remove, re-introduced one field deeper. The cost is one more cursor-shaped type on the public surface; the gain is that the save boundary needs no conversion at all, so the previous task's assertion is deleted rather than re-documented as permanent. `LastSync` is not forked or re-meant for anyone else, and both directions of the seam agree (a fake that echoes back what it was handed now types cleanly). Touches: `@etherfold/core`'s public types, every keeper implementation and fake, and `@etherfold/browser`'s re-export list. Recorded as ADR-0060.

**Where the two cursor types meet, they meet by CONSTRUCTION, not by widening `replay`.** The alternative was widening `IndexerGeneration.replay`'s second parameter (and `_feed` / `promiseToFeed` / `container.replay`) to `LastSync<ABI> | StoredLastSync`. Rejected: it spreads a second cursor shape through the engine and the public replay API to express something none of them reads — `replay` uses only the three block numbers and the context, and rebuilds the window by walking the events (ADR-0042). Instead `cursorFromStream` builds the engine's cursor from the stored one with an empty window, which costs nothing (no keeper stores a window, ADR-0035 as amended) and additionally removes the pre-existing in-place mutation of the object a keeper returned. Touches: the three `fetchFrom` call sites in `indexer.ts` only.

**"Writes narrow, reads tolerate, nothing is migrated" is pinned as a rule, not left implicit.** Segments written before this keep `args`/`eventName` forever; a keeper serves them (an extra key on an event is not the "unparseable segment" damage shape) and asserts the stored type at its own storage-readback boundary. The alternative — a rewriting migration — buys a type that is true of the bytes at the price of a full write pass to delete fields nothing reads, on a substrate where a tab can be evicted mid-pass. The accepted trade is that the stored type describes what goes IN, not what is guaranteed on disk. Recorded as ADR-0060 and pinned by `anOldSegmentStillReplays.test.ts`.

**The server's bounded replay port keeps its decoded declaration, with one documented widening.** `readRange` feeds `ReplayChunk`, a *different* seam (ADR-0056) this task does not narrow, so the same raw rows are widened on the way out to it with a comment saying why. Narrowing `ReplayChunk` would be scope creep into the rebuild port; leaving `storedLogOf` claiming `LogEvent` would have been the dishonest half. Touches: `@etherfold/server`'s `storedEmissionReplaySource` path only, no behaviour change (`GenerationRebuild` re-decodes with `reparse` regardless).

**Test fakes: `makeLog` in the core stream-cache world now returns `StoredLogEvent`.** It always built a raw log and laundered it through `as unknown as LogEvent<Abi>`; typing it honestly meant retyping a handful of local annotations in `pause`/`updateIndexer`/`streamCache` tests (which never decode) rather than maintaining two builders. The browser stream tests' `event()` moved off `as never` onto `StoredLogEvent` for the same reason — cheap, and it stops those call sites being blind to the seam's event half (the keeper is still passed as `never` at some call sites there, which is why the compile-time proof lives in core's type test).


wighawag:
APPROVE ✅ — Gate-3 (conductor's diff-vs-criteria review). CI `verify` COMPLETED/SUCCESS. Final task of the spec.

### The thing I said I would check, checked first

**The temporary assertion from #86 is DELETED, not re-documented as permanent.** The call site now reads:

```ts
saved = await keepStream.saveNewEvents(source, {eventStream: toStore, lastSync: lastSyncToStore});
```

No `as unknown as`, no `TEMPORARY` comment. The expand-then-contract completed exactly as the two tasks planned it between them — and better than I expected: the two types #86 defined privately in `strip.ts`, with a note saying "how the keeper SEAM declares its window is decided by [this task], which owns that choice", have been PROMOTED into `types.ts` and are now the published seam types. The handoff was designed and honoured rather than improvised.

### Acceptance criteria

1. **The seam takes only the stored event** — met: `StreamFetcher` and `StreamSaver`, and so `ExistingStream` and `StreamReader` and every keeper, are declared over `StoredLogEvent` / `StoredLastSync`.
2. **No implementation re-typed to a SUPERTYPE — neither the base event type nor `EmittedLog` — and the seam not widened to a union re-admitting the decoded half** — met. I grepped the whole `src` diff for both escape hatches and for `as unknown as`: none. `EmittedLog` keeps its own meaning and its own callers on the emission-append path, untouched, exactly as your decision required.
3. **Segments written before this keep working untouched** — met, and comprehensively: an old segment is SERVED rather than refused, is NOT rewritten (the decoded half is still on disk after a read), replays out of the cache so an upgrade costs no re-index, has its decoded half IGNORED rather than refused or trusted, and new events append in the new shape beside the old ones.
4. **`CONTEXT.md`'s stream/keepStream entry names the stored-event type** — met.
5. **A changeset** — met.
6. **Tests in the repo's style** — met.

I widened the saver's `eventStream` back to `StoredLogEvent[] | LogEvent<ABI>[]` — the union re-admitting the decoded half that criterion 2 forbids. **Six typecheck errors**, including the real segment keeper in `stream/segments.ts` and three fakes. The seam cannot be widened without the repo going red.

### The two decisions I looked at hardest

**`StoredLastSync` on the seam, rather than leaving the window typed `LastSync<ABI>` with a permanent cast.** This is the right call and the reasoning is the strongest in the diff: under the cheaper option the seam's own type would LIE ABOUT THE VALUE — core strips the window at runtime, so an implementor reading the declared, non-optional `unconfirmedBlocks[].events[].args` would find it always `undefined`. That is the same defect class this whole spec exists to remove, re-introduced one field deeper. The cost is one more cursor-shaped type on the public surface; the gain is that the save boundary needs no conversion at all, which is precisely why #86's assertion could be deleted rather than made permanent.

**The two cursor types meet by CONSTRUCTION, not by widening `replay`.** Widening `IndexerGeneration.replay` (and `_feed`, `promiseToFeed`, `container.replay`) to `LastSync<ABI> | StoredLastSync` would have spread a second cursor shape through the engine and the public replay API to express something none of them reads — `replay` uses three block numbers and the context, and rebuilds the window by walking the events (ADR-0042). Building the engine's cursor from the stored one with an empty window costs nothing, since no keeper stores a window, and it incidentally removed a pre-existing in-place mutation of the object a keeper returned.

### One documented widening, correctly scoped

The server's bounded replay port keeps its decoded declaration: `readRange` feeds `ReplayChunk`, a DIFFERENT seam (ADR-0056) this task does not own, so the raw rows are widened on the way out with a comment saying why. Narrowing `ReplayChunk` would be scope creep into the rebuild port; leaving `storedLogOf` claiming `LogEvent` would have been the dishonest half. No behaviour change — `GenerationRebuild` re-decodes with `reparse` regardless.

### ADR-0060 earns its place

It records the part a reader finding only the types would get wrong, as one rule in three parts: **writes narrow from here on, reads tolerate a decoded half and ignore it, nothing is migrated.** The migration argument is the honest one — rewriting would charge every existing deployment a full write pass to delete bytes that are already ignored, on a substrate where a tab can be evicted mid-pass — and the accepted trade is stated plainly: the stored type describes what goes IN, not what is guaranteed on disk. Pinned by `anOldSegmentStillReplays.test.ts` rather than asserted in prose.

Merging. That completes all five tasks of `the-stream-stores-only-what-the-node-said`.

PR #88, opened 2026-09-06
tasking(a-generation-can-be-seeded-from-a-published-artifact): A genera…

wighawag:
Task: `work/tasks/done/a-generation-can-be-seeded-from-a-published-artifact.md`

Auto-tasked `a-generation-can-be-seeded-from-a-published-artifact` into 5 backlog tasks.

## Tasks

- **decide-what-a-mismatched-seed-digest-does** — Decide what happens when a seed and the client disagree about which stream it is
  - covers: US-4
  - blockedBy: measure-what-a-published-stream-costs-to-install-and-pick-its-shape
- **decide-who-verifies-a-stream-seed-and-against-what** — Decide who verifies a stream seed, and against what
  - covers: US-3
  - blockedBy: decide-what-a-mismatched-seed-digest-does
- **emit-the-sliced-build-plan-for-seeding** — Emit the sliced build plan for seeding a generation, and say whether it is buildable
  - covers: US-5
  - blockedBy: decide-who-verifies-a-stream-seed-and-against-what
- **measure-what-a-published-stream-costs-to-install-and-pick-its-shape** — Measure what a published stream costs to install, and pick its wire shape on that evidence
  - covers: US-2
  - blockedBy: pin-the-seam-a-published-stream-arrives-through
- **pin-the-seam-a-published-stream-arrives-through** — Pin the seam a published stream arrives through, and how it enters a keeper
  - covers: US-1

## Dependency graph

- Keystone: pin-the-seam-a-published-stream-arrives-through
- decide-what-a-mismatched-seed-digest-does ← measure-what-a-published-stream-costs-to-install-and-pick-its-shape
- decide-who-verifies-a-stream-seed-and-against-what ← decide-what-a-mismatched-seed-digest-does
- emit-the-sliced-build-plan-for-seeding ← decide-who-verifies-a-stream-seed-and-against-what
- measure-what-a-published-stream-costs-to-install-and-pick-its-shape ← pin-the-seam-a-published-stream-arrives-through


wighawag:
APPROVE ✅ — Gate-3 on the TASKING. CI `verify` COMPLETED/SUCCESS. The review loop converged after ONE pass.

### The decomposition

Five tasks, one per story, in a strictly linear chain — which is the right shape for an exploration spec, where each decision is an input to the next:

`pin-the-seam` (story 1) → `measure-and-pick-the-wire-shape` (2) → `decide-what-a-mismatched-seed-digest-does` (4) → `decide-who-verifies-a-stream-seed` (3) → `emit-the-sliced-build-plan` (5).

The two reorderings against the spec's numbering are justified rather than arbitrary: whether a seed carries its own digest is a WIRE-SHAPE question, so the mismatch rule follows the shape; and what you verify depends on what identity the seed carries, so verification follows the mismatch rule.

Every task carries `promptGuidance.testFirst: false`, which is correct and shows the tasker understood the spec KIND: the deliverable here is a decision captured where the next author finds it, not shipped capability, so test-first would be a category error.

### It caught the drift — including one I had missed

I flagged one stale premise before tasking (the spec says `replayStream` returns an `ExistingStream`, which stopped being true when `the-stream-stores-only-what-the-node-said` landed). The tasker found that AND a second one I had not checked:

1. **`keepStateOnIndexedDB(name, remote)` is DELETED** (ADR-0037). The spec's story 1 offers it as one of three candidate seams. The task names the live precedent instead — `bootstrapFromSnapshot` in `@etherfold/processor-entities`, with its `SnapshotLocation` / `BootstrapOptions` / `NotBootstrappedReason` types.
2. **`replayStream` returns a `StreamFixtureReader`, not an `ExistingStream`** (ADR-0059), and the keeper seam now takes only `StoredLogEvent` (ADR-0060).

What impresses me is that it did not merely CORRECT the second one — it turned it into the task's central design constraint: *a fixture cannot be handed to an indexer as its `keepStream`, so seeding necessarily WRITES into a keeper*. That is the real consequence of last week's work for this spec, and it sharpens story 1 from "pick a loader" into "pick a loader AND decide the install". The task then pins what the install must decide: how a fixture's decoded events reduce to stored events, which hierarchical stream address they land under, how the batch is cut into segments, and what is written beside them so a seeded client does not re-scan from its start block.

It also tells the builder to RE-VERIFY rather than trust that paragraph, and to route to needs-attention if a third drift turns up. That is the right posture for a launch snapshot.

### The measurement-honesty requirement, honoured properly

The spec is unusually insistent that story 2's measurement not be oversold, and the task delivers it in the body AND as an acceptance criterion: desktop Playwright is a PROXY, every number must be labelled as one, and what it under-reports must be stated — real mobile memory pressure and eviction, slower storage, background tab suspension — or a real device must be named. "Do not present a throttled laptop as a phone."

It also names what to measure rather than leaving it vague: fetch, `JSON.parse` with the tagged-bigint revive (noting format 2 tags bigints, so the revive is not free), the reduction to stored events, and the keeper writes — for both the single-document and chunked variants, plus what resumability costs and whether a partially installed seed is coherent or must be discarded. And it says why peak memory matters as much as wall-clock: 20.5 MB parsed into JS objects is several times its own size in the heap, which is the number that decides whether a low-memory device survives.

### Story 5 is framed honestly, which is the whole point of an exploration spec

`emit-the-sliced-build-plan-for-seeding` is titled "and say whether it is buildable", and the body means it: the follow-on draft lands in `work/specs/proposed/` (staging, so no auto-tasker races a human), every story in it must be committed and build-taskable, and **if part of the capability is still not confidently buildable after the four decisions, do NOT pad the spec with it** — say so plainly and leave it out. "That is a legitimate, useful outcome, and it is far cheaper than a story an agent builds convincingly and wrongly."

That is exactly the failure mode this spec was reframed to avoid, carried all the way into the last task.

Merging. The five land staged and ungated; I will promote them into the pool.

PR #95, opened 2026-09-07
feat(the-stored-event-strip-is-exported-from-core): The stored-event st…

wighawag:
Task: `work/tasks/done/the-stored-event-strip-is-exported-from-core.md`

## What I did

**Checked the task's premises first (no drift).** `storedEventOf` / `storedStreamOf` are still only in `packages/core/src/internal/stream/strip.ts`, nothing outside core imported them, ADR-0060 still governs the keeper seam (`StreamSaver` / `StreamFetcher` speak `StoredLogEvent` / `StoredLastSync`), and ADR-0063 still names the export as a build item. The spike's copy is still there (`docs/spikes/pin-the-seam-a-published-stream-arrives-through/install.mjs`), untouched.

**Test first, from outside core.** `packages/browser/test/storedStripIsPublished.test.ts` imports both functions plus every type in their signatures (`LogEvent`, `StoredLogEvent`, `IndexingSource`, `StoredLastSync`, `StreamSaver`) from `@etherfold/core`, i.e. through the package `exports` and `dist/` exactly as an installed dependency resolves them. It ran red first (`storedStreamOf is not a function`, the ESM link failing on the closed entry). It pins: the decoded half is dropped and the raw half kept whole; a NEW object is built so the caller's event (the one a processor is about to fold) still carries `args` / `eventName`; every element of a batch is rebuilt, not just the array; and the strip's output is handed to a real `StreamSaver<TestABI>` with no cast, which is the shape an installer outside core writes through (ADR-0063). Half the file is run by `pnpm typecheck`, so an unpublished type in either signature would fail here.

**The export.** `packages/core/src/index.ts` re-exports the two functions out of `internal/stream/strip.js`, in the stream group, with a WHY-carrying JSDoc in the entry's established tone: what justifies it (an installer/producer outside core must apply the same rule, ADR-0060), what does not (the loader lives inside core), what stays internal and why, and where the gate is.

**Changeset:** `.changeset/the-stored-event-strip-is-exported.md` (`@etherfold/core` minor, `@etherfold/browser` patch).

**Unchanged:** nothing inside core switched implementation; `indexer.ts` still calls the internal module, `storedLastSyncOf` is untouched, and no test that pins "the stream stores only what the node said" moved.

Gate: `pnpm format:check`, `pnpm check:adr`, `pnpm build`, `pnpm typecheck`, `pnpm test` all green (browser 21 files, core 50 files, whole workspace passing). `pnpm changeset status --since=main` cannot see untracked files, so it reports the pre-commit state as uncovered; both changed packages (`core` via `src/index.ts`, `browser` via the new test) are named in the changeset, which is what it checks once committed. Working tree holds only the three intended paths.

## Decisions

- **The cursor strip `storedLastSyncOf` is NOT exported.** The seam's cursor half is built by an installer rather than stripped from a live one: the spike (and ADR-0063's three block rules) constructs the `StoredLastSync` literally, with an empty window and the capture's own block numbers, so an outside caller has no cursor to strip. Publishing it would offer a tool for a job nobody outside core has, and would invite an installer to strip the engine's live window, which is the exact hazard `storedLastSyncOf`'s own JSDoc warns about. Alternative considered: export all three now (the task allows it as an in-scope decision). Touches: the next task's seed PRODUCER, which is the caller that could reverse this by needing it; widening later is one line plus a changeset, so this is cheap to reverse.
- **Re-exported from `internal/`, module not moved.** `packages/core/src/index.ts` already publishes several helpers straight out of `internal/` with a WHY block (`resolveStreamConfig`, `streamConfigHashOf`, `sameWireContext`, `cursorSyncedThrough`), and a move would have touched `indexer.ts`'s import and split the strip away from `storedLastSyncOf`, which stays internal and shares its rationale. Alternative considered: move `strip.ts` to `src/stream/` and `export *`, which would then have published the cursor strip too, i.e. the opposite of the decision above. Touches: nothing outside core.
- **A named export (`export {storedEventOf, storedStreamOf}`) rather than `export * from './internal/stream/strip.js'`.** The star form is what most of this entry uses, but here it would publish the third function silently and make the surface follow whatever the internal module grows next. Touches: future additions to `strip.ts`, which now have to be published deliberately.
- **The reachability gate lives in a CONSUMER's suite (`@etherfold/browser`), not in core's published-type / typecheck-coverage machinery.** Those two workspace-level tests answer different questions (do published `.d.ts` files import only declared dependencies; is every package typechecked), and neither can see whether a named symbol is reachable through the entry. The existing home for that claim is the consumer imports themselves (`packages/cli/test/streamConfigAgreement.test.ts` imports `resolveStreamConfig` from `@etherfold/core`); browser is the consumer that owns a keeper and is the deployment the seeding spec targets, and the spike used its keeper. Alternative considered: a core-suite test parsing `dist/index.d.ts` for the two names, which would gate the same fact without proving a consumer can actually resolve and type it.
- **The changeset lists `@etherfold/browser: patch` for a test-only change.** `pnpm changeset status --since=main` treats any file under a package directory as a change to it, so a new test in browser needs an entry or the gate goes red at land time. Precedent for a test-only entry: `.changeset/the-gate-does-not-assume-an-idle-machine.md` patches every package for a vitest-timeout change. Touches: the release notes for `@etherfold/browser`, which will carry a patch with no runtime difference; the changeset says so explicitly.


wighawag:
APPROVE ✅ — Gate-3 (conductor review, fresh-context reviewer subagent + my own pass)

CI `verify`: SUCCESS. Reviewed the diff against each acceptance criterion of `the-stored-event-strip-is-exported-from-core`, and against ADR-0060 / ADR-0063.

Per criterion:

1. **Importable from the package entry** — `packages/core/src/index.ts` adds a named `export {storedEventOf, storedStreamOf} from './internal/stream/strip.js'`. `packages/core/package.json` `exports["."]` maps to `./dist/index.js` with no `development` condition, so a consumer really does reach the built entry.
2. **A gate outside core catches un-publication** — `packages/browser/test/storedStripIsPublished.test.ts`. This was the criterion most likely to be faked by a path alias, so it was checked empirically rather than by inspection: `packages/browser/vitest.config.ts` sets no `resolve.alias`, `tsconfig.base.json` has no `paths`, and `import.meta.resolve('@etherfold/core')` from `packages/browser` resolves to `packages/core/dist/index.js`. Run against `main`, `storedEventOf` is `undefined` there — the test fails without this change, so it is a real gate and not a tautology.
3. **JSDoc states the WHY** — it names ADR-0060, the committed `docs/spikes/pin-the-seam-a-published-stream-arrives-through/install.mjs` duplication as the evidence, and explicitly declines to justify the export by the loader (which lives inside core). Matches the in-place-WHY pattern already used four times in this entry.
4. **Every type in the exported signatures is itself published** — the strip names exactly `Abi`, `LogEvent<ABI>`, `StoredLogEvent`; all three are on the entry's export surface, as are every annotation the new test imports. `tsconfig.typecheck.json` covers `test/`, so this is enforced rather than asserted.
5. **Existing behaviour unchanged** — no file under `packages/core/src` other than `index.ts` is touched; the strip is re-exported, not moved, so `indexer.ts` still reaches it internally and no "stores only what the node said" test changes meaning.
6. **Changeset present** — `@etherfold/core: minor`, `@etherfold/browser: patch` (the latter for the test-only file, which `changeset status --since=main` requires).

On the one decision worth checking across tasks: withholding `storedLastSyncOf` is correct. The next task's seed envelope carries the stored context verbatim (no strip needed) and ADR-0063 has the installer *construct* its `StoredLastSync` with an empty window, so the producer needs `storedStreamOf` and not the cursor strip. Widening later is one line plus a changeset.

Non-blocking, recorded not acted on: this test's meaning is build-order dependent (it reads `dist/`), which the repo's `build → typecheck → test` sequencing already handles and `CONTEXT.md` already documents.

PR #96, opened 2026-09-07
feat(a-generation-runs-with-no-stream-keeper-at-all): A snapshot-seeded…

wighawag:
Task: `work/tasks/done/a-generation-runs-with-no-stream-keeper-at-all.md`

Done. The task's premises held against current reality, so I built it.

## What I checked first (drift)

Every premise the task rests on is still true in the tree: `keepStream` is optional on `IndexerGeneration`'s config and every read/write/clear path is guarded by `if (this.config.keepStream)` (`packages/core/src/indexer.ts`, both load branches, `promiseToFollow`, `promiseToSave`), `promiseToSave` answers `'skipped'` with no keeper, `bootstrapFromSnapshot`/`openAndBootstrap` ship in `@etherfold/processor-entities`, the stream address is still `['stream', <indexer-name>, <streamDigest>, <ordinal>|'cursor']`, and the browser suite still drives the real IndexedDB keeper on `fake-indexeddb` with shared helpers that build the hook with or without one. No path assumes a keeper, so nothing needed repairing and there was no design call to route out.

## What I built

**`packages/browser/test/snapshotOnlyMode.test.ts`** (new) — the mode NAMED (`snapshotOnlyClient`, in a file and a `describe` a reader finds by searching "snapshot-only") and asserted across three cases, each with its own fresh state database and its own indexer name:

1. **First run** — a snapshot published from a publisher tab at block 102, a client bootstrapped from it through `openAndBootstrap` and indexed forward over the fake chain with NO keeper. It lands on exactly the applied rows of the control (the neighbouring mode: same snapshot, same events, a real `keepStreamOnIndexedDB` under it), every row applied once, and it resumes from the snapshot's cursor.
2. **Reorg inside the finality window** — branch A to tip 105, then branch B: block 104 is asserted applied and *then* retracted, `0xb104` replaces it, and the revert stops at the snapshot's own block instead of reaching under it. Same state as the kept-stream run.
3. **Reload** — the same database reopened by the same boot path (`kept-local`, snapshot origin still reported), state intact, then continuing onto a new block.

The emptiness claim is proved by READING KEYS, never a spy: no segment at the address a keeper would have used, no cursor record (asserted separately, since a cursor with no segments is a stream too), and the whole `['stream', ...]` keyspace byte-for-byte unchanged across the run, which catches a write under any other name. Each case first asserts the control DID write there, so the instrument is shown to be capable of reading something. I verified the assertions are not vacuous by temporarily wiring a keeper into `snapshotOnlyClient`: all three cases fail, and fail on exactly the emptiness assertion.

**`packages/browser/test/utils/applied.ts`** — `appliedIn` is now typed on the listing it makes (`Pick<EntityStateView, 'listCurrent'>`) so a case can read the state of a store before building a hook over it (the reload check). No behaviour change.

**`.changeset/the-snapshot-only-mode-is-asserted.md`** — see the decision below.

Gate: `pnpm format:check`, `pnpm check:adr`, `pnpm changeset status --since=main`, `pnpm build`, `pnpm typecheck`, `pnpm test` all green (the changeset check verified in a throwaway clone with the work committed, since `changeset status` ignores untracked files). Working tree holds only the three intended files.

## Decisions

**A patch changeset for `@etherfold/browser`, although no behaviour changed.** The task says a changeset only if a published package's behaviour actually changed, and nothing's did — this is tests plus a test-helper type widening. But two enforced repo rules meet here: `pnpm changeset status --since=main` (in the gate) fails outright once a package's files change with no changeset added since main, and `packages/core/test/pendingChangesets.test.ts` REFUSES an empty front matter ("`changeset version` consumes the file and writes it nowhere, so the note is silently lost"). So the escape hatch changesets itself prints (`changeset add --empty`) is closed in this repo by design. I first wrote the empty changeset, found that test, and switched to a patch naming `@etherfold/browser`, whose body states plainly that no runtime code changed and why the entry exists. Alternatives considered: the empty changeset (refused by the repo's own test), and no changeset at all (red gate at land time, a bounce). Precedent for this exact shape: `.changeset/the-gate-does-not-assume-an-idle-machine.md`, a test-config-only change that took a patch across every package with the same disclaimer. What it touches: the release plan (`@etherfold/browser` 0.8.0 → 0.8.1 whenever a release is next cut) and nothing else; it also means the acceptance criterion's "changeset only if behaviour changed" reads as unsatisfiable in this repo for any change that touches a package's `test/` folder, which a future task may want to reconcile in `CONTEXT.md`'s Conventions.

**The snapshot is published at block 102, below the block the reorg replaces.** A snapshot carries no history under its own block, so `revertTo` under the snapshot origin is refused (`RevertBeyondSnapshotError`). The reorg case retracts block 104 (revert to 103), so a snapshot at the branch tip — which is what `entityBootstrap.test.ts` publishes — would have made the reorg case assert that this mode cannot survive the very reorg it has to survive. 102 is also inside the finality window the publisher's own cursor carries, which is what lets the resumed client re-read blocks 100 and 102 without applying them twice. What it touches: nothing in the product; it is a fixture choice, documented at `SNAPSHOT_TIP`. It does however state a real constraint on the mode that the documentation task (`the-snapshot-only-mode-is-documented-with-its-trade`) may want to carry: a publisher must take its snapshot at least the finality depth behind the tip, or a client bootstrapped from it cannot absorb a reorg reaching under it. `bootstrapFromSnapshot`'s optional `finalityDepth` is the client-side half of that and is already built.


wighawag:
APPROVE ✅ — Gate-3 (conductor review, fresh-context reviewer subagent + my own pass)

CI `verify`: SUCCESS. Reviewed against each acceptance criterion and against ADR-0028.

The central claim of this task is an EMPTINESS assertion, and the failure mode that matters is a vacuous one: a probe aimed at the wrong database, store or key prefix finds nothing and passes while proving nothing. So that was attacked directly rather than read:

- The probe (`snapshotOnlyMode.test.ts:200-239`) reads real keys via `idb-keyval`'s `allKeys()`/`get()`, and its key shape (`['stream', name, digest, ordinal]` plus the cursor address) matches what `packages/browser/src/storage/stream/OnIndexedDB.ts` actually writes and what the existing `streamSegments.test.ts` asserts on a POPULATED keyspace.
- The file carries its own control: `expectStreamStoredUnder` runs the same reader over a kept-stream run and asserts it finds segments and a cursor. The instrument is shown reading something before it is used to read nothing.
- Decisive: mutation runs. Wiring a keeper into the snapshot-only client fails all three cases at the segments assertion; wiring one under a name the case never chose fails all three at the keyspace-delta assertion. Both halves are load-bearing.

Other criteria: the equivalence claim is a real comparison run (`keptStreamClient` shares the snapshot and differs only in the keeper, asserted `toEqual`), not an eyeballed value; the reorg-inside-finality and reload cases both drive the real paths and both end in the emptiness assertion; the mode is NAMED (`snapshotOnlyClient`); database names are per-case fresh, and the only shared substrate (idb-keyval's default store, which the keeper hardcodes) is asserted as a before/after DELTA rather than an absolute, so isolation holds. No file under any `src/` is touched, so the "adds no seam, changes no behaviour" constraint is met.

On the one real judgement call — publishing the snapshot at block 102 rather than at the branch tip: this is a true property, not a fixture dodging a defect. Setting it to the tip makes the reorg case throw `RevertBeyondSnapshotError`, which is ADR-0028's documented producer rule ("take snapshots at least the finality depth behind the tip") enforced at `packages/state-store/src/snapshot.ts`. It is a property of the snapshot-aware STORE, not of keeper-absence: the kept-stream control in the same case fails identically. Publishing at the tip would have made this file assert a store-level constraint under the banner of a keeper-absence claim.

`test/utils/applied.ts`: the `EntityStateView` → `Pick<..., 'listCurrent'>` change is a contravariant parameter widening with an unchanged body; all 8 existing call sites still pass and cannot be weakened.

Non-blocking, recorded not acted on:
1. The fixture's publisher takes its snapshot at its own observed tip, and the client does not pass `finalityDepth`, so the consumer-side `inside-reorg-window` guard is never exercised — the mode a reader copies from this file omits half of ADR-0028's two-sided defence. Filed as an observation; it is also material to the documentation task that follows.
2. The patch changeset for a test-only change is unavoidable here: `changeset status --since=main` requires an entry once a package's files change, and `packages/core/test/pendingChangesets.test.ts` refuses the `--empty` escape hatch by design. The Decisions block notes that this makes the criterion's "a changeset only if behaviour changed" unsatisfiable in this repo for any change touching a package's `test/`, which is worth reconciling in CONTEXT.md's Conventions one day.

PR #97, opened 2026-09-07
feat(the-snapshot-only-mode-is-documented-with-its-trade): The snapshot…

wighawag:
Task: `work/tasks/done/the-snapshot-only-mode-is-documented-with-its-trade.md`

Done. Summary of what I did and found.

## Drift check (asked for first)

No drift. The blocking task `a-generation-runs-with-no-stream-keeper-at-all` landed exactly what this task assumed (commit `16bdd1a`): `packages/browser/test/snapshotOnlyMode.test.ts`, which names the mode (`snapshotOnlyClient`), asserts state equivalence against a kept-stream control, proves the empty stream keyspace by **reading keys**, and covers a reorg inside the finality window and a reload. `CONTEXT.md`'s `seeding` entry, ADR-0028, ADR-0063/0064/0065/0066 and the finding all still say what the task quotes. So I described reality, which happens to match.

## What changed

`docs/guide/indexing-in-a-browser-app/index.md`, one new section placed after the two hazards and before the tx-inclusion section (hazard before recipe, per the page's voice):

- **`## Starting from a published snapshot: the snapshot-only mode`** — what the mode is in the project's own register (snapshot-seeded fold, `keepStream` absent, nothing under the stream keyspace, the path most apps should take), with a pointer to the assertion that backs it (`snapshotOnlyMode.test.ts`, and what it proves and how).
- **`### The trade, before the recipe: the generation is a leaf`** — ADR-0028's floor plus no stream beneath, so a later processor-only change is not free and waits for a republished snapshot computed by the new processor; plus what the alternative buys and what it does not.
- **`### How long that wait is`** — the measured cadence table (8,198 publishes / 357 days, median 1.0 h, p90 1.2 h, p99 1.9 h, worst 50.9 h, 99.74% / 99.90%), the 1,802 to 91,527 blocks at 2.000 s/block, cited to `work/notes/findings/what-a-published-stream-seed-costs-to-install.md` **for the numbers only**; then a paragraph, in my own words and citing nothing, that labels the client-side half (that the deployment ran on the snapshot alone) as the maintainer's account rather than a measurement. The page nowhere claims a stream seed is what makes a browser app start.
- **`### Wiring it`** — one copyable block (`openAndBootstrap` inside `createState`, ordered locations with a build-embedded last resort, `finalityDepth`, and no second argument to `createIndexerState`, whose absence is the mode), then the location-risk paragraph (caller owns the location and its risk, `?snapshot=`-style overrides accept an untrusted state source, omission is undetectable) and the `finalityDepth` / publish-below-the-tip paragraph.

Gate-relevant: `pnpm format:check` green, `pnpm check:adr` green, `pnpm changeset status --since=main` green (no workspace package touched, so no changeset is due). No production code changed. All relative ADR links and GitHub blob links verified to resolve; the in-page anchor matches VitePress's slug.

## Out-of-scope signal captured

`work/notes/observations/a-spike-readme-dead-link-breaks-the-vitepress-build.md`: running `pnpm docs:build` to validate my links surfaced a **pre-existing** dead link in `docs/spikes/measure-what-a-published-stream-costs-to-install-and-pick-its-shape/README.md` (a relative link out of the site into `work/`) that fails the site build. Not mine, not fixed, not in the gate.

## Decisions

**The mode is a section INSIDE the browser-app guide, not a new guide page.** The prompt names "the existing browser-app guide under `docs/guide/`" as the destination, and the acceptance criterion asks for "reachable from the browser-app guide", which being in it satisfies maximally. Alternative considered: a new `docs/guide/starting-from-a-published-snapshot/` page linked from the browser-app page and registered in the sidebar. Rejected because it would fork the "wiring an app" reading surface into two pages and require touching `docs/.vitepress/config.mts`'s nav and sidebar (what it touches: nothing, as built). If the page later grows a stream-seed counterpart (slice C), splitting both out into their own page is the natural moment, and this section moves whole.

**No new vocabulary: the page uses "snapshot-only mode", the term `CONTEXT.md`'s `seeding` entry and the landed test already use.** Coherence check done against the glossary, ADR-0028/0063/0066 and the test: `snapshot-only`, `seeding`, `snapshot`, `leaf`, `retention`, `generation` and `stream seed` all keep their existing meanings, and I introduced no flag, key or status. What it touches: nothing new; it is deliberately a reuse.

**The page states that adding `keepStream` on top of a snapshot seed does NOT buy back the free re-fold, and links the `readOnlyStream` self-clear observation.** The criterion asks for "what the alternative buys", and the obvious reader's move ("then also keep a stream") is both ineffective (a stream kept by a snapshot-seeded tab starts at the snapshot's block, so a successor still has nothing to fold below it, and a snapshot is processor-version-keyed anyway) and the exact shape of a known latent defect (`work/notes/observations/a-follower-can-self-clear-the-writers-stream-through-the-read-only-view.md`, and the spec's Further Notes say so). So I named the real alternative as *a stream that reaches back to the source's start block*: self-indexed from that block, or installed from the unbuilt stream seed. Alternative considered: stating only the positive ("a stream underneath makes the change free") and staying silent on the neighbouring combination, which would have left the guide quietly recommending a defect path. What it touches: the future stream-seed slice C documentation (which will describe the mode this section defers to), and `readOnlyStream` / ADR-0044, whose defect the guide now references from a user-facing page. If that defect is fixed or the observation is renamed, this link needs updating.

**The recipe shows `finalityDepth` being passed and states the producer-side rule, rather than mirroring the landed test's fixture.** `work/notes/observations/the-snapshot-only-mode-test-never-exercises-the-inside-reorg-window-guard.md` records that the test publishes at its producer's observed tip and omits `finalityDepth`, so the consumer-side guard never runs there; it explicitly suggests the documentation state the producer rule instead of reproducing the fixture. I followed that. What it touches: the mode's test (a future task may align the fixture with what the guide now documents), and any publisher script, which now has a written rule to meet.


wighawag:
APPROVE ✅ — Gate-3 (conductor review, fresh-context reviewer subagent + my own pass), with two fixes applied on the branch first.

CI `verify`: SUCCESS on the fixed tip. This is a docs PR, so the review went after FACTUAL defects rather than style.

Checked and correct:

- **Every number in the cadence table** verified line by line against `work/notes/findings/what-a-published-stream-seed-costs-to-install.md`: 8,198 publishes / 357 days, median 1.0 h, p90 1.2 h, p99 1.9 h, worst 50.9 h, 99.74% / 99.90% within 2 h / 4 h, 2.000 s per block, 1,802–91,527 blocks. All eight match the source verbatim, attribution included.
- **The leaf claim** matches ADR-0028's retention floor and `CONTEXT.md`'s `seeding` entry, and the cross-reference to the existing "same edit without the blank app" section is faithful to what that section actually says.
- **"Adding `keepStream` on top does not buy it back"** is true at the code, not just plausible: `createSegmentedStream`'s `carryForward` sets `startBlock = lastSync.lastFromBlock` on the first save, so a snapshot-seeded tab's kept stream begins at the snapshot's window and a successor still has nothing to fold below it. The `readOnlyStream` self-clear observation the page links describes exactly that shape, so the hazard is characterised correctly and in scope.
- **The refusal vocabulary is real**: `NotBootstrappedReason` is `'no-locations' | 'unreachable' | 'processor-mismatch' | 'inside-reorg-window'`, and `insideReorgWindow` genuinely only runs when `finalityDepth` is passed — so "omit it and the check never runs" is accurate.
- **The recipe compiles.** The snippet was typechecked verbatim against the built `dist` types with a real processor and source: every imported symbol exists with that name, `openAndBootstrap(store, locations, {processor, finalityDepth})` is the real 3-arg shape, and `keepStream` really does live in `createIndexerState`'s second argument, which is what makes "there is no second argument, and that ABSENCE is the whole of the mode" precisely right rather than merely evocative.
- **The two-tier honesty criterion**, which is the heart of this task, is met exactly: the publisher cadence is stated as measured and cited to the finding; the client-side claim is a separate paragraph explicitly labelled the maintainer's account and "testimony rather than data"; and the finding is cited only for the numbers, never as the source of the distinction it does not itself draw.
- **The prohibition is honoured** — startup is attributed to the snapshot throughout, and the stream seed appears only as what buys the free re-fold, "designed and not built".
- **Links**: all five ADR filenames exist, the in-page anchor matches a real heading, and the absolute GitHub paths exist on main. `vitepress build` was run on this branch AND on `origin/main`: both fail with exactly one dead link, the spike README's, which landed in `00805ae8` and is therefore pre-existing. This PR adds none. The builder filed an accurate observation for it.
- **Zero files under `packages/`**, so the "ships documentation" constraint holds.

Two findings raised by the review and FIXED on the branch before merge (commit `7126490`):

1. The recipe's outcome-union comment omitted `from` from the `bootstrapped` variant. The real type is `{status:'bootstrapped'; at; from}`. Narrowing still worked, but the union was presented as complete when it was not — on a page whose subject is stating things honestly, that is worth the one word.
2. "The wait a processor-only change pays on that cadence is about an hour" understated it. The measured cadence bounds republication of the snapshot the SAME processor computed; a processor-only change additionally waits for the publisher to deploy the new processor and re-index before any snapshot under it exists. Rewritten to say the cadence is the floor on the wait and bounds the last step rather than the first.

Non-blocking, recorded not acted on: ADR-0063 is now superseded in part by ADR-0067, but only on the resumability half, which this page does not invoke; and `pnpm docs:build` is red on `main` with nothing in the acceptance gate running it, which the builder's observation captures and which is worth a task of its own.

PR #98, opened 2026-09-07
feat(a-publisher-emits-a-stream-seed-from-a-captured-stream): A publish…

wighawag:
Task: `work/tasks/done/a-publisher-emits-a-stream-seed-from-a-captured-stream.md`

All acceptance criteria met, gate green (`build` / `typecheck` / `test` / `format:check` / `check:adr` all exit 0).

## Drift check (done first, as asked)

- The strip **is** exported (`packages/core/src/index.ts`, `storedEventOf` / `storedStreamOf`, landed in `4da7b27`); the producer uses it rather than a copy.
- **No** seed envelope existed (no `StreamSeed` / `STREAM_SEED_FORMAT` anywhere).
- ADR-0063 still says `STREAM_FIXTURE_FORMAT` is untouched (its only supersession, ADR-0067, withdraws the *resumability* conclusion, which this task does not rest on); ADR-0064 is unchanged; ADR-0065 is superseded on the trust anchor and byte domain by ADR-0066 exactly as the task body's own "Note on the build pin, superseded" already records. Nothing stale, so I built.

## What I did

**`@etherfold/core`** (`src/stream/seed.ts`, new, exported from the entry): `StreamSeed` with its own `STREAM_SEED_FORMAT` (1), carrying the stored half, the resolved config, the digest label, the coverage, the seed's own stored context, `chainHeadAtCapture` and a typed `StreamSeedProducer` (`kind: 'capture' | 'stored-stream'`, the field ADR-0065's retraction rule is stated against). No index signature anywhere. Plus `serializeStreamSeed` (compact), `streamSeedPayloadOf` (the octets), `streamSeedContentHash` (SHA-256 over those octets, `sha256:<hex>`), and `parseStreamSeed` (throws). `streamDigestOfSourceHashes` is the existing digest rule rephrased over the entries a seed carries, with `streamDigestOf` now defined in terms of it, so nothing about what is hashed, its width or its rendering moved. `StreamFixtureProvenance` gains two **optional** typed keys (`chainHeadAtCapture`, `capturedBy`); `STREAM_FIXTURE_FORMAT` is still 2.

**The producer** (`packages/conformance-workload-stratagems/src/stream-seed.ts`, outside core): `streamSeedFrom` applies the exported `storedStreamOf` and **refuses** to emit unless `streamConfigHashOf` of the config it was handed equals the capture's `lastSync.context.config`; `saveStreamSeed` writes one compact gzipped document (level 9) and returns the digest, the content hash and both sizes. `scripts/emit-stream-seed.ts` (`pnpm --filter @etherfold/conformance-workload-stratagems emit:seed`) prints them.

**The artifact**: `packages/conformance-workload-stratagems/fixtures/stratagems-alpha1.seed.json.gz`, 31,332 events, coverage 12,082,307 → 23,400,000, head at capture 50,968,313. **0.53 MiB gzipped, 20.11 MiB raw** against the finding's 0.54 / 20.1 (the first emit read 0.60 MiB because zlib's default level is 6, not the measurement's 9; fixed rather than accepted). `streamDigest 222f9f7167d2edf981cdcbe9e10c50ec`, `contentHash sha256:956652dae87511829b67d743abe815aa83a11271199ba80a316b36ff40c919ad`.

**Tests**: `packages/core/test/streamSeed.test.ts` (envelope, refusals, hash domain and rendering, digest-rule equivalence), `test/stream-seed.test.ts` (strip, resolved config, the config refusal, disk shape, all temp-file), `test/reference-seed.test.ts` (the committed artifact: parses, digest recomputed, coherence against its own declaration, stored-half-only, capture match, byte-identity with what the producer emits, hash reproduced with `node:crypto` against the printed literal, size band). Changeset in `.changeset/a-publisher-emits-a-stream-seed.md`; `fixtures/README.md` and the `CONTEXT.md` seeding glossary entry name the artifact and its path.

## Decisions

- **Where the producer lives**: `packages/conformance-workload-stratagems/src/stream-seed.ts`, with `scripts/emit-stream-seed.ts` as the emit script and a private `emit:seed` package script. That package already owns the committed capture and its gzip-aware fixture IO, so the seed's file convention (gzip by `.gz` extension) is the same convention stated once. Alternatives considered: a new tooling package (a package for one script), and a CLI verb (explicitly out of scope). Touches: nothing published, and it is deliberately not a CLI command surface; the published capability stays the envelope + the exported strip + the digest rule.
- **Where the artifact landed**: `packages/conformance-workload-stratagems/fixtures/stratagems-alpha1.seed.json.gz`, beside the capture, in exactly one place, and reachable as `ALPHA1.seedPath` (a new optional field on `WorkloadFixture`) so the loader and admission tasks find it without a path literal. Touches both of those tasks.
- **Format number 1**: a fresh sequence, not a continuation of `STREAM_FIXTURE_FORMAT`. Nothing preceded the seed shape, and continuing the fixture's numbering would suggest the two are versions of one format when the point of the split is that they are not.
- **Content hash algorithm and rendering**: SHA-256 (viem's, the primitive `streamDigestOf` already uses; no second primitive introduced), rendered **`sha256:<64 lowercase hex>`**. ADR-0065's withdrawn amendment had said "no prefix", but that amendment is withdrawn wholesale and the task requires a self-describing form; a bare hex literal pasted into a build cannot say which function produced it. `sha256:` is the OCI-digest form and splits on one character. Alternative considered: the SRI form `sha256-<base64>` (rejected: base64 is harder to compare by eye and invites confusion with an SRI attribute the loader does not implement). **This is a contract with the loader and the admission tasks**: they must pin and split exactly this rendering.
- **Content hash byte domain**: ADR-0066's, and not mine to choose: SHA-256 over the **decompressed payload octets**, after transfer decoding and before `JSON.parse`. Enforced in the type, not only in prose: `streamSeedContentHash` takes a `Uint8Array`, so hashing a re-serialised parse is not expressible, and the function lives in **core** rather than in the producer precisely so the loader computes the same bytes with the same code instead of writing a second implementation neither task's tests could catch disagreeing.
- **An unreadable format is a THROW, not data.** `parseStreamSeed` throws, exactly as `parseStreamFixture` does; the **loader catches and reports `unreadable-format`** as data, which is the conversion ADR-0064 already describes for the fixture parser. Alternative considered: returning a discriminated result, rejected because it would put half the refusal vocabulary in core's reader and half in the loader's outcome type. The loader task consumes this: catch around the parse, one refusal for a truncated file, a fixture handed to the wrong reader and a future format alike.
- **Fields added beyond the ones listed**: none on the envelope. `StreamSeedProducer` has three (`kind`, `name`, `at`), and `at` dates when the **events** were produced rather than when the file was written, which is what makes the emit deterministic (a re-emit of an unchanged capture is byte-identical, so a re-emit never moves a pinned hash and any diff is a real one). `StreamFixtureProvenance` gains `chainHeadAtCapture?` and `capturedBy?`, optional and unable to become required.
- **One existing test's allow-list changed**, and it is worth flagging: `packages/core/test/streamIdentity.test.ts`'s "is the ONLY implementation of ITSELF" scans `src/` for files mentioning `sha256`, so adding `streamSeedContentHash` tripped it. The guard exists to stop a **second implementation of the stream ADDRESS**; a content hash over a document's octets addresses nothing and truncates nothing, so I admitted `stream/seed.ts` with that justification **and tightened the guard** by additionally asserting that only `stream/identity.ts` truncates to `STREAM_DIGEST_LENGTH`. The digest rule itself, its inputs, width and rendering are untouched and every other stream-identity assertion passes unmodified.
- **Gzip level 9 for the artifact**: a published artifact is written once and downloaded by everybody, and it is what the wire-shape measurement used, so the emitted size reproduces the finding (0.53 MiB) instead of sitting 12% above it (0.60 MiB at zlib's default) for no stated reason. Touches the reference artifact's bytes, hence nothing pinned (the hash is over the decompressed octets and is level-invariant).


wighawag:
APPROVE ✅ — Gate-3 (conductor review, fresh-context reviewer subagent + my own pass), with one real defect fixed on the branch first.

CI `verify`: SUCCESS on the fixed tip.

**The criterion that looked violated, and is not.** "The digest rule is unchanged ... existing stream-identity tests still pass unmodified", yet `packages/core/test/streamIdentity.test.ts` IS modified. Judged rather than waved through:

- The one modified hunk sits entirely inside the `it` titled "is the ONLY implementation of ITSELF", which asserts *which source files mention the string `sha256`*. It is a meta-guard about source scanning, not a claim about the digest. Every digest-rule assertion in the file (preimage, dedupe, sort, 32-hex width, collision corpus, `0x`-free rendering) is untouched.
- The rule itself is byte-identical, and this was measured rather than reasoned: `streamDigestOf` on **main** over the real capture gives `222f9f7167d2edf981cdcbe9e10c50ec`; the committed artifact declares the same value through the new `streamDigestOfSourceHashes`; and an independent reimplementation of the rule over the seed's own 25 context entries gives the same value again. The extraction is exact — the moved body is character-for-character the old one with `sourceHashesOf(source)` replaced by `entries`.
- The widening is unavoidable once a second, legitimately non-addressing SHA-256 use lands in core, it was disclosed in `## Decisions` rather than smuggled, and the added "only `identity.ts` truncates" assertion is strictly additive (the `hashing` list is still an exact `toEqual`, so no new file can mention `sha256` without reddening the gate).

**The artifact, measured on the committed bytes** rather than taken from the emit log: 559,835 B gzipped (0.534 MiB) against the finding's predicted 0.54, and 21,090,182 B raw (20.113 MiB) against 20.1 — within 1.1% and 0.06%. 31,332 events, matching the capture exactly; **0 of 31,332** carry `args`/`eventName`/`decodeError`, and 0 differ from the capture's event minus those three keys; coverage `{12082307, 23400000}` equals the capture's `lastFromBlock`/`lastToBlock` with the last event at 23,303,136, so the coverage end genuinely sits above the last event-bearing block; `context` deep-equals the capture's. Nothing materially different to report. (Worth recording: the finding's "MB" are MiB — verified against the committed capture's own rows — so the prediction and the result are being compared in the same unit.)

**The reproducibility criterion is not vacuous.** The test pins a *literal* and recomputes with `node:crypto` over the gunzip of the committed file — a different implementation from the `viem` one under test. Reproduced entirely outside the repo: `sha256sum` of the decompressed artifact is `956652da…c919ad`, matching the printed value. Re-running `emit:seed` left the `.gz` byte-identical, so the determinism claim holds and a re-emit cannot silently move a pinned hash.

Also verified: the emit-time config refusal is unconditional, ahead of envelope construction, with no override, and its test drives a real mismatch (`finality: 17` against a capture taken at 12); `STREAM_FIXTURE_FORMAT` is still `2` with `chainHeadAtCapture?`/`capturedBy?` added OPTIONAL, so no fixture-format bump is forced; the producer imports `storedStreamOf` from `@etherfold/core` and contains zero occurrences of the three-key destructure; the envelope carries all seven required fields with no index signature; the changeset covers `@etherfold/core` and correctly omits the private conformance package; and every checkable claim added to `CONTEXT.md`'s `seeding` entry matches what landed.

**One defect found and FIXED on the branch (commit `c5ce7c0`), because it would have been handed straight to the admission task:**

`parseStreamSeed` validated `producer`/`streamConfig`/`streamDigest`/`coverage`/`context`/`eventStream` but **not** `chainHeadAtCapture`, nor the two numbers inside `coverage`. A document lacking the head parsed cleanly and returned an object typed `chainHeadAtCapture: number` holding `undefined`. The next task's capture-depth check is `chainHeadAtCapture - coverage.toBlock < finality`, which on an absent head is `NaN < finality` — **false** — so a seed with no observed head would have been silently ADMITTED by the very check written to refuse it. That is the wrong-but-compiling failure this whole set is meant to avoid, and it is invisible to both tasks' own tests. Now a `typeof` check at the door (not truthiness: `0` is a legal block number), with two tests pinning it, including one asserting a genesis-block seed is still admitted. The committed reference artifact still parses and all conformance tests still pass under the stricter reader.

Also amended **ADR-0066** to pin the content-hash RENDERING as `sha256:<64 lowercase hex>`. The ADR fixed the byte domain but left the rendering unstated, and ADR-0065's withdrawn amendment had said bare-hex-no-prefix — so a strict reader of the live ADRs could have built the loader to expect a bare hex string while the producer prints a prefixed one, and every pinned install would refuse in production and in none of the tests. The rendering is half of what the two ends must agree on; it is now written down where both tasks read it.

Non-blocking, recorded not acted on: the `## Decisions` block says hashing a re-serialised parse "is not expressible", which overstates it — `streamSeedContentHash(streamSeedPayloadOf(parseStreamSeed(text)))` is expressible, the signature discourages it rather than preventing it. The block is the builder's verbatim report so it stays as written; the code's own JSDoc is already honest on this point ("It takes BYTES and not a string, deliberately"). Separately, the new "only identity.ts truncates" regex matches text including comments and is evadable by rephrasing the slice, which is the known hazard in `work/notes/observations/source-scanning-gates-still-match-text-including-comments.md`; it weakens nothing, so it is left alone.

PR #100, opened 2026-09-07
feat(a-seed-that-is-not-for-this-build-is-refused-before-any-write): A…

wighawag:
Task: `work/tasks/done/a-seed-that-is-not-for-this-build-is-refused-before-any-write.md`

All green. Here is what I built.

## What landed

**`packages/core/src/stream/seedInstall.ts`** — every admission check a client can make on its own, all of them before the first `saveNewEvents`, on the one refusal path and the one outcome type the previous task landed. `NotInstalledReason` gains `chain-mismatch`, `stream-config`, `seed-covers-more`, `seed-covers-less`, `integrity-mismatch`, `incoherent`, `inside-reorg-window`. No new status, no throw for an ordinary condition, and no location-based refusal (ADR-0066).

Order inside one location: fetch → integrity (when pinned, over the decompressed octets, **before** the parse) → parse → identity → reach-back → coherence → capture depth → write. The subtree-emptiness gate (ADR-0067) still precedes all of it and precedes any fetch.

- **Identity**: the digest label is *recomputed* from the artifact's own context + resolved config and refused as `incoherent` when the document disagrees with itself; then chain, then config, then the direction from `sourceInvalidationOf(...).stream` taken at the seed's coverage end. Nothing infers "you are out of date".
- **Coherence**: one O(n) pass, two maps, with the ordering/duplicate rules stated over *applications* so a `stored-stream` seed carrying an apply/retract pair stays admissible while a `capture` carrying one is refused.
- **Capture depth**: `chainHeadAtCapture - coverage.toBlock < finality` (the client's own resolved finality).

**`packages/core/src/stream/seed.ts`** — an optional `chain` declaration (`StreamSeedChain`) and `pinnedStreamSeedContentHash`, the read-side of the rendering `streamSeedContentHash` writes.

**`packages/core/src/index.ts`** — `installStreamSeed` + its types are now a public entry point, with the trust contract stated in its JSDoc (caller owns the locations including any runtime override; the hash is optional and only an immutable release-tied artifact can carry one; omission is undefended, so the named host is trusted like the build pipeline).

**Tests**: new `packages/core/test/streamSeedAdmission.test.ts` (28 cases: both directions, each coherence rule with its own failing artifact, the legitimate retraction admitted, the reorg-window synthetic, a table asserting the keyspace is empty after every fetched-document refusal, and the populated-subtree case asserting segments + cursor byte-identical with nothing downloaded). `streamSeedInstallReference.test.ts` gains the pin round-trip against the literal the producer printed (`sha256:956652da…`, not recomputed here), transport invariance on the real artifact, and a wrong-pin refusal. The reference artifact still installs and folds through every check.

Changeset added; `format:check`, `check:adr`, `build`, `typecheck` and the full `test` run are green. (`changeset status --since=main` only sees a *committed* changeset, so it will go green with the runner's commit.)

## Decisions

- **How a caller names the trusted locations: unchanged.** The ordered `readonly string[]` the previous task landed, walked by failover, with a relative hostless path a first-class member. I added no allowlist, no origin predicate and no location-derived refusal (ADR-0066 is explicit that such a reason has no mechanism). What is new is only that the loader's JSDoc now *states* the contract, including that an app accepting a runtime override (`?snapshot=`) has made that choice and the library cannot see or judge it.

- **The seed envelope gains an OPTIONAL `chain` declaration, and the producer was NOT changed.** ADR-0064 requires `chain-mismatch` reported *separately*, but the envelope the producer landed carries no chain: `chainId`/`genesisHash` live only inside the one-way block-0 skeleton hash, which also covers contract addresses and `startBlock`s, so "the block-0 entry moved" cannot truthfully be *named* a chain mismatch. Alternatives considered: (a) report a skeleton difference as `chain-mismatch` — rejected, it re-means the term and would tell a developer who changed a contract address that they pointed at the wrong chain; (b) make the field required and re-emit the reference artifact — rejected, that moves a **published** content hash (`work/tasks/done/a-publisher-emits-a-stream-seed-from-a-captured-stream.md`) and would have made this task both produce and pin the value whose round trip it is supposed to prove. So: optional, checked only when present, and only to *name* a disagreement the digest already decided. **Cost, stated rather than hidden**: nothing in-tree emits it, so the committed reference artifact cannot exercise the sharp reason; a seed omitting it is refused as a direction instead (asserted). **Touches**: the producer (`packages/conformance-workload-stratagems/src/stream-seed.ts`), which should populate it the next time the artifact is legitimately re-emitted, and any future seed publisher.

- **How the direction is computed**: `sourceInvalidationOf(clientSourceHashes, clientConfigHash, seed.coverage.toBlock, seed.context).stream` — valid ⇒ `seed-covers-more`, invalid ⇒ `seed-covers-less`, exactly ADR-0064's rule, with `lastToBlock` = the seed's coverage end so "an added entry at or below the seed's coverage" is what the verdict actually asks. The verdict is *not* the admission rule (exact digest equality is); it only names the direction after the digests have already disagreed. Chain and config are checked first because each is subsumed by the digest and each has an obvious remedy.

- **Which bytes the integrity check hashes**: `streamSeedContentHash(payload)` where `payload` is what `streamSeedPayloadFrom` returns — the decompressed octets after transfer decoding and before `JSON.parse`, which is the domain `streamSeedPayloadOf`/`streamSeedContentHash` (the same functions the producer used) are defined over. It runs **before** the parse, so a corrupted document reports `integrity-mismatch` rather than `unreadable-format`. Match with what the producer printed is asserted end to end, not by construction: the reference test pins the literal `sha256:956652da…` the emit script printed, and the workload package's own suite independently reproduces that literal with `node:crypto`.

- **A malformed pin RAISES rather than refusing** (`pinnedStreamSeedContentHash`, in `seed.ts` beside the function that writes the rendering). A pin is a build constant, so a bad one is a mistake in the caller's source; reported as data it would arrive as `integrity-mismatch` from every location and point a developer at the artifact and the host. It raises before anything is fetched. Alternative considered: comparing literally and letting it refuse — rejected for that misdirection. **Touches**: any caller passing a bare-hex pin from ADR-0065's withdrawn rendering; that is precisely the case it catches.

- **Refusal reasons added beyond the recorded vocabulary: none.** Two *uses* of `incoherent` are mine and worth naming: a seed whose `streamDigest` label disagrees with its own context+config, and one whose `context.config` disagrees with the resolved config it declares. Both are "this document contradicts itself", which is what `incoherent` means, and folding them there avoids two reasons an application would render identically. Which rule failed is logged, not returned.

- **The refusal MESSAGE detail is logged, not carried in the outcome.** The outcome stays `{status, reason}` with nothing else, which is also what makes "the loader never infers you are out of date" assertable (`Object.keys(outcome)`). Alternative considered: a `detail` field — rejected as new public vocabulary for a publisher's debugging problem.

- **How the checks stay O(n)**: the client's own identity (`sourceHashesOf` + config hash + digest) is computed **once per install**, not per location; coherence is a single pass holding two maps (blockNumber → blockHash, `(blockHash, logIndex)` → still-standing), so no nested scan and no sort; capture depth is arithmetic on two numbers the artifact carries.

- **The ordering/duplicate coherence rules are stated over APPLICATIONS, not over every event.** Read literally over all events, ADR-0065's "strictly increasing `(blockNumber, logIndex)`" and "no duplicate `(blockHash, logIndex)`" would ban the apply/retract pair the same ADR refuses to ban. So a retraction is exempt from those two and held to its own rule (a standing application of the same coordinate before it, plus a producer that admits one); the block-hash rule and the coverage rule still apply to every event. Recorded because it is an interpretation of the ADR's list, not a transcription of it.

- **The emptiness gate stays FIRST, so no new check is ever reached against a populated subtree.** That is why the non-destructive case asserts `subtree-not-empty` with nothing downloaded and the rows byte-identical, rather than a digest-mismatch refusal against a populated subtree — which the ordering makes unreachable. Alternative considered: running admission before the probe so an existing-stream client learns *why* the seed is unusable — rejected, it would break the previous task's asserted behaviour (a client that already holds a stream pays no download) and it would gain nothing: the remedy is the same, clear the subtree deliberately. I inherited the `PROBE_FROM_BLOCK` probe and added no second inspection.

- **`CONTEXT.md` corrected on touch**: the seeding entry said `installStreamSeed` is "deliberately NOT exported from the package entry until the admission checks land". That is now false; it names it as public instead. No other glossary text changed.


wighawag:
APPROVE ✅ — Gate-3, after an initial **BLOCK**. The fresh-context reviewer blocked this PR; the defect is fixed on the branch (commit `0a33843`) and CI `verify` is SUCCESS on the fixed tip.

**The block, which was worth having.** This PR added two new admission consumers — `context.source` (iterated to recompute the publisher's digest) and each event's `blockNumber`/`logIndex`/`blockHash` (compared by the coherence pass) — without extending the parse door that the PREVIOUS PR had explicitly written the invariant for ("the fields an admission check does arithmetic on are typed at the door… a comparison that silently passes is the one failure a shallow reader must not hand downstream"). Three JSON-parseable `format: 1` documents were driven through `installStreamSeed`:

| document | before |
|---|---|
| `context.source` a string, not an array | **THREW** `TypeError: entries.map is not a function` out of a public entry point |
| `eventStream: [null]` | **THREW** `TypeError: Cannot read properties of null` |
| an event with no `blockNumber` | **`{status: 'installed', events: 1}`** — rows written |

The third is the serious one. Every coherence rule is a COMPARISON, and a comparison against `undefined` is false, so coverage containment, the ordering rule and the duplicate rule all passed **vacuously** and the document installed. A stored stream is re-folded by every later generation, so that install is permanent — precisely the wrong-but-compiling outcome this set of checks exists to prevent. The first row was also a regression this PR introduced: nothing iterated `context.source` before it.

Fixed by typing the shape where each is consumed: an entry that is not an event is refused `incoherent` before anything is compared to it, and a non-list `context.source` is refused at the parse door as `unreadable-format`, consistent with how the door already types `chainHeadAtCapture` and the coverage numbers. Verified by re-running all three exploit documents: they now return `unreadable-format` / `incoherent` / `incoherent`, as DATA, with the keyspace empty in every case.

Two coverage gaps closed alongside it: the block-number half of the ordering rule had no failing artifact (the only out-of-order fixture varied `logIndex` inside one block, so deleting the block-number comparison left the suite green), and the malformed-event shapes now have their own cases. The changeset was also raised `patch` → `minor`, matching how this repo versions new public surface.

**What held up under attack:**

- **The pin round trip passes end to end**, computed independently with `node:crypto` over the gunzip of the committed artifact: `sha256:956652da…c919ad`, matching both the producer's printed literal and the loader's pinned literal. The gzip-file hash (`dd16001b…`) matches nothing, which is correct under ADR-0066's byte domain. The pin is load-bearing: flipping one character fails two reference tests.
- **Transport invariance is asserted, not assumed**, both on synthetic artifacts and on the real 21 MB one under the producer's literal.
- **Every check is load-bearing**, verified by neutering each in turn: identity → 7 failures, coherence → 7, capture depth → 3, chain → 2, stream-config → 2, and one failure per coherence rule.
- **The direction mapping is not inverted** (`sourceInvalidationOf(...).stream` valid ⇒ `seed-covers-more`), both directions asserted on the same source pair in both orientations, and inverting it fails 5 tests. Nothing infers "you are out of date": the outcome is asserted to be exactly `{status, reason}` via `Object.keys`.
- **The retraction rule is stated against the DECLARATION rather than banned**: a `capture` carrying one is refused, an orphan retraction is refused, and a legitimate apply/retract pair from a `stored-stream` is ADMITTED.
- **The existing test file was strengthened, not weakened** — the only change to `streamSeedInstall.test.ts` replaces a placeholder digest with a real one, so every install case in it now has to pass identity to reach its own assertion.
- **The reference artifact still installs and folds** after the checks land, 31,332 events with `eth_chainId` the only node call.

**On the criterion that looked unsatisfiable, and my judgement.** The task asks for a refusal "on a digest mismatch, a bad pin, incoherence or capture depth" against a populated subtree, but the emptiness gate runs FIRST, so those are unreachable there — a populated subtree can only produce `subtree-not-empty`. I judge the shipped ordering CORRECT and decline to force the criterion's literal wording: running admission before the probe would break the previous task's asserted contract (a client already holding a stream pays no download), and emptiness-first makes the non-destructive property **structural** rather than per-check — no admission check can ever be reached against a populated subtree, so none of them can be what deletes a stream. That is a stronger guarantee than testing each individually. The criterion itself is met by a case that offers a seed failing on identity *and* an unmatchable pin, then asserts `subtree-not-empty`, byte-identical rows, and nothing downloaded.

**On the optional `chain` field.** Sound, and the honesty is the good part: `chainId`/`genesisHash` live only inside the one-way block-0 skeleton hash, which also covers contract addresses and `startBlock`s, so a skeleton difference genuinely cannot be *named* a chain mismatch without telling a developer who changed an address that they pointed at the wrong chain. Optional, not required, so the published content hash does not move — the alternative would have made this task both produce and pin the value whose round trip it is supposed to prove. The stated cost (nothing in-tree emits it, so only synthetic artifacts reach that reason) is real and disclosed, and a seed omitting it is asserted to be refused as a direction instead.

Also amended the observation filed against the previous PR: its claim that the degrading-read hazard is unreachable is WRONG — `createSegmentedStream` wraps its own return in `degradingStream`, so every keeper the install is handed goes through it. What keeps the hazard narrow is that the corrupting window needs `readCursor` to fail inside `fetchFrom` and succeed inside `saveNewEvents`, which is a probabilistic argument rather than a structural one, and that is worth knowing.

Non-blocking, recorded not acted on: `genesisDiffers` is untested; `pickReason`'s ten-entry precedence is never exercised with two competing reasons across two locations; the new non-destructive case writes under one source and installs with another, which reads as if it exercises address derivation and does not (the equivalent case in the previous PR uses one source throughout, so the property is covered); and `StreamSeedChain`'s JSDoc slightly overclaims, since a forged context matching the client is admitted — which is exactly the trust residue ADR-0066 accepts.

PR #101, opened 2026-09-07
feat(the-seeding-outcome-reaches-the-browser-status-surface): The seedi…

wighawag:
Task: `work/tasks/done/the-seeding-outcome-reaches-the-browser-status-surface.md`

Done. Everything is green: `pnpm format:check`, `pnpm check:adr`, `pnpm build`, `pnpm typecheck`, `pnpm test` (whole monorepo) all pass, and the working tree holds only this task's changes.

## What I built

**`packages/browser/src/IndexerState.ts`** (the surface):

- **`SyncingState.streamSeed`**, a new optional field beside `error` and `nonCanonicalGenerations`, carrying `StreamSeedState`: `{status:'installing'}` | `{status:'seeded', at, reachesBackTo, from, events, segments}` | `{status:'refused', reason, direction?}` | absent. Additive; `error` is untouched and no existing field changes meaning.
- **`StatusState.state` gains `'InstallingStreamSeed'`**, so the boot phase is visible on the enum apps already switch on. The phase leaves that value at the terminal outcome (back to `Idle` until the load moves it on).
- **`createIndexerState(..., {seed})`** (`BrowserStreamSeedOptions`: ordered `locations`, optional `expectedContentHash`, `reachBackTo`, `maxEventsPerBatch`, injectable `fetch`) runs `installStreamSeed` inside `init`, before the generation is built, publishing `installing` and then exactly one terminal state. A refusal never gates the boot; a throw (malformed pin, keeper failure) propagates out of `init`.
- `dispose()` clears the field. `direction` is the refusal reason narrowed to `seed-covers-more`/`seed-covers-less`; nothing infers "you are out of date".
- `packages/browser/src/index.ts` re-exports the two core types the public surface now names (`NotInstalledReason`, `StreamSeedLocation`). `installStreamSeed` itself is deliberately not re-exported.

**`packages/browser/test/streamSeeding.test.ts`** (11 cases, real IndexedDB under `fake-indexeddb`, one database/indexer name per case): the published sequence (`installing` → `seeded at N`) read off a subscription; failover to a build-embedded relative path; a three-segment install publishing exactly two states (no byte progress); an app that never seeds seeing nothing change; the no-keeper raise; a refusal with reason **and** direction while state comes up from a published state snapshot and indexes forward (`error` untouched, keyspace empty, nothing deleted); the other direction; a reason with no direction and no invented fields; and the direct path asserted **before** `init()`, **after** `init()`, and refused as `subtree-not-empty` once indexing has started with the existing stream left intact.

**Docs**: a new "Installing a published stream seed, and rendering what it did" section in `docs/guide/indexing-in-a-browser-app/index.md` covering both wirings, what reaches the stores, the trust contract, and the no-progress decision; and the stale "designed and not built" line in the neighbouring section corrected on touch.

**Changeset**: `.changeset/the-seeding-outcome-reaches-the-browser-surface.md` (`@etherfold/browser`: minor).

I checked the task against reality first: the loader (`installStreamSeed`), its `StreamSeedInstallOutcome` and its `NotInstalledReason` vocabulary have all landed as the task describes, including ADR-0067's self-sufficient install, so nothing had drifted.

## Decisions

- **Field name `streamSeed`, not `seed`.** `CONTEXT.md` defines **seeding** as covering BOTH published-artifact shapes (a state **snapshot** and a **stream seed**), and a browser app gets the snapshot outcome from `openAndBootstrap` inside its own `createState`. A field called `seed` on the syncing store would silently mean one of the two. Alternative considered: `seed` (shorter, matches the option name). Touches: the option is still `createIndexerState({seed})` because there it sits under a `keepStream` that already scopes it to the stream; only the published field is qualified.
- **Type name `StreamSeedState`, and the `direction` field is DERIVED.** `StreamSeedState` follows the file's own `SyncingState`/`StatusState` naming; `StreamSeedProgress` was rejected because "progress" is exactly what this surface deliberately does not report, and `StreamSeedOutcome` because `installing` is not an outcome. `direction` duplicates the reason for the two directional reasons (`seed-covers-more`/`seed-covers-less`), typed as `Extract<NotInstalledReason, ...>` so it can never drift from the loader's vocabulary. Alternative: report the reason alone and let each app know which reasons are directional. Touches: nothing outside this package; ADR-0064's "the refusal names a direction" is satisfied either way.
- **A `seed` with no `keepStream` RAISES at `init` rather than being a refusal.** It is a wiring mistake in the caller's own source and no location makes it right; reported as data it would surface as `unreachable` from every mirror and point the developer at the host. This is a new ERROR on a user-visible path, hence recorded. Alternatives considered: silently ignoring the option (invisible), or adding a core refusal reason (would put a browser wiring fault into the loader's published vocabulary). Touches: `createIndexerState`'s `init` contract only.
- **The install runs in `init`, before `openIndexer`, and the phase returns to `Idle` afterwards.** Before the generation is built means the fold finds the stream already there; ordering is not what makes it CORRECT (ADR-0067), which the tests assert by driving the direct path both before and after `init()`. Leaving `InstallingStreamSeed` standing after a terminal outcome would be the one certainly-untrue value; `setupIndexing` moves it on to `Loading` at the next call. Touches: `StatusState` consumers (none in this repo switch exhaustively on it).
- **A refusal leaves the app fully running, and the field is NOT cleared on a reconfigure.** Per ADR-0064 the refusal is reported alongside an otherwise-normal boot: `init` carries on, state comes up from whatever the app's `createState` bootstrapped, and the tab indexes forward (asserted end to end against a published snapshot). `clearSyncingStateForReconfigure` is shared by `updateProcessor`, `updateIndexer` and promotion, and the common reconfigure (a processor change) keeps the very stream this field describes, so clearing there would drop a true report. The accepted residue, documented at the field: after an `updateIndexer` that moves the indexer to a DIFFERENT stream, the field goes on describing the stream the app booted on. Making it per-stream would need a per-generation seed surface, which is beyond this task. Touches: `updateIndexer` callers.
- **On a THROW out of the loader the field is left saying `installing`.** `init` rejects (the caller's own error path), and there is no terminal outcome to report, so inventing one, or clearing the field, would both claim something untrue. Alternative: map it onto `syncing.error`, rejected on the same ground that keeps a refusal out of `error`.


wighawag:
APPROVE ✅ — Gate-3 (conductor review, fresh-context reviewer subagent + my own pass), with three fixes applied on the branch first.

CI `verify`: SUCCESS on the fixed tip.

**The load-bearing behavioural claim holds, and the test is stronger than it looks.** ADR-0064 says a refused seed must not stop the app. The end-to-end case brings a second store up through the real `openAndBootstrap` from a published snapshot at block 102, has the hook refuse a wider publisher's seed (`seed-covers-more`), then indexes to tip and asserts every applied row has `times === 1`. That last clause is what makes it a real test rather than a field check: `applyingProcessor` increments `times` per application and the snapshot already carries blocks 100 and 102, so a fold that restarted from the start block instead of resuming from the snapshot's cursor would read `times: 2`. It also asserts the stream keyspace is empty and `syncing.error` is `undefined`. Making a refusal throw after publishing fails three tests, so "never gates the boot" is enforced rather than decorative.

Also verified: `error` and `nonCanonicalGenerations` are untouched, `clearSyncingStateForReconfigure` is byte-identical, and an app passing no `seed` takes a byte-identical path (asserted by a case that checks the publication sequence is `[undefined]` and that `InstallingStreamSeed` never appears); nothing in the repo switches exhaustively on the widened phase union; all three ordering cases exist and really drive their ordering; both directions are reported as data with no staleness inference anywhere in the diff, including the docs; there is no byte-level progress, pinned by a case using `maxEventsPerBatch: 1` so the install writes three segments and the surface still says nothing; success names the block; no new reactive mechanism; per-case database names on the real `fake-indexeddb` substrate with no reference artifact anywhere near it; and the guide's code samples compile against the real signatures.

**On the two judgement calls I asked to be attacked:**

*The field stays `installing` when the loader throws.* I would ship it. A throw is reachable (a malformed `expectedContentHash` throws synchronously), and the resulting pair reads contradictorily — `status.state === 'Idle'` alongside `streamSeed.status === 'installing'`. But the vocabulary was fixed by the human in "The two wiring questions, ANSWERED", so a fourth terminal is out of bounds; nothing is swallowed, since `init` rejects with a message naming the mistake; and the documented spinner keys on the phase, which does return to `Idle`, so no app written to the guide spins forever. Both alternatives are worse inside the decided vocabulary: clearing says no seed was asked for, which is false, and a synthetic terminal needs a reason the loader has no word for.

*The field is not cleared on reconfigure.* Acceptable. The overwhelmingly common reconfigure is a processor change, which keeps the very stream the field describes, so clearing in the shared helper would discard a true report in the common case to fix a stale one in the rare case. Nothing in the library branches on the field, `dispose()` does clear it, and the residue is disclosed at the field itself. I do think the builder overstated the cost of narrowing it — `updateIndexer` already recomputes the stream digest, so clearing only when the identity actually moved is a few lines, not the per-generation surface the Decisions block describes. Worth a follow-up, not a hold.

**Three fixes applied (commit `7713d6d`):**

1. **The documented recipe would have shown a failure banner to every healthy returning visitor.** The install writes only into an empty subtree, so it succeeds ONCE; every later page load comes back `not-installed` / `subtree-not-empty`. That is truthful and costs no download (the emptiness gate precedes any fetch), but the guide's sample was `if (seed?.status === 'refused') show(explain(...))`, which renders that to a user whose seeding is working perfectly — the exact inverse of "render something true rather than 'loading'". The sample now excludes that one reason and a paragraph explains that it is the steady state and the signal seeding worked, not something to explain to a user.
2. **The `installing` member's JSDoc promised "replaced by a terminal state"**, which is not true on the throw path. It now states the exception, why neither alternative is truthful, and that the phase does clear — which is what an app author needs, and the type member is what they read.
3. **A test asserted half its own name.** "installs BEFORE `init`, and the generation folds the stream that is already there" checked `times === 1` but never that the fold found the installed stream; the case would have passed against a generation that backfilled from block 100. It now pins `chain.ranges` to the tail above the seed's coverage end, so the no-backfill claim is enforced.

Non-blocking, recorded not acted on: `installStreamSeed` costs every browser app ~8.5 KB minified (~2-3 KB gzipped) whether or not it seeds, because `init` references it at runtime so it cannot be tree-shaken — that is the price of the convenience option the human chose, worth knowing rather than acting on; and `CONTEXT.md` is correctly untouched, since it is a concept glossary that names neither `error` nor `nonCanonicalGenerations` and its seeding entry already carries the refusal/direction vocabulary.

PR #106, opened 2026-09-08
feat(the-stream-digest-is-pinned-against-an-absent-flag): Find out whet…

wighawag:
Task: `work/tasks/done/the-stream-digest-is-pinned-against-an-absent-flag.md`

Full gate green (`format:check`, `check:adr`, `changeset status --since=main`, `build`, `typecheck`, `test` — 55 core test files, 943 passed). Working tree holds exactly two new files.

## The answer

**Yes — the digest survives the deletion.** For a deployment that never set `alwaysFetchTimestamps` or `alwaysFetchTransactions`, removing both fields from `ProvidedStreamConfig` leaves the stream digest **byte-identical**: no fork, nothing orphaned, no history re-fetched.

**Drift check: the task's premises still hold.** `resolveStreamConfig` (`packages/core/src/internal/engine/utils.ts:998`) still copies only entries whose value is not `undefined`, so an unset flag contributes **no key**; and `streamDigestOfSourceHashes` (`packages/core/src/stream/identity.ts`) still takes `canonical_form({rule, filter, config: streamConfig})` over the resolved config. Measured: `resolveStreamConfig(undefined)` is `{finality: 17}` and `Object.keys(...)` is exactly `['finality']`, before and after the deletion. A field that puts nothing into the preimage takes nothing out of it when it goes.

**The other half, stated rather than engineered around:** a deployment that *did* set a flag stored its stream at a different address, so a post-deletion build finds nothing there and re-fetches from the source's start block, leaving the old subtree where it is. That is acceptable per ADR-0073 and the spec's withdrawn story 8, and it is already asserted (`packages/core/test/streamIdentity.test.ts`, "MOVES on a stream-config change"; `packages/browser/test/streamIdentity.test.ts`, "leaving the old one intact"), so I did not duplicate it.

**One rider for the deleting task:** the property holds only while the deletion leaves `STREAM_DIGEST_RULE` (`'etherfold/stream/1'`) alone. Bumping that tag re-addresses every stream in existence, which is exactly what the recorded constants catch — I verified they go red under a rule bump, and reverted.

## What I wrote

`packages/core/test/aDeletedStreamFlagDoesNotMoveTheDigest.test.ts` (3 tests):

1. **Recorded-value digest.** Two literal constants for one representative source (`96c7b55ba8b23635afb9f5622d2ba78f` with no config at all, `629df318e950e318c65936713d19f1ec` with only `finality` set), with a docstring saying why they are literals and never recomputed, and a note that they are not to be updated to match a new answer.
2. **The mechanism, plus the post-deletion digest computed without the deletion:** `Object.keys(resolveStreamConfig(...))` is `['finality']`, and the same recorded bytes come out of resolved configs written by hand as `{finality: N}` — which is exactly what a `ProvidedStreamConfig` narrowed to `{finality, parse}` produces.
3. **Stored-stream round trip.** A stream written under a no-flag config through a keeper that **addresses by digest**, then read back through the ordinary `IndexerGeneration.load()` path: the load asks the node for nothing, the state resumes at its cursor, one subtree stays occupied at the recorded digest, `clears` is 0, no event is re-appended, and indexing on never goes below the start block.

Falsification checked both ways: pointing the reload at a flag-setting config makes it fail (cursor back to 0, full re-index), and a `STREAM_DIGEST_RULE` bump reddens all three.

`.changeset/the-stream-digest-is-pinned-against-an-absent-flag.md` — patch, `@etherfold/core`, recording the answer.

## Decisions

**A `patch` changeset for a test-only change, rather than `changeset add --empty` or none.** The gate runs `pnpm changeset status --since=main`, and I verified empirically that once a file under `packages/core/` is *tracked* it errors with "Some packages have been changed but no changesets were found" — so "none" is a guaranteed red gate at land time. Between the remaining two: an empty changeset is the semantically exact instrument, but there is no precedent for one in the 96 files in `.changeset/`, so it would introduce a new repo convention as a side effect of an informative task. A patch bump costs nothing today ("NOTHING IS PUBLISHED YET") and puts the answer into the release notes of the release that will also carry the deletion, which is where a reader of that breaking change will look for it. Touches: `@etherfold/core`'s next version number, and the changeset convention in `CONTEXT.md` (`## Conventions`).

**Both halves in ONE new self-contained test file, rather than extending `streamIdentity.test.ts` and `streamLoad.test.ts`.** The task declares itself a deliberate orphan that "should be dropped rather than defended if it ever costs more than that", and a single file is the droppable unit; splitting it across two established files would make dropping it a surgical edit of tests that are not orphans. It also keeps the existing assertions untouched, satisfying the "not duplicated" criterion by construction. Alternative considered: putting the literal beside the existing literal in `streamIdentity.test.ts` (where a reviewer might look first) — mitigated by the new file's header naming ADR-0073 and pointing at `streamIdentity.test.ts` for the assertions it deliberately does not repeat. Touches: nothing outside the new file.

**A digest-ADDRESSED memory keeper, defined in the test file, instead of the shared `memoryStream`.** `packages/core/test/utils/streamCacheWorld.ts`'s `memoryStream` holds exactly one stream and serves it whatever it is asked for, so a moved digest reads back the same events and "no fork" is unassertable through it. The new wrapper keys one `memoryStream` per digest and takes the resolved config through `setStreamConfig`, mirroring `keepStreamOnIndexedDB`. I deliberately did **not** modify `streamCacheWorld.ts` (e.g. adding a fifth positional stream-config parameter to `makeIndexer`), because that helper is shared by several suites and widening it for one informative test is a cost the other suites pay; the new file composes it instead. Touches: `packages/core/test/utils/streamCacheWorld.ts` is unchanged, so no other core suite is affected.


wighawag:

Every acceptance criterion is met, and the two that are easy to fake are met honestly.

- **Literal recorded digest + why** — `DIGEST_WITH_NO_CONFIG_AT_ALL` / `DIGEST_WITH_ONLY_FINALITY_SET` are string constants compared against `streamDigestOf(...)`, never against a second computation. The docstring states the trap in the task's own terms (both sides move together when the digest function moves) and says the constants are not to be updated to match a new answer. This was the whole task and it is right.
- **Stored-stream round trip, no fork, no clear** — `digestAddressedStreams()` earns its keep: the shared `memoryStream` serves one stream whatever it is asked for, so a fork is structurally unassertable through it. Keying subtrees by digest makes `occupied()` a real fork assertion. `reloadChain.ranges` empty, `clears === 0`, `lastToBlock === TIP`, and the post-reload ranges all above `START_BLOCK`.
- **Nothing deleted** — diff is `+242 −0`. The task said the test must run against the tree as it stands, and it does.
- **Existing assertions not duplicated** — `packages/core/test/streamIdentity.test.ts` is untouched; the header names what it deliberately does not re-assert (the digest MOVING on a real config change).
- **The result is reported** — yes, the digest survives, with the mechanism named (`resolveStreamConfig` omits `undefined` keys, so an unset flag contributes no key to the preimage) and carried into the changeset where the deletion's release notes will be read.
- **Style** — the `logEventFetcher` poke mirrors `streamCacheWorld.ts:357`; vitest describe/it as elsewhere.

Two riders I am recording rather than blocking on:

1. **The rider for the deleting tasks is correct and worth carrying forward**: the property holds only while `STREAM_DIGEST_RULE` (`'etherfold/stream/1'`) is left alone. `alwaysfetchtransactions-…` and `alwaysfetchtimestamps-…` must not touch that tag, and these constants are the thing that catches it if they do.
2. **A `patch` changeset for a test-only change** is a version bump for no shipped behaviour. Justified in the Decisions block (the gate's `changeset status --since=main` reds without one, and an empty changeset would invent a convention that has no precedent in 96 existing files). Correct call for this repo today.

Merging.

PR #107, opened 2026-09-08
feat(a-timestampless-log-is-refused-at-the-fetch-boundary): A log with…

wighawag:
Task: `work/tasks/done/a-timestampless-log-is-refused-at-the-fetch-boundary.md`

All checks are green (`format:check`, `check:adr`, `build`, `typecheck`, full `test` across every package). Here is what I did.

## What I built

**The fetch-boundary refusal** (`@etherfold/core`), added beside the fallback, deleting nothing:

- `TimestamplessLogError` (`packages/core/src/errors.ts`): carries `blockNumber`, `blockHash` and `retryable: false`, and a message naming the standard (`execution-apis#639`), the minimum implementations that serve it including `@nomicfoundation/edr >= 0.20.0`, and the four node-level causes with their distinct fixes (old node / Hardhat bundling an older EDR with the override, not a wait / forked node predating the change / stale EDR `rpc_cache`). Its docblock records why the machinery is permanent and why it does not replace the fold-time guard.
- `assertLogsCarryTimestamps` (`packages/core/src/internal/engine/utils.ts`), beside `assertAscendingByBlock`: one rule, one implementation, for both deployment shapes of ADR-0003.
- Called at both fetch sites, one round trip in and before anything is enriched, folded, stored or pushed: `IndexerGeneration.fetchLogsFromProvider` (`indexer.ts`) and `LogFetcher.fetchAndPush` (`logFetcher.ts`, outside the retrying fetch, before `assertChain('after')` and the push). Both are gated on `!alwaysFetchTimestamps`, with a comment saying that condition dies with the flag in the later task.
- The tolerant read (`parseLogBlockTimestamp`) is untouched: "unreadable" and "absent" reach the refusal as one `undefined`, and neither becomes a number. The `NumberifiedLog.blockTimestamp` docstring, which claimed a caller "still needs the `alwaysFetchTimestamps` fallback", now says the engine refuses such a node instead.

**Tests** (written before the production code):

- `packages/core/test/aTimestamplessLogIsRefusedAtTheFetchBoundary.test.ts`: the refusal and its message content; `retryable === false`; one timestampless log in a range is enough; an unreadable value refuses exactly as an absent one; no refusal while `alwaysFetchTimestamps` is set; a compliant node makes no extra call and behaves exactly as before; the split shape refuses BEFORE pushing (the receiver is handed nothing) and pushes normally on a compliant node.
- `packages/processor-entities/test/the-two-guards-are-not-redundant.test.ts`: a captured stream replayed into an `EntityEventProcessor` with no node in the loop is still refused by `blockPointer`, naming the BLOCK, and folds normally once the capture carries timestamps. `blockPointer` itself is untouched.

**Fixture sweep** (the cost of the guard): 13 test files' fake nodes were pre-#639 nodes and now correctly failed. Fixed at the two shared helpers rather than per test: `makeLog` in `packages/core/test/utils/streamCacheWorld.ts` (new `timestampOf`), `rawLog` in `packages/core/test/logFetcher.test.ts` (plus a `servesTimestamps: false` chain option, used by the existing fallback test that genuinely models an old node), and the one fake node in `packages/browser/test/txInclusion.test.ts`.

**Changeset**: `.changeset/a-timestampless-log-is-refused-at-the-fetch-boundary.md`, prose, `@etherfold/core` minor plus two patch entries for the test-only touches (`changeset status --since=main` counts those packages as changed). Note that the gate's `changeset status --since=main` reads changesets through `git diff`, so it stays red until the runner commits the untracked file; everything it will then see is covered.

## Decisions

- **The guard sits at the two deployment-shape fetch sites, not inside `LogEventFetcher.getLogEvents`.** `LogEventFetcher` receives the FETCH config and the PARSE config; `alwaysFetchTimestamps` lives in the STREAM config, which is hashed into the wire identity and is deliberately not visible to the decoder. Threading it in would have given one call site (and would have covered `captureStream` for free) at the price of putting a stream-config policy inside the decoder, and of a constructor parameter the later deletion task would have to remove anyway. Alternative considered and rejected: placing it inside `enrichEvents`, which is the other shared helper both shapes call, but that whole module is deleted by `alwaysfetchtimestamps-is-deleted-with-the-enrich-path` and the guard must survive it. Touches that task: it deletes two `if (!...alwaysFetchTimestamps)` conditions rather than one parameter.
- **`captureStream` is deliberately NOT guarded**, although it is a fetch path. A capture is a snapshot of what a node said, and the fixture-replay path is precisely one of the two reasons the fold-time refusal exists (ADR-0073); refusing at capture would make a pre-#639 chain uncapturable while adding nothing, since a replay of such a capture is refused by `blockPointer`. Alternative considered: guard it too, for symmetry. Touches: capture/fixture tooling, and the later deletion task if it wants to revisit.
- **The new message does not recommend `alwaysFetchTimestamps`,** although `blockPointer`'s existing message does. The flag is deleted by a later task in this spec, and pointing a second message at it would create a second stale-advice coupling for that task to clean. The escape hatch is still reachable and documented on the flag itself. Touches: `alwaysfetchtimestamps-is-deleted-with-the-enrich-path`, whose message-coupling criterion stays a one-site fix.
- **The 13 broken fixtures were migrated by making their fake nodes SERVE `blockTimestamp`, not by setting `alwaysFetchTimestamps: true` on them.** Setting the flag would have gone red again the moment the flag is deleted; a fake node that serves the field models a supported node and is what the deletion task already asks for on the conformance workload. The one fixture that keeps a timestampless node is `logFetcher.test.ts`'s fallback case, which now says so explicitly (`servesTimestamps: false`). Touches: the deletion task inherits a much smaller migration.
- **Naming**: `TimestamplessLogError` / `assertLogsCarryTimestamps`. Checked against `CONTEXT.md`'s glossary, the ADRs and existing code: neither term already means something, and the error vocabulary (`SuspectedTruncationError`, `UnexpectedChainError`, structural `retryable`) is reused rather than forked, so `@etherfold/fetcher-host` classifies it as `fatal` with no change there.
- **`docs/adr/0073`'s `status: accepted, not yet implemented` was left alone.** Only the fetch-boundary half of that ADR has landed; the deletion it is mostly about has not, and flipping the status now would overstate what shipped.


wighawag:

Checked hardest on the two things a plausible diff gets wrong here: whether the two-guard design survived, and whether a compliant node pays anything. Both hold.

- **The refusal fires at the fetch boundary, with all four causes** — `TimestamplessLogError` names `execution-apis#639`, the serving implementations including `@nomicfoundation/edr >= 0.20.0`, and the four causes with their *distinct* fixes (old node / Hardhat 3.16.0 bundling edr 0.19.0 with the override as the fix rather than a wait / forking a node that predates the change / stale EDR `rpc_cache`). That specificity was the point of the task, not decoration.
- **Gated on the flag, at both deployment shapes** — `IndexerGeneration.fetchLogsFromProvider` and `LogFetcher.fetchAndPush` both wrap the call in `if (!…alwaysFetchTimestamps)`, each with a comment saying the condition (not the guard) dies with the flag. That is the correct handoff to `alwaysfetchtimestamps-is-deleted-with-the-enrich-path`: two `if`s to remove, and the refusal survives the enrich module's deletion because `assertLogsCarryTimestamps` lives in `internal/engine/utils.ts` beside `assertAscendingByBlock`, not in `enrich.ts`. Putting it in `enrich.ts` would have been the trap; it was explicitly considered and rejected.
- **A compliant node is entirely unaffected** — the guard is a `for` loop over logs already in hand. No new request, no new failure mode, nothing on the wire.
- **`blockPointer` is untouched and shown to be non-redundant** — `packages/processor-entities/src/` is not in the diff at all. `the-two-guards-are-not-redundant.test.ts` replays a captured stream into the processor with no node in the loop and asserts the fold-time refusal still fires naming `block 100`, plus the negative case (a capture that carries timestamps folds normally). The two messages are asserted to be distinct: one names the block, the other names the node.
- **Tolerance preserved** — `parseLogBlockTimestamp` is unchanged; absent and unreadable both arrive as `undefined` and both refuse. Neither is defaulted to a number, which is the failure that would be silent and permanent.
- **`retryable: false`** is right and structurally consistent with the existing error vocabulary, so `@etherfold/fetcher-host` classifies it as fatal with no change. A node does not grow a field on a retry.
- **The fixture sweep is honest** — 13 test files' fake nodes were pre-#639 and were fixed at the two shared helpers (`makeLog`, `rawLog`) plus one browser fake, not papered over per-test. The existing fallback test keeps modelling an old node via an explicit `servesTimestamps: false`, which is exactly where a pre-#639 node still belongs.
- **Changeset** present and in prose.

`captureStream` being deliberately unguarded is the right call and is well argued: a capture is a record of what a node said, refusing there would make a pre-#639 chain uncapturable, and a replay of such a capture is caught by the fold-time guard, which is one of the two reasons that guard exists.

Merging.

PR #108, opened 2026-09-08
feat(alwaysfetchtransactions-is-deleted-with-the-transaction-field): De…

wighawag:
Task: `work/tasks/done/alwaysfetchtransactions-is-deleted-with-the-transaction-field.md`

All checks green: `format:check`, `check:adr`, `build`, `typecheck`, and the whole test suite (57 core test files, everything passing).

## What I did

**Drift check first.** Nothing reads `event.transaction`. The only writer was `enrichEvents` itself, and the only consumers of `LogTransactionData` were the fetchers under it. No real consumer exists, so the task's premise holds and I proceeded.

**Test first** (`packages/core/test/noTransactionDataIsFetched.test.ts`, new): a node that *throws* on any method other than `eth_chainId` / `eth_blockNumber` / `eth_getLogs`, driven through both deployment shapes of ADR-0003; the pushed/folded events are asserted on their KEYS (not `toEqual`, which ignores undefined-valued properties). Plus two deletion assertions: `@ts-expect-error` lines that fail `pnpm typecheck` if `alwaysFetchTransactions` or `event.transaction` ever compile again, and a source scan of `src/` for `eth_getTransaction|transactionFetcherFor|LogTransactionData|alwaysFetchTransactions`. The scan was the red one before the deletion (7 offending files).

**The deletion** (`@etherfold/core`):
- `types.ts` — `alwaysFetchTransactions?: boolean` out of `ProvidedStreamConfig`; `transaction?: LogTransactionData` out of `BaseLogEvent`.
- `internal/engine/ethereum.ts` — `LogTransactionData`, `getTransactionData`, `getTransactionDataFromMultipleHashes` deleted (both `eth_getTransactionReceipt` call sites).
- `internal/engine/enrich.ts` — `TransactionFetcher`, `transactionFetcherFor`, the `getTransactions` option, the hash-gathering pass and the assignment loop; the timestamp half is untouched and its docstring narrowed to what it now does.
- `indexer.ts` / `logFetcher.ts` — the `getTransactions` bound method and both call sites, so neither shape can make the call.

**Reach outside core:** `@etherfold/fetcher-host` no longer reads `STREAM_ALWAYS_FETCH_TRANSACTIONS`; `platforms/nodejs-fetcher/README.md` no longer documents it; prose in `stream/identity.ts`, `stream/seedInstall.ts`, `cli/src/folding.ts` and `CONTEXT.md`'s stream-identity glossary entry no longer names the flag as a thing that changes what is stored.

**Tests updated** rather than deleted where the flag was being used as a config axis: `streamIdentity.test.ts`, `updateIndexer.test.ts`, `utils.test.ts`, `cli/test/streamConfigAgreement.test.ts`. The recorded digest literals in `aDeletedStreamFlagDoesNotMoveTheDigest.test.ts` are untouched and still green, which is the sibling task's pin confirming this deletion moved no digest.

**Changeset**: `.changeset/the-transaction-half-of-the-enrichment-is-deleted.md` — states what was removed, that there is no replacement, what a processor needing `from` should do instead (get it elsewhere; the durable fix is emitting it in the event on the contract side), and explicitly no migration guide.

## Decisions

**1. `STREAM_ALWAYS_FETCH_TRANSACTIONS` is removed outright, not refused.** The criterion says the flag goes from "every resolved-config path", and the env read in `@etherfold/fetcher-host` is one. I deleted the read and the README row rather than keeping the variable alive as an error ("this variable no longer exists"). Alternative considered: refuse on sight, on the strength of the repo's "nothing is accepted and ignored" rule quoted in `cli/src/folding.ts`. Rejected because that rule is about a variable the code *reads and discards*, not about one it no longer knows; `resolveFetcherHostConfig` has no unknown-variable validation at all, so an operator's stale `STREAM_ALWAYS_FETCH_TRANSACTIONS` is now ignored exactly as any typo'd variable is. Adding a refusal would be a new user-visible error surface the task explicitly rules out ("no substitute, no stub, no deprecated alias"). What it touches: `platforms/nodejs-fetcher` docs, `streamConfigFor` in the CLI, and the sibling timestamps task, which will do the same to `STREAM_ALWAYS_FETCH_TIMESTAMPS` and can now follow this precedent.

**2. `transactionFetcherFor` is deleted here, although the sibling timestamps task names it in its own sweep list.** That task's criterion is "a repo-wide search for `enrichEvents`, `blockFetcherFor`, `transactionFetcherFor` and `BlockTimestampCache` returns nothing", which this satisfies early rather than contradicts. The alternative — leaving a transaction fetcher standing with no caller — would leave an `eth_getTransactionReceipt` call site in the package, which is exactly what my structural test forbids and what ADR-0073's "the engine makes one data call" claims. The genuinely shared machinery (`enrichEvents`, `blockFetcherFor`, `BlockTimestampCache`, `providerSupportsETHBatch`) is left standing for that task, as instructed.

**3. Two `streamIdentity.test.ts` cases that used the flag as a second config axis were re-pointed at `finality` rather than dropped.** "Two DIFFERENT config changes are two different streams" and the collision corpus were both testing the digest, not the flag, so deleting them would have quietly narrowed digest coverage while the diff looked like a tidy removal. The corpus keeps four config variants (`undefined`, `alwaysFetchTimestamps`, `finality: 5`, `finality: 64`) so the collision count is unchanged.


wighawag:

I reviewed this against the failure mode a deletion diff most often has: taking the surviving read path down with the dead one. It did not.

- **The timestamp half is intact and untouched** — `enrichEvents` still holds `blockFetcherFor`, the `alwaysFetchTimestamps` branch, the block-hash timestamp cache and the batched block fetch. Only `LogTransactionData`, `getTransactionData`, `transactionFetcherFor` and the `eth_getTransactionReceipt` calls are gone. That is exactly the split the task specified so `alwaysfetchtimestamps-is-deleted-with-the-enrich-path` can take the rest.
- **Removed repo-wide, not just where the builder happened to be reading** — `@etherfold/core` (types, indexer, logFetcher, enrich, ethereum, identity, seedInstall), `@etherfold/cli` (`folding.ts`), `@etherfold/fetcher-host` (`STREAM_ALWAYS_FETCH_TRANSACTIONS`), `platforms/nodejs-fetcher/README.md` and `CONTEXT.md`. The only surviving mentions in the tree are ADR-0073 and `packages/core/CHANGELOG.md` (both historical records, correctly left alone) and tests that assert the absence.
- **Both deployment shapes stopped** — the calls are gone from `IndexerGeneration` and from the split `LogFetcher`, as ADR-0003 requires them to be symmetric.
- **No shim, and it is asserted rather than promised** — `noTransactionDataIsFetched.test.ts` pins the removal with a `@ts-expect-error` on `{alwaysFetchTransactions: true}` (so an alias sneaking back in fails the build) and a regex assertion that no `eth_getTransaction`/`transactionFetcherFor`/`LogTransactionData` reference survives. No deprecation window, no runtime warning, no migration guide, per ADR-0073.
- **No test double answers a transaction method any more** — `eth_getTransactionReceipt` has zero hits across `packages/`, `platforms/` and `examples/`.
- **The changeset is right** — it states the capability removal plainly, says there is no replacement, and gives the honest answer for `from` (emit it in the event contract-side; an indexer cannot mint data the chain did not log) without dressing it as a migration path.

The detail I want to record, because it is the drive's first real evidence rather than an assumption: **this PR modifies `aDeletedStreamFlagDoesNotMoveTheDigest.test.ts` by exactly one comment block and leaves both recorded digest constants untouched, and the gate is green.** That is the pinning test from #106 doing its job on the first deletion that could have moved the digest: the property held empirically, not just in argument. The constants must keep being left alone by `alwaysfetchtimestamps-…` too.

Merging.

PR #109, opened 2026-09-08
feat(alwaysfetchtimestamps-is-deleted-with-the-enrich-path): Delete alw…

wighawag:
Task: `work/tasks/done/alwaysfetchtimestamps-is-deleted-with-the-enrich-path.md`

The gate is green (`format:check`, `check:adr`, `build`, `typecheck`, `test` — 0 errors, 0 failures), and the working tree holds only this task's changes.

## Drift check (done first)

Both blocking tasks had landed as assumed. `a-timestampless-log-is-refused-at-the-fetch-boundary` shipped `TimestamplessLogError` / `assertLogsCarryTimestamps` at both deployment-shape fetch sites, so deleting the fallback left a guard rather than a hole. `alwaysfetchtransactions-is-deleted-with-the-transaction-field` had already taken `transactionFetcherFor` (it says so in its own Decisions block), so that symbol was gone before I started — recorded here rather than reported as drift, because it satisfies my sweep criterion early rather than contradicting it. Every other named symbol (`enrichEvents`, `blockFetcherFor`, `BlockTimestampCache`, `getBlocks`) existed exactly as described.

## What I did

**Deleted.** `packages/core/src/internal/engine/enrich.ts` in full (`enrichEvents`, `blockFetcherFor`, `BlockFetcher`, `BlockTimestampCache`); `getBlockData` / `getBlockDataFromMultipleHashes` / `ExtendedEIP1193Provider` from `ethereum.ts`; the `getBlocks` bound method and `blockTimestampCache` field on `IndexerGeneration`; the `blockTimestampCache` field on `LogFetcher`; `alwaysFetchTimestamps` from `ProvidedStreamConfig`, which is now `{finality, parse}`; the `STREAM_ALWAYS_FETCH_TIMESTAMPS` env read in `@etherfold/fetcher-host` and its README row in `platforms/nodejs-fetcher`; `packages/core/test/fetchLogs.test.ts`, whose every test exercised the enrichment dedup.

**Made unconditional.** Both `if (!...alwaysFetchTimestamps)` conditions around `assertLogsCarryTimestamps` are gone, in both shapes.

**Kept, with a test holding the line.** `parseLogBlockTimestamp` and its hex/decimal tolerance, and `blockTimestamp?: number` staying optional. `blockTimestampFromLog.test.ts` was rewritten as the surviving reading-path file: absent, unreadable, negative, fractional and NaN all yield `undefined`, never a number; the decoder does not carry a neighbouring log's timestamp onto a bare one.

**Message coupling.** `blockPointer` no longer recommends `stream: {alwaysFetchTimestamps: true}`; it names the four node-level causes, says there is no fallback, and still refuses rather than guessing. `the-two-guards-are-not-redundant.test.ts` now asserts the message does *not* contain the flag and *does* still refuse. Other prose swept: root `README.md`, `packages/processor-sqlite/README.md`, `CONTEXT.md`'s stream-identity glossary entry, ADR-0002's block-timestamp bullet, `docs/design/historical-state-database.md`, plus docstrings in `LogEventFetcher`, `stream/identity.ts`, `stream/seedInstall.ts`, `EntityEventProcessor` and `cli/src/folding.ts`.

**New test** `packages/core/test/noBlockTimestampIsFetched.test.ts`, mirroring the sibling task's `noTransactionDataIsFetched.test.ts`: both shapes make only `eth_getLogs` (against a node double that *throws* on anything else) and still deliver the timestamps, plus a structural sweep of `src/` for `eth_getBlockByHash|eth_batch|enrichEvents|blockFetcherFor|[Bb]lockTimestampCache|alwaysFetchTimestamps`.

**Conformance workload** migrated by dropping the flag; both fixtures already carry `blockTimestamp` on 100% of their logs (42/42 and 31,332/31,332), verified directly, so no fixture regeneration was needed.

**Load-bearing check.** Two mutations, each reverted: restoring the `if (!...alwaysFetchTimestamps)` guard in `indexer.ts` reds the new "fires UNCONDITIONALLY" test; restoring the flag recommendation in `blockPointer` reds the two-guards test.

**Changeset** `.changeset/the-timestamp-fallback-and-the-enrichment-path-are-deleted.md`, stating the removal and the minimum node requirement (`@nomicfoundation/edr >= 0.20.0`, satisfied today by a package-manager override rather than a Hardhat release). `changeset status --since=main` errors only because the new changeset is untracked and changesets diffs against git; verified green by running it against a throwaway `GIT_INDEX_FILE` copy with `add -A` applied, leaving the real index byte-identical.

## Decisions

- **`getBlockData`, `getBlockDataFromMultipleHashes` and `ExtendedEIP1193Provider` were deleted from `internal/engine/ethereum.ts`, although the task names only the enrich-path symbols.** Their sole caller was `blockFetcherFor`. Leaving them would leave a live `eth_getBlockByHash` (and `eth_batch`) call site in a published package with no caller, which is exactly what ADR-0073's "the engine makes ONE data call" claims is gone, and what my structural sweep test forbids. This follows the precedent the sibling transaction task set when it took the same helpers for receipts. Alternative considered: leave them for the `providerSupportsETHBatch` task. Rejected because that task's premise is that the flag has *no live readers* by the time it runs, which is true of the flag but would not have been true of the `eth_batch` request builder. Touches: `providersupportsethbatch-is-deleted-and-adr-0002-is-corrected`, which now has strictly less to delete.

- **`providerSupportsETHBatch` itself is left standing (dead) rather than deleted here.** It is declared on `ProvidedIndexerConfig`, `ProvidedLogFetcherConfig`, `FetcherHostConfig` and documented as `PROVIDER_SUPPORTS_ETH_BATCH` in `platforms/nodejs-fetcher`, and a dedicated ready task owns removing all three at once *plus* correcting ADR-0002's batch bullet. Deleting it here would have silently absorbed that task and taken a documented operator knob out of a published package under a changeset that does not mention it. I did update its now-false docstring in `fetcher-host/src/config.ts` (it said "which the enrichment fetches use") so no comment asserts a mechanism that no longer exists, and pointed it at the task that removes it. Touches: `providersupportsethbatch-is-deleted-and-adr-0002-is-corrected`.

- **ADR-0002's BLOCK-TIMESTAMP bullet is rewritten here; its BATCH bullet is deliberately untouched.** The timestamp bullet asserted "the fallback stays for now", which this change makes false, and the repo's own spec calls three documents asserting a stale external fact the drift worth cleaning. The batch bullet becomes false only when `providerSupportsETHBatch` goes, and that task explicitly claims it. Alternative considered: leave both to that task. Rejected because it would leave ADR-0002 describing a fallback that does not exist for however long that task waits. Touches: that task, which should now find only the batch bullet to reconcile.

- **`docs/adr/0073`'s `status: accepted, not yet implemented` was left alone.** The deletion has landed but the `providerSupportsETHBatch` consequence has not, so flipping the status now would overstate what shipped. Same call the previous task in this spec made. The last task in the chain is the natural place to flip it.

- **The stream-config test axes that used `alwaysFetchTimestamps` as "a second config that is not `finality`" were re-pointed at `parse`, not dropped.** `streamIdentity.test.ts` ("two DIFFERENT config changes are two different streams", key-order independence, the 1,800-case collision corpus), `updateIndexer.test.ts` (the genuinely-moved-config table, the idempotence loop) and `utils.test.ts` (the resolver's explicit-undefined rules) were all testing the *digest and the resolver*, not the flag, so deleting them would have quietly narrowed coverage while the diff looked like a tidy removal. The collision corpus keeps four config variants (`undefined`, `finality: 12`, `finality: 5`, `finality: 64`) so the count is unchanged. Same reasoning the sibling task recorded. One pinned literal had to move with it: `simple_hash({finality: 17, alwaysFetchTimestamps: true}) === 'ht6tzx8'` became `simple_hash({finality: 17, parse: {}}) === 'hg7dav3'`, computed from the built package. That assertion pins `canonical_form` over a TWO-key object, and it is the object that changed, not the hash function — the one-key and empty-object pins beside it (`h10lkzm2`, `h28y`) are untouched, which is what shows the bytes did not move.

- **The CLI's "agree on the boolean setting" test became "agree on a variable that names a DELETED flag".** Rather than deleting it, it now pins that BOTH halves ignore a stale `STREAM_ALWAYS_FETCH_TIMESTAMPS` and reach the same digest as an environment that never mentioned it. A half-migration where one side still read the variable would put a key in one digest and not the other, and `WireContextMismatchError` is not retryable, so the process would exit. Following the sibling task's precedent, the variable is removed outright rather than refused: `resolveFetcherHostConfig` has no unknown-variable validation at all, so it is now ignored exactly as a typo'd variable is, and adding a refusal would be a new user-visible error surface this task does not ask for.

- **`docs/design/historical-state-database.md` got a one-sentence "superseded on the fallback half" correction rather than a rewrite or nothing.** It told an implementer to fall back to `eth_getBlockByHash` for the blocks whose logs carried none, which is now false. It is a dated design snapshot, so I did not touch its empirical client table (kept as the record it is) and did not restate the whole argument. Alternative considered: leave it entirely, on the grounds that a `Status:` header marks it as historical. Rejected because the header says "design complete, not built", not "not maintained", and an unbuilt component's design doc is precisely the thing someone reads before building it.


wighawag:

This is the biggest deletion of the drive (`+590 −883`, `enrich.ts` gone entirely), so I checked the surviving read paths one by one rather than trusting the summary. Nothing load-bearing went out with the fallback.

**The guard from #107 survived and is now unconditional, which was the whole handoff.** `assertLogsCarryTimestamps` is still called at both deployment shapes (`indexer.ts:1792`, `logFetcher.ts:307`) and the `if (!…alwaysFetchTimestamps)` wrapper is gone from both. That is exactly the two-line change #107 set up: the condition died, the refusal did not. The test flipped with it (`does NOT fire while alwaysFetchTimestamps is set` → `fires UNCONDITIONALLY: no stream config quiets it, not even the deleted flag`), plus a `@ts-expect-error` case pinning that the flag has no alias behind it.

**The surviving chain calls are all intact.** I grepped every `provider.request` in `packages/core/src`: `eth_chainId` (load + both per-cycle sites), `eth_getBlockByNumber` for the genesis check at `indexer.ts:1091`, and `eth_blockNumber` for the tip. Only `eth_getBlockByHash` and the `eth_batch` path under it are gone, which is the enrichment and nothing else. A deletion this size taking the genesis check's block read with it would have been easy and invisible; it did not.

**`blockPointer`'s fold-time refusal still fires** (`processor-entities/src/stream.ts:62`), and the message was correctly *updated* rather than left lying: it no longer offers `stream: {alwaysFetchTimestamps: true}` as a remedy, because that remedy no longer exists. It still names `execution-apis#639`, the serving implementations and the causes. `the-two-guards-are-not-redundant.test.ts` still holds the seed/fixture-replay case, so the two guards remain demonstrably non-redundant with only one of them now reachable from a fetch.

**The recorded digest constants are untouched again** (`96c7b55…`, `629df31…`) and the gate is green, so the second and final flag deletion also did not move the digest. The property ADR-0073 asserted now has both halves measured rather than argued.

**Removed repo-wide**, not per-package: core, cli, fetcher-host (`STREAM_ALWAYS_FETCH_TIMESTAMPS`), conformance, browser tests, processor-sqlite fixtures, plus the prose in `README.md`, `CONTEXT.md`, ADR-0002 and `docs/design/historical-state-database.md`. The only surviving mentions are `CHANGELOG.md` history (correctly immutable) and tests asserting the absence — including `noBlockTimestampIsFetched.test.ts`'s regex sweep for `eth_getBlockByHash|eth_batch|enrichEvents|blockFetcherFor|BlockTimestampCache|alwaysFetchTimestamps`, which will fail if any of it creeps back.

No alias, no deprecation window, no runtime warning, no migration guide. Changeset present and in prose.

Merging.

PR #110, opened 2026-09-08
feat(providersupportsethbatch-is-deleted-and-adr-0002-is-corrected): Re…

wighawag:
Task: `work/tasks/done/providersupportsethbatch-is-deleted-and-adr-0002-is-corrected.md`

Done. Working tree holds only this task's changes.

## What I did

**Premise check first (repo-wide, both spellings).** `grep -i` for `providerSupportsETHBatch`, `PROVIDER_SUPPORTS_ETH_BATCH` and `eth_batch` across every package, platform, example, doc and archive. Live readers found: `@etherfold/core` (two config types), `@etherfold/fetcher-host` (`FetcherHostConfig`, the env read, the pass-through in `host.ts`), `platforms/nodejs-fetcher` (README env table). Exactly the three the task names. One extra hit, `archive/ethereum-indexer-server`, is outside the workspace globs (see Decisions). No surviving reader the task does not name, so no needs-attention route. Also checked ADR-0002 for drift: the previous task rewrote only the **block-timestamp** bullet and explicitly left the **batch** bullet, which is what I found, so there was nothing to reconcile.

**Tests first (red, then green).** Three files, each red before the deletion:
- `packages/core/test/theEngineHasNoOpinionAboutBatch.test.ts`: `@ts-expect-error` on both `ProvidedIndexerConfig` and `ProvidedLogFetcherConfig` plus a structural sweep of `src/`. I verified the type half really bites by temporarily re-adding the field: `tsc` then reports `TS2578: Unused '@ts-expect-error' directive`.
- `packages/fetcher-host/test/batchKnobRemoved.test.ts`: `PROVIDER_SUPPORTS_ETH_BATCH` in the environment produces a resolved config with no such **key** (four spellings incl. `false`), `@ts-expect-error` on the overrides type, and a sweep of `src/` in **both** spellings.
- `platforms/nodejs-fetcher/test/batchVariableRemoved.test.ts`: neither spelling in the README or `src/index.ts`, with a self-check that it is really reading the env table.

**The deletion.** Field removed from `ProvidedIndexerConfig` (`types.ts`) and `ProvidedLogFetcherConfig` (`logFetcher.ts`); removed from `FetcherHostConfig` with its docstring, from `resolveFetcherHostConfig`, and from the `LogFetcher` construction in `host.ts`; the now-unused `readBoolean` helper deleted with it; the README env-table row removed.

**ADR-0002.** The batch bullet is rewritten, not struck, and keeps its role as the standing correction: it leads with "batch RPC is not prohibited", says the correction still stands, then says the allowance is now **moot** because the engine makes no batchable request (one `eth_getLogs` per range, one `eth_blockNumber`, one `eth_chainId`), notes a caller's provider may still batch transparently, and closes by framing this as the strongest form of the ADR's own cost constraint rather than a reversal.

**Changeset** `.changeset/the-batch-support-knob-is-deleted.md`, minor on all three packages, leading with "a documented deployment variable is removed, not an internal flag", and recording that it is digest-neutral (the flag was a sibling of `stream`, never inside the hashed stream config).

**Gate:** `pnpm build && pnpm typecheck && pnpm test` all green repo-wide, plus `pnpm format:check` and `pnpm check:adr`.

## Decisions

- **`archive/ethereum-indexer-server/src/server/simple.ts` still names `providerSupportsETHBatch`, and I left it.** `archive/README.md` states the tree is deliberately outside `pnpm-workspace.yaml` and is "not built, typechecked, tested, versioned or published", kept as the reading material `docs/reviews/server-cli-batch.md` analyses. It already names several deleted things (`EthereumIndexer`, `ProcessorFilesystemCache`), so editing it would falsify a frozen historical record to satisfy a grep rather than remove a reader. Alternative considered: delete the option there too, which compiles nothing and would have made that file diverge from the review doc that reads against it. Touches: the acceptance criterion "a repo-wide search confirms nothing still names either the flag" is satisfied for live code; `work/` items (the spec, two done tasks, a dropped spec) likewise still name it as history and are records I must not edit.
- **`docs/adr/0073`'s `status: accepted, not yet implemented` was left alone.** Two more tasks from this spec are still in `work/tasks/ready/` (`the-engine-declares-its-method-set-and-a-test-holds-it-to-it`, which is blocked by this one, and `the-edr-requirement-and-the-scope-of-the-claim-are-documented`), so flipping the status now would overstate what shipped. Same call both predecessor tasks made; the last task in the chain is the place for it.
- **The generated API docs are not regenerated in this commit, because they are not committed.** `docs/api` is typedoc output and is gitignored (`.gitignore:53`), produced by `pnpm publish-typedoc` during `docs:build`. Since the flag is gone from the source types, the next build emits docs without it; there is no stale committed artifact to sweep. Recorded because the acceptance criterion offers "regenerated, or noted as regenerating on build" and this is the second branch.
- **`readBoolean` was deleted rather than kept for a future boolean variable.** It had exactly one caller. Keeping it would leave an unused helper whose only purpose was the removed knob, against CONTEXT.md's "delete what has stopped earning its keep". Touches: any later task adding a boolean env var to `@etherfold/fetcher-host` re-adds a six-line parser (`readNumber` next to it is the template).
- **The removed variable gets no warning path, and I added no test asserting one is absent.** The task forbids a deprecation warning, and `streamConfigFromEnv`'s existing docstring already establishes the house rule that a retired variable is "ignored like any other unrecognised variable". I stated that rule in the changeset and in the fetcher-host test's header comment instead of encoding it as a new mechanism. Touches: `platforms/nodejs-fetcher` operators, who see the variable simply gone from the docs.


wighawag:

Every criterion met, and the two judgement calls in it were made deliberately and explained rather than fallen into.

- **Gone from all three homes**, which is the criterion that a diff can pass locally and fail in fact: `@etherfold/core` (`types.ts`, `logFetcher.ts`), `@etherfold/fetcher-host` (`config.ts`, `host.ts`) and `platforms/nodejs-fetcher` (the env table). Not just the package the builder happened to be reading.
- **`PROVIDER_SUPPORTS_ETH_BATCH` is out of the README table, not merely unread**, and each of the three packages carries a test that greps its own source for both spellings, so a reintroduction fails rather than rots.
- **ADR-0002's bullet was corrected, not deleted** — it keeps the original correction it existed to make (batch was never prohibited) and adds why the allowance is now moot (there is no request left to batch), naming the surviving surface. It also makes the right distinction: a caller's provider may still batch transparently; the engine simply no longer asks. That is a stronger reading of ADR-0002 than a reversal of it.
- **The surviving env readers are intact** — `STREAM_FINALITY`, `MAX_EVENTS_PER_FETCH`, `MAX_BLOCKS_PER_FETCH`, `REQUESTS_PER_SECOND`, `RETRY_*`, `SUSPECT_RESULT_COUNT` all still read through `readNumber`. Only `readBoolean` went, and only because the knob was its last caller. I checked this specifically: a config-deletion taking a sibling reader with it is the classic version of this mistake.
- **Changeset** names it as removing a *documented deployment variable*, so it cannot be read as an internal cleanup.

The two calls I want to endorse explicitly:

1. **Leaving `archive/ethereum-indexer-server/src/server/simple.ts` naming the flag is correct.** That tree is outside `pnpm-workspace.yaml`, is not built or typechecked, already names other deleted symbols (`EthereumIndexer`, `ProcessorFilesystemCache`), and is the reading material `docs/reviews/server-cli-batch.md` analyses against. Editing it would falsify a frozen record to satisfy a grep, with no reader removed. The repo-wide criterion is met for live code; the remaining mentions are ADRs, `work/` history and this archive, all records.
2. **Not regenerating `docs/api` is correct because it is not committed** — `.gitignore:53`, emitted by `pnpm publish-typedoc` during `docs:build`. The criterion explicitly allowed "noted as regenerating on build", and there is no stale committed artifact. I verified `docs/api` has no tracked hits on either the branch or `main`.

Also right to leave ADR-0073 at `accepted, not yet implemented`: two tasks of this spec are still open, and the last one to land is where that status flips.

Merging.

PR #111, opened 2026-09-08
feat(the-engine-declares-its-method-set-and-a-test-holds-it-to-it): The…

wighawag:
Task: `work/tasks/done/the-engine-declares-its-method-set-and-a-test-holds-it-to-it.md`

## What I did

**Drift check first.** Both blockers have landed (`work/tasks/done/alwaysfetchtimestamps-is-deleted-with-the-enrich-path.md`, `.../providersupportsethbatch-is-deleted-and-adr-0002-is-corrected.md`), and the enrichment machinery is gone from `packages/core/src`. But the task's method list is **stale by one**: the engine makes a fourth call, `eth_getBlockByNumber` with `['earliest', false]`, in `IndexerGeneration.promiseToLoad` (`packages/core/src/indexer.ts:1091`) whenever a source declares a `genesisHash` and `skipGenesisCheck` is not set. It is identity, not data, and once per load, so I declared **four** rather than widening the guard quietly. See Decisions.

**The declared set, stated once** (`packages/core/src/providerSurface.ts`, exported from the package index):

- `ENGINE_PROVIDER_METHODS` = `eth_getLogs`, `eth_blockNumber`, `eth_chainId`, `eth_getBlockByNumber`
- `declaredMethodsOnly(provider)` is the wrapper: it **records** every method (`methodsRequested`, which is what tests assert the subset against) and **refuses** anything else with `UnexpectedProviderMethodError` (`retryable: false`, message naming the method and the cost). Idempotent, so a reconfigure does not stack wrappers.
- The one method that can name a block is narrowed to the genesis probe: the chain's first block passes, a height is refused.

**Applied at every provider intake in the engine**, so one wrapper sees every call: `IndexerGeneration.reinit` (which is what `LogEventFetcher` is then built from, and covers the container and the browser package), `IndexerGeneration.updateIndexer`'s pre-reinit chain-id probe, `LogFetcher`'s constructor, and `captureStream`.

**Tests** (`packages/core/test/theEngineDeclaresItsMethodSet.test.ts`, 9 cases, written before the wiring and in the sibling style of `noBlockTimestampIsFetched.test.ts`): the subset assertion over a driven cycle in both deployment shapes of ADR-0003; the genesis probe recorded as the identity read it is; the refusal, over a deliberately **permissive** node double that would have answered `eth_getBlockByHash`; the block-read narrowing; idempotence; the receiving half (no provider parameter exists, asserted with `@ts-expect-error`, and it folds a batch while a guarded node beside it records zero); and the README check.

**The claim** is stated in `README.md` and `packages/core/README.md`, each behind a `<!-- provider-surface: ... -->` anchor, and the test holds both **in both directions**: every method named must be declared and every declared method must be named.

Also: a `CONTEXT.md` glossary entry for the new term, a prose changeset (`.changeset/the-engine-declares-its-method-set.md`), and a stale docblock in `LogFetcher` corrected (it still promised "the block/transaction reads a stream config asks for").

## The demonstration (the guard bites)

I reintroduced a per-block call in `IndexerGeneration.fetchLogsFromProvider` (`for (const blockHash of new Set(...)) await this.provider.request({method: 'eth_getBlockByHash', ...})`) and ran the core suite:

- **with the guard: 126 tests across 16 files failed**, every printed failure an `UnexpectedProviderMethodError` naming `eth_getBlockByHash` and the declared set. The call never reached any node double.
- **counterfactual, guard disabled, same reintroduced call: 128 failed** (the two extra being my own refusal tests). The failing set is otherwise identical, because core's ~20 hand-written provider doubles happen to be strict today and each throws its own `unexpected method` string.

So the honest reading: the property was resting on twenty doubles each remembering to throw, and on **nothing at all in production**, where a real node answers `eth_getBlockByHash` cheerfully and nothing fails. The guard makes it one structural fact, and my own permissive-double test is the case that isolates it (the node answers, the wrapper refuses, `answered` never contains the method). The demonstration was reverted; `packages/core/src/indexer.ts` is byte-identical to before it.

## Gate

`pnpm format:check`, `pnpm check:adr`, `pnpm build`, `pnpm typecheck`, `pnpm test` all green (core 966 passed, full monorepo green including `browser`, `cli`, `server`, `examples`). `changeset status --since=main` still reports "no changesets found" only because the changeset file is untracked; `git diff` against the merge base does not see untracked files, and it will be satisfied once the runner commits with `git add -A`.

I also left one out-of-scope signal in `work/notes/observations/an-orphaned-extra-ts-holds-the-only-eth-call-in-the-core.md`: `packages/core/src/internal/utils/extra.ts` is imported by nothing and holds the only `eth_call` in the package.

## Decisions

- **The declared set is FOUR methods, not the task's three.** `eth_getBlockByNumber` is added because `IndexerGeneration.promiseToLoad` reads the chain's first block to check a source's `genesisHash` (real, user-supplied configuration read by `@etherfold/utils` and hashed into the stream identity). It is IDENTITY like `eth_chainId`, asked once per load and never per block, so it belongs in the set rather than being routed around the wrapper. Alternatives considered: declaring three and calling the genesis check on an unguarded raw provider (a documented hole at exactly the method that reads blocks), or deleting the check (out of scope, and it is a real feature). Touches: the README claim and `CONTEXT.md`, which now say four.

- **The guard REFUSES in production rather than only recording in tests.** This is what makes the subset assertion run "across the engine's existing test suite" without editing 20 test files or inventing a global test-side registry: there is no seam a vitest setup file could reach into, because providers are plain objects constructed inside each test. It introduces a new error type (`UnexpectedProviderMethodError`), but no configuration and no node can provoke it: the only code that reaches a guarded provider is this engine, so it is an internal invariant, not an operator-facing refusal. Alternatives considered: a module-level recording registry read by a global `afterEach` (production code carrying a test-only global, and it retains objects), and a per-test opt-in wrapper (drifts the moment someone writes a new test). Cost is one `Set` lookup per request. No ADR written: ADR-0073 already decides the claim and this is its enforcement, hard to be surprised by given that ADR, and one deletion to reverse.

- **`eth_getBlockByNumber` is narrowed to the genesis probe by its PARAMS, not just its name.** A set of method names alone leaves a per-block-shaped hole at the one declared method that can read a block: a reintroduced enrichment loop written against `eth_getBlockByNumber` would pass a name-only guard. The narrowing states "identity, not data" in the only place it can be stated. Trade-off: the guard now enforces slightly more than the "set" the tests read, which I documented at the site and in the README claim.

- **The probe predicate accepts BOTH `earliest` and `0x0`, and this touches another ready task.** `work/tasks/ready/the-genesis-check-asks-for-block-zero-not-the-earliest-tag.md` replaces the tag with the explicit block number (because `earliest` means lowest-available, not genesis). Keying my predicate on `earliest` alone would have made that one-line fix fail here with a confusing refusal, so both spellings pass; both are the bottom of the chain, which is the property the predicate actually defends. Note for that task: it also asks that no `earliest` remain in the repo, and `packages/core/src/providerSurface.ts` is now a second site holding the string, deliberately, with a comment pointing at the task.

- **The claim lives in TWO READMEs, anchored and checked both ways.** The root README is where "relies only on EIP-1193" is claimed to a user; `packages/core/README.md` is the engine's own doc. Rather than pick one and leave the other vaguer, both carry the claim behind an HTML-comment anchor and the test asserts set equality against `ENGINE_PROVIDER_METHODS` in each. Duplication cannot drift silently, and declaring a fifth method now requires writing it where a user reads it. Side effect worth knowing: `eth_*` names may not appear inside those anchored paragraphs unless declared, so the history of a deleted call belongs in the ADR, which is where it already is.

- **`captureStream` is guarded too.** It is a third place the engine talks to a node (one `LogEventFetcher`, `eth_getLogs` only). Including it costs one line and keeps "every provider the engine holds is behind the declared surface" true without exceptions; excluding it would have made the rule "every intake except the capture one".

- **New vocabulary: "the declared provider surface"** (`ENGINE_PROVIDER_METHODS`, `declaredMethodsOnly`, `MethodDeclaringProvider`, `UnexpectedProviderMethodError`, `isGenesisProbe`). Checked against `CONTEXT.md` and the ADRs: no existing term meant this ("chain-facing surface" was prose in ADR-0073 and ADR-0002 with no code behind it, and this names exactly that, at the engine layer where the calls are). The error name follows the existing `UnexpectedChainError` / `UnexpectedFromBlockError` family and carries `retryable: false` per the errors-file convention. I added a glossary entry to `CONTEXT.md` so the term has one home.


wighawag:

All seven criteria met, and the demonstration criterion — the one most likely to be asserted rather than performed — was genuinely performed, with a counterfactual I did not ask for and would not have thought to.

- **Declared once, readable by tests** — `ENGINE_PROVIDER_METHODS` in `packages/core/src/providerSurface.ts`, with the reasoning that a test carrying its own copy of the list is a second source of truth. No duplicate list anywhere.
- **The wrapper asserts subset, and it runs across the whole existing suite** — because the engine holds its provider behind `declaredMethodsOnly()` at all three intakes (`IndexerGeneration`, `LogFetcher`, `captureStream`), every existing test that constructs an engine is now covered without one of them being edited. That is a better answer to "not in a single dedicated test" than a vitest setup file could have given, since the doubles are plain objects built inside each test with no seam to reach.
- **The reintroduction demonstration is real and reverted** — 126 failures across 16 files, each an `UnexpectedProviderMethodError` naming `eth_getBlockByHash`; `eth_getBlockByHash` appears nowhere in `packages/core/src` on the branch. The counterfactual (guard disabled, same call: 128 failures) is the valuable part, because it shows the property was previously resting on ~20 hand-written doubles each remembering to be strict, and on **nothing whatsoever in production**, where a real node answers that call cheerfully. The permissive-double test isolates exactly that gap.
- **Both READMEs match the declared set exactly**, behind anchors, asserted set-equal in both directions, so a fifth method cannot be added without writing it where a user reads it.
- Changeset present, prose, and `CONTEXT.md` carries the glossary entry for the new term.

Three calls I specifically endorse:

1. **Declaring four methods rather than the task's three.** The genesis probe is real, user-supplied configuration hashed into the stream identity, and it reads a block. The alternative — running it on an unguarded raw provider — would have left a documented hole at precisely the method that can read blocks. Declaring it as identity and saying why is the honest version.
2. **Narrowing `eth_getBlockByNumber` by its PARAMS, not just its name.** Without this the guard has a per-block-shaped hole at its one block-reading method: a reintroduced enrichment loop written against `eth_getBlockByNumber` would sail through a name-only set. This is the difference between a guard and a gesture.
3. **Refusing in production rather than recording in tests.** It adds `UnexpectedProviderMethodError`, but nothing an operator or node can provoke: only this engine reaches a guarded provider, so it is an internal invariant. One `Set` lookup per request, one deletion to reverse. Correct trade.

Carried forward for `the-genesis-check-asks-for-block-zero-not-the-earliest-tag`: `isGenesisProbe` deliberately accepts **both** `earliest` and `0x0`, so that task's one-line fix cannot fail here — but `providerSurface.ts` is now a second site holding the string `earliest`, which collides with that task's "no `earliest` remains in the repo" criterion. I am planting that as a note on the task before it is dispatched, so its builder narrows this predicate rather than being caught by its own grep.

The off-path finding (`packages/core/src/internal/utils/extra.ts` is imported by nothing and holds the only `eth_call` in the package) was correctly captured as an observation rather than fixed in place. It is picked up in this drive's end-of-run triage.

Merging.

PR #112, opened 2026-09-08
feat(the-edr-requirement-and-the-scope-of-the-claim-are-documented): Do…

wighawag:
Task: `work/tasks/done/the-edr-requirement-and-the-scope-of-the-claim-are-documented.md`

Done. All acceptance criteria are met and the full gate (`format:check`, `check:adr`, `changeset status`, `build`, `typecheck`, `test`) is green.

## What I did

**Re-verified the version facts first (they had not drifted, but one adjacent fact had).** Against the live registry on 2026-09-08: `@nomicfoundation/edr` latest is `0.20.0` (published 2026-09-02, first release carrying `blockTimestamp` on logs per `edr#1644`); `hardhat@3.16.0` (latest, published 2026-09-07) pins `@nomicfoundation/edr` at exactly `0.19.0`, and the `hh2` line (`2.29.1`) pins `0.12.0-next.23`. So no released Hardhat bundles edr >= 0.20.0 and the override stays the primary instruction rather than a fallback. I also read the upstream 0.19 -> 0.20 release notes to state the delta accurately instead of quoting the ADR's summary from memory.

**Rewrote the root `README.md` Caveats section** (`README.md`, the only source change):

- The timestamp paragraph now leads with the requirement being on the log, states `@nomicfoundation/edr >= 0.20.0` as a minimum requirement, and keeps the honest reason the field stays optional on the wire (a forked pre-spec remote, plus an absence EDR's RPC cache recorded from one).
- A separate paragraph for the Hardhat audience: the requirement is on the EDR VERSION, you are not waiting for anyone, with the exact pinned versions as of today, followed by four copy-pasteable override snippets (pnpm 11+ `pnpm-workspace.yaml`, pnpm 10 `package.json`, npm `overrides`, yarn `resolutions`).
- The override's caveat is stated rather than implied: an untested pairing, verify it in your project, plus the two actual 0.19 -> 0.20 behaviour changes to look at (interval-mining range validation, `InlineConfigDirectiveError.function` widened to `string | undefined`) and the `rpc_cache/v2` move.
- A paragraph scoping the claim: "a fold over logs" is about the ENGINE's calls, completeness is with respect to the node's LOG INDEX, with the bloom-omission case named, the Polygon block 74,614,768 numbers (848 vs 8 more from `eth_getTransactionReceipt`, 856 from an indexed source), the finding referenced with its provenance caveat, and the ADR-0004 collision (a stable absence is undetectable rather than noisy) said plainly.

**No migration section, deliberately** (per the task's explicit fence), and no compatibility path, no "who is affected" enumeration.

**Checked the two sweeping criteria rather than assuming:** no document instructs a reader to set `alwaysFetchTimestamps` / `alwaysFetchTransactions` (remaining mentions are historical records: ADRs, changesets, CHANGELOG, done tasks, tests asserting their absence), and the README's provider-surface claim still matches `ENGINE_PROVIDER_METHODS` (`packages/core/test/theEngineDeclaresItsMethodSet.test.ts`, 9 passed).

**One observation note** (out-of-scope signal, captured not fixed): `work/notes/observations/edr-cache-segmentation-claim-reads-backwards.md`.

No changeset was added: the change touches no package directory, so no published package's API moved and `changeset status --since=main` is clean.

## Decisions

- **The pnpm override snippet is given in TWO files, not one.** ADR-0073 and the task both say "`pnpm.overrides`", which is the pnpm 10 form; pnpm 11 and later ignore the `pnpm` field in `package.json` entirely and read `overrides:` from `pnpm-workspace.yaml`. This repo's own `pnpm-workspace.yaml` documents that exact trap for its `abitype` override, so a package.json-only snippet would have been a snippet that silently does nothing for a reader on current pnpm, failing the "working override snippet" criterion. Alternatives considered: the pnpm 11 form alone (leaves anyone still on pnpm 10, which is what this repo pins, with nothing) or the ADR's wording verbatim (wrong for the current major). Touches: nothing in code; it does mean the README is more precise than ADR-0073's sentence on this point.
- **The override value is `>=0.20.0` rather than a pin at `0.20.0`.** The requirement is a floor, and pinning would freeze a reader on today's release; a reader who wants the conservative form can pin. Touches: the same string appears in the `TimestamplessLogError` message (`override @nomicfoundation/edr to >=0.20.0`), so the two agree.
- **The stale-cache clause was restated rather than repeated.** ADR-0073, ADR-0002 and the `TimestamplessLogError` docstring say EDR's version-segmented cache means "entries written before the change keep answering without the field"; the 0.20.0 release note says the opposite about those entries (cache moved to `rpc_cache/v2`, everything else ignored). The surviving, accurate hazard is an absence RECORDED from a forked pre-spec remote and replayed until `rpc_cache` is dropped, so the README says that. The conclusion (the refusal is permanent machinery) is unaffected. I did not edit the ADRs, the changesets or the error docstring: they are records and out of this task's fence, so the discrepancy is captured as an observation instead.
- **The two additions live in the ROOT README only.** The task names the README's Caveats as the home; `packages/core/README.md` and `packages/processor-sqlite/README.md` already carry the shorter engine-side statements, and duplicating the override snippets into a published package README would create a second place to keep in step with pnpm's file moves. Touches: a reviewer who expects the completeness-scope sentence in `@etherfold/core`'s README will not find it there.
- **No new doc-scanning test was added.** The repo does pin some prose with tests (the provider-surface anchor, the retired-knob scanners), but those exist where prose asserts something the CODE can contradict. A version floor and an external node limitation have no in-repo referent to diff against, so a test would only assert that a string is present, which pins wording rather than truth and goes stale exactly when the version does. The provider-surface criterion is already covered by the existing test, which I ran.


wighawag:

All six criteria met, and this is the only doc-only task of the drive that came back with a *correction to the repo's own record* rather than a transcription of it.

- **The requirement is stated against the EDR version, not the Hardhat version**, with the point made explicitly: no published Hardhat bundles 0.20 (`hardhat@3.16.0` pins `0.19.0`, `hh2` pins `0.12.0-next.23` as of today), and that is not a wait because EDR is an ordinary dependency. Override snippets given in four forms, not three: pnpm 11 (`pnpm-workspace.yaml`), pnpm 10 (`package.json`), npm and yarn.
- **The caveat is stated, not implied**, and is better than the task asked for: it names the actual published 0.19→0.20 behaviour delta (interval-mining range validation now rejecting `[0, N]`, the `InlineConfigDirectiveError.function` widening) so a reader can size the risk rather than being told to be careful.
- **The completeness claim is correctly scoped and correctly weighted.** It is scoped to the node's log index, names the bloom-omission case with the captured Polygon numbers (848 vs 856 on block 74,614,768) and cites the finding *including its provenance caveat* (a third party's dated captures, not re-run here). Crucially it lands the point on **ADR-0004**: a bloom-omitted log is a STABLE absence, not a flapping one, so it never presents as a noisy reorg, it presents as a log that never existed. That is the correct reading of why this is a correctness disclosure and not a performance footnote, and it is the framing I most wanted to see here.
- **No migration guide, no compatibility path**, and no document still tells a reader to set a deleted flag — I grepped every non-record `*.md`: zero hits for all three retired names outside CHANGELOGs, ADRs, `work/` and `.changeset/`.
- **The provider-surface claim is still machine-checked** — the README edit sits alongside the anchor from #111, so the set-equality test would have reddened on a mismatch and did not.

The judgement call worth endorsing: **re-verifying the EDR facts live turned up that three of our own documents state the cache mechanism backwards.** ADR-0073, ADR-0002's bullet and the `TimestamplessLogError` docstring all say pre-change `rpc_cache` entries "keep answering without the field", while the 0.20.0 release note says the cache moved to `rpc_cache/v2` and everything else is ignored. The right moves were made on both axes: the README states the accurate surviving hazard (an absence recorded from a forked pre-spec remote, replayed until `rpc_cache` is dropped), and the ADRs/docstring were left alone as out-of-fence records with the discrepancy captured in `work/notes/observations/edr-cache-segmentation-claim-reads-backwards.md`. The conclusion those documents draw is unaffected — the forked-node cause stands alone, so the refusal is still permanent machinery — which is exactly why this is an observation and not a block.

That observation is task-worthy and is carried into this drive's end-of-run triage: three source documents state a mechanism that the upstream release note contradicts, and only one of the three has been corrected.

Not adding a doc-scanning test was also right: a version floor and an external node limitation have no in-repo referent, so such a test would pin wording rather than truth.

Merging.

PR #113, opened 2026-09-08
feat(a-provider-refusal-is-read-from-its-data-before-its-prose): Read a…

wighawag:
Task: `work/tasks/done/a-provider-refusal-is-read-from-its-data-before-its-prose.md`

All gates green (`BUILD OK`, `TYPECHECK OK`, `TEST OK`, format + ADR check clean).

## What I did

**Drift check first.** `getNewToBlockFromError` in `packages/core/src/internal/engine/RangeLogFetcher.ts` was exactly as the finding describes: `-32005` or `-32602`-behind-`looksLikeRangeHint`, then `/\[.*\]/gm` over the message, with `error.data` consulted only on the separate `-32603` path in `getLogs`. No drift; the premise held.

**Re-verified the captures live** (the finding says to, and it had moved). Probed 60+ keyless public endpoints on 2026-09-08. Reproducible script + raw output + analysis in `docs/spikes/a-provider-refusal-is-read-from-its-data-before-its-prose/` (`capture-refusals.sh`, `refusal-shapes.md`). Headlines: the `publicnode` archive refusal is byte-identical three months on; `rpc.mevblocker.io` no longer sends the structured `{from, limit, to}` shape the finding quotes (it span-caps now) and `eth.merkle.io` no longer serves the method at all; Nethermind chains (Gnosis, Chiado, Fraxtal) put the *entire* hint in `data` behind a bare `"invalid params"` message, which the finding did not have; and `-32000` really is a range code on eleven endpoints.

**Implementation** (test written red first, then the code; mutation-checked afterwards):
- `suggestedToBlockFromStructuredData` reads `data.to`, requiring a numeric `limit` so a request echo cannot be mistaken for a refusal descriptor.
- `data` is also read as prose, then the message, each gated on its own text.
- `-32000` accepted under the same gate as `-32602`; the two markers deliberately unchanged.
- `looksLikeRangeHint` kept, given a JSDoc saying what it earns its keep on.

**Tests** (`packages/core/test/rangeLogFetcher.test.ts`, 26): every parse path driven by a verbatim provider body with the endpoint and date named, plus two end-to-end `RangeLogFetcher.getLogs` tests, one asserting halving is unchanged against an uninformative provider and one asserting a named range steers the retry instead. Verified by mutation: removing the gate fails one test, removing `-32000` fails another, ignoring `error.data` fails four.

**Also:** a changeset (`.changeset/a-provider-refusal-is-read-from-its-data-before-its-prose.md`), and one observation note that the finding has partly gone stale (`work/notes/observations/the-getlogs-refusal-finding-has-partly-gone-stale.md`) — I did not edit the finding itself.

## Decisions

**1. `error.data` is read when it is PROSE, not only when it is a `{from, to, limit}` object.** The task's change 1 says "prefer `error.data` over prose"; I read `data` as prose too. Why: the object shape's named source (mevblocker) no longer reproduces, while four live endpoints (Gnosis, Chiado, Fraxtal, plus Ronin) return `{"code":-32602,"message":"invalid params","data":"Query returned more than 50000 results. Try with this block range [...]"}` — a complete machine-readable answer discarded today because the *message* fails the gate. This was the single highest-value real-world win the re-verification found. Alternative considered: object-only, which would have left those four unread and made change 1 nearly unobservable today. What it touches: nothing outside this function, but it is a fourth behaviour beyond the task's three. The gate is applied **per text source** (data and message each gated on their own content), specifically so a hint in `data` cannot license lifting a bracketed pair out of an unrelated message — the composite gate was the tempting simplification and it re-opens exactly the hole change 3 protects.

**2. The structured path requires `limit`, and takes no prose gate.** `data.to` alone is not enough: some providers put the request body in `data`, and reading `to` out of an echo would hand the retry the range that was just refused. `limit` is the node's own cap, so its presence is what makes the object a refusal descriptor. Given that, no `looksLikeRangeHint` check is applied to the structured path, because that gate exists to disambiguate prose. Alternative considered: gating the structured read on the message too, which would have made a machine-readable answer depend on the prose beside it — the thing the task is removing.

**3. `-32000` is widened, but no test can show it producing a value from a captured body, and one test is therefore synthetic.** Across every endpoint probed, not one `-32000` carried a bracketed pair or a structured descriptor: every real `-32000` range refusal states its cap in prose only ("max range: 10000", "maximum is set to 2048"). Extracting a stated number is user story 3 and a different task, so under this task's two mechanisms the widened code changes no answer *today* — it stops the hint being discarded on its code, and it is what story 3 will build on. Consequence: acceptance criterion "`-32000` with a range hint is honoured" and criterion "every parse path has a test whose input is a captured real-world response" are in tension given the evidence. I resolved it by asserting every real `-32000` shape verbatim (four gate-passing, five gate-failing including Cronos's `"maximum [from, to] blocks distance: 2000"`, which is a real bracketed pair that must not be parsed) plus **one clearly-labelled synthetic test** — the Infura body re-coded to `-32000` — whose only job is to fail if someone drops `-32000` from the accepted codes again. Without it, change 2 would have zero coverage. It is labelled synthetic in the test, with the reason, pointing at the spike. Alternative considered: no test at all for change 2 (leaves it silently revertible), or inventing a plausible provider message (worse: it would read as captured). A reviewer may prefer to drop that one test; the rest stand on captures.

**4. The gate's mutation test stays constructed.** Criterion 4's archive assertion (`publicnode`'s `-32602`, verbatim) is present and asserted directly, but it would *not* fail if the gate were deleted, because that message contains no brackets. The test that actually bites is the repo's pre-existing `'invalid argument 0: expected one of [0x1, 0x2]'`, which I kept and re-commented. No provider in the sweep put a bracketed hex pair in an unrelated invalid-params message, so the input that proves the gate bites has to be written; the real refusals the gate rejects are asserted verbatim beside it.

**5. Codes carrying suggestions that this task does not accept are asserted as `undefined` rather than widened.** Alchemy's public endpoint and blastapi return a bracketed suggestion under `-32600`, and Base/Optimism/Blast use `-32614`, Cloudflare `-32047`, Taiko `-32012` with a nested `data.details.maxAllowedRange`. Widening to those is not in this task's scope and would be a new user-visible behaviour; they are recorded in the spike for the cap-extraction task and pinned as currently-unread, so the omission is deliberate rather than an oversight.


wighawag:

The criterion I checked hardest is change 3, because it is the one a tidying diff silently removes: **the `looksLikeRangeHint` gate survived, and it is now defended rather than merely present.**

- The archive refusal is asserted directly (`-32602` "Archive requests require a personal token", `ethereum-rpc.publicnode.com`, re-captured 2026-09-08 and byte-identical to the 2026-06-30 capture) and yields no range hint.
- Beside it is an explicit **mutation test**: delete the gate and `invalid argument 0: expected one of [0x1, 0x2]` starts returning `0x2`, so the fetcher would shrink to a two-block range against an endpoint whose complaint had nothing to do with ranges. The comment says exactly that. That test is honestly labelled as constructed rather than captured, with the reason (no provider in the sweep puts a bracketed hex pair in an unrelated invalid-params message), which is the right call: a criterion asking for real captures should not be satisfied by inventing a fake capture.

The subtle correctness point the diff gets right and that I would have flagged if it had not: **the hint gate is applied to each text source independently** ("never on the strength of the other"), so a hint present in `data` cannot license lifting a bracketed pair out of an unrelated `message`. And the structured `{from, to, limit}` path correctly takes no hint gate at all, because that gate disambiguates *prose*, with the required `limit` doing the guarding instead.

Other criteria:

- **`data` before prose**, in both its shapes: the Infura `{from, to, limit}` descriptor, and — the find of this task — Nethermind's *prose in `data`* behind a bare `"invalid params"` message (Gnosis, Fraxtal), which the old message-only reader discarded entirely while halving blindly against a node that had already given the answer.
- **`-32000` accepted under the same gate**, with the honest footnote that no reachable `-32000` currently carries a machine-readable suggestion — every real one states its cap in prose only, which is story 3's job, not this task's. The widening is what makes that hint readable at all.
- **The unknown-endpoint case is not made worse** — halving is untouched and asserted end to end, including two "states a cap but suggests no range" cases that must fall through to it.
- **Captures are real and provider-named**, and were re-run rather than copied: the task asked for that and it actually happened.

The observation `the-getlogs-refusal-finding-has-partly-gone-stale.md` is exactly the right output of that re-run: `rpc.mevblocker.io` no longer sends the structured shape (it now enforces a 10,000-block span and answers `-32602`), and `eth.merkle.io` no longer serves `eth_getLogs` at all, so two rows of the finding's cap table no longer reproduce. The shape itself is still real (Infura, verbatim in `ethers-io/ethers.js#4703`). Captured as a note rather than fixed in place, correctly, and the full re-run is preserved in `docs/spikes/` alongside the repo's ten existing spike directories.

That note is task-worthy — the finding's `source:` and cap table want amending, and three downstream tasks in this drive read that finding — so it is carried into the end-of-run triage.

Merging.

PR #114, opened 2026-09-08
feat(an-archive-refusal-is-terminal-not-halved): An archive refusal is…

wighawag:
Task: `work/tasks/done/an-archive-refusal-is-terminal-not-halved.md`

Done. Everything is green: `pnpm build && pnpm typecheck && pnpm test` all pass, `prettier --check` clean, `check:adr` clean, working tree holds only the intended changes.

## Drift check (done first)

- **Blocking task landed**: `d92021c` is in `main`'s history, and `getNewToBlockFromError` in `packages/core/src/internal/engine/RangeLogFetcher.ts` has the structured-`data` path, the `-32000` widening and the pinned `looksLikeRangeHint` gate, whose archive test (`packages/core/test/rangeLogFetcher.test.ts`) already asserts the captured refusal yields no range hint. That is exactly the classification this task builds on.
- **The retryable mechanism still works as described**: `RetryableError` (`packages/core/src/errors.ts`) is a structural `retryable` property, read as `err.retryable !== false` by `logFetcher.ts` (`isRetryable`, feeding `withRetries`) and by `@etherfold/fetcher-host`'s `isRetryable` (feeding the backoff/fatal split). No `instanceof` anywhere. I reused it; I invented nothing.
- **Finding vs spike**: read both. The archive capture is byte-identical in `docs/spikes/a-provider-refusal-is-read-from-its-data-before-its-prose/refusal-shapes.md` (2026-09-08), so the premise this task rests on is the part of the finding that did NOT go stale. I did not amend the finding.

## What I built

Test first, then production code.

- `packages/core/src/errors.ts`: new **`ArchiveRefusedError`** (`retryable = false`), carrying `fromBlock`, `toBlock` and the provider's verbatim words. Its message names archive access as the cause, says no range size fixes it, and names the three remedies (archive endpoint, authenticate this one, start from a block it serves).
- `packages/core/src/internal/engine/RangeLogFetcher.ts`: new exported **`archiveRefusalFromError(error)`**, returning the offending text or `undefined`. It reads `error.data`, `error.data.message`, then `error.message` (the same order the range parser reads them in) and classifies only when the text mentions `archive` **and** carries an entitlement word (`token`, `api key`, `plan`, `upgrade`, `unsupported`, `not supported`, `not enabled`). `getLogs`'s catch checks it before the retry budget and before any hint is read, and throws. `getNewToBlockFromError` is untouched.
- `packages/core/src/logFetcher.ts`: one doc line, adding the new error to the list of fatal outcomes a host should stop on.
- Tests (`packages/core/test/rangeLogFetcher.test.ts`): the captured `publicnode` refusal stops the fetcher after **one** request (was 1000 → 499 → 249) with a non-retryable, archive-naming error; an archive mention it cannot classify (`archive node is syncing`) still halves 1000 → 499 → 249 end to end; a dropped socket spends the whole budget and reaches the caller with **no** `retryable` opinion, asserted directly; plus classifier negatives for the real ankr credential refusal, four real range refusals and malformed/absent errors. Every input is a captured body with its endpoint named, except three labelled `CONSTRUCTED` ones (no captured archive refusal puts its text in `data`, and no captured refusal is an archive-transient).
- `packages/fetcher-host/test/classification.test.ts`: the new error added to the list that asserts the host stops on every refusal core throws.
- `.changeset/an-archive-refusal-is-terminal-not-halved.md`: prose changeset, `@etherfold/core` minor + `@etherfold/fetcher-host` patch.

Note on the gate: `changeset status --since=main` reports "no changesets found" while the changeset file is still **untracked** (it diffs against git, which does not see untracked files). It resolves once the runner commits.

## Decisions

**1. The classifier is `archive` AND an entitlement word, not either alone.** This is the width judgement the prompt flags, and both halves are load-bearing. `archive` alone swallows "archive node is syncing" and "archive backend temporarily unavailable", which are transient, and a false terminal is the worse failure (grinding is slow and visible; a false terminal is fast, wrong and permanent). An entitlement word alone swallows the captured `rpc.ankr.com` "You must authenticate your request with an API key", which is about the endpoint rather than about serving history, so every deployment that briefly loses a key would stop. Alternatives considered: (a) substring `archive` only, rejected as too eager; (b) an exact match on the one captured `publicnode` sentence, rejected as matching nothing but one provider's current wording (they revise these: two rows of the finding's own table stopped reproducing in three months); (c) a negative list of transient words, rejected because a deny-list drifts and fails open, while the positive conjunction fails closed onto today's halving. The entitlement words beyond `token` (`api key`, `plan`, `upgrade`, `unsupported`, `not supported`, `not enabled`) are a deliberate, documented widening beyond the single capture. **What it touches**: it introduces a new user-visible terminal failure on the fetch path, so it is reversible only in the sense that widening/narrowing a regex is; the two negative tests (`archive node is syncing`, the ankr message) are what pin the shape.

**2. No error CODE gate.** The captured refusal is `-32602`, but the 2026-09-08 sweep found range refusals under seven different codes, so a code set here would be a list that goes stale the way the finding's cap table did. The identifying evidence is the text; the code adds nothing. Alternative considered: restricting to `-32602`/`-32000`, which would silently miss the same refusal from any provider that codes it differently, and would look like a safety rail while being an arbitrary one.

**3. Read from `error.data` before `error.message`, though no captured archive refusal puts it there.** Parity with `getNewToBlockFromError`, which had to learn this because Nethermind puts its whole complaint in `data` behind a bare `"invalid params"` (Gnosis, Chiado, Fraxtal, captured). Rather than have two functions read one error in two different orders, both read the same three places. The test for it is labelled CONSTRUCTED with that reasoning. Alternative considered: message-only, which is what the evidence strictly supports but re-opens exactly the hole the blocking task closed for the range path.

**4. Thrown BEFORE the retry-budget check, so the last attempt reports it too.** If an endpoint starts gating history mid-backfill on the final retry, the operator still gets the real cause instead of a range error. Costs nothing; noted because it is a deliberate ordering inside the catch.

**5. `ArchiveRefusedError` is a new PUBLIC type in `@etherfold/core`, and it is named for the refusal, not for the remedy.** It sits in `errors.ts` with the other fetch-path refusals (`SuspectedTruncationError`, `NoFetchProgressError`, `TimestamplessLogError`) and is exported through `index.ts`, which is what makes it readable structurally across a package boundary. Alternatives considered: `ArchiveRequiredError` (states a remedy the fetcher cannot be sure of: the endpoint may have an archive and simply not sell it to this connection) and `HistoryUnavailableError` (too close to the `BlockUnavailableError` family, which is about a STORE's retention, not a node's entitlement, and reusing that shape of name would blur two unrelated concepts). Coherence check: `archive refusal` is the task's and the finding's own vocabulary, collides with no `CONTEXT.md` term, and duplicates no existing concept. **What it touches**: `@etherfold/fetcher-host`'s classification list (a test-only change, hence its patch entry in the changeset) and one doc line in `logFetcher.ts`.

**6. No ADR.** The trade-off is real but the decision is cheap to reverse (one regex and one class, both covered by tests that state the hazard in each direction) and it is recorded in the changeset and at the choice site. It does not meet the "hard to reverse" half of the ADR gate.


wighawag:

Every criterion is met, and the classifier is built with the right asymmetry in mind.

- **The captured archive refusal stops the loop** — `RangeLogFetcher.getLogs` throws `ArchiveRefusedError(fromBlock, toBlock, refusal)` instead of halving, asserted end to end ("stops on an ARCHIVE refusal instead of halving through the retry budget").
- **The error names archive access as the cause**, and quotes the provider's own words back verbatim rather than paraphrasing a node, which is why `archiveRefusalFromError` returns the text instead of a boolean.
- **Structurally non-retryable** — `retryable: false` on the error, and `packages/fetcher-host/test/classification.test.ts` pins that the host stops on it, including the case where the error comes from a *second copy of core* (structural classification, not `instanceof`). No caller matches on message text.
- **An unclassifiable refusal still halves exactly as before** — asserted twice, including the specific "mentions an archive but cannot be classified" case.
- **A transient network failure is asserted directly not to be terminal** — "lets a transient NETWORK failure exhaust the budget and stay retryable".
- Changeset present, in prose.

The design call that makes this correct rather than merely working: **the classifier requires the word `archive` AND an entitlement word** (`token`/`api key`/`plan`/`upgrade`/`unsupported`/`not enabled`), and the docstring states the asymmetry it is protecting. `archive` alone would make "archive node is syncing" and "archive backend temporarily unavailable" terminal, and an entitlement word alone would make Ankr's unauthenticated-request refusal terminal on any deployment that briefly loses its key. The reasoning given is the one that matters: **grinding is slow and visible, a false terminal is fast and wrong**, so the eager direction is the dangerous one. Everything ambiguous is left on the halving path it has always taken.

Also correct that it requires no error CODE: the spike found range refusals under seven different codes, so the identifying evidence here is the text, and the captured archive body happens to be a `-32602`.

And it composes cleanly with #113 rather than fighting it: `looksLikeRangeHint` already rejects the archive body (a `-32602` mentioning neither `results` nor `block range`), so the range parser and the archive classifier cannot both claim the same refusal. There is an explicit test that the range parser still yields no hint for it, so the two guards stay disjoint by assertion and not just by construction.

Merging.

PR #115, opened 2026-09-08
feat(a-stated-numeric-cap-becomes-a-ceiling): A cap a provider states i…

wighawag:
Task: `work/tasks/done/a-stated-numeric-cap-becomes-a-ceiling.md`

Done. All checks green (`pnpm build && pnpm typecheck && pnpm test`, `format:check`, `check:adr`).

## What I did

**Drift check first.** The blocking task landed (`d92021c`, "Read a provider refusal from its structured data before its prose"), and the ceiling field is exactly as the task describes it: `RangeLogFetcher.foundNumBlockToHigh` is the discovered upper bound learned by being refused, `safeNumBlock` is the largest span successfully fetched. No drift; I read both the finding and the fresher spike (`docs/spikes/a-provider-refusal-is-read-from-its-data-before-its-prose/refusal-shapes.md`) and took every captured body from the spike, which is where the two disagree.

**`packages/core/src/internal/engine/RangeLogFetcher.ts`**

- New exported `statedBlockCapFromError(error)`: the third reader of one refusal, beside `getNewToBlockFromError` (shrink this retry) and `archiveRefusalFromError` (stop). It extracts a block-span cap a provider wrote out in prose, reading `error.data` before `error.data.message` before `error.message`, like both siblings.
- Extraction is a table of eight tight patterns, each taken from a captured refusal and each carrying its own BLOCK anchor, plus a plausibility filter (whole, `1..10_000_000`). `2K` / `10,000` suffix and separator handling included. The lowest plausible candidate in a refusal wins.
- New `RangeLogFetcher.lowerBlockCeilingTo(cap)`: the ONLY writer of `foundNumBlockToHigh`. The `-32603` "block range too wide/too large" path was re-routed through it, so there is now no expression anywhere that can raise the ceiling; it is one `Math.min`.
- The catch block lowers the ceiling from a stated cap before any branch computes the next range, so the cap bounds the halving fallback AND a suggested range (Alchemy's suggestion honours its 10K log cap and ignores the 2K block cap it states in the same sentence).

**`packages/core/test/rangeLogFetcher.test.ts`** (written before the production code): 20 new assertions' worth of unit tests over captured bodies with the endpoint named in a comment (Immutable, zkEVM, Merlin, Avalanche, Flare, Sonic, Cronos, Kava, 1rpc, Celo, Mantle, Base, Alchemy public, blastapi, Cloudflare, Ronin, plus Alchemy/#4703 and #1816), negative cases for result caps and non-range refusals, implausible numbers, and four end-to-end fetcher tests: the ceiling bounds a suggested range (asks 1,999 rather than 10.9M), a later larger cap never raises it (asserted in the field and in the next cycle's span), a capless refusal halves exactly as before, and the one-block edge. Mutation-checked: removing the wiring fails 3 tests, replacing the `Math.min` with an assignment fails the only-lower test.

**`.changeset/a-stated-numeric-cap-becomes-a-ceiling.md`** — prose changeset, `@etherfold/core: minor`.

## Decisions

- **No error CODE gates the cap reader.** It reads any error's text, unlike `getNewToBlockFromError`, which accepts only `-32005/-32602/-32000`. Chosen because the 2026-09-08 sweep found stated caps under `-32000`, `-32602`, `-32614`, `-32600` and `-32047`, and the identifying evidence is the text, exactly as `archiveRefusalFromError` (which also takes no code) argues. Alternative considered: gate on the same three codes as the range parser, which would drop Base, Alchemy's public endpoint and Cloudflare for no safety gain, since a wrong read here can only lower a ceiling. Touches: a cap is now read under codes no other path in this file accepts; it does NOT widen the codes `getNewToBlockFromError` accepts.

- **A stated cap becomes the ceiling DIRECTLY, so the fetcher asks for cap − 1 blocks.** `foundNumBlockToHigh` means "this many blocks was too high", and a stated cap means "this many is allowed", so the faithful encoding would be `cap + 1`. I chose the conservative one: no message says whether "up to a 2K block range" is inclusive, and one block of throughput buys never having to know. Alternative (`cap + 1`, asking exactly the stated cap) rejected as a guess that can over-ask. Touches nothing outside this file; documented at the choice site.

- **Plausibility bounds `1..10_000_000`, whole numbers only.** The widest cap in the whole captured corpus is 10,000 and the fetcher's own default `maxBlocksPerFetch` is 100,000, so a seven-figure "cap" is a block number, a byte count or a result total a pattern sat next to. This is the acceptance criterion's "clearly not a block count" made concrete; the upper bound is a judgement call and the number is a named constant. Note the bound is load-bearing beyond hygiene: setting the ceiling at all switches the post-success sizing from log-density to ceiling bisection, so a bogus huge ceiling is not a no-op.

- **Deliberate MISSES where a provider does not name the unit.** `range 16777216 exceeds limit of 10000` (Linea, mevblocker) and `GetLogs query must be smaller than size 1024` (Harmony) are real block-span caps that this reader does not extract, because nothing in the sentence says the unit and the neighbouring shape (`logs matched by query exceeds limit of 10000`, Arbitrum) is a RESULT cap. A miss costs the halving path we would have taken anyway; a false positive pins the fetcher to a small range for the process lifetime, slowly and invisibly. Recorded because a later task may reasonably want to widen these.

- **The `-32603` ceiling path was re-routed through the new single writer** (behaviour-preserving: same `Math.min`, same fallback to `maxBlocksPerFetch`). This is what makes "structurally impossible to raise" true rather than merely intended, per the task's headline property. It is a small refactor of code this task did not otherwise have to touch.

- **The range asked for is now floored at ONE block at both ceiling-clamp sites.** A ceiling of 1 previously computed a zero-width, backwards range (`toBlock = fromBlock - 1`). It was already reachable via the `-32603` path when a two-block span was refused; my change makes it reachable from a parsed cap of 1 (QuickNode's discover plan really does state `limited to a 5 range`, so tiny caps are real). Two `Math.max(1, ...)` wrappers, behaviour-identical for every ceiling ≥ 2, asserted by a test. I fixed it rather than filing an observation because my feature widens the door to it; flagging it here since it is a change to pre-existing sizing arithmetic.


wighawag:

All six criteria met, with real captures throughout and the one genuinely dangerous confusion in this area handled explicitly.

- **Both required phrasings parse** — Alchemy's `up to a 2K block range` (quoted verbatim from `ethers-io/ethers.js#4703`, with the `K` suffix handled) and `Exceed maximum block range: 5000` → `5000`.
- **The cap only ever LOWERS the ceiling**, asserted with the case that actually bites: two captured bodies of one shape with different numbers (`rpc.merlinchain.io` 2000, then `zkevm-rpc.com` 10000). The test then checks the *next* cycle is still bounded by 2000, with the exact span sequence `[100000, 1999, 999, 1499]`, so a raise would show up as a concrete different number rather than as a soft assertion. The comment states why this matters: a bigger number arriving after a smaller one is what a misparse looks like from the inside.
- **A refusal stating no cap is unchanged** — `rpc.pulsechain.com`'s numberless complaint (the majority case) still produces the halving sequence block for block, `[1000, 499, 249]`.
- **Implausible numbers are ignored** — zero, negative, non-whole, and absurdly large each rejected, and the two constructed inputs are labelled CONSTRUCTED with the captured phrasing they are built on rather than being passed off as captures.
- **Captures are real and provider-named** throughout, dated 2026-09-08.
- Changeset in prose.

The part that earns the approval beyond the checklist: **a RESULT cap is never read as a BLOCK cap**, with its own describe block. That is the failure this whole area invites, because the two kinds are incompatible and both are stated in the same grammar — and the Alchemy message states *both in one sentence* (a 2K block span and a 10K log count). Reading the 10K as a span there would have the fetcher ask for a range the endpoint has already refused. Arbitrum's log-counting refusal, Gnosis's hint-in-`data`, PulseChain's numberless complaint and the archive refusal from #114 are each asserted to yield no block cap.

Composes correctly with the two PRs before it: the archive body still classifies as terminal, not as a cap, and the range-hint gate is untouched.

Merging.

PR #116, opened 2026-09-08
feat(the-result-cap-is-discovered-not-guessed-by-an-operator): The resu…

wighawag:
Task: `work/tasks/done/the-result-cap-is-discovered-not-guessed-by-an-operator.md`

## What I did

`suspectResultCount` is now resolved in three tiers — `configured` → `reported` → `default` — instead of being a number an operator had to guess. Tests were written first at the two seams (the refusal reader, and the fetcher's resolution/precedence), and both were confirmed to fail before the production code went in.

**Reality check first (the task is a launch snapshot).** The blocking task landed (`d92021c`), and `getNewToBlockFromError` already gates its structured path on `typeof data.limit === 'number'`, so a reported limit does reach this code — but only from the Infura `{from, to, limit}` body (ethers-io/ethers.js#4703, keyed endpoint). **No endpoint in the 2026-09-08 keyless sweep produces that shape any more** (`docs/spikes/.../refusal-shapes.md` records mevblocker dropping it). So a structured-only implementation would be a path nothing in the fresh capture exercises. What IS live there is providers *counting the cap out in words*: Gnosis and Chiado (50000), Fraxtal (20000), zkSync Era and Abstract (10000), Arbitrum One and Nova (10000), plus Alchemy's `a cap of 10K logs`. I therefore read both shapes (decision recorded below).

**Changes**

- `packages/core/src/internal/engine/RangeLogFetcher.ts`: new exported `reportedResultCapFromError` — the fourth reader of one refusal, structured (`{to, limit}`, the two fields gating each other) before prose (patterns anchored on `results`/`logs`, never on a block-span phrasing). Lowest plausible candidate wins; zero/negative/fractional/absurd is ignored **and logged**. `recordReportedResultCap` is the single writer and only ever lowers; `reportedResultCap` exposes it. Recorded *before* the retry budget is consulted, so a refusal on the last attempt still teaches the number. `plausibleBlockCap` was split into a notation-only `capToken` + its own bounds (behaviour-identical, asserted by the existing tests).
- `packages/core/src/logFetcher.ts`: `suspectResultCount` is now a public getter returning `{count, source}`, re-resolved per fetch (a cap learned from the refusal that provoked it applies immediately); new `defaultSuspectResultCount` option for the third tier; log lines when a report takes effect and when configuration overrides one.
- `packages/core/src/errors.ts`: `SuspectedTruncationError` carries and names the `source`, with the fix that follows from it (a REPORTED count is overridden by configuring one).
- `packages/fetcher-host`: `FetcherHostConfig.suspectResultCountSource` (`'configured' | 'default'`), passed through so the host's *default* no longer masquerades as an operator's assertion; the startup line says which tier the number is.
- Docs: `platforms/nodejs-fetcher/README.md` (the three tiers), `CONTEXT.md` glossary bullet, and a prose changeset covering `@etherfold/core` (minor), `@etherfold/fetcher-host` (minor), `@etherfold/platform-nodejs-fetcher` (patch).

**Gate:** `format:check`, `check:adr`, `build`, `typecheck`, `test` all green (1025 core tests, 52 fetcher-host, full suite passing). `changeset status --since=main` still reports "no changesets" only because the file is untracked (changesets reads `git diff` against the diverged commit); it resolves when the runner commits. Working tree holds only the intended changes.

## Decisions

- **A reported cap is read from PROSE as well as from the structured `{from, to, limit}` descriptor the task names.** The task points at `limit` in structured error data, but the fresh capture shows that shape is no longer reproducible on any keyless public endpoint, while seven endpoints count their result cap out in words. Reading only the structured form would have shipped a feature exercised by no provider in the current sweep. The prose patterns are tight and each anchors on what is being COUNTED (`more than N results|logs`, `logs matched by query exceeds limit of N`, `cap of N logs`), which is the same unit-safety rule `statedBlockCapFromError` uses in the opposite direction. Alternative considered: structured only, plus a note that nothing exercises it. Touches: nothing else — `statedBlockCapFromError` (task `a-stated-numeric-cap-becomes-a-ceiling`) deliberately refuses result caps, so this reader takes exactly the shapes that one leaves, and neither was widened.

- **A reported cap only ever LOWERS what was learned (lowest-wins, within a refusal and across refusals).** Mirrors `lowerBlockCeilingTo`, with a sharper asymmetry: a suspect count *below* the node's real cap costs a re-fetched half-range, one *above* it misses the truncation entirely and the receiver deletes state (ADR-0004). One URL can also front several backends with different caps, and the smallest is the only number safe against all of them. Alternative considered: latest-report-wins, which keeps up with a provider that revises its cap upward but can carry a stale-high number into a missed truncation. Touches nothing outside `RangeLogFetcher`; documented at the writer.

- **A new core option, `defaultSuspectResultCount`, exists so a HOST can state the third tier without asserting it.** `@etherfold/fetcher-host` resolved its default (10000) into the same field an operator's value goes in, so *every* deployment looked `configured` to core and no reported cap could ever have filled the gap — the feature would have been dead on the only production path. Core's own fallback (`fetch.maxEventsPerFetch ?? 10000`) is unchanged, per the acceptance criterion, and the host still refuses to let the suspect count follow `maxEventsPerFetch`. Alternatives considered: dropping core's `maxEventsPerFetch` fallback (violates "the existing default is unchanged"), or a boolean flag on core's config (says less and reads worse). Touches: `FetcherHostConfig` gains `suspectResultCountSource`, which is a new required field on that type, and `FetcherHost` now passes `suspectResultCount` conditionally.

- **`LogFetcher.suspectResultCount` changed from a private number to a PUBLIC getter returning `{count, source}`.** It is the acceptance criterion's visibility surface, and the only existing reader was a fetcher-host test reaching in through a cast, which now asserts the public shape. Alternative considered: keeping a number getter plus a second `suspectResultCountSource` getter — two accessors for one answer that must never be read apart. Touches: `packages/fetcher-host/test/classification.test.ts`.

- **`SuspectedTruncationError`'s constructor takes the source as a required third argument.** Criterion 6: the refusal that names this knob must still name it correctly, and "the count this fetcher was **configured** to treat as suspect" is false for a discovered one — worse, the fix differs (a reported count is overridden by configuring one). Alternative considered: an optional parameter defaulting to `'configured'`, rejected because a silently wrong provenance in this particular message is exactly the confusion it exists to prevent. Touches: two test call sites (core, fetcher-host).

- **The runtime `source` is NOT yet in the fetcher host's status/report surface.** It is visible through the getter, the startup line, a log on every change, and the truncation error. Putting the learned values into a status surface is user story 7 of the spec (`the range the fetcher has learned is readable in the status surface`) and belongs to that task, not this one.


wighawag:

All seven criteria met, and the refusal message is now better than it was rather than merely still correct.

- **Precedence is `configured` > `reported` > `default`**, implemented as a discriminated `{count, source}` rather than a bare number, and asserted at each rung directly (`{count: 4, source: 'configured'}` beating a report, `{count: REPORTED_CAP, source: 'reported'}` when nothing is configured, `{count: 10000, source: 'default'}` when neither).
- **An explicitly configured value wins, asserted directly** — and the README states *why* in the right terms: setting it is an assertion about your node, and a number parsed out of an error message is weaker evidence than that.
- **With neither, the default is unchanged** at `10000`.
- **The effective value and its source are visible** — logged when a provider reports a cap, when a reported cap is adopted, and when a configured value overrides one.
- **Nonsensical reports are ignored and logged** — zero, negative, fractional and absurd counts all rejected.
- **The refusal message still names the knob correctly, and now names its PROVENANCE.** `SuspectedTruncationError` takes the source and renders both where the number came from (`configured by this deployment` / `REPORTED by this provider … this deployment having configured none` / `this fetcher default, which nothing configured and no provider reported`) and a fix that differs per source. It keeps the warning against raising `maxEventsPerFetch` to escape it, which is the trap that makes truncation *more* likely. An operator hitting this now knows which of three situations they are in, which is the same standard the timestampless refusal was held to earlier in this drive.
- `SUSPECT_RESULT_COUNT` still reads, still validates, and the README table now says the value is discovered where a provider reports one.

The find worth calling out, because the feature would have been inert without it: **`@etherfold/fetcher-host` was resolving its own default into the same field an operator's value goes into**, so every deployment looked `configured` to core and no reported cap could ever have filled the gap. Splitting `suspectResultCountSource` out of the number is what makes discovery reachable on the deployed path at all. That is in scope — criterion 1 is unsatisfiable in a real deployment otherwise — and it is the kind of thing only visible when you go to wire the feature up.

The two-kinds-of-cap discipline from #115 holds here too: a reported RESULT cap feeds `suspectResultCount`, never the block-span ceiling.

Merging.

PR #117, opened 2026-09-08
feat(the-genesis-check-asks-for-block-zero-not-the-earliest-tag): The g…

wighawag:
Task: `work/tasks/done/the-genesis-check-asks-for-block-zero-not-the-earliest-tag.md`

## Summary

**Drift check first:** both call sites existed exactly as described (`packages/core/src/indexer.ts:1107` live, `:1647` commented-out), and the re-grep confirmed the third site the prompt warned about (`providerSurface.ts`'s `isGenesisProbe`) plus prose in `README.md`, `packages/core/README.md`, `CONTEXT.md` and one unconsumed changeset. No drift that blocks the task.

**Test first** (`packages/core/test/theGenesisCheckAsksForBlockZero.test.ts`, new): a fake node that answers `earliest` and `0x0` DIFFERENTLY (a pruned/regenesis node), plus a genuine mismatch, a node that serves no block 0, an outright failing request, and the `skipGenesisCheck` escape hatch. It went red (4/5) before the production change and green after.

**The fix** (`packages/core/src/indexer.ts`):
- `GENESIS_BLOCK = '0x0'` with a docstring saying why the tag is wrong (lowest block the client HAS; pruned nodes and regenesis chains), so it does not get tidied back.
- The live check moved into `checkGenesisHash`, which separates the three conditions: `GenesisHashMismatchError` (wrong chain, names expected + received, `retryable: false`), `GenesisBlockNotServedError` (could not CHECK, names the two remedies, `retryable: false`), `GenesisCheckUnavailableError` (request failed, carries `cause`, `retryable: true`). None uses another's wording; the test asserts the two non-mismatch messages never say "different/wrong chain".
- The commented-out per-cycle genesis check is DELETED, replaced by a comment saying why and pointing at `one-chain-identity-check-per-cycle-not-two.md`.

**Around it:** `isGenesisProbe` narrowed to `0x0` alone (its `earliest` arm dropped, per its own comment); `theEngineDeclaresItsMethodSet.test.ts` now pins `earliest` as REFUSED; `skipGenesisCheck` got the docstring it never had; the two READMEs, the `CONTEXT.md` glossary entry and the unconsumed method-set changeset were corrected where they still claimed the probe may use `earliest`. New changeset: `.changeset/the-genesis-check-asks-for-block-zero.md`.

**Gate:** `pnpm build && pnpm typecheck && pnpm test` green (core: 1031 passed, 1 skipped, 60 files; all other packages green), `pnpm format:check` clean, `pnpm check:adr` clean, `changeset status --since=main` lists `@etherfold/core` at minor. Working tree holds only the intended changes.

## Decisions

- **The `eth_chainId` checks are left alone, deliberately.** The task offers extending the three-way separation to the neighbouring chain-id check and says to say so if I leave it: I left it. The forward-pointer puts that work on the AFTER-fetch call, which `work/specs/proposed/one-chain-identity-check-per-cycle-not-two.md` (story 4) is going to rewrite anyway to name expected-vs-received; doing it here would put two edits on the same lines for no coverage this task's acceptance asks for, and the load-path chain-id refusal (a bare `Error`, also uncaught) is a third site neither this task nor that spec scopes. Touches: that spec's stories 1 and 4, and `UnexpectedChainError` in `errors.ts`, which is the type that check would eventually reuse.
- **Three error CLASSES, not one class with a `reason` discriminant.** The alternative (the shape `UnexpectedProviderMethodError` uses) collapses three different remedies into one type a caller has to unpack; the repo's stated convention is that the TYPE tells a host what to do and `retryable` is read structurally. Naming avoids the existing vocabulary deliberately: `GenesisBlockNotServedError` rather than anything in the `BlockUnavailableError` family (`NoSuchBlockError` / `BlockNotRetainedError`), which is about what a STORE can answer, not a node; `GenesisCheckUnavailableError` mirrors `IngestionUnavailableError`, the existing retryable-transport member of the same file. Touches: `@etherfold/core`'s public exports (`errors.ts` is re-exported from `index.ts`), so this is API surface and is described in the changeset.
- **`GenesisBlockNotServedError` is `retryable: false` and `GenesisCheckUnavailableError` is `retryable: true`.** A node does not acquire genesis while a caller waits, so retrying the first is a spin; the second learnt nothing about the chain and is exactly the flaky-endpoint case. This is a user-visible policy for a host that loops on `retryable`, hence recorded rather than buried.
- **The commented-out per-cycle check is DELETED rather than repaired.** The task allows either. Dead code cannot be tested, so a repaired copy would rot the same way; and reviving a per-cycle genesis read is a cost-per-cycle decision belonging to the neighbouring spec. The comment left behind names that spec and `checkGenesisHash` as the one place the question is asked.
- **A comment was judged sufficient; no ADR written.** The fix is small, local and single-sited, and the reason now lives in three places a would-be tidier hits: `GENESIS_BLOCK`'s docstring, `isGenesisProbe`'s docstring, and the guard test that refuses `earliest`. I did not disagree with the task's judgement after reading the code.
- **"No `earliest` remains in the repo" is read as: the engine never ASKS with the tag.** Taken literally it contradicts the criterion one line above it (a comment must state why `0x0` is used, which cannot be written without naming the tag). So every remaining occurrence in `packages/` is prose explaining why the tag is NOT used, plus the guard test that asserts it is refused. Historical `work/` records that quote the old spelling are untouched, as is `work/specs/proposed/node-log-api.md`, where `earliest` is a block tag a future read API would resolve, not this check.
- **I edited another task's unconsumed changeset** (`.changeset/the-engine-declares-its-method-set.md`, one parenthetical) along with the two READMEs and the `CONTEXT.md` glossary entry. Narrowing `isGenesisProbe` made their "(`earliest` or `0x0`)" claim false, and a changeset is unconsumed release intent rather than history, so shipping release notes describing a rule the code no longer follows seemed worse than the minimal edit. Touches: the `the-engine-declares-its-method-set` done record's decision 3, which anticipated this narrowing.


wighawag:

All eight criteria met, and both forward-pointers on this task were honoured — including the one that asked for a judgement rather than an edit.

- **The live check asks for `0x0`**, with a comment at the site stating specifically why (`earliest` is the lowest block the client HAS, which is genesis only when the client has genesis; pruned nodes and regenesis chains break it), so a future tidy-up has the reason in front of it.
- **The commented-out per-cycle variant is DELETED, not repaired** — replaced by a comment saying why deletion beat repair (dead code cannot be tested, so it would go stale again) and pointing at `one-chain-identity-check-per-cycle-not-two.md` as the place that decision belongs. That is the better of the two options the task allowed.
- **A node whose lowest available block is not genesis now loads**, covered by a fake provider that answers `earliest` and `0x0` differently — the exact test the task asked for.
- **A genuine mismatch still refuses, naming both hashes.**
- **The adjacent defect is properly fixed**: three distinguishable outcomes, three types — `GenesisHashMismatchError` (wrong chain), `GenesisBlockNotServedError` (the check could not be MADE), `GenesisCheckUnavailableError` (the request failed; retryable, carrying what the node said). Each asserted directly, and none reported in another's wording. This was the part that mattered most: a flaky endpoint at startup was previously indistinguishable from being pointed at the wrong chain.
- `skipGenesisCheck` now has a docstring placing it in the load-time deployment-gate family beside `strictProcessorDrift`, and the escape hatch is asserted to ask the node nothing at all.
- Changeset in prose.

**On "no `earliest` remains in the repo":** met in the sense that matters and correctly not in the literal sense. No code asks a node for the tag anywhere. What remains is (a) ordinary English uses of the word ("the source's earliest `startBlock`"), (b) the comments the *other* criterion explicitly requires, explaining why the tag is not used, and (c) one test that passes `earliest` in order to assert the guard now REFUSES it. A diff that had stripped those to satisfy a grep would have deleted the protection this task exists to install.

**The forward-pointers, both honoured:**

1. `isGenesisProbe`'s `GENESIS_BLOCK` is narrowed to `{'0x0'}`, with the `earliest` arm removed and its history recorded in the docstring — so #111's guard and this fix agree, and the engine now refuses a spelling it no longer uses. The unreleased changeset from #111 was also corrected (`earliest` or `0x0` → `0x0`), which is right: it is unconsumed, and shipping release notes describing a predicate that no longer exists would be a small lie in the record.
2. The optional `eth_chainId` extension was **declined, with reasoning that reads the pointer correctly**: that work belongs on the after-fetch call, which `one-chain-identity-check-per-cycle-not-two` story 4 is going to rewrite anyway, so doing it here would put two edits on the same lines for coverage this task's acceptance does not ask for. It also names a third site (the load-path chain-id refusal, a bare uncaught `Error`) that neither this task nor that spec currently scopes. That third site is worth carrying into the spec when it is tasked.

Merging.

PR #118, opened 2026-09-08
feat(the-learned-range-is-reported-and-can-be-configured-back-in): The…

wighawag:
Task: `work/tasks/done/the-learned-range-is-reported-and-can-be-configured-back-in.md`

## What I built

**Drift check first (the task's own instruction).** Both premises hold: `RangeLogFetcher` still tracks exactly the three in-memory fields the task describes (`foundNumBlockToHigh`, `safeNumBlock`, `numBlocksToFetch`), and the status surface is still `/status` in `@etherfold/server` reporting reorg counters plus an injected `cursor` envelope. Nothing had drifted, so I built it.

**Report it.**
- `LearnedRange` (`@etherfold/core`, `RangeLogFetcher.ts`, re-exported from `types.ts`): `{ceiling?, safeSpan?, nextSize}`, all in blocks, absent rather than zero where nothing has been learned. `RangeLogFetcher.learnedRange` reports it; `LogFetcher.learnedRange` and `LogFetcher.limits` publish it, the latter beside the existing `suspectResultCount` (the previous task in this chain explicitly deferred putting that on a status surface to this one).
- `GET /status` gains `fetcher`, from a new injected reporter `ServerOptions.getFetcherLimits` (carried through `@etherfold/platform-nodejs`): `{reported: true, learnedRange, suspectResultCount}` or `{reported: false, reason}`, and **no field at all** on a host holding no fetcher. `etherfold run` injects one; `index`, `serve` and the Worker host do not.
- The `CeilingProbe` test subclass that existed only to read the ceiling through a protected field is deleted; those tests now read the public getter.

**Accept it back.** `fetch.learnedRange` on the core config, seeded through the existing single ceiling writer (`lowerBlockCeilingTo`), bounded by `maxBlocksPerFetch`; `LEARNED_RANGE` (JSON, the reported object pasted back) on `@etherfold/fetcher-host`, so `run`, `build` and `fetch` all read it, with the startup line saying when a range was remembered.

**Persist nothing.** No store write anywhere; ADR-0074 records the decision and why persisting-in-the-fetcher and pushing-to-the-receiver were both rejected. `LogFetcher`'s class docstring now says out loud that the learned range fails its own "losing it costs one request" test, and that report-and-configure is the answer to that rather than an exception to it.

**Tests** (written at the seam before the production code): 8 new cases in `packages/core/test/rangeLogFetcher.test.ts` (reports the three numbers; starts from a reported range instead of re-paying discovery; adapts down from a refused configured size; lowers a stale configured ceiling and never wedges; ignores implausible/incoherent/over-bound values), 2 in `logFetcher.test.ts`, a new server suite, a new fetcher-host suite, and 2 CLI `run` cases over real HTTP. Docs: ADR-0074, changeset, `CONTEXT.md` glossary (log-fetcher + the `/status` bullet), and the three READMEs. `pnpm build && pnpm typecheck && pnpm test`, `pnpm format:check` and `pnpm check:adr` are all green; `changeset status --since=main` reports my changeset once it is committed (it is untracked, and changesets reads that list from git).

**One observation filed** (not fixed, out of scope): `work/notes/observations/a-tightened-ceiling-leaves-a-safe-span-above-it.md` — a provider that tightens its cap can leave `safeNumBlock` above the ceiling, and the error-path bisection (`floor((ceiling - safe)/2)`, a step with no base, unlike the success path) then asks for a single block before climbing back. Pre-existing; my seeding just makes it easy to reach, and the test that meets it says so.

## Decisions

- **The status surface is `/status`, through a SECOND injected reporter (`getFetcherLimits`) and a new top-level `fetcher` field.** The spec's story 7 says "readable in the status surface", and that surface belongs to the receiving half, so only the COMBINED `run` can report it; a split `etherfold fetch` process has no HTTP surface and gets the startup line and logs only. Alternatives considered: growing the value inside the existing `cursor` envelope (rejected — it re-means a field that is about the cursor, and ADR-0047 reserved growth inside it for the generation dimension), and giving the fetcher its own status surface (a much bigger, separate question). Touches: `@etherfold/server` `ServerOptions`, `@etherfold/platform-nodejs` `StartOptions`, and the CLI's `run` (only `run` injects one; `index`/`serve` deliberately do not).
- **That field is TYPED rather than carried verbatim, unlike the cursor beside it.** ADR-0047 makes the cursor opaque because its meaning lives behind the storage seam and is a processor's (ADR-0027); a learned range is `@etherfold/core`'s (a dependency `@etherfold/server` already has) and is three numbers, so the server names the keys and there is no serialisability probe to run. What is kept from ADR-0047 is the degrade rule: a reporter that throws or has nothing to say yields `{reported: false, reason}` rather than an omission, because "no fetcher here" and "the reporter is broken" are different news. Alternative considered: an opaque `value` blob for symmetry, rejected as hiding a shape both ends already publish. Recorded in ADR-0074.
- **ONE environment variable carrying the whole reported object (`LEARNED_RANGE`), not three.** The round trip is "read the reported object, paste it back", and three variables would make an operator take a report apart and let two thirds of one arrive. Alternative considered: `LEARNED_CEILING` / `LEARNED_SAFE_SPAN` / `LEARNED_NEXT_SIZE`. Touches: the fetcher host's published variable set (documented in `platforms/nodejs-fetcher/README.md`), which `run`, `build` and `fetch` all read; no CLI flag was added, per that command's rule that everything a fetcher deployment tunes stays in the environment.
- **A new startup REFUSAL, but deliberately asymmetric: unreadable is refused, unrecognised is ignored.** Malformed JSON, a non-object, or a member that is not a positive whole number of blocks raise `FetcherConfigError` naming the field (the same class as a malformed `INDEXING_SOURCE`), because that is a misconfiguration an operator can fix and silently doing nothing would leave them watching a rediscovery they thought they had avoided. An unknown KEY is ignored, so a report that grows a field later cannot turn a supervisor that pastes it into an outage — this is a performance hint, and refusing to start over one is disproportionate. Inside core the same values are only ever ignored, never refused, which is defence in depth for a hint arriving through a human.
- **A configured `safeSpan` at or above the configured `ceiling` is dropped rather than clamped or believed.** A span cannot be both known-safe and at-or-above a width that was refused, and the ceiling is the half a refusal backs. I deliberately did NOT make the discovered path coherent the same way (clamping `safeNumBlock` inside `lowerBlockCeilingTo`): that changes the halving arithmetic every deployment depends on, which is outside this task's fence, so it is the observation note above instead. Consequence, documented on the getter: a REPORTED range may carry that incoherent pair honestly, and the configuration side is where it is dropped.


wighawag:

All six criteria met, and the one that a large diff most easily violates — "nothing is written to any store" — is met and then defended.

- **The status surface reports it** (`/status`'s `fetcher` field, `LogFetcher.limits`), with the shape documented in the server and fetcher READMEs and in the new ADR. The reporting path degrades properly: a throwing reporter yields a reason rather than failing the page, a host holding no fetcher carries no field, and a fetcher that has learned nothing reports that rather than inventing a ceiling.
- **It can be handed back** as `fetch.learnedRange` / `LEARNED_RANGE` and is honoured as the starting point, with three good robustness properties I did not ask for: a PARTIAL range is accepted (a deployment may know only its provider's cap), an unknown key is ignored (a newer report can be pasted into an older build), and an unreadable value is refused naming the variable.
- **A configured value is a starting point, never a promise** — asserted three ways: it adapts DOWN from a refused size, it lowers a configured ceiling the provider refuses "and never wedges on a stale one", and it is still bounded by `maxBlocksPerFetch` whatever was configured. Nonsensical configured numbers and an incoherent safe-span/ceiling pair are dropped rather than believed.
- **Nothing is persisted.** I grepped the diff for store writes and there are none. The decision is recorded in ADR-0074 with the alternatives it rejects, which is exactly right for a decision whose whole risk is being quietly reversed later: the note that "with no way to read the value, *remember it* and *write it down inside the fetcher* are the same sentence" is the observation that makes the report-and-configure answer obviously correct rather than merely chosen.
- **A run configuring nothing behaves exactly as today** — "leaves a fetcher told nothing exactly as it was: rediscovering from the starting range".
- Changeset in prose; `CONTEXT.md`'s log-fetcher entry updated to carry the new vocabulary.

Writing ADR-0074 was in scope and correct: the task's own framing is that persisting nothing is *the decision this task makes explicit*, and it is argued from ADR-0003 (the chain-facing half holds no state worth losing) and ADR-0004 (the receiver is authoritative about the only durable number, and a second store is how split brain returns) rather than from taste.

The observation `a-tightened-ceiling-leaves-a-safe-span-above-it.md` is a genuine pre-existing defect found while wiring this up: `lowerBlockCeilingTo` can put `foundNumBlockToHigh` below an established `safeNumBlock`, and the error-path bisection `Math.floor((foundNumBlockToHigh - safeNumBlock) / 2)` then goes negative, so `Math.max(1, …)` fires and the next request asks for a single block. Correctly **not** fixed here — it costs round trips only, it recovers on the following call, and changing that arithmetic moves the halving path every deployment depends on, which is well outside this task's fence. It is carried into the end-of-run triage as task-worthy.

Merging.

PR #119, opened 2026-09-08
feat(the-request-planner-issues-its-list-concurrently): The request pla…

wighawag:
Task: `work/tasks/done/the-request-planner-issues-its-list-concurrently.md`

Done. All checks green.

## What I did

**Drift check first:** confirmed the task's premises hold. `packages/core/src/internal/engine/ethereum.ts` still holds the planner (`generateLogRequestForTopicsAndFiltersCombinations`, unchanged) and `getLogsWithVariousFilters`, which had the single-request early return followed by a `for` loop with an `await` inside it, then the sort and the adjacent-pair de-duplication. No drift. I also verified the sequential path's mid-loop failure behaviour: the first rejection threw out of the function, the later requests were never issued, and nothing partial was returned.

**Test first** (`packages/core/test/theRequestListIsIssuedConcurrently.test.ts`, 9 cases): 3 were red before the production change (concurrent issuance, the bound, the short-circuit) and 6 green (the behaviours that had to survive: the union's exact order and de-duplication, the single-request path, the no-topics path, failing rather than returning a partial).

**Production change** (`packages/core/src/internal/engine/ethereum.ts`):
- Added `MAX_CONCURRENT_LOG_REQUESTS = 4` and `issueRequestsBounded`, a small index-claiming worker pool. Results are written per request INDEX and flattened in request order, so the array the sort and de-duplication are fed is byte-for-byte the array the sequential loop built, whatever order the node answers in.
- The single-request path (and the `eventNameTopics === null` path) is untouched: still one call returned as the node answered it, no sort, no de-duplication.
- Failure: a worker swallows its rejection into a `failures` list and stops claiming work, `Promise.all` over the workers therefore SETTLES (no orphaned promise, no rejection with nowhere to go), and only then is the lowest-indexed failure rethrown. Never a partial union, per ADR-0004.

Also updated one now-false word ("sequentially") in the docstring of `packages/core/test/fetchFilter.test.ts`'s `requestsMade`, and added `.changeset/the-request-planner-issues-its-list-concurrently.md` (prose, `@etherfold/core: patch`).

Gate: `pnpm build`, `pnpm typecheck`, `pnpm test` (whole workspace), `pnpm format:check`, `pnpm check:adr` all pass. `changeset status --since=main` still reports the missing changeset only because the file is untracked (it compares `git diff` against the diverged commit, which cannot see an untracked file); it will be seen once the runner's `git add -A` commit lands.

## Decisions

- **The bound is `4`, a module constant, not a `LogFetcherConfig` knob.** Chosen because a browser allows six connections per origin on HTTP/1.1 and the engine has three other declared methods sharing that pipe, so 4 leaves headroom while still collapsing the common filtered configurations (2-3 requests) into one round trip of latency. Alternatives considered: `Promise.all` over the whole list (rejected: the list length is caller-controlled via `filters`, so it is an unbounded fan-out at a public provider, and the rate-limit refusal that comes back is then mis-read by `RangeLogFetcher` as a range hint and halved against); a `LogFetcherConfig.maxConcurrentRequests` knob (rejected: it is a user-visible default nobody can currently pick better than this, nothing measures it yet, and it would sit next to `learnedRange` as a second, unmeasured performance surface). Touches: `LogParseConfig.filters` (the thing that decides how long the list is) and, if it ever becomes a knob, `LogFetcherConfig` and the `/status` fetcher report.
- **When several requests fail, the error that surfaces is the LOWEST-INDEXED one, not the first to reject.** The pool claims indices in order, so the started set is always a prefix, which makes the lowest-indexed failure exactly the error the sequential loop would have thrown. This is load-bearing rather than cosmetic: `RangeLogFetcher.getLogs` READS that error for an archive refusal, a stated block cap, a reported result cap and a suggested `toBlock`, so a non-deterministic "whichever lost the race" would make the adaptive path depend on network timing. Alternative considered: rethrow the first rejection observed (simpler, rejected for the above).
- **No new request is started once a failure is known, and the in-flight ones are awaited before the error is rethrown.** The first half preserves "the loop never reached the requests after the failing one" (no wasted round trips on an answer that cannot be returned); the second half is what stops a concurrent gather leaving orphaned rejections behind it, since `Promise.all` short-circuits on rejection while its siblings keep running. Asserted directly in the test (`settled === started` after the rejection).


wighawag:

All six criteria met. Criterion 5 was the one I came to check hardest, because a concurrency change that returns a partial union is the failure this repo is least able to absorb — and it is handled better than the criterion asked for.

- **A partial answer is never returned.** A worker swallows its rejection into a `failures` list, `Promise.all` then *settles* rather than short-circuiting, and the error is rethrown only after every issued request has finished. So nothing is orphaned, no rejection is left unhandled, and no caller ever sees the logs that did arrive. The docstring states the reason in ADR-0004 terms rather than as a style preference: a range delivered with logs missing is read by the receiver as an ABSENCE, an absence is concluded as a REORG, and a reorg reverts state. That is the correct weighting — this is correctness work wearing a performance change's clothes.
- **The error that surfaces is the LOWEST-INDEXED failure**, not the first to reject. This is the detail that would have been easy to get wrong and expensive to debug: `RangeLogFetcher.getLogs` *reads* that error for a stated range cap, a result cap and an archive refusal (all three landed earlier in this drive), so surfacing whichever request happened to fail first would have made the hint the fetcher adapts on non-deterministic. Workers claim indices in order, so the started set is always a prefix and the lowest-indexed failure is exactly what the sequential loop would have thrown.
- **No new request is started once a failure is known**, mirroring the loop never reaching requests past the failing one. Asserted directly.
- **The result list is identical** — one array per request in REQUEST order, never arrival order, so the union is fed the same array the loop built. Pinned by "does not depend on which request answers first" as well as by the sort/de-duplication test.
- **The single-request path is untouched**, asserted directly: the node's answer comes back as given, with no sort and no de-duplication introduced, and the no-topics path likewise.
- **Bounded by a stated constant** (`MAX_CONCURRENT_LOG_REQUESTS = 4`), not by the request list, and the reasoning is sound: the list length is caller-controlled, so twenty filters would otherwise be a twenty-way burst at a public endpoint, whose rate-limit refusal the fetcher would then misread as a range hint and halve against. Four also leaves headroom under the browser's six-per-origin HTTP/1.1 limit given the engine's three other declared methods. Keeping it a constant rather than a `LogFetcherConfig` knob is right by ADR-0048 and by this repo's own "measure before optimising" sibling spec: nothing has reported a deployment it is wrong for.

Changeset in prose.

Merging. This is the last task of the drive.

PR #120, opened 2026-09-10
feat(the-browser-indexing-loop-schedules-its-prune): The browser indexi…

wighawag:
Task: `work/tasks/done/the-browser-indexing-loop-schedules-its-prune.md`

All checks green. Summary of what I did:

## What I built

**`packages/browser/src/IndexerState.ts`** — the indexing cycle now schedules the prune its retention implies:

- `advanceOnce` (the one cycle every drive verb goes through: `indexMore`, `indexMoreAndCatchupIfNeeded`, `indexToLatest`, the auto-index tick) runs `pruneScheduled()` **after** the advance has landed and been published, so no `applyBlock` is ever on the stack when a delete runs.
- `pruneScheduled` prunes the state of **every generation the container currently holds**, unconditionally (ADR-0022: a prune is a no-op wherever there is no floor, and the capability report carries no depth, so a host cannot tell a `revert-only` store with a floor from one without). One bounded pass per store per cycle; whatever a pass could not finish is the next cycle's, and an unfinished pass says so through `PruneReport.complete` at info level.
- The store of each generation is recorded in the `createProcessor` wrapper (the first moment both halves of a generation's identity exist), keyed by `{stream, processor}`, so a dropped generation's store stops being pruned and a hot processor swap keeps working. First write wins, mirroring the container's resolve-rather-than-duplicate rule. Cleared on `dispose`.
- New `createIndexerState(..., {pruneBudget})`, defaulting to the exported `DEFAULT_PRUNE_BUDGET = 1000`, validated at construction through the seam's own `pruneBudget()`.

**Tests** — `packages/browser/test/scheduledPrune.test.ts` (written before the production code, red for the right reason first): a window prunes, `revert-only` **with a depth** prunes (asserted explicitly), `unbounded` and `revert-only` with no depth delete nothing, the live version of a row last written 30+ blocks below the floor survives, a budget of 1 bounds each pass and the loop comes back for the rest, a zero budget is refused at the line that configured it, and a seam-watching decorator asserts no `prune` is ever in flight during an `applyBlock`. Everything is asserted on stored version counts and on what a read returns.

**Real-engine evidence** — `browser/cut.ts` gained a `prune` case and `browser/indexing.spec.ts` a test that counts what a real IndexedDB is physically holding (10 unbounded → 6 windowed → 6 revert-only, same state on all three). Passes on chromium; firefox/webkit binaries for this Playwright version are not installed in this environment, so that run is chromium-only evidence.

**Fixture** — `browser/workload.ts` gained `BRANCH_A_LATER` (branch A plus one event 96 blocks later, so a 64-block window's floor sits above every close and below several live versions), `EXPECTED_A_LATER`, and `versionCount(databaseName)`.

**Docs/changeset** — the `retention` option and the browser README now say that what a deployment states it keeps is what it reclaims; `.changeset/the-browser-indexing-loop-schedules-its-prune.md` (minor, `@etherfold/browser`).

No default changed: `retention` is still `unbounded` everywhere. Gate: `format:check`, `check:adr`, `check:refs`, `build`, `typecheck`, `test` all green (`changeset status --since=main` only sees the changeset once it is committed, since it reads `git diff`).

## Decisions

- **The prune attaches to `advanceOnce` (every drive verb), not to `_auto_index` alone.** The README tells apps to call `indexMore` / `indexMoreAndCatchupIfNeeded` on each new head rather than using the timer, so scheduling only inside the auto-index loop would leave the recommended wiring never pruning. Alternative considered: `_auto_index` only (rejected for that reason); inside `Indexer.indexMore` in core (rejected: that is the container's apply path, and the scheduling decision is the host's per ADR-0022). Touches every caller of the hook's drive verbs: a cycle now costs one tip read plus a bounded delete pass per generation.
- **A user-visible default budget, `DEFAULT_PRUNE_BUDGET = 1000`, plus a `pruneBudget` option on `createIndexerState`.** A pass has to be bounded or a backlog blocks a cycle, so a number had to be chosen; 1,000 is argued from the measured IndexedDB prune rate (6.3 s at 62,553 versions on the full-scan prototype) against the loop's 4 s resting interval, and drains the measured workload's whole unbounded footprint in ~30 cycles. Alternatives: no budget at all (one unbounded delete on the cycle that first meets a large backlog), and no knob (a host with a different responsiveness budget could not express it, and the "comes back for the rest" behaviour would be untestable). It reuses the seam's existing vocabulary (`PruneOptions.maxVersions`, `pruneBudget`, `d1PruneBudget`) rather than inventing a word, and it is deliberately **not** an off switch: `retention` is where a deployment says it wants nothing dropped.
- **A misconfigured budget is a REFUSAL at `createIndexerState`.** `pruneBudget({maxVersions})` from `@etherfold/state-store` is called once at hook construction, so `0` throws in the seam's own words where the app wrote it. Alternative: let it throw on the first cycle inside the try/catch, which would turn the seam's deliberate loud refusal into one log line per cycle for ever while the store grew. Touches anyone passing the new option; nothing else can reach the new throw.
- **A prune that throws is logged and the cycle continues.** Indexing is what the tab is for, and a failed delete leaves a store larger than it asked to be rather than a wrong answer. Alternative: propagate, which would stop indexing on a transient IndexedDB failure (and, under `_auto_index`, turn it into the retry-with-backoff path). Deliberately **not** published on `syncing`: whether a store's retention is actually enforced is the sibling task `a-store-reports-whether-its-retention-is-enforced`'s concept, and inventing a second report here would fork it.
- **Which stores get pruned: the generations the container currently holds, keyed by generation identity.** `HeldGeneration` deliberately does not carry the store, so the hook records the pair in its own factory wrapper. Alternative: a flat set of every store the hook ever built (simpler, but goes on pruning a generation dropped on promotion, whose database the registry's `dropState` may have deleted). Known limit, documented at the map: a store handed in already built through `updateProcessor` was not built by a factory here, so a swap onto a genuinely different store is prunable again after the next `init` (the ordinary hot-reload case rebuilds the processor over the same store).

None of these met the ADR gate: each is local to `@etherfold/browser`, reversible in a line, and ADR-0022 already carries the load-bearing decision they implement.


wighawag:

Reviewed the diff against the nine acceptance criteria. Each is met behaviourally, not by assertion-of-intent.

| Criterion | Evidence |
|---|---|
| Store WITH a floor physically drops versions, observed as a falling count | `scheduledPrune.test.ts:83` asserts `versionCount` falls `UNPRUNED → RETAINED`; the real-engine run asserts `windowedVersions === 6` against `unboundedVersions === 10` |
| `revert-only` + `finalityDepth` prunes (the case a binary window-or-not impl gets wrong) | `scheduledPrune.test.ts:112`, asserted explicitly, plus `revertOnlyVersions === 6` in the engine run |
| `unbounded` and `revert-only`-with-no-depth delete nothing; host may call unconditionally; no new seam read invented | `it.each` at `:126` covers BOTH no-floor cases; `pruneScheduled` calls `prune` unconditionally with no capability branch — the ADR-0022 reasoning is in the docstring |
| LIVE version survives however old | `:155` — tokens last written at 102/104 under a floor of 136, read back through the store itself and not only the hook handle |
| Bounded per pass, continues until `complete` | `:178` counts passes at a budget of 1; the continuation is across CYCLES, which is what the criterion's own "never blocks one cycle" requires |
| Apply path performs no deleting | `:240`; `pruneScheduled()` is called at the end of `advanceOnce`, after `reportGenerationProgress`, outside every apply (ADR-0022) |
| Behavioural assertions | counts + returned reads throughout, no mock-call assertions |
| IndexedDB evidence from the real engine, not `fake-indexeddb` | new Playwright case in `packages/browser/browser/indexing.spec.ts` + `cut.ts`/`workload.ts` prune case; the docstring correctly names WHY only the engine can make this claim (`upper: null` is not a valid IndexedDB key) |
| Changeset | `.changeset/the-browser-indexing-loop-schedules-its-prune.md` |

Constraints the task carried, all honoured:
- **Floor, not window.** The trigger is `retentionFloor` via an unconditional `prune`, never `retention.kind === 'window'`. The `revert-only` trap is the subject of its own test.
- **No new seam read** to ask whether a store has a floor. None added.
- **Out of scope respected.** `BrowserStateStore.ts` is 10+/0- and docstring-only; no default changed and the browser's default retention is NOT bounded. `pruneBudget` is a new optional knob, not a change to an existing default.

Two judgement calls I checked rather than waved through: the budget default (1000) is justified from the recorded 6.3 s / 62,553-version measurement rather than picked, and validation happens once at configuration via the seam's own `pruneBudget` so a bad value fails where it was written instead of once per cycle for ever. Both are the right call.

Gate 1 (acceptance) green in a fresh worktree rebased onto main. Gate 2 off by repo config for this drive.

PR #121, opened 2026-09-10
feat(the-cli-schedules-the-prune-its-retention-implies): The CLI schedu…

wighawag:
Task: `work/tasks/done/the-cli-schedules-the-prune-its-retention-implies.md`

The gate is fully green. Summary of what I did and verified:

## What I did

This was a continuation. The prior run's implementation was already on `work/task-the-cli-schedules-the-prune-its-retention-implies` and was bounced by the acceptance gate at `pnpm changeset status --since=main`, before `build`, `typecheck` or `test` ever ran.

**The fix (one line):** deleted `'@etherfold/platform-cf-worker': patch` from `.changeset/the-cli-schedules-the-prune-its-retention-implies.md`. That package is `private: true` and `.changeset/config.json` sets `privatePackages: false`, so changesets treats it as ignored and refuses any single changeset spanning both sides of that line. It needs no entry at all: it is never published, and the change to `platforms/cf-worker/src/d1.ts` is a JSDoc block with no executable change. The changeset prose is unchanged; it still describes the D1 docstring update in its closing lines.

**Then I ran the whole gate for the first time against this work.** All of it is green:

- `pnpm format:check` — clean
- `pnpm check:adr` — 74 ADRs, no duplicates
- `pnpm check:refs` — all work/ references resolve
- `pnpm changeset status --since=main` — now accepted (`@etherfold/state-store` minor, `etherfold` minor, plus dependents at patch)
- `pnpm build` — all 20 projects
- `pnpm typecheck` — all 19 projects, tests included
- `pnpm test` — every package passes, including the two new suites

No production code needed changing: nothing behind the changeset refusal was broken.

## Acceptance criteria, re-checked against what is actually on the branch

- **A CLI deployment with a floor drops versions, asserted on stored counts** — `packages/cli/test/scheduledPrune.test.ts` counts rows in the canonical namespace's tables: 16 unpruned versus 11 with `--retention 20`, and the windowed store answers identically to the unbounded one.
- **`revert-only` with a `finalityDepth` prunes** — asserted; the CLI always states a depth (the stream's own), so `retentionFloor` returns one and the same five closed versions go.
- **The live version survives however old it is** — token 0, written 50 blocks below the floor, is still the current state; all 8 live token rows survive.
- **The loop takes its budget as a PARAMETER and loops on `complete`** — `pruneMore(states, {maxVersions})` in `@etherfold/state-store` (not in the CLI), so `@etherfold/platform-cf-worker` can drive it with `d1PruneBudget(plan)` without depending on the CLI; the `d1.ts` docstring now shows that call.
- **Budgeted passes reach the same end state as one unbounded pass** — asserted directly in `packages/state-store/test/pruning.test.ts`, on surviving rows and on a follow-up `prune()` reporting nothing left, in both stores.
- **Nothing in the ingest or apply path deletes** — asserted with spies: `prunesDuringAnApply` is 0 while the fold runs, and in `entityStore.test.ts` every `prune` invocation ordinal is greater than every `applyBlock` one.
- **`@etherfold/server` gains no state-store dependency** — confirmed: its deps are still `@etherfold/core`, `hono`, `named-logs`, `remote-sql`. The host is `packages/cli` (`folding.ts` builds the store, `driveCycles` is the loop).
- **A changeset accompanies the change** — present and now accepted by the gate.

Working tree holds only the intended change (`git status`: one modified changeset file, no untracked strays).

## Decisions

- **The changeset drops `@etherfold/platform-cf-worker` rather than splitting into two changesets.** A private package is never published, so it needs no changeset entry, and the only change to it is a JSDoc block with no executable change. The alternative (a second changeset for the private package) would create a release note for something that never releases, and `privatePackages: false` means changesets would refuse to version it anyway. Touches: nothing beyond this change's release notes. The general trap is captured at `work/notes/observations/a-changeset-mixing-private-and-published-packages-fails-late.md`.

The decisions below are the prior run's build-time rationale, restated because they are the rationale for the work being landed and each was re-verified as still accurate against the code and the now-green gate:

- **`build` prunes too, at its exit, not only `run`.** The task's prompt names `run.ts`'s loop, but the spec's tasking amendment says what ships is "unconditional pruning in every host this project ships", and a `build` database is a publishable artifact for which "prunes eventually" is not a property. Since `catchUpDelayMs` defaults to `0`, the between-cycles hook barely fires on a one-shot, so `build` drains with bounded passes *after* the loop ends (skipped after a `fatal`, and skipped when an external `deps.signal` stopped it: a caller asking a process to stop is not asking it to finish a delete). Alternative considered: `run` only, which would leave `build --retention 50000` emitting an artifact holding every version. This is what required editing the existing `entityStore.test.ts` assertion `expect(prune).not.toHaveBeenCalled()`; it now asserts the ADR-0022 property directly (every prune call is after the last `applyBlock`). Touches: `build`, `oneShot`/`equivalence` behaviour under a floor (unbounded default is a no-op, so those stay identical).
- **`etherfold index` is deliberately NOT covered.** The receiving half is fed over the wire and has no cycle to prune between; the only in-band place would be the ingest path, which ADR-0022 refuses. Giving it a timer or an admin route is a design decision with its own surface, so it is left out, said so in its README row, and the residual gap is captured at `work/notes/observations/etherfold-index-schedules-no-prune.md`. Touches: the `index` command and anyone reading "every shipped host prunes".
- **A new concept in `@etherfold/state-store`: `pruneMore` / `ScheduledPruneReport`.** Checked against the glossary: `prune` keeps its exact meaning (one store's call), `retention` is untouched, and the name mirrors `rebuildMore` — the shape `CONTEXT.md` already names for "bounded work per call, reporting whether it finished, with the host looping". It sits at the seam layer rather than in the CLI because AC4 asks for a loop a D1 host can drive without reimplementing it, and `@etherfold/platform-cf-worker` cannot depend on the CLI. Alternative considered: CLI-local, which would have made the Worker claim aspirational. It also promotes `@etherfold/state-store` from a devDependency to a direct dependency of `etherfold`.
- **`DEFAULT_PRUNE_BUDGET = 10_000`, a constant with no flag.** A pass must be bounded or the first cycle meeting a backlog pays for all of it; the number is argued from the measured 1.1 s at 62,553 versions (≈0.2 s per pass) against a 4 s poll interval, and drains the measured 29,393-version footprint in three cycles. No `--prune-budget` flag was added: the CLI's command table treats every input as owned/optional/refused per command and a new flag is a surface change bigger than this task, while a local database imposes no allowance of its own (the platform that does passes `d1PruneBudget` to the same parameter). It is exported so a caller can see it, and it is not an off switch — `--retention` is where a deployment says what it wants kept. Touches: `run`/`build` cycle cost, and any future `--prune-budget` task.
- **The budget is spent across the states held, not per state.** A host mid-upgrade holds two generations; a per-store budget would do twice the work it budgeted for, which on a platform where the budget IS the request's query allowance is the difference between a scheduled prune and a rejected invocation. Touches every caller of `pruneMore`, including the Worker example now in `d1.ts`.
- **A failed prune is logged and the cycle continues** (both commands), mirroring the rebuild chunk beside it: indexing is what the process is for, and a delete that could not run leaves a store larger than it asked to be rather than a wrong answer. A `build` whose final prune fails still exits `0`, because it folded everything it was asked to fold.

None of these met the ADR gate: each is local and reversible in a line, and ADR-0022 already carries the load-bearing decision they implement.


wighawag:

This task bounced once on the acceptance gate and was resolved back into the loop; this review covers the branch as it now stands, and deliberately re-checks the criteria against what is on it rather than trusting the earlier run.

**The bounce is genuinely closed.** The gate had stopped at `pnpm changeset status --since=main` on a mixed changeset naming the private `@etherfold/platform-cf-worker` beside two published packages, which meant `build`, `typecheck` and `test` had never run against this work at all. The changeset now names only `@etherfold/state-store` and `etherfold`, the prose is intact, and the full gate has now run green in a fresh worktree rebased onto main. `platforms/cf-worker/src/d1.ts` remains JSDoc-only (no executable line in the diff), so dropping its entry lost nothing.

| Criterion | Evidence |
|---|---|
| CLI deployment with a floor drops versions, asserted on counts | `cli/test/scheduledPrune.test.ts:163` |
| `revert-only` + `finalityDepth` prunes, because it is a floor | `:184` |
| Live version survives however far below the floor | `:197` |
| Loop takes its budget as a PARAMETER and loops until `complete` | `pruneMore(states, {maxVersions})` in `state-store/src/pruning.ts`; `:53`/`:62` cover both report states |
| Budgeted passes reach the same end state as one unbounded pass, **asserted** | `state-store/test/pruning.test.ts:85`, exactly the criterion |
| Nothing in the ingest or apply path deletes | `cli/test/scheduledPrune.test.ts:248` — "reaches a prune only from the host, never from the path that applies a block" |
| `@etherfold/server` NOT given a state-store dependency | `packages/server/package.json` untouched; the dependency moved dev→prod in `packages/cli` only, correctly, since the CLI is the host that now imports `pruneMore` at runtime |
| Changeset | present, and now well-formed |

The premise the task asked to have MEASURED rather than assumed — that the host is the CLI and `@etherfold/server` constructs no store — holds: I confirmed independently that `server`'s dependencies are `@etherfold/core`, `hono`, `named-logs`, `remote-sql`, and that the only `new VersionedStateStore(...)` on the server side is `packages/cli/src/folding.ts:297`.

Three things I checked rather than waved through:

- **The budget is spent ACROSS states, not per state.** That is the difference between a Worker's query allowance being respected and being exceeded N-fold on a host holding several generations, and it is asserted (`pruning.test.ts:115`), as is pruning a state held twice exactly once (`:135`).
- **`run` and `build` loop differently, on purpose.** One bounded pass in the cycle gap (`index.ts:435`) versus passes until the floor is reached for a one-shot whose database is a publishable artifact (`:481`). `pruneHeldUntilComplete` guards termination by returning on a zero-progress pass, so a store that cannot drain cannot spin.
- **Failures do not fail the fold.** `pruneMore` deliberately does not catch, leaving the decision to the host; the CLI cycle logs and continues.

The honest gap is recorded rather than hidden: `work/notes/observations/etherfold-index-schedules-no-prune.md` notes that `etherfold index` still schedules no prune, which is correct scoping, since it is fed over the wire and a prune in the ingest path is what ADR-0022 refuses.

PR #122, opened 2026-09-10
feat(the-prune-claims-in-the-docs-match-the-code): The prune claims in…

wighawag:
Task: `work/tasks/done/the-prune-claims-in-the-docs-match-the-code.md`

Done. Both files amended; working tree holds only those two changes.

## What I verified before writing

- `work/notes/findings/sqlite-in-the-browser.md:147` carried the claim that `@etherfold/state-store-sqlite` "has no pruning at all, so every backend it ships today is effectively `unbounded`".
- `packages/state-store-sqlite/src/store.ts:383` `prune()` is real: `pruneBudget`, `retentionFloor` off the tip, row ids SELECTed in `maxRowsPerStatement` chunks then deleted, never touching a live version or a block row, `logger.info` of what it dropped, and a `complete` probe. No `VACUUM`, by documented choice.
- `packages/state-store-indexeddb/src/store.ts:460` `prune()` is a range scan over `UPPER_INDEX` (`IDBKeyRange.upperBound(floor)`), oldest close first, budgeted, with the same `complete` probe.
- Hosts: `packages/browser/src/IndexerState.ts` (`pruneScheduled`, per cycle) and the CLI's `run`/`build` (`packages/cli/src/pruning.ts`, driven from `packages/cli/src/index.ts:435,481`) DO schedule a prune today. `etherfold index` does not (already captured in `work/notes/observations/etherfold-index-schedules-no-prune.md`), and a bespoke Worker host still wires its own via `d1PruneBudget` (`platforms/cf-worker/src/d1.ts`).

## What I changed

- `work/notes/findings/sqlite-in-the-browser.md`: the superseded sentence is left VERBATIM (it is the traceable record of what was true on 2026-08-22) with an inline dated `[SUPERSEDED ... amended 2026-09-10]` marker on the same line, followed by a `> **Amended 2026-09-10** by the-prune-claims-in-the-docs-match-the-code` block that separates the backend half (now false: both backends prune, with the mechanism named) from the host half (now partial: which hosts schedule, which do not), and says explicitly that the footprint row measured the spike's prune-plus-`VACUUM`, not the shipped one. No measurement, table or conclusion touched.
- `docs/spikes/sqlite-in-the-browser/src/store/versioned-sql.ts`: the same correction in the `prune` docstring, as an `AMENDED 2026-09-10` paragraph quoting the old wording. Spike BEHAVIOUR unchanged (comment-only diff, confirmed by `git diff`).

## Checks

- `grep -rn "no pruning at all" work/notes docs/spikes` → exactly two lines, each one carrying its dated amendment marker on the matched line itself. `work/tasks/done/` records untouched.
- `pnpm check:refs` green (838 files, 202 artifacts, all refs resolve). `pnpm check:adr` green (74 ADRs, no duplicates).
- `pnpm format:check` / `changeset status` could not run here (no `node_modules` in this worktree), but the change touches no workspace package (`docs/spikes/*` is outside `pnpm-workspace.yaml` and `private`), so no changeset is owed, and `.prettierignore` excludes `docs/` and `*.md` anyway.
- Two notes on the task's own coordinates, for the runner: the body is at `work/tasks/backlog/the-prune-claims-in-the-docs-match-the-code.md` (not `ready/`) and the spec at `work/specs/tasked/a-configured-window-is-actually-pruned.md` (not `ready/`). I read both and touched neither.

## Decisions

- **The amendment does NOT say "no host called it", because that is no longer true, and I proceeded rather than stopping.** Acceptance criterion 2 asks the correction to distinguish "the backends implement pruning, and no host called it", and the task's What-to-build proposes the narrower claim "no HOST in the repository schedules a prune, so retention is advisory in every DEPLOYMENT". That second premise has itself gone stale since the task was written: `the-browser-indexing-loop-schedules-its-prune` (e4755317) and `the-cli-schedules-the-prune-its-retention-implies` (27b6e65e) both landed, so the browser loop and the CLI's `run`/`build` do schedule one. Writing the task's suggested sentence would have put a NEW false claim into a finding, which is exactly the failure this task exists to repair. I therefore satisfied the criterion's INTENT (separate what the backend can do from what a host actually calls) with the verified current state: backends prune; browser and CLI `run`/`build` schedule it; `etherfold index` and bespoke hosts do not. Alternative considered: STOP and route to needs-attention as drift. I judged it not load-bearing and trivially reversible: it is a prose amendment in a note, the core premise of the task (the finding's backend claim is false and must be corrected in place) is intact and confirmed against the code, and the task's own Prompt instructs me to confirm the current truth myself before writing. This touches no code, no flag and no other task; it does mean a reviewer comparing the amendment to criterion 2 word-for-word will see a deliberate divergence, which is why it is recorded here.
- **The superseded sentence is retained verbatim with an inline marker rather than rewritten into past tense.** A finding's value is that a claim can be traced to when it was measured, so the amendment adds and never overwrites. The practical consequence is that the grep phrase still exists in `work/notes/`, which is why the dated marker sits on the SAME LINE as the phrase: any future grep for the stale claim lands on text that immediately declares itself superseded. Alternative considered: editing the sentence to past tense (would have removed the phrase from the grep but made the old claim unquotable and the diff a silent edit).
- **No observation note opened for the spec-level repetitions.** `work/specs/tasked/one-processor-everywhere.md:66` and `work/specs/tasked/a-configured-window-is-actually-pruned.md:14` still carry the old claim, but both specs are explicitly launch snapshots ("records intent at creation, NOT maintained"), so that is contract-sanctioned staleness rather than drift worth a signal, and the grep criterion scopes them out on purpose.


wighawag:

This is one of the two prose-only tasks in this drive where the acceptance gate is nearly blind: `.prettierignore` excludes `docs/` and `*.md`, so `format:check` never sees these edits and an empty diff would have passed everything else. The grep criterion is the real check, so I ran it myself against the branch rather than trusting the gate.

**The criterion, executed:** `grep -rn "no pruning at all" work/notes docs/spikes` returns exactly 2 matches, and 2 of 2 carry a dated amendment marker (`SUPERSEDED; amended 2026-09-10` in the finding, `AMENDED 2026-09-10` in the spike docstring). No bare occurrence of the superseded claim survives in either scanned tree.

| Criterion | Evidence |
|---|---|
| Corrected in place as a DATED AMENDMENT, not a silent edit | The finding's original sentence is preserved VERBATIM and flagged inline as superseded, with the amendment in a block quote beneath it. A reader meeting the old claim cannot miss that it moved. |
| Distinguishes the two things now different | Explicit, and it is the amendment's organising point: the sentence "bundled two separate things, a BACKEND that cannot prune and a DEPLOYMENT that therefore keeps everything, and they have come apart" |
| Same correction applied to the spike docstring | `docs/spikes/sqlite-in-the-browser/src/store/versioned-sql.ts`, amended and cross-linked back to the finding |
| No measurement, table or conclusion altered | Diff is 9+/1- in the finding, confined to the pruning paragraph; every number, the footprint row and all conclusions are untouched |
| Grep scoped on purpose; `work/tasks/done/` NOT rewritten | Confirmed: the only `work/tasks/done/` file in the diff is this task's own done record. The historical records carrying the same phrase are left alone, which is exactly right. |
| `check:refs` / `check:adr` green | Green in the gate; the new cross-reference to `work/notes/observations/etherfold-index-schedules-no-prune.md` resolves, since that note landed with PR #121 |

Two things raise this above merely satisfying the criteria:

- **It preserves the finding's value rather than tidying it away.** The superseded sentence stands as written because it records what was true on 2026-08-22, and the amendment says so explicitly. That is the correct handling for a finding, whose worth is that its claims trace to when they were measured.
- **It is current as of THIS drive, not as of when the task was written.** The task's own framing said "no HOST in the repository schedules a prune", which was true when authored and stopped being true two PRs ago. The amendment catches that: it records that the browser loop and the CLI's `run`/`build` now schedule a pass, while `etherfold index` and bespoke hosts do not. The builder measured the world instead of transcribing the task, which is what was asked for.

Spike BEHAVIOUR is unchanged: the `prune(floor)` body is untouched and only its docstring moved, so the exhibit those numbers were taken with still is that exhibit.

PR #123, opened 2026-09-10
feat(revert-only-is-the-documented-way-to-ask-for-no-history): revert-o…

wighawag:
Task: `work/tasks/done/revert-only-is-the-documented-way-to-ask-for-no-history.md`

Done. Prose-only change, exactly as the task scoped it.

## What I changed

**`packages/browser/src/storage/state-store/BrowserStateStore.ts`** (the point of configuration, `BrowserStateStoreConfig.retention`): added a section "Reorg safety and NO history: write `revert-only`, not a small window" saying the three things the criteria ask for:
- `revert-only` is the way to ask for reorg safety with no history, it is a `retention` value so the **default IndexedDB backend takes it too** (not only the light store that is `revert-only` by construction, ADR-0023), and it reports `capabilities.asOf === false` so a caller learns at startup rather than from a refused read later, with the reason `asOf` is a separate field (the two fail differently).
- Why a small window is not a substitute, measured: retention counts BLOCK NUMBERS and never updates (ADR-0019), and event-bearing blocks on the real stream are median **429 blocks apart, max 1,226,194** (`work/notes/findings/sqlite-in-the-browser.md`), so `{blocks: 64}` typically holds exactly one of them (the tip's) and often none: a store that refuses almost every historical read while looking configured for history.
- What `revert-only` still guarantees: revert keeps working, because its floor is the finality depth, which is exactly the set of superseded versions a revert reopens (`retentionFloor`). The `finalityDepth` field's own docstring now says why it belongs beside `revert-only` and not only beside a window.

**`packages/browser/README.md`**: the same paragraph, beside the existing "What you say you KEEP is what the tab keeps", which is the other place a browser app developer meets the choice.

**`packages/state-store/src/retention.ts`**: per-arm docs on `RetentionSetting`'s `'revert-only'` and `'unbounded'` members (the `{blocks: N}` arm already had one), so an editor showing the union at the point of choice describes them. No type change: comments only.

**`packages/browser/test/browserStateStore.test.ts`**: one added test pinning the claim the docstring now makes on the DEFAULT backend, `{retention: 'revert-only', finalityDepth: 64}` reports `{retention: {kind: 'revert-only'}, asOf: false}`. It passes on the code as it stands (this task changes no behaviour); its job is to make the prose checkable, so it cannot drift silently.

**`.changeset/revert-only-is-the-documented-way-to-ask-for-no-history.md`**: patch for `@etherfold/browser` and `@etherfold/state-store` (the standing convention; `changeset status --since=main` would otherwise flag both packages at land time).

## Checks

`pnpm format:check`, `pnpm check:adr`, `pnpm check:refs` green; `pnpm build` and full `pnpm typecheck` green; `@etherfold/browser` (25 files, 160 tests) and `@etherfold/state-store` (9 files, 138 tests) green. `changeset status --since=main` cannot see an untracked changeset locally (it diffs against the merge base), so it reports the two packages as uncovered until the runner commits; the changeset parses and resolves to both packages under a plain `changeset status`.

Working tree holds only the four edits plus the changeset. No behaviour, default or type changed; no git operations performed.

One note for the runner, not a task problem: the item body is at `work/tasks/backlog/revert-only-is-the-documented-way-to-ask-for-no-history.md`, not `work/tasks/ready/` as the prompt stated. I read it there and left it untouched.

No in-scope design decisions were made, so there is no `## Decisions` block: every choice here was either dictated by the acceptance criteria or was a factual statement verified against the code (`IndexedDBStateStore.capabilities` reporting `asOf: false` for `revert-only`, and `retentionFloor`'s `revert-only` arm flooring at the finality depth).


wighawag:

The second prose-only task in this drive, and the same caveat applies: `.prettierignore` excludes `docs/` and `*.md`, so the gate barely covers it and an empty diff would have passed. I checked the substance directly.

| Criterion | Evidence |
|---|---|
| Browser retention config documents `revert-only` as the way to say reorg safety with no history, and says it reports `asOf: false` | `BrowserStateStore.ts`, in the `retention?:` docstring — the point where an app actually chooses. It also explains WHY `asOf` is reported separately from `retention`: they fail differently, one refusing outside a window and the other everywhere. |
| States the measured reason a small window is not a substitute | Cited with provenance, not as folklore: block numbers rather than updates (ADR-0019), median **429 blocks** between event-bearing blocks and max 1,226,194 on the real stream, attributed to `work/notes/findings/sqlite-in-the-browser.md`. The consequence is drawn concretely: `{blocks: 64}` typically holds exactly ONE, "a store that refuses almost every historical read while looking configured for history". |
| States what `revert-only` still guarantees | The reorg half survives, because its floor IS the finality depth, so the versions a revert reopens are exactly the ones kept. This is also why the expanded `finalityDepth` docstring now says it is load-bearing beside `revert-only` rather than only beside a window. |
| No behaviour, default or type change | Verified against the diff: the only executable additions are one TEST and docstrings. The `RetentionSetting` union is byte-identical in its members, with doc comments inserted between them; `BrowserStateStoreConfig`'s shape is unchanged; no default moved. |
| `check:refs` / `check:adr` green | Green in the gate, including the new citation of the finding |

Two judgement calls worth naming:

- **It documents in two places, and both are right.** The task asked for the point of configuration, which is the browser config type. The builder also annotated the `RetentionSetting` union members in `retention.ts`, which is what a developer sees on hover in any package. That is a real widening of scope, but it is the same sentence at the other end of the same choice, and the task's goal ("legible at the point of configuration") is better served by it. Not scope creep.
- **The added test is not a behaviour change.** `browserStateStore.test.ts` asserts that a `revert-only` store reports `{retention: {kind: 'revert-only'}, asOf: false}`. It pins the exact claim the new prose makes, so the documentation cannot silently drift from the code it describes. Given this task is prose whose gate is nearly blind, that is the most valuable thing in the diff.

The task warned "if you find yourself changing a type or a default, you are in the wrong task". The builder did not.

PR #124, opened 2026-09-10
feat(every-mutating-path-carries-a-writer-token): Every mutating path c…

wighawag:
Task: `work/tasks/done/every-mutating-path-carries-a-writer-token.md`

Done. Here is what I built.

## What landed

- `src/writer.ts` (new): the writer-token concept, and `writerToken()` — an opaque value only one claim could produce (nothing orders or parses it).
- `errors.ts`: `StoreWriterChangedError`, declared once at the seam beside `BlockNotRetainedError`, carrying which mutating path was refused and no token.
- `capabilities.ts`: `StateStoreCapabilities.singleWriter`, documented against the stream keeper's existing "one-writer rule" so the two are not confused.
- `store.ts`: the guard obligation stated on the interface, with `migrate` explicitly excluded.
- `MemoryStateStore` and `PatchStateStore` report `singleWriter: false` honestly (their storage is an instance field; a token there compares only with itself).

**IndexedDB** — new `writer` object store (schema 3, `contains`-guarded), and `claimOrCheck` first inside every mutating transaction: `applyBlock`, `revertTo`, `writeCursor`, `clearCursor`, `prune`. `prune` was widened to read the tip *inside* the `readwrite` transaction that deletes against it.

**SQLite** — `_writer` inside the table namespace; every mutating statement ANDed with `COALESCE((SELECT token …),'') = ?` (inserts become `INSERT … SELECT … WHERE guard`), the claim first and the token read back last in the same batch, ADR-0054's shape verbatim. Guarded: `applyBlock`, `applyBlocks` (per batch), `revertTo`, `writeCursor`, `clearCursor`, `prune`, `drop`.

**Conformance** — `cases/single-writer.ts` (12 cases) gated on the claim, plus an optional `twoWriters: {sharingStorage, addressedApart}` option, wired into the SQLite, IndexedDB and real-browser runs. `the-suite-catches.test.ts` gained a store that *claims* a single writer and lets a second one write, and asserts the new cases go red.

**Evidence it is not decoration** — `state-store-sqlite/test/two-writers.test.ts` lands a rival's committed write in the exact window between the tip read and the delete it decided (wrapped handle, no timers); neutering the guard turns 6 of its 8 cases red, which I ran. `state-store-indexeddb/test/two-writers.test.ts` starts two writers in the same tick and pins that `prune`'s tip read is in the deleting transaction; neutering the check turns 2 of 3 red. Chromium browser run: 5/5 green, 312 conformance cases, multi-tab showing one winner (20 writes) and three refused (19 each), audit `missing: 0, surplus: 0`.

`pnpm format:check && check:adr && check:refs && build && typecheck && test` all green (2 900+ tests). ADR-0075 and a changeset are in. `changeset status --since=main` only fails locally because the changeset file is still untracked; it resolves on the commit.

## Decisions

**`singleWriter` is a REQUIRED capability field, not optional.** Chosen so a new backend must answer the question rather than inherit a default; the alternative (optional, absent = false) would have let a backend that *does* enforce it silently claim nothing, and a backend that does not silently look honest. Touches every `StateStoreCapabilities` literal (six test files updated) and any future backend. On coherence: the repo already has a **one-writer rule** meaning "only the indexing generation appends to a stream" (ADR-0044, stream keeper). This is a different seam and a different mechanism, so I kept the word and documented the distinction at both ends (`capabilities.ts`, CONTEXT.md) rather than inventing a third term.

**A backend that claims `singleWriter` and supplies no `twoWriters` FAILS a conformance case saying so.** Considered: skipping the chapter silently (what every other optional affordance would do). Rejected because the cases are selected on the CLAIM, and skipping the only cases that can catch a fiction is how the report stops meaning anything. Touches any backend author outside this repo: the suite's third argument is now effectively mandatory for such a backend.

**`prune` claims/checks even when there is no floor to prune at.** A no-op prune still takes the claim, on both backends, so the rule is uniform ("every mutating path refuses a lost writer") and a conformance case can state it without knowing the store's retention. The alternative (check only when something will be deleted) makes an `unbounded` store's prune the one mutating path a lost writer may still call. Cost: on SQLite the check rides the tip read it already made, so no extra round trip; on IndexedDB it is inside the transaction already open. Touches hosts that prune on a timer: the pruner now claims the store.

**`drop` (SQLite) is guarded by a compare-and-swap on the claim's RELEASE, not by the DDL.** `DROP TABLE` takes no `WHERE`, so `DELETE FROM _writer WHERE token = ?` plus the read-back in one batch decides it, and the drops follow. Considered a "poison pill" statement that makes the batch error when the token differs (obscure and engine-dependent) and leaving `drop` unguarded (a hole in a promise about the storage). Consequence stated in the ADR: the drops are not in the same transaction as the check. An UNCLAIMED handle still takes over and may drop — necessary, because the server registry's `dropState` builds a fresh handle, and refusing would make disposing of a generation impossible.

**D1's prune arithmetic moved, in another package.** The guard is one more bound parameter on `dropVersionsStatement`, whose default sat exactly on D1's 100-parameter cap, so `prune` now names one fewer row id per statement; and a round costs three queries rather than two, so `QUERIES_PER_PRUNE_ROUND` went 2→3 and `d1PruneBudget` returns smaller budgets on both plans. Leaving the constant at 2 would have been a documented derivation that had quietly become false. Touches `platforms/cf-worker`.

**A refused writer stays refused; a writer whose FIRST write failed re-claims.** Both backends keep `token` plus a `claimed` flag set only when the claim's transaction/batch commits. Re-claiming after a loss would let two writers take the store in turns (the corruption at half speed); not re-claiming after a *failed* first write would leave a handle permanently refusing itself after an ordinary duplicate-height error. Touches the demotion task, which acts on the first half.

**`browser/multi-tab.spec.ts` had to change, and I made the minimal correction rather than the contention case.** That spec wrote from four tabs into one database concurrently, which the guard now forbids, so it would have gone red. It now counts a `StoreWriterChangedError` as an OUTCOME, asserts every tab still OPENED the shared database (the ADR-0024 claim it exists for), that `wrote + refused` accounts for every attempt, that nothing failed for any other reason, and that a fifth connection finds exactly the rows the tabs were told they wrote. The real contention case (same heights, one winner per height, torn-state audit, three engines, results recording, the no-guard experiment) is untouched and remains `work/tasks/backlog/two-tabs-contending-for-one-height-leave-one-winner.md`. Note for whoever picks that up: its "Blocked by" says `StoreWriterChangedError` is created by the demotion task, but it is created here.

**Browser result artifacts were left as they were.** I ran the Chromium project to verify (numbers above), but Firefox and WebKit binaries for the pinned Playwright version are not installed here, so refreshing only Chromium would have left `docs/spikes/indexeddb-row-backend-browser-default/results/` internally inconsistent for cross-engine comparison (timings and pass counts from different dates). I restored all three files; recording a three-engine run belongs to the follow-on task, which already owns that path.


wighawag:

The task this drive exists for: it closes the corruption hole while changing no caller. 45 files, +1848/-146, and all fourteen criteria are met. I checked the ones most likely to be faked hardest.

| Criterion | Evidence |
|---|---|
| All SEVEN mutating paths guarded; `migrate` not guarded and does not claim | `applyBlock`, `revertTo`, `writeCursor`, `clearCursor`, `prune` (+ `applyBlocks`, `drop` on SQLite) all route through `claimOrCheck`. `migrate()` is `await this.database()` and nothing else, so the several-tabs-all-open case is untouched. |
| `StoreWriterChangedError` declared ONCE, in `@etherfold/state-store` | `errors.ts:145`, beside the other seam errors, so `instanceof` holds across the package boundary |
| Refused call leaves the store byte-identical; `applyBlocks` per-batch semantics stated AND tested | ADR "Consequences" states it; `sqlite/test/two-writers.test.ts:105` asserts earlier batches applied and later ones not |
| The check is inside the same atomic unit as the write; `prune`'s tip read closed | The IndexedDB prune transaction is now `[VERSIONS, BLOCKS, WRITER]`, so the tip is read inside the transaction that deletes against it. This was the specific hole named in the task. |
| Token opaque | `writerToken()` is a UUID; the ADR rejects a monotonic revision precisely because a counter cannot distinguish a win from a loss on read-back |
| Guard on the TOKEN, never the cursor VALUE | Explicitly considered and rejected in the ADR: guarding the cursor would make the store parse what a cursor MEANS (ADR-0027), "and a reorg legitimately rewinds a cursor. The guard asks WHO is writing, never WHAT." |
| A store reports whether it enforces a single writer; memory + patch honest | `StateStoreCapabilities.singleWriter`; `MemoryStateStore` and `PatchStateStore` return `false` with the reasoning inline |
| Two DIFFERENT storage identities both write, neither refused | `addressedApart` in the conformance chapter, asserted as hard as the refusal half |
| Two generations addressed apart both write concurrently | Same chapter; both handles' writes land |
| An abandoned writer is taken over with no clear and no waiting | The claim has no lease and no expiry; the first mutation claims unconditionally. Asserted in conformance. |
| Single-writer behaviour unchanged, existing suites unmodified | See note below |
| Conformance parameterised by factory and gated on the reported claim | `if (!capabilities.singleWriter) return []` |
| A rival lands in the EXACT window; removing the guard turns it red; no `setTimeout` races | See note below |
| ADR records rationale, the ADR-0054 relationship, and BOTH rejected alternatives | ADR-0075 |

Three things I want to record, because each is a place this could have been faked and was not:

**1. "Existing suites unmodified" is honoured in substance.** Several existing test files DO appear in the diff, which is the sort of thing that should make a reviewer suspicious. Every one is the same mechanical edit: `toEqual` is an exact-match assertion, so widening the capability report with `singleWriter` forces every capability literal to gain the field. No test's expected BEHAVIOUR changed, and the new field's value is per-backend correct (`true` for IndexedDB, `false` for memory/patch). Nothing was loosened to make a suite pass.

**2. The no-timer criterion is honoured, and the one `setTimeout` in the diff is a comment saying why there isn't one.** The SQLite window test uses a `RivalBeforeSQL` handle that lands a rival's committed write precisely between the reading batch and the writing one, with the comment "a `setTimeout` race would pass or fail on scheduling, while this one puts the rival's committed write precisely between". The ADR also reports the mutation result concretely: removing the guard turns six of that file's eight cases red. On IndexedDB the stronger claim is made instead, that there is no window to land in.

**3. The suite is protected against its own vacuity.** `the-suite-catches.test.ts` adds a `LyingSingleWriterStore` that CLAIMS `singleWriter: true` over a `MemoryStateStore` and enforces nothing, then asserts the conformance suite FAILS it. And a backend that claims the capability but supplies no two-handle affordance fails a case saying exactly that, rather than silently skipping the only cases that could catch a fiction. That is ADR-0054's anti-vacuous discipline applied to this task's own test suite.

Consequences are disclosed rather than buried, which I checked individually: `drop` cannot put its DDL in the claim's transaction, so a rival claiming in the gap is dropped out from under, and the ADR says so; the D1 per-statement parameter cap moved by one, so `QUERIES_PER_PRUNE_ROUND` went 2 → 3 and a prune names one fewer row id per statement; the IndexedDB schema moved to version 3, `contains`-guarded so an existing database keeps every row.

Gate 1 green on the rebased tip. Gate 2 off by repo config for this drive. The repository is materially safer than it was this morning.

PR #125, opened 2026-09-10
feat(applyblock-refuses-a-height-below-the-recorded-tip): applyBlock re…

wighawag:
Task: `work/tasks/done/applyblock-refuses-a-height-below-the-recorded-tip.md`

Done. Everything is green (`format:check`, `check:adr`, `check:refs`, `build`, `typecheck`, `test` across all 19 packages/platforms/examples).

## Premise check (done before building, as the prompt asked)

I traced every path that can reach `applyBlock` and none applies at or below an existing tip:

- **`applyEventStream`** (the one production caller) computes the fork over the whole stream, `revertTo(fork)` first, then applies blocks that are all at or above `fork + 1`.
- **Fetch** (`generateStreamToAppend`): a delivered block below the store's tip requires the blocks above it to have been contradicted, which emits retractions, which lowers the tip below it via the revert. The one theoretical exception (a node that under-delivered a range earlier) is now a loud refusal instead of quiet corruption; I captured that in `work/notes/observations/a-fetch-can-in-principle-offer-a-block-below-the-tip.md` rather than weakening the check.
- **Replay / rebuild** (`generateStreamFromReplay`, `GenerationRebuild.more`): the window is rebuilt by walking the stream, which de-duplicates what the fold already applied; a follower folds into its own namespace.
- **Bootstrap** (`bootstrapFromSnapshot` → `SnapshotAwareStateStore.bootstrap`): installs one block at `takenAt`, and only when the local position is strictly behind it, so it is either an empty store or a store whose tip is below.

Empirically confirmed too: the whole suite (core 1064, server 286, cli 198, browser 160, stratagems workload) is green with the tightening in.

## What I built

Test-first: the three conformance cases went in and went red before any production code.

- **Seam** (`@etherfold/state-store`): new `blockNotAboveTip(number, tip)` in `blocks.ts` (one message, four backends), the rule stated on `StateStore.applyBlock`, and the check in `MemoryStateStore`.
- **`@etherfold/state-store-patch`**: same check against its `tip`.
- **`@etherfold/state-store-indexeddb`**: the tip is read with one backwards cursor over `blocks` **inside the same `readwrite` transaction** as the write; a refusal aborts, so the store is byte-identical. Pinned by a transaction-recording test (no extra `readonly` transaction is opened).
- **`@etherfold/state-store-sqlite`**: no read-inside-a-transaction exists, so it is ADR-0054's shape: `aboveTipGuard` (`NOT EXISTS (SELECT 1 FROM _blocks WHERE number > ?)`) rides **every** statement of the block (versions and cursor included, so a refused block moves nothing), and a `latestBlockStatement` opening the same batch supplies the number the message names. `WriterGuard` generalised to a composable `StatementGuard` so one predicate carries both preconditions.
- **Conformance**: three cases in `a block is one atomic unit` (refused below the tip even where the height is free; empty store admits any height; a revert makes a height applicable again), plus an `AccommodatingStore` in `the-suite-catches.test.ts` that rewinds instead of refusing, so the new case is proven to bite.
- Docs (three READMEs), the affected browser-harness prose, and a changeset.

Existing duplicate-height and duplicate-hash refusals are untouched (on SQLite still the primary-key violation, deliberately: `>` not `>=`).

## Decisions

- **The refusal is a plain `Error` carrying a shared message, not a new error type.** It means the same thing as the duplicate-height refusal beside it (the CALLER is wrong: revert first, or stop writing), and `StoreWriterChangedError` stays the only typed one on this path because it means the opposite (a lost race whose correct response is demoting to a reader) — spec user story 3 rests on that distinction, and a second type would blur it. Alternative considered: a `BlockNotAboveTipError`, rejected because nothing branches on it. It touches `a-refused-writer-demotes-itself-to-a-reader` (backlog), which must keep telling the two apart by TYPE — it still can.
- **The message lives at the seam (`blockNotAboveTip`) rather than being authored per backend.** The acceptance criterion allows backends to differ, and SQLite turned out to be able to name both heights (the tip read rides its batch), so all four say the same thing from one place. The existing duplicate-height message is copied in three backends and is exactly the drift this avoids. It touches any future backend: it inherits the message by calling the helper.
- **`applyBlocks` (SQLite-only, no production caller) now requires its blocks to ASCEND, refused locally before any I/O.** Each block is judged against the tip its predecessors left, so a descending pair would silently apply to nothing. This is a NEW REFUSAL on a public method of that backend: it mirrors the engine's existing `assertAscendingByBlock` rather than inventing a rule, and the alternative (silently dropping, or one read-back per block) was worse in both directions. It touches `the-seam-splits-into-a-readable-and-a-writable-store` and `the-seam-narrows-and-a-reader-cannot-write` (backlog), which keep `applyBlocks` on the SQL tier's own class.
- **`applyBlocks` sends its LOWEST block in a batch of its own, carrying the tip read.** That makes the refusal whole (nothing applied) rather than "the rest of the first batch landed", at the cost of one extra round trip per CALL, not per block. Alternative considered: reading the tip in the first packed batch (cheaper, but a refusal would leave the higher blocks of that batch applied) and one read-back per batch (needs a batch→block mapping `planBatches` deliberately does not carry). It touches `platforms/cf-worker`'s D1 limits test, which now asserts two batches with the same packing bound.
- **`packages/state-store-indexeddb/test/persistence.test.ts` ("interleaves writes from both") was rewritten to interleave in ASCENDING order**, and the multi-tab browser harness now counts a tip refusal as an outcome beside `StoreWriterChangedError`. Those tests encoded "each tab owns its own heights", which the source spec explicitly says was never evidence that two indexers can share a database; a store's blocks are one sequence. No coverage was dropped: the property they exist for (neither handle caches state the other contradicts; four tabs can OPEN one database) is asserted unchanged.
- **`.wrangler/` added to `.gitignore`.** Running the cf-worker tests writes `platforms/cf-worker/.wrangler/tmp`, which is untracked, unignored, and would otherwise be swept into the commit by `git add -A`.


wighawag:

Context worth recording: a FIRST attempt at this task ended in an empty diff with no reason, and the runner surfaced a dispose-defaulted question. I disagreed with that judgement on evidence — `applyBlock` on main checked the writer token, a duplicate height and a duplicate hash, and never compared against the tip — and requeued rather than cancelling. This branch vindicates that: the tightening was genuinely absent and is now present.

| Criterion | Evidence |
|---|---|
| Refuses a height not strictly above the recorded tip | IndexedDB: backwards cursor on BLOCKS inside the writing transaction. SQLite: `aboveTipGuard` = `NOT EXISTS (SELECT 1 FROM blocks WHERE number > ?)` on every statement, which combined with the primary key on `number` is exactly "strictly above". |
| An EMPTY store admits any height | `tip === undefined` on IndexedDB; the `NOT EXISTS` predicate is trivially true on an empty table. Asserted in conformance. |
| The tip is read inside the same transaction as the write | IndexedDB reads it in the same `readwrite` transaction, so a rival's revert cannot lower it between read and write. SQLite prepends `latestBlockStatement` to the same guarded batch, ADR-0054's shape unchanged. |
| The refusal names both heights where the substrate can compose it | Shared `blockNotAboveTip(number, tip)` in `@etherfold/state-store`; on SQLite the verdict is a conditional write plus a read-back, with the message assembled after the batch. No uniform message was forced on a substrate that cannot produce one. |
| Existing duplicate-height and duplicate-hash refusals still hold | The tip check is ordered AFTER both, deliberately, so re-applying a block keeps the message that names it as a caller bug |
| Existing suites green, including replay, rebuild and bootstrap | Full gate green on the rebased tip; those paths install into a store with no tip, which is the admits-any-height case |
| Conformance carries the case on every backend | `cases/block-atomicity.ts`, covering the refusal, the empty store, and re-admission after a revert |
| Changeset | Present, and it documents the new `applyBlocks` ASCEND requirement |

**The premise check was actually performed, which is the thing this task most needed.** The task asked the builder to confirm for itself that no caller legitimately applies at or below the tip, and to STOP rather than weaken the check if one existed. The builder found the one shape that can in principle produce it — `generateStreamToAppend` selects new blocks by MEMBERSHIP of the retained window rather than by a height threshold (ADR-0051), so a node that under-delivered a range earlier could return those logs on a later re-fetch — reasoned that every reorg shape reaches `applyBlock` after a `revertTo` that lowered the tip (which is why the suite stays green), and judged that refusing beats the previous behaviour, which was applying a version whose `_lower` sat under the live one. That is the correct call: the old behaviour was quiet corruption. It captured the residue as `work/notes/observations/a-fetch-can-in-principle-offer-a-block-below-the-tip.md` rather than expanding scope, and left the open question honestly stated — whether a wedged fold is the response we want for an under-delivering node, or whether that deserves its own diagnosis.

Two consequences I checked rather than accepted:

- **`applyBlocks` now requires ASCENDING blocks**, which is a new caller-visible refusal. It is justified (each block is judged against the tip the one before it left), refused before anything is sent, and documented in the changeset. The lowest block rides its own batch carrying the tip read, so a refusal leaves NOTHING applied at the cost of one extra round trip per call rather than per block.
- **The suite is again protected from vacuity.** `the-suite-catches.test.ts` adds an `AccommodatingStore` that rewinds to make room for a block the tip has passed, and asserts the suite fails it — while ALSO asserting the empty-store and post-revert cases do NOT fail, so the chapter cannot pass by simply refusing everything.

The `.gitignore` addition (`.wrangler/`, vitest-pool-workers scratch) is unrelated housekeeping found in passing, and is a runtime artifact rather than source. Benign.

Gate 1 green. Gate 2 off by repo config for this drive. With PR #124, the corruption hole named by the spec is now closed from both directions.

PR #126, opened 2026-09-10
feat(a-store-reports-whether-its-retention-is-enforced): A store report…

wighawag:
Task: `work/tasks/done/a-store-reports-whether-its-retention-is-enforced.md`

All acceptance criteria are met and the full gate is green (`format:check`, `check:adr`, `check:refs`, `build`, `typecheck`, `test`; `changeset status --since=main` verified against a scratch index, since it only sees tracked files and the runner commits before verifying).

## What I built

**The read** — `StateStore.readRetentionEnforcement()`, a twelfth verb at the seam (`packages/state-store/src/store.ts`), answering one of three things from `packages/state-store/src/enforcement.ts`:

- `{kind: 'no-floor'}` — `unbounded`, or `revert-only` with no declared depth: nothing to enforce.
- `{kind: 'never-pruned', floor}` — a floor, and no pass has ever run. The bespoke host.
- `{kind: 'pruned', floor, prunedTo}` — a floor, and a pass ran at `prunedTo`.

`floor` (as it stands now) and `prunedTo` (the last pass) are reported apart, because the gap is the diagnostic. Whether a store has a floor is computed from the SETTING (`retentionFloor` asked at block 0), so a configured store that has applied no block yet reports `never-pruned` rather than hiding as `no-floor`, and `revert-only` **with** a depth is reported like any other floored store.

**The durability** — the record rides the cursor port under `RETENTION_ENFORCEMENT_KEY`, exactly as `SNAPSHOT_ORIGIN_KEY` does: durable, unversioned, never reverted, never pruned. One more key rather than a new table on four backends. On IndexedDB it commits in the *same* transaction as the deletion it describes (the prune transaction gained `CURSORS`; the pinned transaction assertion in `two-writers.test.ts` was updated with the reason). On SQL and the two in-memory backends it is written after the deletion, so a crash under-claims enforcement rather than over-claiming it.

**Every backend implements it**: memory, IndexedDB, SQLite, patch, plus pass-through on `SnapshotAwareStateStore` (delegated whole and deliberately *not* narrowed the way `capabilities` is — the snapshot floor drops nothing, so `no-floor` under a reported window is true of it).

**`capabilities` is untouched** — same synchronous getter, same type, same pre-open readability. Only its module docstring gained a pointer.

**Conformance** — a new chapter (`cases/retention-enforcement.ts`) asked of every backend. It cross-checks rather than expecting a fixed answer, because only the store knows whether it has a floor: a pass with a floor must leave the store reporting `pruned` **at that floor**, a pass with none must leave `no-floor`. Where a backend can open two handles on one storage it also asserts the second reads back the first's pass (the suite's closest thing to a restart) and that reporting never claims the store. Two new lying backends in `the-suite-catches.test.ts` prove the chapter has teeth.

**Also**: ADR-0076, the `workerd` finding note (`work/notes/findings/a-worker-cannot-hold-a-timer-across-requests.md`, quoting the two Cloudflare pages retrieved today), a changeset, and glossary/README updates in `CONTEXT.md`, `packages/state-store/README.md` and `packages/state-store-conformance/README.md`.

Tests were written before the production code at the seam (`packages/state-store/test/retention-enforcement.test.ts`), then per-backend reload tests and the conformance chapter.

## Decisions

- **The new concept is named `retention enforcement`, and the verb is `readRetentionEnforcement`.** Checked against the glossary: `retention` keeps its exact meaning (a distance in BLOCK NUMBERS), `prune` keeps its exact meaning (the call that drops), and "enforce" was already this repo's word for exactly this pair — `capabilities.ts` already said "enforcing it has two halves". It does not overlap `capabilities` (a synchronous CLAIM) or `PruneReport` (what ONE pass did). The verb is `read*` to match `readCursor`, since it is a read and not a getter. Alternative considered and rejected: a field on `capabilities`, which the task's own analysis rules out and ADR-0076 records. Touches: the seam interface, so every `StateStore` implementation including the two test-side decorators.

- **The three kinds are `no-floor` / `never-pruned` / `pruned`, with `no-floor` first-class rather than a degenerate "never pruned".** `unbounded` is the default and probably most deployments; reporting them as "never pruned" would drown the one report that matters. Alternative: a two-arm boolean-ish shape with a nullable floor, rejected because it makes the healthy default indistinguishable from the defect. Touches: anyone who later surfaces this on a status endpoint.

- **A pass is recorded even when it deleted nothing.** The evidence is that a pass RAN at a floor, not that rows went, because a host pruning on a schedule deletes nothing on most cycles. Alternative: record only on `versionsDeleted > 0`, rejected because it makes the healthy case the alarm and makes a quiet chain indistinguishable from a broken host. Cost, accepted and documented: one small cursor-port write per prune call that has a floor (in the same IndexedDB transaction; one extra guarded batch on SQL). Touches: the browser and CLI prune schedules, which now write once per pruning cycle.

- **An unreadable record reads as `never-pruned` rather than throwing** — deliberately the opposite of the snapshot origin beside it, which throws. That marker is a SAFETY floor whose absence has a store claim history it never received; this one is a DIAGNOSTIC, so under-claiming asks a human to look at a healthy store while a throw would take down a store whose data is in no way suspect. This introduces no new error path anywhere. Touches: nothing outside `enforcement.ts`; recorded because "no refusal" was a choice, not an oversight.

- **Nothing SURFACES the report: no `/status` field, no `syncing` field, no host wiring.** The task's criteria ask for a read and a conformance obligation, and `the-browser-indexing-loop-schedules-its-prune` deferred publishing an enforcement state to this task's concept precisely so it would not be forked — publishing it now would mean choosing an audience, a shape and a cadence for two different host surfaces, which is a surface decision with its own scope. `EntityStateView` / `VersionedStateView` (the processor-side read handles) likewise do not expose it. Touches: any future task that wants operators to SEE this; the seam read is what it would build on.

- **`packages/state-store-conformance/README.md` gained a bullet for the pruning chapter as well as the enforcement one.** That chapter existed and was undocumented before this task (a pre-existing gap); documenting it in the same bullet as the new one was cheaper and more coherent than describing half a subject. Touches: docs only.

- **`@etherfold/browser` is listed in the changeset as `patch`.** Its only change is a test decorator implementing the new seam verb, so nothing it ships changed — but the acceptance gate's `changeset status --since=main` fails a changed package with no entry, and that is the exact bounce `the-cli-schedules-the-prune-its-retention-implies` hit. The changeset prose says plainly why it is there. Alternative: leave it out and take the red gate at land time.


wighawag:

| Criterion | Evidence |
|---|---|
| An ASYNC read distinguishing the three states | `RetentionEnforcement` = `{kind: 'no-floor'}` \| `{kind: 'never-pruned', floor}` \| `{kind: 'pruned', floor, prunedTo}`, via `readRetentionEnforcement()` on every backend |
| Durable across a reload or restart | The record is written into the durable cursor slot under `RETENTION_ENFORCEMENT_KEY`, and crucially INSIDE the prune's own transaction, so a store cannot report a pass it did not complete, nor forget one it did |
| The synchronous `capabilities` getter is UNCHANGED | Verified against the diff: `capabilities.ts` gained comments only, zero executable change. Every existing consumer keeps compiling and keeps its pre-open readability, which was the explicit constraint. |
| Conformance asks every backend the same question | `cases/retention-enforcement.ts`, wired into the suite |
| No default and no configuration shape changed | Confirmed across the diff |
| An ADR records why a report replaced the refusal, naming stories 2 and 3 | ADR-0076, which states in terms: "**Stories 2 and 3 were answered differently rather than delivered**", and explains why the record must outlive a task body that moves to `done/` |
| The `workerd` constraint recorded as a FINDING | `work/notes/findings/a-worker-cannot-hold-a-timer-across-requests.md`, with a proper `source:` naming both Cloudflare pages, their URLs and their retrieval date — which is what makes a finding correctable later |
| Changeset | Present |

The task's central instruction was a prohibition — do NOT put this on the `capabilities` getter, because it is synchronous and documented as readable before `migrate` and before the database is open, so a durable value cannot honestly be produced there. The builder respected it exactly: a separate async read, additive, and the sync getter untouched. The reasoning is recorded where the next person will meet it.

The anti-vacuous discipline holds for the third PR running: `the-suite-catches.test.ts` adds a `SilentPrunerStore` (prunes and reports that it never has) and an `EagerEnforcementStore` (no floor, claims a pass at one), and asserts the suite fails BOTH — so the chapter catches lying in either direction rather than only the convenient one.

**One nit, recorded and not blocking.** The enforcement record lives in the shared cursor namespace under a reserved key, and the protection against a caller choosing that key is documentation ("A caller choosing its own key should avoid those two") rather than a refusal in `writeCursor`. I checked whether this task invented that hazard: it did not. `SNAPSHOT_ORIGIN_KEY` already occupies the same namespace on the same advisory terms, and this change documents the convention where a caller would actually look, which is an improvement on what it found. The blast radius of a collision is a wrong DIAGNOSTIC, not corrupted data. Making reserved keys a refusal rather than a convention is worth considering as a follow-up across both keys at once; it is out of scope here and no acceptance criterion asks for it. I am carrying it into this drive's cross-drive triage rather than blocking on it.

PR #128, opened 2026-09-10
feat(a-refused-writer-demotes-itself-to-a-reader): A refused writer dem…

wighawag:
Task: `work/tasks/done/a-refused-writer-demotes-itself-to-a-reader.md`

**Done.** A writer whose mutation is refused now demotes itself to a reader instead of raising at the application, and the tab that lost keeps showing correct data.

## What I built

**`packages/browser/src/demotion.ts` (new)** — the single exported `demoteToReader(writer, reason)`: it stops the fold, drops the in-memory cursor, narrows every store the writer was folding into with `openForReading`, warns through `named-logs`, and returns a `Demotion {reason, reading}`. `DemotionReason` is `'write-refused' | 'lease-lost'` — one code path for both ways of losing. The port (`DemotableWriter`: `stopFolding` / `forgetCursor` / `stores`) exists so the ORDER is the function's guarantee, not each caller's to remember. `isStoreWriterChanged` is module-internal (not published).

**`packages/browser/src/IndexerState.ts`** — one refusal handler (`whileWriting`) that every driver goes through, because the refusal can surface from any step that writes (`load()` re-folds; an empty cycle still writes the cursor). On a refusal it demotes and answers `undefined`; on anything else it throws exactly as before. The auto-index loop is not re-armed, `startAutoIndexing()` answers `false`, `setupIndexing` no longer publishes `error: FAILED_TO_LOAD` for it, a refused prune is re-raised rather than logged, `dispose()` clears the demotion, and the hook exposes `demoteToReader(reason)` plus `syncing.demotion`.

**Tests** — `packages/browser/test/demotion.test.ts` (REAL contention: a second `createBrowserStateStore` on the same `databaseName`, no timers in the guard path): reports instead of raising; drops the cursor; issues no mutation on any guarded path however driven (counted through a recording proxy, not inferred); stops fetching; keeps answering reads (`EXPECTED_A`); is not a follower; stops the auto-index loop; the lease-loss caller reuses the same function; the exported function's order and its read handles. `demotionWarns.test.ts` asserts the `named-logs` warning (own file: the factory has to be hooked before the module builds its logger). `packages/state-store/test/writable-seam.test.ts` asserts `StoreWriterChangedError` stays outside the `BlockUnavailableError` family. The "raised on each guarded path" half was already covered by `@etherfold/state-store-conformance`'s `a second writer writes nothing` chapter plus each backend's `two-writers.test.ts`.

**Also**: ADR-0078, a `demotion` glossary entry in `CONTEXT.md`, browser README + guide sections, and the changeset (`@etherfold/browser` minor, `@etherfold/state-store` patch).

Gate: `format:check`, `check:adr`, `check:refs`, `build`, `typecheck`, `test` all green (`changeset status --since=main` only sees committed changesets, so it will pass once the runner commits; the two changed packages are both named).

I also captured one out-of-scope signal in `work/notes/observations/an-auto-index-loop-retries-a-tip-refusal-for-ever.md` (a tab whose store was moved far ahead *before* it ever wrote gets the tip refusal, not the writer refusal, and the loop retries it at ~90 `eth_getLogs`/s for ever).

## Decisions

- **The demotion lives in `@etherfold/browser`, not in `@etherfold/core` beside `pause`.** A demoted writer must HOLD ITS STORE AS A READER, and the core has no store: on both containers the state is an opaque type parameter from a caller's `createState`, and `@etherfold/core` does not depend on `@etherfold/state-store` (ADR-0016's direction). The browser hook is the only place with all three halves (the stores it built, the loop that fetches, the surface an app subscribes to) and is where the lease-loss caller will live. Alternatives: core with the store half re-implemented per runtime (two demotions with one name), or the seam (which knows nothing about generations or fetching). **Touches**: the receiving container (server/CLI, `receivingContainer.ts`) gets NO demotion — deliberate, since a second writer there is an operator misconfiguration rather than a tab that must keep rendering, and the honest server behaviour is the refusal reaching the host that scheduled the call. Recorded as ADR-0078.
- **`indexMore` / `indexMoreAndCatchupIfNeeded` / `indexToLatest` now answer `Promise<LastSync | undefined>`; `undefined` means DEMOTED and nothing else.** A user-visible, breaking signature change. Alternatives: throwing (rejected — both browser loops swallow exceptions and retry on a timer, so the refusal would be retried for ever) and returning the last cursor (rejected — that is the very value the demotion drops, and `checkTxInclusion` answers from it). **Touches** every app driving the loop and the browser test fixture `indexToTip`, which now raises naming the demotion. Nothing is published, so it costs a changeset.
- **`syncing.demotion` is reported for the INDEXER, not per generation.** A claim is scoped to one unit of STORAGE (ADR-0075) and the shipped pattern hands one store to every generation (ADR-0077), so a per-generation `demoted` flag would report a state no backend can produce. **Touches** the acceptance wording "a reported state on the generation" — read as the reported state of the thing that lost the claim.
- **Named `demoteToReader`, never `demote`.** `promote` already means moving the canonical pointer BETWEEN generations, and a bare `demote()` beside it on the same object would read as its inverse. The `...ToReader` suffix names the axis; the JSDoc, the glossary entry, the README and the ADR all state the distinction explicitly. The word itself is not new: ADR-0075/0077 and the source spec already say "demoted writer".
- **`'lease-lost'` is in the reason union although nothing in the tree produces it yet.** The lease spec (`one-tab-indexes-and-the-others-read`) needs a reason to pass and would otherwise invent one; it is asserted by a test that calls the function that way, not deferred to a future spec. **Touches** that spec only.
- **`startAutoIndexing()` on a demoted tab returns `false` and starts nothing** rather than throwing a new error. `false` is the answer it already gives for "a loop is already running", so no new refusal type was introduced on a user-visible path.
- **A refused `prune` inside the scheduled pass is re-raised rather than logged** (that catch otherwise swallows everything), so the one refusal handler demotes. A swallowed refusal would leave a tab that had learnt it lost and carried on indexing until its next write said so again.
- **The refusal is recognised by `instanceof` OR `error.name === 'StoreWriterChangedError'`.** Same reasoning as `isOutOfSpace` in `@etherfold/core`: a bundled app can hold two copies of `@etherfold/state-store`, and an unrecognised refusal is retried for ever.


wighawag:

| Criterion | Evidence |
|---|---|
| A demoted generation folds NOTHING and holds a store opened for reading | `demotion.test.ts:256` — hands back the stores as READ handles, and stops folding BEFORE it does, which is the right order |
| NOT a follower, and for the stated reason | `:210` "is NOT a follower: a follower re-folds and would be refused on every block". The trap the task warned about is tested, not just avoided. |
| No further mutation on any guarded path, asserted | `:134`, and pointedly "however it is driven" — so it holds for the manual path and the auto loop alike |
| Demotion is OBSERVABLE: reported state plus a `named-logs` warning | `demotionWarns.test.ts` asserts the warning names the reason (`write-refused`), the new state (`ANSWERING READS`) and that there is `nothing to retry` |
| NOT a member of the `BlockUnavailableError` family | `StoreWriterChangedError extends Error`; `writable-seam.test.ts` adds "is not a read this store cannot answer" |
| Drops the in-memory cursor, stops fetching, answers reads, does not propagate outward | `:120` (the cursor "is now a lie"), `:153` (stops FETCHING), `:166` (goes on showing correct data), `:101` (reports rather than raising at the application) |
| Does not resume writing on its own | One-way by construction: nothing re-claims, consistent with ADR-0075's rule that a backend never re-mints a claim it has committed. Becoming a writer again means building a new store, which forces the re-read. |
| A SINGLE exported function, used by the refusal handler AND callable by a caller | `demoteToReader` in `packages/browser/src/demotion.ts`, and `:231` proves reuse by demoting on a LEASE LOSS "through the one function the refusal handler uses" — which is what `one-tab-indexes-and-the-others-read` will need |
| Tests cover the refusal and a generation demoting rather than throwing | Both, plus `:176` covering the auto-index loop, "which is the path an app actually drives" |
| Changeset | Present |

The reuse criterion deserves a note because it is the one most easily satisfied on paper. The task asked for a single function callable from a caller-supplied hook so a later lease-loss caller reuses it, and warned against asserting that a future spec will like it. The test does the right thing: it actually drives a lease-loss demotion through the same entry point, so the reuse is demonstrated rather than promised.

**A real off-path finding was captured and correctly NOT fixed here.** `work/notes/observations/an-auto-index-loop-retries-a-tip-refusal-for-ever.md` records that a tab whose store was moved far ahead by another writer BEFORE it ever wrote gets `blockNotAboveTip` (a plain `Error`) rather than `StoreWriterChangedError` — correctly, since a store that never claimed claims on its first write, so the writer guard passes and the tip check refuses. `_auto_index` treats that as transient and re-fetches for ever: measured at ~90 `eth_getLogs` calls per second of wall clock, cursor pinned at 0, nothing published on `syncing.error`. The builder scoped it out with the right reasoning — that refusal says the CALLER is wrong rather than "you lost a race", which is exactly the distinction PR #125 established — and pointed at the existing `RetryableError` / `err.retryable === false` vocabulary that this loop reads nothing of. That is golden-rule-3 behaviour: capture, don't fix in place. I am carrying it into this drive's triage as task-worthy.

PR #129, opened 2026-09-10
feat(two-tabs-contending-for-one-height-leave-one-winner): Two tabs con…

wighawag:
Task: `work/tasks/done/two-tabs-contending-for-one-height-leave-one-winner.md`

## What I did

**The contention case now exists** as a second test in `packages/state-store-indexeddb/browser/multi-tab.spec.ts` (the harness the task points at), with three new cases in `browser/cut.ts` (`contention-claim`, `contention`, `contention-audit`). Four tabs of one app, one `databaseName` (the storage identity the token is scoped to), all offering **every height in one range at once**.

What it asserts:
- exactly one write lands per height (`won` across all tabs, sorted, equals the contended range);
- every loser is refused **by name** (`refusedBy` is `{StoreWriterChangedError: n}` and nothing else; any other failure lands in `unexpected`, which must be empty);
- an independent connection afterwards (every tab has closed its handle; `migrate` never claims) finds no half-applied block (a block record at every height), no row from a refused writer (the row at each height names the tab that was *told* it wrote it), and no cursor behind its data (`lastSync` is the last landed height and its winner).

**Green on three engines.** `pnpm --filter @etherfold/state-store-indexeddb test:browser`: 18/18 on Chromium, Firefox and WebKit. Results are written to `docs/spikes/indexeddb-row-backend-browser-default/results/contention-<engine>.json`. Which tab wins is engine-decided and observably differs (chromium tab 3, firefox tab 0, webkit tab 0).

**The results record says what only this run covers.** The spike README gained a `## Contention counts only in this run` section: `fake-indexeddb` cannot demonstrate cross-tab serialisation at all, so the node suite asserts the RULE and is not evidence about tabs; and the case is deliberately not in the acceptance gate for the same reason as the rest of the browser run (three browser binaries a clean checkout lacks). The node `test/two-writers.test.ts` header now says the same thing from its own side and points here.

**The no-guard experiment was actually run, not claimed.** `claimOrCheck` was temporarily made a no-op (the pre-ADR-0075 state), the case turned red on Chromium, and `store.ts` was restored byte-for-byte from a backup (it is not in the diff). The observation, with the quoted output, is recorded under `### Removing the guard turns it red (observed, 2026-09-10, Chromium)`: three of four tabs are refused twelve times each, **not one refusal carries the name `StoreWriterChangedError`**, and the store instead tells them "applying the same block twice is a caller bug", which is the opposite diagnosis. The README is honest that rows stayed coherent in that run and why (`applyBlock`'s duplicate-height check was always a CAS), and that the four paths the token was really built for are the conformance suite's to assert.

**Gate:** `format:check`, `check:adr`, `check:refs`, `build`, `typecheck`, `test` (706 tests in this package, whole-repo suite green). `changeset status --since=main` is red only locally because the changeset file is still untracked; it reads `git diff` against `main`, so it goes green once the runner commits. Changeset added at `.changeset/two-tabs-contending-for-one-height-leave-one-winner.md`.

Two side effects worth naming: the other `results/*.json` were **stale from before ADR-0075** (they still showed four tabs writing 20 blocks each) and my full browser run refreshed them, which is the same command's own output rather than a separate edit; and I ran `playwright install firefox webkit` (the repo's own documented step) because the installed binaries were older than this Playwright expects. `test-results/` scratch was deleted; the tree holds only intended changes.

## Decisions

- **The tabs CLAIM at a barrier before the height race, in a case of their own (`contention-claim`), and the claim is a `writeCursor` under a per-tab key.** Why: a handle that has never written claims UNCONDITIONALLY on its first mutation (`store.ts`), so tabs racing straight from unclaimed handles are refused for offering a **height already recorded** — the duplicate-height caller bug, which means the opposite thing — and the acceptance criterion "every loser is refused with `StoreWriterChangedError` specifically" would be unreachable by construction. `writeCursor` is the shortest mutation that claims and can be refused for no other reason (no block precondition), so every tab's claim commits and the last committer holds the store. Alternatives considered: (a) race unclaimed and accept a mix of refusal kinds — rejected, it fails the criterion and re-tests what the four-tab case already counts; (b) let tabs claim mid-race so the store changes hands, with the new holder re-reading the tip and continuing — rejected because the handover would have to be scheduled by a timer, and this spec family is explicit that the cases must not depend on timing (spec story 13, and `two-writers.test.ts`: "nothing below uses a timer"). **Consequence to be aware of**: the winner is therefore the last claimer, so one tab lands every height and the other three are refused every height. The contention is real (all four offer each height concurrently and the engine serialises them) but the winner does not change hands mid-run. **Touches**: nothing outside this spec file and `cut.ts`; no production code, no flag, no default.
- **Refusals are counted BY ERROR NAME (`refusedBy`) rather than as one `refused` total**, unlike the neighbouring four-tab case which folds two legitimate refusal kinds into one number. This case is the one that asserts the name, so a refusal wearing another name must be visible rather than absorbed; anything not `StoreWriterChangedError` goes to `unexpected` and fails the test. Recognition is `instanceof` with `error.name` as the fallback, matching how `@etherfold/browser` recognises the same refusal (a bundled app can hold two copies of `@etherfold/state-store`). **Touches**: the results schema for `contention-<engine>.json` only; `multi-tab-<engine>.json` keeps its existing shape.
- **The test parses the cursor value it wrote (`<height>:<tab>`).** ADR-0027 keeps the STORE ignorant of what a cursor string means; the test is the caller that wrote it, so reading it back is the caller's business and is what makes "no cursor behind its data" checkable. Stated at the assertion so nobody reads it as the store gaining an opinion about cursor contents. **Touches**: nothing else.
- **A module-level handle (`contender`) in `cut.ts` carries the claim between the two `run` calls.** A claim belongs to a HANDLE and is never re-minted, so the tab must race through the object it claimed with; module state survives because both runs are `page.evaluate` into one loaded page, and a reload correctly clears it. **Touches**: `cut.ts` only, and only the contention cases — no other case keeps state between runs.


wighawag:
APPROVE ✅ — Gate 3 (conductor's diff-vs-criteria review, including an independent run)

This case is deliberately OUTSIDE the acceptance gate, so the gate's green says nothing about it. I ran it myself, in a throwaway clone of this branch, rather than taking the committed results on trust.

**Independent run — 18/18 passed on three engines (34.1s):**

```
✓  2 [chromium] › several tabs contending for the same heights leave one winner and no torn state (602ms)
✓  8 [firefox]  › several tabs contending for the same heights leave one winner and no torn state (2.5s)
✓ 14 [webkit]   › several tabs contending for the same heights leave one winner and no torn state (2.3s)
```

**And I reproduced the guard-removal experiment**, because a recorded manual result is exactly the kind of claim that rots. Making `claimOrCheck` return without claiming or checking (the pre-ADR-0075 state of this backend) and re-running turns the case RED with precisely the recorded signature: `"refusedBy": {}`, `"won": []` for the losing tabs, and `unexpected` filled with "block N is already recorded: applying the same block twice is a caller bug". Not one refusal carries `StoreWriterChangedError`. The README's account is accurate.

| Criterion | Evidence |
|---|---|
| Several tabs contend for the SAME heights; exactly one write lands per height | The winning set is asserted to equal the full height list, and every tab's `won + refused` equals what it attempted, so nothing is unaccounted for |
| Every loser refused with `StoreWriterChangedError` SPECIFICALLY | `refusedBy` is asserted as `{StoreWriterChangedError: <count>}` — refusals are counted BY NAME, which is what makes the guard-removal experiment discriminating |
| An independent connection afterwards finds a coherent store | A fifth connection that took no part in the race audits every height: recorded, owner matching the actual winner, no row from a refused writer |
| Runs on Chromium, Firefox and WebKit; output to the spike results directory | Verified by my own run; `contention-{chromium,firefox,webkit}.json` committed |
| NOT added to the acceptance gate, and that is stated where the results are kept | README: it needs `playwright install` and three browser binaries, and "a gate that cannot run on a clean CI checkout is a gate that gets skipped" |
| The recorded result includes the guard-removal observation, recorded rather than claimed | README §"Removing the guard turns it red (observed, 2026-09-10, Chromium)", with the actual failure output. Reproduced by me. |
| Changeset | Present |

**The most valuable thing in this PR is a piece of honesty that a weaker version would have omitted.** The README records that in the UNGUARDED run the row-level audit stayed coherent anyway, because one tab happened to win every height by speed and `applyBlock`'s duplicate-height check was already a compare-and-swap. So this case's evidentiary value is the DIAGNOSIS the losers receive — "you lost a race and wrote nothing" versus "applying the same block twice is a caller bug" — and not the coherence. It then says plainly that the four paths the token was actually built for (`writeCursor`, `clearCursor`, `revertTo`, `prune`) CANNOT be observed from this case and are asserted by the conformance suite instead. That stops a future reader over-claiming this test as proof the whole guard works.

It is equally careful about the boundary the task warned of: `fake-indexeddb` cannot demonstrate cross-tab serialisation at all, so the node suite asserts the RULE and this run is the evidence that two TABS behave, with "Do not read the green node suite as covering it" stated where a reader will meet it.

Gate 1 green. Gate 2 off by repo config for this drive. Contention verified on three real engines by me, not inferred.

PR #130, opened 2026-09-10
feat(the-seam-narrows-and-a-reader-cannot-write): The seam narrows so a…

wighawag:
Task: `work/tasks/done/the-seam-narrows-and-a-reader-cannot-write.md`

The seam is narrowed and every suite in the repository is green.

## What I did

**Measured first.** Grepping `StateStore` across `packages/*/{src,test,browser}`, `platforms/*/src` and `examples/*/src` found ~150 mentions, most of them READS that need no change. The mutating set was wider than the task's list in two ways it did not predict: `@etherfold/processor-sqlite` is affected (it hands a concrete store to `EntityEventProcessor`, whose constructor is the atomic trigger), and so are both `examples/*/browser/main.ts` (the shipped `createState` shape). The three named exceptions were all real and all handled.

**Test first, at the seam.** `packages/state-store/test/writable-seam.test.ts` now types its refusals against `StateStore` itself; before the production change the five `@ts-expect-error` lines went red as unused directives (`error TS2578` ×5) plus a missing export. That file is what `pnpm typecheck` evaluates, which is the repo's only way to assert a compile error. I added two lines to it: a backend is not a claim, and a reader cannot re-open its own handle for writing.

**The shape.** `StateStore` is now the reads (the readable interface folded in). `StateStoreBackend extends StateStore` carries the five mutating verbs. `WritableStateStore extends StateStoreBackend` adds the `token`. `ReadableStateStore` and `StateStoreMutations` are deleted. Backends declare `implements StateStoreBackend`; their surfaces are untouched, `createD1Store` still returns the concrete class, and `platforms/` needed no change at all.

**Consumers narrowed:** `EntityEventProcessor`/`fromEntityProcessor`/`applyEventStream` take `WritableStateStore`; `pruneMore`, `IndexerState`'s `createState`/`createProcessor`/`statesByGeneration`/prune set, `demotion`'s `stores()`, the CLI's container/`store`/prune path, and the stratagems workload's `replayIntoStore`. The claim is taken where a store is BUILT (the CLI's `createState`, the browser app's `createState`, each harness's own factory). The conformance suite lost `throughAClaimedWriter`: 622 cases → 321, which is exactly the duplicated pass (301) and nothing else.

**CONTEXT.md**: the `StateStore` entry now describes the three names, `prune` is titled `WritableStateStore.prune`, the conformance-suite entry says `StateStoreBackend`, the `readable store / writable store` entry is folded rather than duplicated (including the opposite-arbiter warning against conflating this with ADR-0044's stream rule), and **storage identity** is pinned as its own entry. Both package READMEs and the browser usage example follow.

**Gate**: `format:check`, `check:adr`, `check:refs`, `build`, `typecheck`, `test` all pass (3,853 tests, 0 failures). `pnpm changeset status --since=main` fails **locally only** and not because of coverage: `@changesets/git` finds changesets with `git diff --name-only`, which cannot see an untracked file, so it reports "no changesets were found" for any new changeset before it is committed. Verified by probe: a changeset naming every published package produces the same error, and a clean tree produces none. It will pass in the runner's post-commit gate, as every previous task's changeset did.

## Decisions

- **A THIRD seam name, `StateStoreBackend` (= the reads plus the five mutating verbs), against ADR-0077's "three names during the migration and two afterwards".** ADR-0077 assumed only the shapes a CONSUMER holds need naming; three kinds of code hold the implementor's surface and can hold neither consumer shape — a backend's `implements` clause (left as `implements StateStore`, a backend that forgot `prune` would compile and fail at runtime), every factory (`StateStoreFactory`, `WorkloadStoreFactory`, `BrowserStateStoreFactory`, `createBrowserStateStore`), and `openForWriting` plus the two full-surface wrappers. Alternative considered: spell `StateStore & StateStoreMutations` at each — rejected on the count (fifteen sites, a name that says nothing about who holds it, nowhere to document what a backend owes). Coherence: "backend" is the word this repo already uses for exactly this (the conformance suite is "what a backend must pass"), it names an IMPLEMENTOR while the other two name HOLDERS, and it REPLACES two scaffolding names rather than adding a third to them. Touches: every backend class, every store factory, `CONTEXT.md`'s StateStore + conformance-suite entries. Recorded in **ADR-0079**.
- **`openForWriting` takes a `StateStoreBackend`, not a store already narrowed to its reads.** ADR-0077's sketch wrote `openForWriting(store: StateStore)` when `StateStore` meant the readable shape, which would let a reader widen its own handle back and leave the narrowing saying nothing. Requiring the backend costs no caller (every site that claims is the site that built the store) and makes the narrowing one-way. The alternative kept the sketch's letter and needed a cast to `clearCursor` inside the seam module. Touches: ADR-0078's demotion story, which already says a demoted writer builds a NEW store; asserted by a new `@ts-expect-error` in `writable-seam.test.ts`.
- **A cap-refused generation now leaves an EMPTY namespace behind, and I changed a landed acceptance criterion's test to say so.** `a-changed-context-creates-a-successor-instead-of-clearing` asserted "no orphan tables, no orphan record"; claiming migrates (ADR-0077) and `ReceivingIndexer.add` builds state → processor → record, with the cap enforced at the record, so a host that claims in `createState` performs DDL before the refusal. The order cannot be swapped: the record is keyed on the processor's version hash, which needs the processor, which needs the state (ADR-0043), and a pre-check on the count alone would refuse re-opening a generation the container already holds — the case `create` deliberately resolves. The test now asserts the narrower truth (no registry row; the namespace carries no state; the only row in it is the claim), and `receivingContainer.ts`'s doc sentence, which claimed the state factories create no storage until the first write, is corrected. Alternatives, both in `@etherfold/core` and both out of this fence: a `dropState` on the refusal path, or a cap pre-check taking the identity. Touches: `@etherfold/core`'s `add()` contract and that done task's criterion. Recorded in **ADR-0079**.
- **`VersionedStateEventProcessor` claims on FIRST USE rather than in its constructor.** Claiming is asynchronous and that constructor is not, and it is `new`ed at eighty call sites. Alternative considered: an async factory replacing the constructor — rejected because it changes eighty sites and turns a documented construction-time refusal (a processor with no `version`) into a rejected promise at thirty-two of them. Safe here for a reason specific to this class: the store is one it built and nothing else holds, so there is no rival between the constructor and the first fold. It is still an EXPLICIT claim (`openForWriting`); what the narrowing removed is a mutation nobody claimed for, not a claim at a particular line. `getVersionHash`/`getCodeFingerprint` now call the shared `entityProcessorVersionHash`/`processorCodeFingerprint` (one formula, two callers — not two spellings) because a host reads both before anything folds. Touches: `@etherfold/processor-sqlite`'s public class. Recorded in **ADR-0079**.
- **`createBrowserStateStore` deliberately does NOT claim, so `createState` is now `async () => openForWriting(await createBrowserStateStore(...))`.** This is the user-visible default: a tab that only RENDERS opens the same database, so claiming in the store factory would have every reading tab take the store from the tab that is indexing (spec stories 4 and 6). Building a store and becoming its writer are two acts. Alternative considered: claim inside `createBrowserStateStore` so the documented example was unchanged — rejected for exactly that reason. Touches: `BrowserGenerationSpec.createState`, both examples, the browser README and `IndexerState`'s two doc examples.
- **The conformance factory hands over the BACKEND, so every case mutates the store it is given directly.** The alternative was one shape = the claimed handle, which would have rewritten fourteen case files and lost the undecorated declaration-probe question the suite doc names. Backend test files are untouched, which is what the "concrete classes unchanged" criterion needs. Touches: `StateStoreFactory`, `StorePair`, `WorkloadStoreFactory` (kept matching it, as its doc promises).
- **`openForWriting` / `openForReading` / `WRITER_CLAIM_KEY` are re-exported from `@etherfold/processor-entities`**, beside the bootstrap primitives and for the same stated reason: they are on the same boot path, so an app calling `openAndBootstrap` does not need a second import to say which of the two it is doing. Additive; touches that package's public surface (in the changeset).
- **`packages/state-store-patch/test/sparse-stream.test.ts`'s `award` narrowed to `StateStoreBackend`, not to `WritableStateStore`.** It is a backend suite driving concrete `PatchStateStore` instances, so requiring a claim there would add a claim to a suite whose subject is the class; the seam-typed mutation the task wanted gone is gone either way, because the READABLE type no longer carries `applyBlock`. `pointsOf` beside it stays `StateStore`, which now documents that its reads come through the consumer seam.


wighawag:
APPROVE ✅ — Gate 3 (conductor's diff-vs-criteria review, including an independent browser run)

The only task in this drive that can break a caller, and the one whose five-way `blockedBy` fan-in was the safety property. It landed green on the first attempt, with all five blockers on main beneath it.

| Criterion | Evidence |
|---|---|
| The readable seam carries the reads, `declarations`, `migrate`, capabilities and the cursor read, and NO data-mutating member | `interface StateStore` is exactly: `capabilities`, `declarations`, `migrate`, `readCursor`, `readRetentionEnforcement`, `getCurrent`, `getAsOf`, `listCurrent`, `listAsOf`. `migrate` stays, for the stated reason. |
| Every seam-typed mutating site holds the writable shape | 81 files across the four named packages, plus all three named exceptions: `state-store-patch/test/sparse-stream.test.ts`, `state-store-indexeddb/browser/workload.ts` and `browser/browser/workload.ts` |
| A mutation without a claimed token cannot be expressed | `WritableStateStore = StateStoreBackend + token`, obtainable only via `openForWriting` |
| Concrete backend classes UNCHANGED with full surface; `createD1Store` returns the concrete class | The SQL tier keeps `queryCurrent`/`queryAsOf`/`applyBlocks`/`drop`; backend `src` changes are the `implements` clause. `platforms/cf-worker` is absent from the diff entirely, exactly as the task predicted, because it holds no seam-typed value. |
| Conformance no longer parameterised over two shapes | The expand-phase parameterisation is gone; every chapter is asked once |
| `CONTEXT.md` updated where this made it false | All four: the **StateStore** entry rewritten for the three names; **prune** retitled and re-homed; the **conformance suite** entry now says `StateStoreBackend`; and **storage identity** pinned as its own entry — including the instructed warning that the two rules have OPPOSITE arbiters, a stream's writer being the OLDEST SURVIVING generation while a state store's is the LAST CLAIMANT |
| Every existing suite green, INCLUDING the browser contention case | See below |
| Full gate: format, ADR, refs, changeset status, build, typecheck, test | Green on the rebased tip |
| Changeset | Present, `major` where it must be |

**I re-ran the browser suite myself on this branch**, because that criterion names the contention case explicitly, this PR modifies its harness (`cut.ts`), and the case sits outside the acceptance gate so the green above says nothing about it:

```
✓  2 [chromium] › several tabs contending for the same heights leave one winner and no torn state (571ms)
✓  8 [firefox]  › several tabs contending for the same heights leave one winner and no torn state (2.4s)
✓ 14 [webkit]   › several tabs contending for the same heights leave one winner and no torn state (2.2s)
18 passed (28.0s)
```

The contention guarantee survives the narrowing on all three engines.

Three things I checked rather than accepted:

- **The diff is shaped as the task predicted.** +983/-582 across 81 files, overwhelmingly type annotations and deletions, with the two expand-phase scaffolding names (`ReadableStateStore`, `StateStoreMutations`) removed rather than left behind. No new logic grew, which is the task's own stated failure signal.
- **`docs/spikes/` is untouched.** Frozen evidence, outside the workspace filter, and correctly left alone.
- **The two surprising consequences are disclosed, not buried.** Claiming MIGRATES, so a generation the cap later refuses leaves an empty namespace behind (the cap is keyed on the processor's version hash, which needs the processor, which needs the state) — with the honest statement of what a refusal still guarantees: no registry record and no state. And `VersionedStateEventProcessor` claims on FIRST USE rather than in its constructor, because claiming is asynchronous and that constructor is not; still an explicit claim, and safe because the store is one it built and nothing else holds. Both are in the changeset and in ADR-0079.

The changeset is genuinely good migration documentation: it is organised by what the reader IS (holds a store / implements a backend / wires a browser app / runs the conformance suite) rather than by what changed, which is the difference between a breaking change someone can act on and one they have to reverse-engineer.

Gate 1 green. Gate 2 off by repo config for this drive. `a-second-writer-writes-nothing` is now complete: a reader cannot write because the type says so.

PR #131, opened 2026-09-11
feat(the-cycle-checks-chain-identity-once-after-the-fetch): The in-proc…

wighawag:
Task: `work/tasks/done/the-cycle-checks-chain-identity-once-after-the-fetch.md`

## Drift check (done first)

The task's premises still hold today:
- `packages/core/src/indexer.ts` `promiseToIndex` really had the two `eth_chainId` calls a few lines apart, both hand-rolling the hex parse, the before one carrying the `chainChanged` comment (the repo's only mention of it outside docs).
- ADR-0081 is `accepted, not yet implemented` and says exactly what the task says.
- Both sibling tasks are still in `work/tasks/ready/` (`a-chain-identity-refusal-names-the-chain-it-expected.md`, `the-fetcher-checks-chain-identity-once-after-the-fetch.md`), so the engine still throws bare `Error`s and `logFetcher.ts`'s `assertChain('before'|'after')` pair is untouched. No collision, nothing stale.
- The one thing the task said that needed resolving: the spec is at `work/specs/tasked/…`, not `work/specs/ready/…` (the launch header in the prompt pointed at `ready`). Same file, read it there.

## Changes

`packages/core/src/indexer.ts`
- Deleted the before-fetch `eth_chainId` call and its refusal from `promiseToIndex`.
- The surviving after-fetch call now goes through the existing internal `getChainId` helper instead of re-parsing the hex inline, and carries a banner comment saying why it is the survivor (the fetch is the window that can corrupt a fold, the before call only failed fast), why it must not be restored for symmetry or deleted as the cheaper-looking one, why `chainChanged` is not a second line of defence (never built, no listener, `EIP1193ProviderWithoutEvents` cannot carry a subscription, and an async event cannot replace a check at a known point), and why it has no flag. Cites ADR-0081.
- The refusal's type and message are untouched (`throw new Error('chainId changed after fetch')`), per the scope fence.

`packages/core/test/follower.test.ts` (test written first, watched fail on both counts, then made green)
- Rewrote the section banner: the chain is checked once, after the fetch, and both outcomes plus the call count are pinned.
- Extended the existing `movableChain` harness with an ordered `trace` of `eth_chainId` / `eth_getLogs` and a `forgetCalls()` so a trace is one cycle's; `indexerOn` now also traces `fold` and `stream write` and gives the generation a `memoryStream` keeper, so "nothing reached the stream" is a claim about a real seam.
- New case: a normal cycle's trace is exactly `['eth_getLogs', 'eth_chainId', 'stream write', 'fold']`, with a separate count assertion on `eth_chainId` so a restored before-fetch call fails loudly.
- Rewrote `is REFUSED before it fetches at all when it moved between cycles` into `is REFUSED when it moved BETWEEN cycles too, and the range it wasted is re-derived`: still refused, nothing folded, no stream write, no cursor, and the next cycle asks for the same range and folds it.
- The mid-fetch case (the guard being kept) is untouched and green.

`.changeset/the-cycle-checks-chain-identity-once-after-the-fetch.md`: `@etherfold/core` patch, describing the halved identity cost, what a caller sees change (the between-cycles refusal now happens after the range is fetched, cursor unmoved), and why the survivor has no flag.

## Verification

`pnpm format:check`, `pnpm check:adr`, `pnpm check:refs`, `pnpm check:changesets`, `pnpm build`, `pnpm typecheck`, `pnpm test` all green (core: 1078 passed, 1 skipped; whole workspace green). Working tree holds only the two modified files plus the new changeset.

One gate note, not a problem with the change: `pnpm changeset status --since=main` reports "no changesets were found" while the changeset is still UNTRACKED, because `@changesets/git` finds changeset files with `git diff` against the diverged commit and `git diff` cannot see untracked files. It resolves as soon as the runner's `git add -A` commit lands; `npx changeset status` (no `--since`) parses the file and bumps `@etherfold/core` as expected.

## Decisions

- **The surviving call uses the internal `getChainId` helper rather than keeping the inline hex parse.** Why: the prompt points at it, it is what the fetcher path already uses, and leaving a hand-rolled `parseInt(...slice(2), 16)` beside a helper that does exactly that is the duplication ADR-0081's "the two shapes converge" consequence argues against. Alternative considered: pure deletion, leaving the after call byte-for-byte as it was, to keep the diff minimal for the sibling refusal-type task. Touches: `the-fetcher-checks-chain-identity-once-after-the-fetch` not at all, and `a-chain-identity-refusal-names-the-chain-it-expected` only adjacently (it rewrites the `throw` line below; the `chainId` variable name and its value, a decimal string, are unchanged, so its edit still applies cleanly).
- **The rewritten between-cycles test asserts the after-fetch refusal message (`/chainId changed after fetch/`).** Why: with the before call gone, that provider now trips the surviving guard, and the message is the one the code has today; matching the sibling mid-fetch case keeps one spelling in the suite. Touches: `a-chain-identity-refusal-names-the-chain-it-expected` will have to update this regex along with the mid-fetch one it was always going to update. I deliberately did not soften it to a bare `rejects.toThrow()` to hide that, because an untyped catch-all would stop the test saying which guard fired.
- **The chain-swap harness gained a stream keeper, so both refusal cases now assert nothing was written to the stream.** Why: acceptance criterion 3 is about the fold AND the stream, and the group previously ran with no keeper at all, making the stream half unassertable. Touches: only this `describe`'s own `indexerOn`; the existing mid-fetch case is otherwise unchanged and green.
- **ADR-0081's `status: accepted, not yet implemented` is left as-is.** Why: it covers three tasks and two are still in `ready/`; flipping it now would overstate what landed and would also collide with whichever sibling lands last. Alternative considered: amend it to say the engine half is done, which is a running commentary an ADR should not carry.


wighawag:
APPROVE ✅

Gate-3 (conductor's diff-vs-criteria review). Gate-1 green; Gate-2 off for this drive by configuration, so this is the review of record.

**Per criterion:**

- **One `eth_chainId` per cycle, after the fetch, before the write or the fold** — met, and asserted twice over: as a count (`trace.filter(...) === ['eth_chainId']`) and as a POSITION (`['eth_getLogs', 'eth_chainId', 'stream write', 'fold']`). The order assertion is the stronger half and the criterion only asked for the count; a count alone could not have said where the survivor sits.
- **A count assertion, so restoring a before-fetch call fails loudly** — met. Both assertions above break on a restored call.
- **Mid-fetch swap still refused, nothing reaches fold or stream** — met, and the existing case's own assertions are unchanged. The harness around it gained a keeper, which strengthens rather than weakens it.
- **The between-cycles case rewritten, not deleted** — met, and this was the criterion most at risk. The case is rebuilt around the guarantee that survives (a moved provider reaches neither stream nor fold, and the cursor stays put) and drops only the cost that changed (`chain.ranges` being empty). It then goes one better than asked and drives a second cycle to show the wasted range is re-derived rather than skipped.
- **The cursor does not move on a refusal** — met, asserted directly (`stream.cursor === undefined`) and behaviourally via the re-derived range.
- **The `chainChanged` comment no longer promises a mechanism that does not exist, and the survivor carries why it survived** — met. The replacement comment states the asymmetry, names the deleted call, warns against restoring it for symmetry AND against deleting the survivor as the cheaper-looking of the two, and records that `EIP1193ProviderWithoutEvents` structurally cannot carry a subscription. It cites ADR-0081 rather than restating it.
- **Tests mirror the repo's style; changeset present** — met. `@etherfold/core` is published, patch is right for a behaviour-preserving cost reduction.

**Drift notes and sibling boundaries, checked explicitly:**

- The refusal's TYPE and message are untouched (`throw new Error('chainId changed after fetch')`), as the task required — this task is about which calls happen, not how the refusal reads. `a-chain-identity-refusal-names-the-chain-it-expected` still owns that line.
- No flag, no config field, no option. That was proposed, argued and withdrawn in ADR-0081, and nothing here reopens it.
- ADR-0081's `status: accepted, not yet implemented` is correctly LEFT IN PLACE. It covers three tasks, two of which are still in `ready/`; the terminal task `a-chain-identity-refusal-names-the-chain-it-expected` is the one pinned to remove it. Flipping it here would have overstated what landed, and the Decisions block reasons about exactly this.
- Adopting the internal `getChainId` helper in place of the inline hex parse is in scope and correct: the task's own prompt points at it, and the Decisions block records the knock-on for the sibling task (same variable name, same decimal-string value, so its edit still applies cleanly).

**Note for the sibling task, not blocking here:** the rewritten case pins `/chainId changed after fetch/`, so `a-chain-identity-refusal-names-the-chain-it-expected` will need to update this regex alongside the mid-fetch one it was always going to touch. The Decisions block calls this out and deliberately declines to soften it to a bare `rejects.toThrow()`, which is the right call — a catch-all would stop the test saying which guard fired.

Merging.

PR #132, opened 2026-09-11
feat(the-fetcher-checks-chain-identity-once-after-the-fetch): The split…

wighawag:
Task: `work/tasks/done/the-fetcher-checks-chain-identity-once-after-the-fetch.md`

All checks green. Summary of the work:

**Production change** (`packages/core/src/logFetcher.ts`):
- Deleted the opening `await this.assertChain('before')` from `fetchAndPush`.
- The surviving call sits unchanged in position, between the fetched range and the push, and still goes through `withRetries` (a flaky `eth_chainId` is a provider problem, not a chain swap). Its comment now carries the ADR-0081 reasoning: which of the pair is the guard, why the other is gone, and why the survivor has no flag.
- The class docstring's "made both before fetching and before pushing" is now "made ONCE a cycle, after the fetch and before the push (ADR-0081)".

**`packages/core/src/errors.ts`**: `UnexpectedChainError`'s exported signature is untouched (still `(expected, actual, 'before' | 'after')`); only its now-false docstring line was corrected and a JSDoc added on the `when` parameter saying `'before'` is unreachable from a cycle and why it is kept.

**Tests** (`packages/core/test/logFetcher.test.ts`): new group `a provider that changes chain mid-fetch`, borrowing the shape of the engine's `follower.test.ts` movable-chain harness (a provider wrapper that flips `eth_chainId` from inside `eth_getLogs`, plus a trace that includes the push):
- `costs ONE eth_chainId for a cycle, after the fetch and before the push` — a count assertion plus `trace === ['eth_getLogs', 'eth_chainId', 'push']`, so a restored opening call fails loudly and so does moving the survivor.
- `is REFUSED when it moves DURING the fetch` — `UnexpectedChainError` with both chain ids, message pinned to `checked after fetching`, `receiver.sends === 0`, nothing received.
- The pre-existing wrong-chain case in `what only this side can check` is untouched and still green (it now fetches a range before refusing, and still pushes nothing).

**Changeset**: `.changeset/the-fetcher-checks-chain-identity-once-after-the-fetch.md` (`@etherfold/core: patch`).

Gate: `pnpm format:check`, `pnpm build`, `pnpm typecheck`, `pnpm test` (all packages), `check:adr`, `check:refs`, `check:changesets` all pass. `changeset status --since=main` only sees committed/tracked changeset files, so it reports red until the runner commits; the file is present and correctly named.

Drift check: ADR-0081 is accepted and matches; the sibling engine task (`the-cycle-checks-chain-identity-once-after-the-fetch`) has already landed and touched only `indexer.ts` + `follower.test.ts`, so no collision. `packages/fetcher-host/test/classification.test.ts` still constructs `UnexpectedChainError(..., 'before')` directly, which is exactly why the constructor keeps both values.

## Decisions

- **Collapsed the private helper's `'before' | 'after'` parameter**, so it is now `private async assertChain(): Promise<void>` and passes the literal `'after'` to `UnexpectedChainError` at the single construction site. The alternative was leaving the parameter to mirror the error's own. Chosen because a private one-call-site helper whose parameter has one reachable value is dead flexibility, and a reader seeing `assertChain('after')` would reasonably go looking for the `'before'` call that no longer exists. Touches nothing outside the class: the *exported* error keeps both values (the fetcher host's classification tests construct `'before'` directly, and narrowing a published constructor buys nothing), and the retry label is now the constant string `'checking the chain id (after fetch)'`.
- **Corrected the `UnexpectedChainError` docstring in `errors.ts`** ("checked before it fetches and again before it pushes" was made false by this deletion) and added a JSDoc on the `when` parameter recording that `'before'` is now unreachable from a cycle but deliberately kept. This is the one file I touched outside the fetcher, and a sibling task (moving the engine's bare `Error` refusals onto this type) also edits this class, so a textual merge conflict is possible; the edit is confined to comments, no behaviour or signature.
- **Updated two order assertions outside the fetcher suite** (`test/noBlockTimestampIsFetched.test.ts`, `test/noTransactionDataIsFetched.test.ts`), which pinned the deduped method list as `['eth_chainId', 'eth_getLogs']`. They are about *which* chain calls a cycle makes, not their order, and the order flipped as a direct consequence of the deletion; they are now `['eth_getLogs', 'eth_chainId']` with a one-line comment naming ADR-0081. No assertion about the fetched/pushed data changed.
- **No ADR written**: the rationale is entirely ADR-0081's, already accepted, and the code now cites it at the surviving call site.


wighawag:
APPROVE ✅

**Per criterion:**

- **One `eth_chainId` per cycle, after the range is fetched and before the batch is pushed** — met, and asserted as a count AND a position (`['eth_getLogs', 'eth_chainId', 'push']`). Tracing the receiver's `send` to put `push` in the same trace is the right shape: it makes "before the push" an assertion rather than an inference.
- **A count assertion, so restoring an opening call fails loudly** — met.
- **Wrong-chain provider still refused with `UnexpectedChainError`, still pushes nothing, existing case green with assertions unchanged** — met, and I checked this one specifically rather than trusting the gate. `refuses to fetch from a provider serving another chain` asserts only `toBeInstanceOf(UnexpectedChainError)` and `receiver.sends === 0`; it never asserted the `when` side, which is exactly why the deletion leaves it untouched. The diff does not modify that region.
- **Swapped-during-the-fetch case added** — met. This is the case the surviving call exists for and the one that would silently stop being covered if someone later deleted the wrong call, so it earns its place. It asserts the type, both chain ids, `retryable: false`, that nothing crossed the wire, and the trace stopping at `eth_chainId`.
- **`UnexpectedChainError`'s exported signature unchanged** — met. Still three parameters, `when` still `'before' | 'after'`.
- **Tests mirror the repo's style; changeset present** — met, `@etherfold/core` patch.

**On the two shapes the task deliberately left open:**

Both were decided and both are recorded, which is what was asked. Collapsing the PRIVATE helper's parameter while leaving the EXPORTED error's is the better half of each choice rather than a split difference: the helper has one call site and one reachable value, so the parameter was dead flexibility and `assertChain('after')` would have sent a reader hunting for a `'before'` call that no longer exists, while the error is exported and constructed with `'before'` directly by the fetcher host's classification tests, so narrowing it would be a breaking change buying nothing. The retry policy is preserved on the survivor, as the task asked, and the docstring now says why (a flaky `eth_chainId` is a provider problem, not a chain swap).

**Drift notes and sibling boundaries:**

- Stayed inside the fetcher. The one file touched outside it, `errors.ts`, is comment-only: the class docstring said "checked before it fetches and again before it pushes", which this deletion made false, so correcting it was required rather than scope creep. No behaviour, no signature. The Decisions block flags the possible textual conflict with the sibling task that moves the engine's bare `Error` refusals onto this type; since the edit is confined to comments it will be trivial to resolve.
- The two order assertions in `noBlockTimestampIsFetched` / `noTransactionDataIsFetched` had to flip: they pinned the deduped method list and the order changed as a direct consequence. Neither assertion about fetched or pushed DATA changed, and each flip carries a comment naming ADR-0081.
- ADR-0081's `status: accepted, not yet implemented` correctly left in place; `a-chain-identity-refusal-names-the-chain-it-expected` is the task pinned to remove it.
- No flag, no option on the survivor.

**Two non-blocking nits, recorded for the end-of-drive triage rather than held against this PR:**

1. `refuses to fetch from a provider serving another chain` now slightly misnames itself: after this change the fetcher DOES fetch and then refuses to push. The assertions had to stay unchanged (criterion) and they did, but the title could have been updated without touching them. Cosmetic, and the body is correct.
2. The reflowed `UnexpectedChainError` docstring leaves a short orphan line ("possibly catch: the"). Prettier accepts it, so it is aesthetic only.

Merging.

PR #133, opened 2026-09-11
feat(a-chain-identity-refusal-names-the-chain-it-expected): A chain-ide…

wighawag:
Task: `work/tasks/done/a-chain-identity-refusal-names-the-chain-it-expected.md`

Everything is green. Summary of the work:

**Drift check (first):** all three premises held. The per-cycle site still threw `new Error('chainId changed after fetch')` via `getChainId` (both blocked-by tasks had landed), the load path still hand-rolled the hex parse and threw a bare `Error`, the reconfigure path still threw the indented multi-line template literal, and ADR-0081 still carried `status: accepted, not yet implemented`. Nothing superseded. Proceeded.

**What changed**

- `packages/core/src/errors.ts`: `UnexpectedChainError`'s third parameter widened from `'before' | 'after'` to an exported `ChainIdentityCheckPoint` (`'before' | 'after' | 'cycle' | 'load' | 'reconfigure'`), and the message is now built from a per-check-point table (`where` + `consequence`) held in one place in the module. `'before'` / `'after'` produce byte-identical messages to today, and `retryable` stays `false`, so the fetcher path and `@etherfold/fetcher-host`'s structural classification are untouched.
- `packages/core/src/indexer.ts`: the three engine sites now read the chain id through the existing internal `getChainId` helper and throw `UnexpectedChainError` with their own check point (`'cycle'`, `'load'`, `'reconfigure'`). The reconfigure site lost its hand-rolled hex parse and its indented template literal; the load site lost its second copy of the parse and now prints decimal ids like the others.
- `packages/core/test/aChainIdentityRefusalNamesBothChains.test.ts` (new): the load and reconfigure sites, asserting the TYPE, both ids on fields and in the message, the negative (no claim about a receiver or about nothing being pushed), plus non-vacuous "same chain still works" cases, a reconfigure that resets and therefore never asks, and a pin that the fetcher path's constructor and message are unchanged.
- `packages/core/test/follower.test.ts`: the two per-cycle chain-swap cases now assert by type + ids + "after the fetch" + the negative, instead of `/chainId changed after fetch/`; the section header points at the new file for the two siblings.
- `docs/adr/0081-...md`: the `status: accepted, not yet implemented` frontmatter removed (last task in the chain, per `ADR-FORMAT.md`).
- `.changeset/a-chain-identity-refusal-names-the-chain-it-expected.md`: patch on `@etherfold/core`.

**Gates run:** `pnpm format` + `format:check`, `pnpm build`, `pnpm typecheck`, `pnpm test` (whole workspace), `check:adr`, `check:refs`, `check:changesets`, all green. `changeset status --since=main` reports "no changesets found" only because it reads `git diff` and my changeset file is still untracked; it resolves once the runner commits.

## Decisions

- **The path-specific clause is derived from a WIDENED `when` parameter, not from a new free-text constructor argument and not from a second error class.** `UnexpectedChainError`'s third parameter is now `ChainIdentityCheckPoint` = `'before' | 'after' | 'cycle' | 'load' | 'reconfigure'`, and a module-private table maps each point to both its "where" fragment and its consequence sentence. The task recommended making the path-specific clause a constructor input; a closed enum of check points IS that input, and it is better than passing the clause as a string, because the wording then stays in one table in `errors.ts` (the five refusals cannot drift into five spellings, which is the state ADR-0081 exists to end) and callers cannot inject prose into a published error. Alternatives considered: (a) a `consequence: string` fourth parameter with the fetcher's clause as the default (rejected: prose as public API, and the wording spreads back out to the throw sites); (b) a shared base class with a second name for the engine (rejected: the task pins it as better only if the two paths need different retryability, and they do not, so it would give a host two identities to classify for one condition). Touches: `@etherfold/core`'s public error surface and `@etherfold/fetcher-host`'s classification, which reads `retryable` structurally (`retryable !== false`) and is therefore unaffected: all five points stay `retryable: false`.
- **`when` is re-documented as "the point at which the check was made" rather than "which side of the fetch caught it", and the engine's per-cycle point is `'cycle'`, not `'after'`.** Coherence check: this generalises an existing parameter rather than forking a concept, and the fetcher's `'after'` could not be reused for the engine's post-fetch check, because `'after'` carries the receiver clause that only the split deployment has. Reusing `'before'` for the load path was also rejected: ADR-0081 deleted the before-fetch call, so re-meaning `'before'` as "at load" would resurrect a value the ADR retired with a different meaning. `'before'` stays reachable-by-construction and unreachable-from-production, exactly as `the-fetcher-checks-chain-identity-once-after-the-fetch` decided. Touches: the fetcher-host classification test that constructs `new UnexpectedChainError('1', '137', 'before')`, which keeps working unchanged.
- **The reconfigure refusal keeps the old prose's remedy ("did you forget to pass a new source?") and the load refusal keeps "point at a node for the chain this source names, or correct the source's chainId".** These are the only actionable parts of the messages being replaced, and dropping them while typing the error would have made the refusal catchable and less useful at once. The engine's messages state no consequence about a receiver or about pushing, which the tests assert as a negative on all three sites.
- **No ADR written.** The rationale is ADR-0081's, already accepted, and its consequences section already states this end state; this change is what makes that section true, so the pending-status line went rather than a new decision arriving.


wighawag:
APPROVE ✅

Gate-3 (conductor's diff-vs-criteria review). Gate-1 green; Gate-2 off for this drive by configuration. This is the LAST task in ADR-0081's chain, so it carries the ADR obligation as well as its own.

**Per criterion:**

- **Per-cycle refusal names expected, received, and that it was caught after the fetch** — met (`'cycle'` point, message asserts `/after the fetch/`).
- **Load and reconfigure refuse with the same typed family, each naming both ids** — met, and both now read the chain id through the shared `getChainId` helper. The reconfigure site's multi-line template literal, indentation and all, is gone.
- **No refusal claims a consequence that does not hold on the path that threw it** — met, and this was the wrinkle the task flagged to get right rather than around. It is enforced as a NEGATIVE assertion (`not.toMatch(/receiver|pushed/i)`) at all three engine sites, so the fetcher's receiver clause cannot silently migrate back onto an engine path.
- **Catchable by TYPE, not by message** — met. Note the follower suite's two cases were upgraded from `rejects.toThrow(/chainId changed after fetch/)` to instance-and-fields assertions, which is a strict improvement and exactly the handoff the previous task's Decisions block predicted.
- **Chain id read through the existing shared helper, not a fourth hex parse** — met at all three sites.
- **`UnexpectedChainError` remains exported and constructible as today** — met. Widening the third parameter from `'before' | 'after'` to a five-value union is source-compatible for every existing caller, so the fetcher path and the fetcher host's classification test that constructs `'before'` directly both keep working untouched. I checked the classification itself reads `retryable` structurally and all five points stay `retryable: false`, so nothing crosses to the retry side.
- **Tests assert type and both ids for each of the three engine sites** — met. Two sites in the new file, the third pinned where the guard's other properties already live, with a comment in each location pointing at the other. Each site also has a non-vacuous positive case (same harness, one digit apart, gets through).
- **ADR-0081's `status: accepted, not yet implemented` REMOVED** — met, and removed correctly: the whole frontmatter block goes rather than the status being reworded, which matches the repo's convention (implemented ADRs such as 0079 and 0071 carry no frontmatter at all; only a non-default status gets one). I verified no `not yet implemented` reference to 0081 survives anywhere in `docs/adr/`.
- **Changeset present** — met, `@etherfold/core`.

**On the error-shape decision, which the task asked to be argued rather than assumed:**

The recommendation was to make the path-specific clause a constructor input. A closed `ChainIdentityCheckPoint` enum plus a module-private table IS that input, and it is the better reading of it: a `consequence: string` parameter would have made prose part of a published API and scattered the wording back to the five throw sites, which is precisely the five-spellings state ADR-0081 exists to end. The shared-base alternative was correctly rejected on the task's own stated condition (better only if the paths need different retryability, and they do not). Declining to reuse `'after'` for the engine's post-fetch check is the subtle call and it is right: `'after'` carries the receiver clause, and re-meaning `'before'` as "at load" would have resurrected a value ADR-0081 retired under a different meaning.

Keeping the two actionable remedies from the old prose ("did you forget to pass some new source?", "correct the source's chainId if the node is the one you meant") is the right instinct. Typing an error should not cost the only part of it an operator could act on.

**Scope:** the reconfigure site was in scope by the task's own explicit inclusion, and the extra case pinning that the check is skipped when the new source resets the indexer anyway is a genuine clarification of what the guard protects, not scope creep.

With this merged, ADR-0081 is fully in the code and no longer claims otherwise. Merging.

PR #134, opened 2026-09-11
feat(the-indexer-is-hosted-in-a-dedicated-worker): The indexer is hoste…

wighawag:
Task: `work/tasks/done/the-indexer-is-hosted-in-a-dedicated-worker.md`

Task complete. The gate is green (`format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test`), and the out-of-gate browser run passes on all three engines.

## Drift check

The task still matches reality: ADR-0082 is `accepted, not yet implemented`, `CONTEXT.md`'s **indexer host** entry is as quoted, `packages/browser/src` contained no `Worker`, `createIndexerState` is as described, the seam (`openForWriting` / `StateStore`) landed as ADR-0079 says, and the six dependent tasks are all still in `ready/`. Nothing stale; I built.

## What I built

**The host seam** (`packages/browser/src/host/`, six modules):

- `envelope.ts` — one typed request/response with correlation, surfaces multiplexed as CASES (`PortCases`), plus `HostProgress` and a protocol namespace tag so a foreign message on a shared scope is ignored rather than answered.
- `clone.ts` — `assertClonable`, called before every `postMessage` on both ends. Refuses functions, symbols and live class instances NAMING THE PATH (`the 'read' response.rows[1].store`).
- `endpoint.ts` — `MessageEndpoint` (the untyped duplex) and `HostAccess` (how a port is obtained: the one thing a hosting shape differs in), plus `executionScopeName()`.
- `serve.ts` — `serveIndexerHost`: opens the container, drives it, answers the port. Names no worker.
- `port.ts` — `connectToIndexerHost`: what the tab holds. No store, no mutating verb.
- `dedicatedWorker.ts` — the shape: `dedicatedWorkerHost(worker)` (tab) and `hostIndexerInThisWorker(spec)` (worker entry helper, which refuses to run in a document). The only code in the package that names `Worker`.

**Evidence**: `browser/indexer.worker.ts` (the ~8-line entry an app writes), a `hosted-in-a-worker` case in `cut.ts`, `browser/hostedInAWorker.spec.ts` (real browser, real worker, 3 engines: asserts `scope === 'DedicatedWorkerGlobalScope'` vs `tabScope === 'Window'`, the tip reached, `EXPECTED_A` — the same constant the main-thread case asserts — and the port's full surface). Node-side: `test/aHostFoldsAndATabAsksHowFar.test.ts` (host over a real `MessagePort`) and `test/onlyWhatCanCrossIsPosted.test.ts`. `bundlesForABrowser.test.ts` gained a case bundling the worker entry for a browser, so the in-gate run covers the entry shape. Browser results are not kept as files: the run is `pnpm --filter @etherfold/browser test:browser`, reported by the `list` reporter, outside the gate per the existing convention.

`createIndexerState` is untouched. A changeset is included. One observation note captured (`work/notes/observations/a-pre-fetch-cursor-publishes-a-nonsense-percentage.md`) for a latent NaN in the main-thread progress arithmetic I did not fix.

## Decisions

**The envelope's shape.** One `{protocol, kind, id, case, payload}` request and one `{protocol, kind, id, case, ok, value|error}` response, with the surfaces declared in a `PortCases` map (case name → `{request, response}`) that a later task adds a KEY to. Alternatives considered: a channel per surface (rejected by ADR-0082), and Comlink-style proxying (the ADR allows it; rejected because the contract is then implicit in a proxy's behaviour instead of in named message shapes a test can assert). The `case` is carried on the response too, so a refusal names the surface rather than an opaque id. Touches every later task on this port.

**No protocol VERSION, only a namespace.** `protocol: 'etherfold/indexer-port'` exists so a host ignores messages another library posted to a shared scope. A version number was considered and rejected: both ends come out of one build (the app's bundler compiles the tab and the worker entry from the same tree), so a version would be a compatibility promise nothing keeps. Touches the SharedWorker task, where one host serves several tabs of the same origin — still one build.

**Where the host/container seam falls.** The host is a NEW function (`serveIndexerHost`) that opens the container directly through `openIndexer`, not a wrapper around `createIndexerState`. It takes `BrowserGenerationSpec` — the exact shape `createIndexerState` already takes — plus what that function takes at `init`, so the last task can fold the two together rather than reconcile two vocabularies. What it deliberately does NOT carry yet: the scheduled prune, the stream-seed install, generation-progress reporting, demotion publication, the reactive triple. Alternative considered: moving `createIndexerState` wholesale now (rejected — the prompt fences it, and it would decide six later tasks by accident). Touches all six later tasks, especially `createindexerstate-becomes-the-main-thread-host`.

**The host auto-starts its driver, and stops on a failure waiting cannot fix.** There is no control surface yet, so a host that needed to be started could not be. It retries `isRetryable` failures on a one-second timer and stops otherwise, reporting `failure` through `progress` rather than throwing (an exception in a worker entry reaches nobody). This makes `start` in `a-tab-controls-the-indexer-across-the-port` an idempotent no-op on a fresh host unless that task changes the default. Alternative: start lazily on the first port call (rejected — it would make "how far have you got" a mutating verb).

**A new type name, `HostAccess`.** The glossary has HOST, CONTAINER, DRIVER, PORT and HOSTING SHAPE but no word for the thing a shape produces. `HostAccess` is `{host, endpoint, close?}` and is documented as exactly `CONTEXT.md`'s phrase "how a port is obtained". It does not re-mean PORT: the port is the typed boundary built OVER it. Alternative considered: overloading `connectToIndexerHost({worker})` with a discriminated options object (rejected — it would put the word `Worker` inside the shape-agnostic module and make the seam a branch rather than a file).

**The clone guard refuses class instances, which structured clone accepts.** Clone copies own properties and drops the prototype, so a store/processor/provider "arrives" and fails on its first call one layer from anything explaining it. Refusing is stricter than the platform and is the point. Risk: a later surface wanting to carry a class-shaped value must send its data instead. Touches the store-proxy task, whose rows are plain objects.

**`HostProgress` carries no derived figure, and reports the container's `0 of 0` honestly.** `blocksBehind` already means "behind the CANONICAL generation" (`GenerationProgress`), so reusing it for "behind the tip" would give one word two meanings; the two raw numbers cross and the derivations are the push task's to place. Consequence documented in the JSDoc and found the hard way: before the first fetch a cursor is `0 of 0`, so equality alone is not "caught up" — it made the Chromium run race the fold until the tests named the fixture's tip.

**`scope` is reported alongside `host`.** `host` is a label the entry passes; `scope` is `globalThis.constructor.name` measured where the answer was computed. It exists so "the UI thread is not doing the fold" is assertable without a timing threshold. Nothing branches on it.

**Two new refusals.** `hostIndexerInThisWorker` throws if called in a document (a stray import would start a second container on the UI thread folding into the same store), and again if the context has no worker messaging. `UnclonableValueError` is a new exported error type (a `TypeError`: the same value is refused identically on every attempt).


wighawag:
APPROVE ✅

Gate-3 (conductor's diff-vs-criteria review). Gate-1 green; Gate-2 off for this drive by configuration. This is the thickest task in the spec and six later tasks build on the seam it sets, so I reviewed it against the criteria AND independently re-ran the browser suite rather than taking the browser claims on trust.

**The browser criteria, verified rather than assumed.** The acceptance gate ends in `pnpm test`, which is vitest only; `test:browser` sits outside it by ADR-0030, so a green gate says nothing about the three criteria here that require a real engine. I cloned the branch to a scratch directory, installed, built, and ran `pnpm --filter @etherfold/browser test:browser`:

```
24 passed (23.0s)
✓ [chromium] browser/hostedInAWorker.spec.ts:34 › folds a real workload in a dedicated worker... (1.1s)
✓ [firefox]  browser/hostedInAWorker.spec.ts:34 › ... (2.0s)
✓ [webkit]   browser/hostedInAWorker.spec.ts:34 › ... (1.0s)
```

The new case passes on all three engines and the 21 pre-existing cases are still green, so the harness change did not cost anything that already worked.

**Per criterion:**

- **An app constructs a hosted indexer by writing a worker entry plus a constructor call** — met and DEMONSTRATED rather than claimed: `browser/indexer.worker.ts` is the entry, and its substance is three imports and one `hostIndexerInThisWorker({...})` call. The processor crosses as an import, which is the ADR-0082 decision this shape exists to honour.
- **The host opens the store for writing; nothing handed to the tab can mutate it** — met, and asserted as a fact about the TYPES a tab can name: the real-browser case pins `portSurface` to exactly `['close', 'host', 'progress']`. There is no mutating verb to forget not to call.
- **The tab asks how far it got and gets a correct answer across the real boundary** — met, and non-vacuously: the case pins the fixture's tip rather than just `lastToBlock === latestBlock`, which would also have been true of a host that had not fetched yet.
- **A real chain fold completes in the worker and matches the main-thread result** — met, against the same `EXPECTED_A` constant `indexing.spec.ts` asserts, so the two paths are compared rather than each being self-consistent.
- **The UI thread is demonstrably not folding, asserted on where the work happens rather than a timing threshold** — met, via `globalThis.constructor.name` measured in the context that computed the answer: `DedicatedWorkerGlobalScope` for the host, `Window` for the tab. No duration anywhere, so the case cannot flake on a loaded machine.
- **Structured-clone safety, refused with a message naming the FIELD** — met, and better than the criterion asked. The refusal carries the whole path (`the response.rows[1].read`), so a large payload says where, and it refuses class instances too, which structured clone would silently accept while dropping the prototype. That strictness is right: a store or provider that "arrives" and fails on its first call lands a layer away from anything that would explain it.
- **Host-agnostic host, with a seam a reviewer can point at** — met, and I checked the strong form of the claim rather than the comment asserting it. Stripping comments, the identifier `Worker` appears in exactly one module's code, `host/dedicatedWorker.ts`, plus a re-export line. `serve.ts` mentions workers only in prose. The SharedWorker and main-thread tasks can therefore add a shape without a second copy of the host.
- **Tested in a real browser with a real worker using the existing harness** — met and independently confirmed above.
- **Changeset present** — met, `@etherfold/browser` minor, correct for additive public surface.

**Vocabulary, which `CONTEXT.md` pins and which is easy to drift:** clean. No `body`, no `shell` anywhere in the new modules; the code uses HOST, CONTAINER, PORT and HOSTING SHAPE as the glossary defines them. The one new name, `HostAccess`, is introduced deliberately and justified in the Decisions block as a word the glossary lacks ("how a port is obtained"), and it explicitly does not re-mean PORT. That is the right way to handle a vocabulary gap: name it, argue it, and say it is new.

**On the decisions six later tasks inherit:** the envelope (`{protocol, kind, id, case, payload}` with a `PortCases` map a later task adds a KEY to) is the shape ADR-0082 asked for, and carrying `case` on the response so a refusal names its surface is a good call. Declining a protocol version is argued correctly (both ends come out of one build, so a version would be a promise nothing keeps) and the argument survives the SharedWorker task, which is still one build. Putting `serveIndexerHost` on the same `BrowserGenerationSpec` that `createIndexerState` already takes is exactly what keeps the terminal task a fold-together rather than a reconciliation of two vocabularies.

**Off-path finding, correctly captured rather than fixed:** `work/notes/observations/a-pre-fetch-cursor-publishes-a-nonsense-percentage.md` records that `setLastSync` divides by `latestBlock` and publishes `NaN` before the first fetch. The note leaves the production arithmetic untouched, fixes only the test-side half that bit this task, and points at `a-tab-sees-sync-progress-pushed-from-the-worker` as the likely owner. That is the right disposition and I will carry it into this drive's end-of-drive triage.

Merging.

PR #135, opened 2026-09-11
feat(a-tab-reads-the-store-across-the-port): A tab reads the store acro…

wighawag:
Task: `work/tasks/done/a-tab-reads-the-store-across-the-port.md`

## What I did

**Drift check first.** The task's premises hold: `the-indexer-is-hosted-in-a-dedicated-worker` landed with exactly the envelope the task assumes (`{protocol, kind, id, case, payload}` plus a `PortCases` map "a later task adds a KEY to"), ADR-0082 is unamended and still describes this surface as a mechanical proxy of `createReadSurface`'s four reads, `createReadSurface` is where the task says with the four reads it says, and `@etherfold/state-store-conformance` is the cases-as-data + thin-vitest-adapter shape it points at. Nothing stale; I built it.

**The port carries the store's four reads (`@etherfold/browser`).**
- `src/host/envelope.ts`: five new cases on the existing envelope (`getCurrent`, `getAsOf`, `listCurrent`, `listAsOf`, plus `declarations`), plus `PortRow`. No transport, correlation or clone-guard change.
- `src/host/serve.ts`: the host records which store each generation folds into and answers reads from the store the **canonical** generation folds into, resolved per read. A read arriving before the store exists waits for it, and is rejected with the failure that stopped the host if one never arrives. Rows are projected in the host by the same `declaredRow` the same-thread surface uses.
- `src/host/reads.ts` (new): `PortStateReads` (the untyped, entity-name-string handle, the port's counterpart of `EntityStateView`) and `createPortReadSurface(port, entities)`, whose result type *is* `ReadSurface<StateStore, D>` — the same type `createReadSurface` produces.
- `src/host/port.ts`: `IndexerPort.reads`. Nothing mutating crosses; the tab holds no store.

**Identical is a test, not a claim.** `browser/readWorkload.ts` holds a second fixture processor (a `transfer` entity keyed on `(blockNumber, logIndex)`, so a prefix listing is a real question) folding the *same* captured logs as `workload.ts`, and a 12-case list as data. `test/aTabReadsTheStoreAcrossThePort.test.ts` runs that list against both surfaces over a real `MessagePort`, plus byte-for-byte row comparisons, the no-mutating-verb check, a **held-chain** case that reads mid-fold deterministically, and the two refusals. `browser/readsAcrossThePort.spec.ts` + `browser/reads.worker.ts` run the same list in a page against a real dedicated worker: passing on Chromium, Firefox and WebKit (`pnpm --filter @etherfold/browser test:browser`, 27/27 across all specs).

**Bundle evidence.** `test/bundlesForABrowser.test.ts` bundles the three imports a tab's half writes and asserts no query runtime (`graphql` module, `queryCurrent`/`queryAsOf`), no store implementation and no engine came with them, and that it costs a fraction of the whole package: 8.2 KB against 152.4 KB minified.

**`@etherfold/state-store`:** `UnknownEntityError` (thrown by `mustGet`, so every backend raises one named refusal) and `assertDeclaredBy` exported from `read-surface.ts` so the port proxy checks declarations with the seam's own rule.

Changeset: `.changeset/a-tab-reads-the-store-across-the-port.md` (both packages, minor). `pnpm format:check && check:adr && check:refs && check:changesets && build && typecheck && test` all green (`changeset status --since=main` only sees a changeset git already tracks, so it resolves at the runner's commit).

## Decisions

**Reads are served from the store the CANONICAL generation folds into, resolved per read.** The host now records `statesByGeneration` exactly as `createIndexerState` does, and each read resolves `container.canonical.record` through it. Alternative considered: capture the one store this host builds today (simpler, and indistinguishable now, since a host opens exactly one generation) — rejected because the generation-control task adds generations and a captured store would then answer from a fold the pointer has moved off, silently. Touches `a-tab-controls-the-indexer-across-the-port` (generation control) and `createindexerstate-becomes-the-main-thread-host`, which will fold the two `statesByGeneration` maps into one.

**Four cases, not one `read` case with a verb inside.** The seam's own names (`getCurrent`/`getAsOf`/`listCurrent`/`listAsOf`) become four keys on `PortCases`. Alternative: a single `read` case carrying a discriminated verb — rejected because the envelope's dispatch already is that narrowing, and one case would make the host hand-narrow a union. Note the previous task's "unknown case" test posted a hypothetical `read` case; it still exercises the refusal path (no case is named `read`). Touches every later task that adds a case.

**The host projects the rows, the tab does not.** `declaredRow` runs where the store is, so "the rows are identical" is one implementation rather than two that agree by inspection, and version columns never cross. Consequence: the rows a tab gets are shaped by the HOST's declarations, which is what makes the declaration check below load-bearing. Touches nothing else on the port.

**A `declarations` case, and the surface's shape check happens on the first read rather than at construction.** `createReadSurface` refuses a declaration the store does not share at construction, naming both; a port cannot ask that synchronously. So `createPortReadSurface` stays synchronous, fires one `declarations` request immediately, and every read awaits it — same rule (`assertDeclaredBy`), same words, later moment. Alternatives: an async factory (rejected — it would stop the port surface mirroring `createReadSurface`'s call shape), or checking only the entity NAME at the host (rejected — a renamed FIELD would then project to `undefined`, the plausible wrong answer this seam refuses everywhere). Touches `@etherfold/state-store`, which now exports `assertDeclaredBy`.

**`UnknownEntityError` is added to `@etherfold/state-store`, not to the browser package.** "Refused in a way the app can act on" needs a name that survives structured clone (the class cannot). Putting it at the seam, raised by `mustGet`, means the refusal a tab gets is the refusal a same-thread caller gets, and avoids two classes of one name in two packages (the reason `BlockNotRetainedError` lives there). Alternative: a browser-local refusal raised before touching the store — rejected as a second vocabulary for one condition. It is a behaviour change for every backend (a subclass instead of a bare `Error`; message unchanged) and is called out in the changeset.

**An error's CLASS does not cross the port; its `name` does, and the shared cases assert on the name.** `errorFromPort` still rebuilds a plain `Error` with the host's name and stack, so `refuses(...)` in the case list compares `error.name` and the same case passes on both surfaces. Alternative considered: reconstructing the seam's error family on the tab side (`BlockNotRetainedError` carries clonable data, and `@etherfold/state-store` is a dependency of `@etherfold/browser`, so it is feasible) — rejected here as a second error vocabulary this task does not need, and because the previous task's recorded decision is that a class is not guessed at across the port. Touches any later task that wants `instanceof` on a refusal from the host.

**A second worker entry and a second fixture processor rather than extending `workload.ts`'s.** Its `token`/`counter` are both single-column ids, so a prefix listing there can never return two rows or truncate; adding an entity to the shared processor would have moved the version counts the existing prune case asserts. `browser/reads.worker.ts` is a sibling of `indexer.worker.ts` rather than a branch inside it, because a worker entry is what an APP writes and no app switches processors on a query parameter.

**`test/utils/port.ts`.** The `MessageChannel` `wire()` helper moved out of `aHostFoldsAndATabAsksHowFar.test.ts` so both host tests use one; that file is otherwise unchanged except for its port-surface assertion, which legitimately grew `reads` (as did the equivalent assertion in `hostedInAWorker.spec.ts`).


wighawag:
APPROVE ✅

**The browser criterion, verified rather than assumed.** `test:browser` is outside the acceptance gate by ADR-0030, so the gate's green says nothing about "tested in a real browser with a real worker". I re-ran the suite off this branch in a scratch clone:

```
27 passed (50.4s)
✓ [chromium] browser/readsAcrossThePort.spec.ts:33 › reads the store across a port to a real worker, and matches a same-thread surface (684ms)
✓ [firefox]  browser/readsAcrossThePort.spec.ts:33 › ... (1.5s)
✓ [webkit]   browser/readsAcrossThePort.spec.ts:33 › ... (3.2s)
```

New case green on all three engines, and the 24 cases that existed before this task are still green.

**Per criterion:**

- **All four reads across the real boundary** — met.
- **Rows identical to a same-thread surface, the same case list run both ways** — met, and met LITERALLY rather than in spirit: `browser/readWorkload.ts` exports one `readSurfaceCases` array, and `aTabReadsTheStoreAcrossThePort.test.ts` iterates that same array twice, once under `a surface over a store on this thread` and once under `a surface over a port to a host holding the store`. Two hand-written lists that happened to agree would have been the easy way to satisfy this wording and would have drifted the first time someone added a case; one list run twice cannot. A separate case pins the rows byte for byte.
- **Nothing on the tab's side can mutate the store, and the proxy exposes no way to try** — met, asserted as a surface shape (four reads per entity and nothing else) rather than as a promise.
- **Reading works while the fold is running** — met, with its own case rather than being folded into the equality cases.
- **No query-language runtime in the tab's bundle, ASSERTED not stated** — met, and this is the best-judged part of the PR. It is a real esbuild build over the three imports an app's tab half actually writes, asserting both the absence of named symbols (`MemoryStateStore`, `openIndexer`, `createIndexerState`, `serveIndexerHost`) and a loose size ratio. Keeping the entry inline rather than in a fixture file is right (the imports ARE the claim), and so is the deliberately loose 4x bound: a byte-exact size assertion is a test that fails on every honest change, and the failure worth catching is a tab-only import silently starting to drag an engine behind it, which 4x catches.
- **A read for an undeclared entity is refused in a way the app can act on** — met, and the reasoning behind `UnknownEntityError` is the sharpest thing in this diff. Same-thread the mismatch is caught at construction and `instanceof` is available; across a port an error's CLASS does not survive structured clone, so the refusal is only actionable if `name` is pinned as a readonly field. The test asserts on `name` rather than `instanceof`, which is the assertion that actually corresponds to what a tab can do. Placing the error at the seam beside `BlockNotRetainedError` is correct for the stated reason (every backend raises it through `mustGet`; two classes of one name in two packages would break `instanceof` across the package boundary), and the note on why it is deliberately NOT a `BlockUnavailableError` will stop someone merging the two families later: one is about a block this store cannot answer for, the other about the declarations and the store disagreeing, and no store configuration makes the second answerable.
- **Changeset** — met, and correctly names BOTH published packages it touches, `@etherfold/state-store` and `@etherfold/browser`, at minor.

**Scope:** the `@etherfold/state-store` edits are in scope rather than creep. Generating a typed read surface across a port is what forced the declaration check to be expressible against a declaration LIST rather than a live store handle, and `assertSameDeclaration` → `assertDeclaredBy(store.declarations, entity)` is exactly that narrowing, with the refusal it raises promoted from a bare error to a named one.

**Vocabulary:** clean. HOST, CONTAINER, PORT used as `CONTEXT.md` pins them; no `body`, no `shell`. The new surface slots onto the existing envelope as a CASE, which is what the previous task's envelope was shaped for, so the boundary is being extended rather than redesigned.

Merging.

PR #136, opened 2026-09-11
feat(a-tab-sees-sync-progress-pushed-from-the-worker): A tab sees sync…

wighawag:
Task: `work/tasks/done/a-tab-sees-sync-progress-pushed-from-the-worker.md`

## What I built

A tab is now PUSHED its host's sync progress, and nothing polls.

**Drift check first.** The task's premises hold. `serveIndexerHost`/`connectToIndexerHost` exist with the envelope the previous two tasks landed, `HostProgress` already carried the two cursors and its own docstring explicitly deferred "the phase an app actually renders" and the derived figures to this task, and ADR-0082's "Status is PUSHED, and the control surface NARROWS" is unamended. `SyncingState`/`StatusState` are where the task says they are. Nothing superseded.

**The envelope** (`src/host/envelope.ts`) gained a third message kind, `kind: 'push'`, carrying no correlation id (nobody asked), plus `PortPushes` as the map a later push adds a key to, and `subscribeToProgress` / `unsubscribeFromProgress` cases. `HostProgress` grew a `phase` (`SyncPhase`: `waiting` / `loading` / `catching-up` / `at-tip` / `refused`) and three derived figures (`blocksBehindTip`, `numBlocksProcessedSoFar`, `syncPercentage`).

**The host** (`src/host/serve.ts`) publishes on applied work: the container's per-batch cursor callback and the driver's own phase moves. A report identical to the last one posted is not posted, so a driver resting at the tip is silent. Nothing is posted until a tab subscribes, and a refusal is pushed with the `failure` that caused it.

**The tab** (`src/host/port.ts`) gained `onProgress(listener) => detach`. The first listener's subscribe answers with the current progress (so a tab attaching mid-fold is correct immediately); the last detach tells the host to stop posting.

**The helper** (`src/host/progress.ts`, new) is `createProgressReadable(port)`: the `Readable` shape this package already publishes stores as, holding the host's last report by reference, `undefined` until there is one.

**Tested** at the seam the task names: `test/aTabSeesSyncProgressPushedFromTheWorker.test.ts` (10 cases, over a real `MessagePort`, with a GATED chain so the push sequence is deterministic and no case waits on a clock) and `browser/progressPushedFromTheWorker.spec.ts` (a real dedicated worker, green on Chromium, Firefox and WebKit). The test was written before the production code and failed for the right reason (it exposed an unrelated engine wedge, noted below).

**Gate:** `format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test` all green (browser package 225 tests; whole workspace green). `changeset status --since=main` reports "no changesets found" only because the new changeset is still untracked (changesets reads `git diff <merge-base>`, which does not see untracked files); it resolves when the runner commits.

**Observation captured, not fixed:** `work/notes/observations/a-fetch-narrower-than-finality-never-advances.md` - configuring `fetch.maxBlocksPerFetch` at or below `stream.finality` wedges the cursor forever (50+ identical `eth_getLogs` ranges in three seconds, `lastToBlock` frozen), because a cycle rewinds by the unconfirmed window before it fetches. Out of scope; my fixture keeps the width above finality.

## Decisions

**What I carried over from `SyncingState`/`StatusState`, and what I left behind.** Carried: `waitingForProvider` (as the `waiting` phase, cleared at the same moment - when the container opens), the catching-up/at-the-tip distinction, the loading state, and `ExtendedLastSync`'s `numBlocksProcessedSoFar` and `syncPercentage` under their existing names and rules, so an app moving between the main-thread hook and a port binds the same words to the same meanings. Left behind: the load's sub-steps (`fetchingLogs`, `processingFetchedLogs`, `FetchingEventStream`, `ProcessingEventStream`, `InstallingStreamSeed`), because across a port each would be a message and none of them is a thing an app changes its screen for; `numRequests`, `autoIndexing`, `error`, `nonCanonicalGenerations`, `streamSeed` and `demotion`, which are other surfaces' (generation control, demotion, seeding) and not this channel's; and `totalPercentage`, which divides by the whole chain height, so a deployment starting at block 20,000,000 reads 99.9% from its first fetch. Alternative considered: mirror the whole triple - rejected, that is exactly what ADR-0082 refuses. Touches `createindexerstate-becomes-the-main-thread-host`, which will have to decide whether the hook's finer fields survive as the main-thread host's extras.

**The distance to the chain tip is `blocksBehindTip`, not `blocksBehind`.** `GenerationProgress.blocksBehind` already means how far a non-canonical generation is behind the CANONICAL one; reusing the bare word would give one term two meanings across one package. Alternative: reuse `blocksBehind` and disambiguate by context - rejected on the coherence rule, and the previous task's envelope note had already flagged this exact collision. Touches nothing else, but it is the name apps will render, so it is hard to change later.

**`at-tip` is the driver's own rest condition (`lastToBlock >= latestBlock` on an ADVANCE), and `createIndexerState`'s 20-block `catchupThreshold` did not come across.** It is literally the same expression the loop rests on, so a phase claiming the tip while the driver went on fetching is unexpressible. The threshold exists on the main thread to stop a UI flickering a few blocks from the tip; an app can apply its own from `blocksBehindTip`, which it has. Alternative: carry the threshold as a `HostedIndexerSpec` knob - rejected as a user-visible default that would make two rules for one question. Touches the main-thread host task, which inherits `catchupThreshold` and must decide whether it keeps it.

**The phase only reaches `at-tip` from an ADVANCE, never from the cursor alone.** A container that has loaded and not yet fetched publishes `0` of `0`, so equality is true before a single log is asked for. After the load the phase is `catching-up` (behind by an unknown amount) until an advance says otherwise. Alternative: compare the cursors wherever they update - rejected, it is the trap the envelope's own docstring names and it would render "live" over an empty database.

**The derived figures are computed in the HOST and are absent until a tip has been learnt.** Same placement as `declaredRow` (the previous task's "the host projects the rows, the tab does not"): one implementation rather than two that agree by inspection, and no fourth block number on the wire whose name would collide with `fromBlock`/`startBlock`. They are absent together below `latestBlock > 0` rather than computed from numbers that do not mean what they look like; `syncPercentage` answers `100` where there is no span rather than dividing by zero. This is the port-side half of `work/notes/observations/a-pre-fetch-cursor-publishes-a-nonsense-percentage.md`; I deliberately did NOT touch `createIndexerState`'s `NaN`, because that is the main-thread host task's surface.

**A push is a CHANGE, not a heartbeat.** A report identical to the last one posted is not posted. Without it, a driver resting at the tip would emit a push per rest interval, which is the polling this channel replaced with the cost moved to the other end of the wire. The consequence an app must know: an idle tab receives nothing, which is why attaching answers with the current value. Alternative: post every advance - rejected as a timer wearing a signal's clothes. Asserted directly (the host fetches more ranges while pushing nothing).

**Two named cases (`subscribeToProgress` / `unsubscribeFromProgress`) rather than one `subscribe` case taking a topic.** The same reason the four reads are four cases: the envelope's dispatch already is that narrowing, and a topic argument would make one case's response type depend on its request's contents. `a-reader-learns-when-the-state-moved` adds its own pair beside them if it wants one. Touches every later task that adds a push.

**Subscribing ANSWERS with the current progress, rather than triggering an immediate push.** It removes the window in which a fresh subscriber holds nothing and the race between that push and a concurrent `progress()` answer. The port caches the host's last word so a SECOND listener on an open subscription is served too (asynchronously, so no listener fires before the call that added it returned). Alternative: push-on-subscribe - rejected for the ordering ambiguity.

**`HostProgress` GREW rather than a second, richer pushed type.** One progress value, two deliveries: `progress()` pulls it, `onProgress` is pushed it. Two types would be two things to keep in step and "what did I miss" would answer differently from "where are we now". Consequence: the existing `progress` case's response now carries the phase and the derived figures too, which is additive.

**`onProgress` is a SUBSCRIPTION returning a detach, not an assignable slot.** It reads like the container's `onLastSyncUpdated` and is deliberately not the same contract (several listeners, each releasing its own), because a port is held by app code in several places while the container's callbacks are the host's private wiring. Documented at the type. Alternative: `subscribeToProgress(fn)` on the tab surface - rejected as noisier for the thing an app writes most.

**The fixture worker entry gained a `fetch` URL parameter.** `browser/indexer.worker.ts` now reads an optional fetch width from its own URL, as it already does for the database name, so the progress case can watch the fixture's five blocks arrive in more than one advance. Unset by default, so `hosted-in-a-worker` and `reads-across-the-port` fetch exactly as before. Alternative: a third worker entry - rejected as a whole file for one number.


wighawag:
APPROVE ✅

**The browser criterion, verified rather than assumed** — re-ran `test:browser` off this branch in a scratch clone:

```
30 passed (21.1s)
✓ [chromium/firefox/webkit] browser/progressPushedFromTheWorker.spec.ts:29
  › pushes progress from the worker to its own tab, and stops when the tab lets go
```

Green on all three engines; the 27 pre-existing cases still pass.

**Per criterion:**

- **A tab renders how far behind, and the number advances without polling** — met, and the "never asked to poll" half is asserted rather than implied.
- **Coarse phase correct at each transition** — met, and asserted as an ORDERED sequence rather than a set of reachable states, which is the stronger reading.
- **Subscribe and unsubscribe, with an unsubscribed tab receiving nothing** — met, and rounded out with the two cases that actually catch regressions here: pushes stop when the LAST listener lets go and resume when one returns, and remaining listeners keep being fed when one of several leaves.
- **The helper binds to a progress display and is demonstrably a view, not state of its own** — met, with an explicit case that it holds nothing of its own and another that it releases the subscription it opened.
- **Cadence tied to applied work, not a timer, and no timing dependence** — met, and pinned from the other side too: `says nothing while the driver rests at the tip, however many times it advances`. That is the case that stops a future timer-based push slipping in, because a timer would make it fail.
- **A tab attaching part-way learns current progress without waiting for the next push** — met (`tells a late one where the fold is`).
- **Changeset** — met.

Two cases beyond what was asked earn their place: a push carried through a REAL structured clone (so the envelope's clone guarantee is exercised on this surface, not just asserted for it), and the refusal being pushed when the driver stops on something waiting cannot fix.

**On the decision this task owned — what crosses from `SyncingState`/`StatusState` and what does not.** This is where ADR-0082 was most likely to be undone by accident, since the temptation is to mirror the three reactive stores across the port, and it was not. The carried set (the `waiting` phase, catching-up versus at-the-tip, loading, and `numBlocksProcessedSoFar`/`syncPercentage` under their existing names and rules) keeps an app's vocabulary stable when it moves between the hook and a port. The left-behind set is argued individually rather than waved away, and each reason holds: the load's sub-steps would each be a message and none is a thing an app repaints for; `numRequests`/`autoIndexing`/`error`/`nonCanonicalGenerations`/`streamSeed`/`demotion` belong to other surfaces; and `totalPercentage` divides by the whole chain height, so a deployment starting at block 20,000,000 would read 99.9% from its first fetch. Renaming the tip distance to `blocksBehindTip` rather than reusing `blocksBehind` is exactly right, since `GenerationProgress.blocksBehind` already means distance behind the CANONICAL generation and one word with two meanings is the thing `CONTEXT.md` keeps stamping out.

**It closed the port-side half of the observation I carried from PR #134, and correctly left the rest.** `a-pre-fetch-cursor-publishes-a-nonsense-percentage` reported that a cursor of `0` of `0` before the first fetch produces a `NaN` percentage. Here the derived figures are ABSENT until a tip has been learnt rather than computed from numbers that do not mean what they look like, and `syncPercentage` answers `100` where there is no span rather than dividing by zero, with a case pinning it (`never calls an unfetched cursor the tip`). `createIndexerState`'s own `NaN` is deliberately untouched, because that surface belongs to `createindexerstate-becomes-the-main-thread-host`. That is the right boundary: fix the half you own, leave the half you do not, and say which is which.

**New off-path finding, correctly captured and NOT fixed:** `a-fetch-narrower-than-finality-never-advances` records that `fetch.maxBlocksPerFetch` at or below `stream.finality` wedges the cursor forever, because a cycle rewinds by the unconfirmed window before it fetches. It is measured (50+ identical `eth_getLogs` ranges in three seconds, `lastToBlock` frozen), the mechanism is named with the file and function, and it names a plausible route in (a public node refusing wide ranges plus a deep finality) and a remedy in the shape `pair-compaction` and `retention` already use. This is a genuine defect with a silent failure mode — a host in that state reports `catching-up` truthfully forever — and I am carrying it to this drive's end-of-drive triage as task-worthy.

**Vocabulary:** clean, no `body`/`shell`; the surface is added to the existing envelope as a CASE.

Merging.

PR #137, opened 2026-09-11
feat(a-tab-controls-the-indexer-across-the-port): A tab starts, stops a…

wighawag:
Task: `work/tasks/done/a-tab-controls-the-indexer-across-the-port.md`

All green: build, typecheck, format, the full node suite (18 packages), and the Playwright run in all three engines.

## What I did

**The control surface crosses the port** (`@etherfold/browser`), as five cases on the existing envelope rather than a new channel:

- `packages/browser/src/host/envelope.ts` — `startIndexing`, `stopIndexing`, `reconfigure`, `generations`, `promotion` cases, plus two new types: `HostGeneration` (the four fields `GenerationProgress` already carries, under the same names and rules, plus `canonical`) and `HostReconfigure` (`{generation, added}`).
- `packages/browser/src/host/serve.ts` — the driver is now start/stoppable over a container that is opened ONCE and held open across a stop, so a stopped host still answers reads. `stopIndexing` awaits the cycle in flight, so no chain request is made after it resolves and the cursor is where a completed cycle would have left it. Rests (tip interval, retry pause) are interruptible. `reconfigure` serialises onto `Indexer.add({source, …hostFactories})`; `generations`/`promotion` report the container's own records, cursors and resolved policy. `onPromoted` drops the retired generation's cursor.
- `packages/browser/src/host/port.ts` — the matching verbs on `IndexerPort`; only the `source` crosses (a fold is code).
- `packages/browser/src/host/errors.ts` — `PortError.details` carries a refusal's own enumerable fields and `errorFromPort` re-attaches them, so a `GenerationCapReachedError` still says which cap, what limit and which generations could be deleted.

**Tests.** `test/aTabControlsTheIndexerAcrossThePort.test.ts` (11 cases over a real `MessagePort`, gated chain, nothing waits on a clock): stop/start honesty, a mid-cycle stop that lets its cycle land and resumes with contiguous ranges, idempotent start/stop plus a concurrent stop+start, reconfigure-beside + promotion + the promoted generation answering reads, generation visibility, the resolved policy (and `immediate`), the cap refusal keeping its name and fields, the clone refusal, structured clone. `browser/controlsTheIndexer.spec.ts` + a `controls-the-indexer` case in `cut.ts` run the same lifecycle against a real dedicated worker; it passes on chromium, firefox and webkit. The worker fixture gained a `generations` URL parameter (one database per generation, the rule `createState` states). Changeset added.

**Out-of-scope signal captured:** `work/notes/observations/a-successor-catches-up-one-range-per-tip-interval.md` (both browser drivers rest on the canonical cursor alone, so a successor catches up one fetch per tip interval; pre-existing, `createIndexerState` included).

## Decisions

**The port's `reconfigure` is the BESIDE path only (`Indexer.add({source})`); the in-place `updateIndexer` did NOT cross.** The task's acceptance criterion pins the beside behaviour ("state is not discarded, the new generation is built beside the live one, promotion follows the policy"), and the in-place verb is the outage-shaped discard the generation model replaced. Alternatives considered: carry both verbs (rejected — one word would mean two things across one boundary, and the hook's two names would both have to cross); carry only the in-place one (rejected — it discards state). **The consequence a reviewer should weigh:** the criterion's other half, "a reconfigure that ... invalidates the source", has no home on the port, because the beside path asks no source question of an existing fold and so produces no `SourceInvalidation` verdict — I deliberately did NOT re-derive one (`CONTEXT.md`: the verdict is reported, never re-derived). What a tab gets instead is `follows` (did the stream survive: no refetch) and `added` (did anything get created). The cost is real and worth naming: an ABI entry appended ABOVE the cursor is FREE through `updateIndexer` (ADR-0034) and, through this verb, moves the stream digest and pays for a full refetch in a new generation. Adding an `updateSource` case later is cheap by design (a key on `PortCases`). Touches `createindexerstate-becomes-the-main-thread-host`, which keeps both verbs on the hook and must decide whether the two surfaces stay asymmetric.

**No `promote` case: the port carries generation VISIBILITY, not pointer control.** The task's "Done means" is start, stop, reconfigure and visibility; `promote` is an operator affordance the server exposes behind its own admin token (ADR-0057). Consequence: an app running `manual` can see a successor catch up from a tab but cannot move the pointer from one. Alternative: add `promote(GenerationId)` — rejected as scope, and reversible as one case. Touches any later task that wants manual promotion or revert from a tab.

**`PortError` grew `details`, and `errorFromPort` re-attaches those fields onto the rebuilt error.** "Refusals keep their type" is unsatisfiable on name alone for the refusal a reconfigure actually meets: `GenerationCapReachedError`'s actionable half is `candidates`/`candidateStreams`, and a name-plus-sentence crossing would leave a tab parsing prose for the list it needs. Fields are clone-filtered per key and dropped individually rather than taking the refusal down. The narrowing axis stays the NAME (a prototype cannot cross), as `errors.ts` already predicted. Touches every later case that can refuse.

**`startIndexing`/`stopIndexing`, not the hook's `startAutoIndexing`/`stopAutoIndexing`, and no per-step advance case.** On a port there is only one kind of indexing — the host's driver — so the hook's "auto" (which distinguishes the timer loop from manual `indexMore`) has nothing to distinguish from. A `indexMore` case was deliberately not added: a round trip per cycle is the polling ADR-0082 replaced. Touches the main-thread-host task, which must reconcile the hook's four verbs with these two.

**Start and stop are idempotent ANSWERS, and a start racing a stop WAITS for it.** They name a state a caller wants, not an edge, so asking twice is not a refusal (this also sets a user-visible convention for every later lifecycle verb). The race matters: `startIndexing` awaiting an in-flight stop is what stops "asked to index, sitting there not indexing" when a stop unwinds after a start looked. Asserted directly.

**A stopped host keeps its PHASE and reports `indexing: false`; no `stopped` phase was added.** `SyncPhase` is a closed set describing where the FOLD is; `indexing` already says whether a driver is advancing it. Adding a sixth member would have made one question two.

**The container is opened once and stays open across a stop.** Stopping the driver is not closing the container: a settings screen switching indexing off must not take the app's data down with the fold. Consequence: `dispose()` remains the only thing that ends a host.

**A request payload that cannot cross now REJECTS instead of throwing synchronously** (`port.ts`). `assertClonable` ran before the promise existed, so `await port.call(x).catch(...)` did not catch the one input it refuses — and the function's own docstring already claimed it rejected. The reads that shipped first carry only strings and numbers, so nothing had reached it; a source is the first structured payload. Small, in-scope, and noted in the changeset.

**The fixture worker keys state per generation behind a `generations` URL parameter rather than by default.** An app that reconfigures must key on `context.stream` unconditionally, and the control case asks for it; the parameter exists only because the cases that predate the control surface read the worker's database back BY NAME from the page and cannot guess a digest. Same precedent as the previous task's `fetch` parameter, and it leaves those cases byte-identical.


wighawag:
APPROVE ✅

```
33 passed (24.8s)
✓ [chromium] browser/controlsTheIndexer.spec.ts:30 › a tab starts, stops and reconfigures the indexer across the port (743ms)
✓ [firefox]  ... (1.6s)
✓ [webkit]   ... (964ms)
```

Green on all three engines, 30 pre-existing cases still passing. The criterion asked specifically for a real-browser reconfigure THAT PROMOTES, which is the part easiest to satisfy in name only, so I checked the case body rather than the case title: it asserts two generations both level with the tip, one canonical and one not, and the resolved promotion policy (`{policy: 'on-catch-up', dropOnPromotion: false}`). That is a promotion demonstrated across a real boundary, not a reconfigure that merely returned without throwing.

**Per criterion:**

- **Start and stop, with no chain request after a stop resolves** — met, and asserted on chain requests after the stop ANSWERED rather than after it was issued, which is the version that actually pins the guarantee.
- **A stop mid-cycle leaves the store consistent and the cursor where a completed cycle would leave it; resume without re-indexing or skipping** — met (`lets the cycle in flight land, and resumes without re-indexing or skipping`). Letting the in-flight cycle land rather than tearing it down is the right shape, and both failure directions (re-index, skip) are asserted rather than just one.
- **Reconfigure preserves existing generation behaviour: state kept, new generation beside the live one, promotion by policy** — met, plus a case for the reconfigure that did not really change the source resolving to the live generation instead of pointlessly adding one.
- **A tab can see which generations exist, their progress, and which answers reads** — met.
- **A refused or source-invalidating reconfigure reports a TYPED outcome an app can branch on, not a message to parse** — met, and the case name says the thing that matters: the refusal keeps its NAME and its FIELDS. Since an error's class does not survive structured clone, name-plus-fields is the only form of "typed" that is real across this boundary, and it is consistent with how `UnknownEntityError` was handled in PR #135.
- **Starting a started indexer or stopping a stopped one is well defined and wedges nothing** — met, explicitly.
- **Changeset** — met.

Two cases beyond the criteria are the right instincts again: a generation report carried through a REAL structured clone, and the port refusing to carry anything that is code while naming where it sat. The clone guard added in PR #134 keeps being exercised on each new surface rather than trusted once.

**New off-path finding, correctly captured and NOT fixed:** `a-successor-catches-up-one-range-per-tip-interval` records that both browser drivers decide whether to rest on the CANONICAL generation's cursor alone, so a successor added by a reconfigure while the canonical generation is already at the tip advances one fetch range per rest interval. Quantified (at the default four-second interval, a successor with a thousand ranges of history takes over an hour of wall clock while the process sits idle between advances) and honest about why the suite does not show it (five blocks, a 0.05-0.25 s interval). It also notes the fix needs care in the other direction, since resting only when every generation is level must not turn into a hot loop for a generation that has not loaded or a follower that cannot advance.

Worth flagging that this one is NOT confined to the new code: `createIndexerState`'s `indexMoreAndCatchupIfNeeded` has the same shape, so it is a pre-existing defect this work surfaced rather than introduced. Leaving it unfixed here is right on scope grounds, and I am carrying it to this drive's end-of-drive triage as task-worthy. It is also worth the terminal task's attention, since that task adapts `createIndexerState` and will be looking at the other half of exactly this loop.

**Vocabulary:** clean; control arrives as CASES on the one envelope, as ADR-0082 requires.

Merging.

PR #138, opened 2026-09-12
feat(checktxinclusion-answers-from-the-tab): checkTxInclusion answers f…

wighawag:
Task: `work/tasks/done/checktxinclusion-answers-from-the-tab.md`

## What I did

**Drift check first.** The task's premises hold: `a-tab-controls-the-indexer-across-the-port` landed (`src/host/{envelope,serve,port}.ts` with the control cases), ADR-0082's envelope-as-cases shape is what is in the tree, `checkTxInclusion` and its verdict types are where the task says, and `CONTEXT.md`'s **indexer host** entry already lists **tx inclusion** as part of the status-and-control surface. Nothing superseded. Built as specified.

**One new case on the existing envelope**, no new channel:

- `packages/browser/src/host/envelope.ts` — `checkTxInclusion: {request: {queries: readonly TxInclusionQuery[]}, response: Record<string, TxInclusionVerdict>}`. Core's verdict type crosses unchanged; the doc states why it must not narrow to a boolean, why `minedAtBlock` is per query, and why it is a snapshot rather than a push.
- `packages/browser/src/host/serve.ts` — answers from the cursor the host is currently reporting plus `container.finalityDepth`, delegating to core's `checkTxInclusion`. It follows the canonical pointer for free: `onPromoted` already drops the retired cursor, so a tab is never answered from a window nobody maintains.
- `packages/browser/src/host/port.ts` — `IndexerPort.checkTxInclusion(queries)`, same name and same argument shape as the main-thread hook, so app code moves across unchanged.

**Tests** (written red first at the agreed seam, then made green):

- `packages/browser/test/checkTxInclusionFromTheTab.test.ts` (new, 7 cases over a real `MessagePort`): whole pending set in one round trip with each verdict whole; `minedAtBlock` per query producing `below-window` (and `ahead-of-cursor` for a receipt above the cursor); the before/after pair with the fold held one block below the transaction (`absent`/`window-miss` → `included`/`window-hit`); both unknowns — `not-synced` with no generation live and `window-not-covering` with a tip 100,000 blocks above the fold; a promotion no longer answering from the retired window; a real `structuredClone` of a verdict.
- `packages/browser/browser/txInclusionFromTheTab.spec.ts` + a `tx-inclusion-from-the-tab` case in `cut.ts`: the same claims in a real browser against a real dedicated worker, in three moments (nothing synced → held below the transaction → at the tip). Passes on chromium, firefox and webkit.
- `packages/browser/browser/indexer.worker.ts` gained two fixture-only gates (`holdChain`, `holdAbove=N`) released by the page posting to the worker directly — which works because the envelope ignores messages that are not its own, so that rule now gets exercised for real.
- `workload.ts` gained `txInBlock(...)`, reading a hash out of the fixture instead of quoting a counter's output; `aHostFoldsAndATabAsksHowFar.test.ts` and four Playwright specs updated for the new port surface key.

**Standing conventions:** changeset added (`.changeset/checktxinclusion-answers-from-the-tab.md`). Gate run end to end (`format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test`) — all green; `changeset status --since=main` only counts committed changeset files, so it resolves at the runner's commit as it does for every task here.

## Decisions

**`checkTxInclusion` does NOT wait for the host's container, unlike the four reads on the same port.** A read awaits `firstState` because "read me the rows" has no honest answer before a store exists. This question does have one: `unknown`/`not-synced` is a verdict an app renders (keep the optimistic update). Alternative considered: await `openContainer()` for symmetry with the reads — rejected because it turns the cheapest true answer into a call that hangs for as long as a provider takes to answer `eth_chainId`, on a UI thread laying out a pending queue, and because acceptance criterion 6 asks for the honest unknown rather than an error or a wait. This touches the read path's documented waiting rule (they now differ deliberately, and both sites say why) and it is what makes the "no generation live" case deterministic in the tests.

**Finality is read as `container ? container.finalityDepth : 0`, mirroring `createIndexerState` exactly.** The `0` is never observable: with no container there is no cursor either, so the `not-synced` branch answers before finality is consulted. Alternative: refuse when no container — rejected for the same reason as above.

**`checkTxInclusion` is on `IndexerPort` only, not on `IndexerHost`.** The host handle carries lifecycle verbs (`startIndexing`, `stopIndexing`) "so an entry point that wants to decide for itself has the verb its tab has"; a worker entry point has no use for the inclusion question, and `createIndexerState` is the main-thread surface for it. Touches the final task of this spec (`createIndexerState` becoming this host): if that task wants one surface for both, it adds the verb there rather than removing a duplicate.

**`CONTEXT.md`'s *tx inclusion* entry was extended, not forked.** It named `createIndexerState(...).checkTxInclusion` as *the* browser surface; there are now two at different layers, so the parenthetical lists both and the entry gains the port's rules (crosses whole, snapshot not subscription, follows the canonical pointer, does not wait). No new term was coined: this is the existing concept reaching an existing boundary. The final task still owns the `createIndexerState`-as-main-thread-host wording.

**Fixture-only gates in `indexer.worker.ts`, released over a non-port message.** Needed because the fake chain answers instantly, so "before the fold reached it" is otherwise a race in a real worker. Alternative considered: a port case to hold the fold — rejected outright as production surface invented for a test. The side channel is confined to `browser/` scaffolding and is off unless the URL asks for it, so every pre-existing case runs exactly as before.


wighawag:
APPROVE ✅

Process note: this task's FIRST dispatch died on a transient infra timeout before producing any code (no work branch, body still resting in the pool, one stale lock). I released the lock with `dorfl requeue` and re-dispatched; this PR is that second run. Nothing was carried over, so there is no half-finished work hiding in this diff.

```
36 passed (24.8s)
✓ [chromium] browser/txInclusionFromTheTab.spec.ts:33 › a tab asks about its pending transactions and a worker answers the verdict whole (285ms)
✓ [firefox]  ... (517ms)
✓ [webkit]   ... (505ms)
```

Green on all three engines, 33 pre-existing cases still passing.

**Per criterion:**

- **One call, several hashes, a verdict per hash** — met.
- **`minedAtBlock` per query producing `below-window`** — met, and demonstrated in the real-browser case too, not only in node.
- **Status AND basis survive the boundary, including the distinct causes of "unknown"** — met, and this is the criterion the whole task turns on. The browser case pins the full matrix over a real `postMessage`: `{status: 'unknown', basis: 'not-synced'}`, then `{status: 'absent', basis: 'window-miss'}` alongside `{status: 'included', basis: 'below-window'}`, then `{status: 'included', basis: 'window-hit', blockNumber: 104, blockHash: '0xa104'}`. Basis is preserved everywhere, so the collapse-to-a-boolean failure that would reintroduce the double-count cannot pass unnoticed.
- **Included / absent / unknown-with-not-synced-basis each demonstrated** — met, and the distinction that actually matters is asserted: a transaction the fold has passed and not seen is `absent`/`window-miss`, NOT `unknown`. Those two rendering differently in an app is the entire reason the basis crosses.
- **A verdict differs before and after the fold advances past the transaction** — met, with its own case, so "answered from current state rather than cached" is demonstrated rather than asserted.
- **No generation live, or a window covering nothing, gives the honest unknown rather than an error or a fabrication** — met, as two separate cases (`says nothing is synced yet rather than refusing`, `says its window covers nothing while it is far behind the tip`). Answering rather than hanging is the right behaviour and is worth the explicit case: unlike the four reads on this port, this call must NOT wait for the host's store, because a host that has synced nothing has an honest answer where "read me the rows" has none.
- **Changeset** — met.

Two cases beyond the criteria are well chosen: a verdict carried through a REAL structured clone, and `stops answering from the window of a generation the pointer has moved off`. The second is the subtle one and it is correct — the retired cursor is dropped at a promotion, so a window nobody maintains must answer nothing rather than answer stale.

**On the CONTEXT.md edit, which I checked specifically because the terminal task also owns part of this entry.** The change is additive and does not trespass: it names `IndexerPort.checkTxInclusion` ALONGSIDE the existing `createIndexerState(...).checkTxInclusion` rather than replacing it, and appends the port-crossing semantics (whole pending set in one call, basis intact, snapshot and never a subscription, follows the canonical pointer, does not wait for the store). `createindexerstate-becomes-the-main-thread-host` still owns reconciling the `createIndexerState` half of this entry with whatever it ships, and this edit leaves that reconciliation possible rather than pre-empting it. Documenting "it is a SNAPSHOT and never a subscription — a watcher asks again when `progress` says the fold moved" in the glossary is a good call, because it is exactly the thing an app author would otherwise assume the wrong way.

**Vocabulary:** clean; the surface arrives as a CASE on the one envelope, as ADR-0082 requires.

Merging.

PR #139, opened 2026-09-12
feat(a-terminated-worker-restarts-and-resumes-from-the-cursor): A termi…

wighawag:
Task: `work/tasks/done/a-terminated-worker-restarts-and-resumes-from-the-cursor.md`

All green: full gate (`pnpm build && pnpm typecheck && pnpm test`) passes, and the Playwright run passes on all three engines (39 tests, including the new one).

## What I built

**First, the drift check.** The task's premises hold: `checktxinclusion-answers-from-the-tab` is in `work/tasks/done/`, every surface the port carries exists, ADR-0082's closing section is unamended, ADR-0027's cursor-with-the-block is in force, and `openForWriting`'s claim neither blocks nor expires (so a corpse's store is simply taken over). Nothing superseded. I built it.

**`packages/browser/src/host/restart.ts`** (new): the vocabulary and the policy. `HostDeath` (cause, which consecutive attempt, whether a replacement is coming, how long until it, how many calls it rejected), `IndexerHostDiedError` (carrying the death and the CASE the lost call was on), and the watch/restart options with their defaults resolved in one place. It also pins the three neighbouring words apart: a **death** (host gone) is not a `HostProgress.failure` (live host, stopped driver) and not a **demotion** (live writer, lost claim — the opposite case).

**`port.ts`** now owns the host's lifetime: it watches (probe on silence), rejects everything in flight by type, tells the app through `onHostDeath`, releases the corpse *before* opening a successor, restarts through the shape's `reopen` with a bounded backoff, and re-opens the tab's own progress subscription against the new host. Nothing tells the new host where to resume: it reads the cursor.

**`endpoint.ts` / `dedicatedWorker.ts`**: `HostAccess` gains `reopen?`, and `dedicatedWorkerHost` now takes `() => new Worker(...)` instead of a worker.

**`envelope.ts` / `serve.ts`**: one new case, `ping`, whose answer is its whole content.

**Tests.** `test/aTerminatedHostRestartsAndResumes.test.ts` (5 cases over a real `MessagePort`: typed rejection + the event, resume asserted on the fetched ranges, every surface after the restart, the bounded loop, and a close that is not a death). `browser/restartsAndResumes.spec.ts` + a `restarts-and-resumes` case: a real dedicated worker terminated **while it is writing a block**, held below the tip so there is a fold to die in the middle of, asserting on the ranges the replacement asked the node for (nothing at or below the start block, union with no gap) and on the state matching `EXPECTED_A`. The worker fixture reports its fetches and its writes to the page, because a worker's memory dies with it and a delivered message does not.

`CONTEXT.md`'s **indexer host** entry carries the death/restart model; a changeset accompanies the change.

## Decisions

**A hosting shape takes the LINE THAT BUILDS a host, not a host.** `dedicatedWorkerHost(worker)` became `dedicatedWorkerHost(() => worker)`, and `HostAccess` gained an optional `reopen`. A shape *is* "how a port is obtained" (ADR-0082, `CONTEXT.md`), so obtaining one again after a death belongs there. Alternative considered: accept `Worker | (() => Worker)`, which keeps every call site compiling but leaves ports that silently cannot restart, with nothing in the types saying the restart half is missing. Touches: the 5 `cut.ts` call sites, `bundlesForABrowser.test.ts`, and the two remaining shape tasks (`a-sharedworker-serves-several-tabs-from-one-host`, `createindexerstate-becomes-the-main-thread-host`) — a shape that genuinely cannot be re-obtained omits `reopen` and its port reports `restarting: false` instead of pretending. I also corrected the code snippet in the earlier task's unreleased changeset (`.changeset/the-indexer-is-hosted-in-a-dedicated-worker.md`), since release notes teaching an API that will not exist is a defect a reader pays for.

**A death is concluded from SILENCE, by a probe on a timer.** New `ping` case; if the host has said nothing for `watch.everyInSeconds` (default 5) the port probes, and an unanswered probe within another interval is a death (so a death is noticed within ten seconds by default). This is liveness polling sitting next to ADR-0082's "status is PUSHED", which is why both the case and the watch say in their own docs that what is polled is LIVENESS and never STATUS. Alternatives considered and rejected: the `MessagePort` `close` event (not available on all three engines this package tests against); the `Worker` `error` event (an uncaught error is not a death — a worker survives it); concluding a death only from a call already in flight (an idle tab, or one resting at the tip, would never learn). `watch: false` is offered and documented as honest-but-silent: it is right only for a host that cannot die independently (the coming main-thread shape).

**The restart policy: 5 consecutive restarts, 0.5s doubling to 30s, and a host alive 60s is SETTLED.** The app is told on every death which attempt it is and whether a replacement is coming, so "this keeps happening" is renderable. Alternative considered: restart unboundedly (a worker that dies on boot becomes a hot loop and the app is never told), or reset the counter on any successful restart (same loop, slower). The settle window is what keeps the budget a crash-loop budget rather than a lifetime quota. All four are configurable; nothing else in the repo reads them.

**A call made between a death and the restart is REJECTED, not queued.** Same typed error as the calls that were in flight. Queueing would be the silent retry ADR-0082 refuses, one step later. Consequence a caller must expect: a waiting loop over `progress()` has to tolerate `IndexerHostDiedError` as "not yet" (`cut.ts`'s `asking` helper shows the shape).

**Kill first, then replace.** The port releases the access it is replacing (for a dedicated worker: `terminate()`) *before* it opens a successor, including when no restart is coming. So a host merely suspected of being dead costs a restart, never a second writer; the writer claim is the guarantee underneath rather than the mechanism. The node test asserts the ordering (`['open', 'close', 'open']`) rather than trusting it.

**`IndexerHostDiedError` is narrowed by `instanceof`, unlike every refusal on `PortError`.** Those are raised in the host and rebuilt from data, so their prototype cannot survive the crossing and the `name` is the narrowing. This one is raised in the tab about a host that is not there, so nothing crossed and the class is exact. The `name` is pinned all the same for code that narrows the whole surface one way.


wighawag:
APPROVE ✅

```
39 passed (28.8s)
✓ [chromium] browser/restartsAndResumes.spec.ts:61 › a terminated worker restarts and resumes from the cursor (822ms)
✓ [firefox]  ... (1.0s)
✓ [webkit]   ... (1.1s)
```

Green on all three engines, 36 pre-existing cases still passing.

**Per criterion:**

- **The app observes the death as an event, without polling** — met. The app polls nothing; the PORT probes liveness internally and hands the app `onHostDeath`. The Decisions block is right to flag that this sits next to ADR-0082's "status is PUSHED" and to insist in both the case's and the watch's own docs that what is polled is LIVENESS and never STATUS. The three rejected alternatives are each rejected for a real reason rather than a stylistic one: the `MessagePort` close event is not available on all three engines this package tests against, a `Worker` error event is not a death (a worker survives an uncaught error), and concluding a death only from a call already in flight would leave an idle tab, or one resting at the tip, never learning.
- **Every call in flight rejects with a typed error naming the cause; none hangs** — met. Rejecting rather than queueing calls made between the death and the restart is the correct reading of the same rule: queueing would be the silent retry ADR-0082 refuses, one step later. The consequence is documented where a caller meets it (a waiting loop over `progress()` must treat `IndexerHostDiedError` as "not yet").
- **Resume from the persisted cursor: no re-index from the start block, no skipped range, state matches an uninterrupted run** — met, and the browser case pins the resulting state to the same `EXPECTED_A` constant the uninterrupted path asserts, so it is an equality against a known-good result rather than internal consistency. The spec is also honest that a resume RE-asks for part of what it already had: overlap is not a skip, and saying so beats a test that quietly pretends the boundary is exact.
- **A termination deliberately induced MID-FOLD, in a real browser** — met, and this is where the case earns its keep. It terminates part way through APPLYING A BLOCK, not merely between cycles, and the JSDoc says why that distinction matters: a test that only terminated between cycles would look correct on a five-block fixture and fail an afternoon of real syncing. The case asserts two workers were built and that the first died mid-write.
- **Calls after the restart work on every surface that existed before** — met, as its own case rather than spot-checked on one surface.
- **A repeatedly dying worker does not loop unboundedly, and the app can tell** — met. The policy (5 consecutive restarts, 0.5s doubling to 30s, alive-60s means SETTLED) is sensible, and the settle window is the detail that makes the budget a crash-loop budget rather than a lifetime quota. The attempt number and whether a replacement is coming are reported on every death, so "this keeps happening" is renderable rather than merely true.
- **At no point are two hosts writing; the writer claim is held by exactly one** — met, and met by ORDERING rather than by hope: the port terminates what it is replacing before opening a successor, including when no restart is coming, and the node test asserts the sequence `['open', 'close', 'open']` rather than trusting it. The consequence is stated plainly and is the right trade: a host merely SUSPECTED of being dead costs a restart, never a second writer.
- **Changeset** — met.

**On the API change, which I checked because it reaches back into merged work.** `dedicatedWorkerHost(worker)` becoming `dedicatedWorkerHost(() => worker)` is the right call and is argued correctly: a shape IS "how a port is obtained", so obtaining one AGAIN after a death belongs there, and the rejected alternative (`Worker | (() => Worker)`) would have kept every call site compiling while leaving ports that silently cannot restart with nothing in the types saying so. Amending PR #134's still-unconsumed changeset to teach the new form is legitimate rather than tampering: that changeset has not been released, and shipping release notes that teach an API which will not exist is a defect a reader pays for. Both packages are unreleased on this surface, so nothing published is broken.

The forward-note for the two remaining shape tasks is exactly what a task at this position should leave: a shape that genuinely cannot be re-obtained omits `reopen`, and its port reports `restarting: false` rather than pretending.

**Also good:** `IndexerHostDiedError` is narrowed by `instanceof` while every `PortError` refusal is narrowed by `name`, and the asymmetry is explained rather than left to look like an inconsistency: the refusals are raised in the host and rebuilt from data so their prototype cannot survive the crossing, whereas this one is raised in the TAB about a host that is not there, so nothing crossed and the class is exact. Pinning `name` anyway for code that narrows the whole surface one way is the right belt-and-braces.

**CONTEXT.md:** the `indexer host` entry gains the death semantics and, valuably, separates a death from its three neighbours (a `HostProgress.failure` is a LIVE host whose driver stopped and which still answers reads; a close is the tab letting go, so nothing is reported and nothing restarts; a demotion is the opposite case, where the store moved on without the writer). The entry correctly RETAINS its `NOT YET BUILT` marker, which is `createindexerstate-becomes-the-main-thread-host`'s to remove.

**Vocabulary:** clean; no `body`/`shell`.

Merging.

PR #140, opened 2026-09-12
feat(a-sharedworker-serves-several-tabs-from-one-host): A SharedWorker…

wighawag:
Task: `work/tasks/done/a-sharedworker-serves-several-tabs-from-one-host.md`

## What I did

**Drift check first.** The task still matches reality: its blocker (`a-terminated-worker-restarts-and-resumes-from-the-cursor`) is in `work/tasks/done/`; ADR-0082 still names dedicated the default and shared the opt-in; `HostingShape` already carried `'shared-worker'`; and the seam the first task's done record described is exactly where it said it was (`serveIndexerHost` takes a `HostAccess` = `{host, endpoint, close?, reopen?}` over a structural `MessageEndpoint`, whose doc already anticipated "a `MessagePort`, which is what a SharedWorker hands each tab"). One path in the launch prompt is stale (the spec is at `work/specs/tasked/…`, not `ready/`); nothing else. No STOP.

**The shape.** `packages/browser/src/host/sharedWorker.ts` (new, the only new production file):
- `sharedWorkerHost(() => new SharedWorker(url, {type:'module', name}))` for the tab: `endpoint` is `worker.port`, `close()` closes only that tab's port, `reopen()` asks the app's factory again.
- `hostIndexerInThisSharedWorker(spec)` for the entry point, taking the *same* spec the dedicated helper takes.
- `oneEndpointOverEveryClient()`, which presents the per-client wires to the host as the one endpoint every shape gives it.

**Nothing inside the host changed.** `serve.ts`, `port.ts`, `envelope.ts`, `endpoint.ts`, `reads.ts`, `progress.ts`, `restart.ts`, `clone.ts`, `errors.ts` are untouched (`git diff` shows only `host/index.ts`, `host/dedicatedWorker.ts` and the test-harness `cut.ts`). The two host-directory edits are an export + module doc, and one extra REFUSAL branch in the dedicated helper (see Decisions). The strongest evidence is in the browser run: one built entry file is loaded once as a `Worker` and once as a `SharedWorker`, and one piece of app code run against both ports reports identical values except `host` and `scope`.

**Tests** (test-first: the node file was written and failing 11/11 before any production code existed):
- `packages/browser/test/aSharedWorkerServesSeveralTabs.test.ts` (11 cases, in the gate): one fold for two clients over real `MessagePort`s, colliding correlation ids answered per client and counted *on the wire*, pushes only to subscribers, one client leaving mid-fold, a later host resuming from the cursor, the same port surface as the dedicated shape, and five refusals.
- `packages/browser/browser/sharedWorkerServesSeveralTabs.spec.ts` (3 tests × 3 engines, outside the gate per this package's convention): two real pages on one real SharedWorker, the lifecycle (one tab closed mid-fold, then the last), and one entry serving both shapes. All 9 pass on Chromium, Firefox and WebKit; the whole browser suite is 48/48.
- Property verified rather than built: same script URL + different `name` is a different host (and folds elsewhere, since the fixture derives its database from `self.name`); a different query string is too. Also observed: all three engines support module shared workers, none accepts the ESM bundle as a classic script, and all three end the worker when its last client goes.

**Gate:** `pnpm format:check && check:adr && check:refs && check:changesets && build && typecheck && test` exits 0. `changeset status --since=main` passes (verified in a throwaway clone, since changesets reads git and my changeset is still untracked). Changeset added. Working tree holds only the intended changes.

**Durable artifact:** `docs/spikes/a-sharedworker-serves-several-tabs-from-one-host/` (README + per-engine result JSON written by the spec), following the IndexedDB multi-tab precedent.

**Out-of-scope signal captured:** `work/notes/observations/adr-0082-still-says-not-yet-implemented.md` (ADR-0082's status line and `CONTEXT.md`'s "NOT YET BUILT" are now false for everything but the last task; I did not touch either, because the spec's banner assigns those docs to `createindexerstate-becomes-the-main-thread-host`).

No election was built and none is assumed. The dedicated shape is untouched: there is no shape *setting* to default, because a shape is a different call (that was the first task's decision), so "dedicated remains the default" is preserved by every pre-existing test passing unchanged and by the docs still leading with it.

## Decisions

**Nothing inside the host had to change, and the one structural addition is in the shape file.** A SharedWorker's entry is handed a wire per CLIENT; the host is written against one wire. `oneEndpointOverEveryClient()` (in `sharedWorker.ts`) makes N wires look like the one every shape provides, so `serve.ts` is byte-identical between the shapes. Alternative considered and rejected: teaching the host to hold several endpoints (that is a second implementation of the host, which is the accident ADR-0082 exists to prevent). Touches every later task on this port: a new case needs no change here, because the fan-in is generic over the envelope.

**Correlation ids are RE-NUMBERED per client, and that is the part that was not optional.** The envelope says an `id` is unique per port and not globally, which is exactly true and becomes load-bearing with several ports: two tabs are two documents each counting from 1, so ids collide by construction. A broadcast wire would have *resolved* one tab's `progress()` with another tab's rows, so each request gets an id in one space on the way in and the answer is re-numbered back and posted to the single client that asked. This is the evidence about the seam a later reader needs: the shared shape did not need the host changed, but it did need this, and a third shape with several clients would need it too. Node test `answers two tabs whose correlation ids collide` asserts it on the wire, not on the shapes of the values.

**Pushes are filtered in the shape, which puts two case names inside a shape file.** The host counts subscriptions in aggregate and posts when the count is non-zero, so WHICH clients asked can only be known here; the fan-in watches `subscribeToProgress`/`unsubscribeFromProgress` and posts a push only to subscribed clients. Alternative: broadcast and let an unsubscribed tab ignore what arrives (rejected: it bills every tab for the one rendering a progress bar, and it would break the host's own rule that nothing is posted to a tab that did not ask). This is the only protocol knowledge in the file. Touches any later PUSH added to `PortPushes`: it will need its subscribe/unsubscribe case names added to this filter, which the module doc says.

**A missing `SharedWorker` is a REFUSAL, not a silent fallback to dedicated.** `sharedWorkerHost` throws before calling the app's factory, naming what is missing and pointing at `dedicatedWorkerHost`. This is a new user-visible error. Alternative considered: falling back automatically (rejected: the shape decides how many writers an app has and whether an election is needed, so swapping it silently changes the deployment under the app; a fallback ladder is `one-tab-indexes-and-the-others-read`'s decision). No predicate is exported either: `typeof SharedWorker === 'undefined'` is the whole check and duplicating it as API would be a second way to ask. The async failure a runtime with `SharedWorker` but no MODULE support produces is logged by the shape and concluded as a death by the port's existing watch; no new death cause was added.

**A second entry helper, and a new cross-refusal in the DEDICATED one.** `hostIndexerInThisSharedWorker` is a sibling of `hostIndexerInThisWorker` rather than scope auto-detection inside one helper, so each shape's file holds both of its ends and the `host` label a tab is told stays a consequence of where the code ran. The cost is that calling the wrong one is possible, and it is SILENT (a shared scope has no `postMessage`; a dedicated one never fires `connect`), so `hostIndexerInThisWorker` now refuses a shared scope by name, pointing at the sibling. That is the one behaviour change to an existing file, and it converts a host-that-never-answers into a sentence. Scope detection is `'onconnect' in globalThis`, measured across all three engines (true only in `SharedWorkerGlobalScope`) rather than `instanceof`, so it is also stateable in a node test.

**No word was coined.** CLIENT is the platform's own word for a connected page (and the task's), HOST / CONTAINER / PORT / HOSTING SHAPE are the glossary's; `body`/`shell` were not reintroduced. The fan-in is described by what it does ("every attached client's wire, presented as one endpoint") rather than given a new noun, and `CONTEXT.md` was deliberately not edited: its **indexer host** entry already describes the shared shape and the URL-plus-name property, and the spec's banner assigns those entries to the final task.

**`close()` cannot take a shared host down, and the restart path differs in a way that is safe.** The dedicated port terminates a suspected-dead host before opening a successor, so two writers are impossible. A `SharedWorker` has no `terminate()` and must not be taken down by one of its tabs, so `reopen()` reconnects to the instance that is still running where there is one. That is safe precisely because a SharedWorker is a singleton by construction (its one prize), so there is no second host to create. Documented at the call site; nothing was added to `HostDeathCause`, whose single member still reflects what a tab can actually observe.

**Test-fixture shape: the harness entry is deliberately BOTH entries.** `browser/indexer.bothShapes.worker.ts` branches on its own scope and calls one of the two helpers, because the harness builds exactly one worker entry per mount, and because it makes "what runs inside the host is unchanged" a fact about ONE built file rather than a comparison of two. An app writes one of the two lines and never branches. The instance identity the spec asserts on is the fixture's own value posted off the port; no `whichHostAreYou` case was added to the envelope, because an app has no use for the answer.


wighawag:
APPROVE ✅

```
48 passed (38.1s)
✓ [chromium/firefox/webkit] sharedWorkerServesSeveralTabs.spec.ts:87  › two tabs attached to one SharedWorker read one fold, and both are pushed its progress
✓ [chromium/firefox/webkit] sharedWorkerServesSeveralTabs.spec.ts:197 › a SharedWorker outlives one tab, and the tab after the last one resumes from the cursor
✓ [chromium/firefox/webkit] sharedWorkerServesSeveralTabs.spec.ts:294 › one entry point serves both hosting shapes, and app code cannot tell them apart
```

All three cases green on all three engines, 39 pre-existing cases still passing.

**This task also RECORDED its results, and did it properly.** Nine JSON artifacts under `docs/spikes/a-sharedworker-serves-several-tabs-from-one-host/results/`, one per case per engine, which is the convention the IndexedDB package already established. That is a real improvement on the earlier tasks in this drive, whose browser evidence was reproducible-on-demand but not committed. The artifacts are also the best evidence in the PR: in `two-tabs-webkit.json` both tabs report the SAME host `instance` (`jdbjlh`) with a third instance distinct from it, which is "exactly one fold is running" as recorded DATA rather than as an assertion I have to trust.

**Per criterion:**

- **The app selects the shape at construction, and app code is byte-identical between shapes** — met, with a dedicated real-browser case (`app code cannot tell them apart`) and a node case (`hands a tab the same surface the dedicated shape hands it`). The recorded `portSurface` arrays are identical across both tabs and list all twelve surfaces this drive built.
- **Two tabs read correct state, both receive progress, exactly one fold** — met, and demonstrated by the shared `instance` id above rather than inferred.
- **Closing one attached tab does not stop the fold for the other** — met.
- **Closing the last tab leaves a consistent store, and the next tab resumes rather than re-indexes** — met.
- **What runs inside the host is unchanged; a reviewer can see only port acquisition differs** — met, and this is the criterion I checked hardest because it is the one ADR-0082 exists to protect. It is demonstrated STRUCTURALLY rather than argued: `src/host/serve.ts` and `src/host/port.ts` are NOT in this diff at all. The shape arrived as one new file, `sharedWorker.ts`, plus its export. The seam established in PR #134 held under exactly the load it was built for, and no second copy of the host appeared.
- **The dedicated shape remains the default when nothing is specified** — met, and consistent with the explicit-selection design: there is no implicit branch to get wrong, and the dedicated shape is named as the default in both docs and the refusal path.
- **A runtime without SharedWorker is told, rather than failing obscurely** — met, and the message is the best-written refusal in this drive. It names the runtime and the scope, gives the exact remedy as copyable code, states that the dedicated shape is the default and works everywhere, and explains why there is deliberately NO automatic fallback: "the shape decides how many writers an app has, so it is the app's choice to make". A silent fallback here would quietly change an app's writer topology, which is precisely the class of bug the writer/reader split exists to prevent.
- **Changeset** — met.

The node case `answers two tabs whose correlation ids collide, each with its own answer and nobody else` deserves a mention: two independent tabs generating overlapping ids is a real hazard the moment one host serves several of them, and it would have been easy to miss until it produced a baffling cross-talk bug in the field.

**Off-path finding, correctly captured and NOT acted on:** `adr-0082-still-says-not-yet-implemented` notes that both ADR-0082's status line and `CONTEXT.md`'s `NOT YET BUILT` marker are now false for everything except the one remaining task. Leaving both alone is right, and the note's only inaccuracy is minor and self-correcting: it says the ADR's status line "is not assigned to anyone", when in fact `createindexerstate-becomes-the-main-thread-host` carries removing it as an explicit acceptance criterion. That task is next and will close both lines together, so the note resolves itself on the next merge.

**Vocabulary:** clean; HOST, PORT, HOSTING SHAPE used as pinned, no `body`/`shell`.

Merging.

PR #141, opened 2026-09-12
feat(createindexerstate-becomes-the-main-thread-host): createIndexerSta…

wighawag:
Task: `work/tasks/done/createindexerstate-becomes-the-main-thread-host.md`

All green: `pnpm format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test` (267 browser tests, whole monorepo suite), plus `pnpm --filter @etherfold/browser test:browser` (51 Playwright tests across Chromium, Firefox and WebKit).

## What I did

**Drift check first.** ADR-0082 still carried `status: accepted, not yet implemented` and its `createIndexerState` paragraph; `CONTEXT.md`'s **indexer host** entry was still marked NOT YET BUILT; the README still led with the main-thread hook; all seven prior tasks of the spec were done and had left `createIndexerState` alone. No stale premise, so I built.

**The third shape.** `createIndexerState` keeps its name, its arguments and every verb it returned; it gained one method, `mainThreadHost()`, which joins a wire to the host that is already there. `src/host/mainThread.ts` is the shape's file (a sibling of `dedicatedWorker.ts` / `sharedWorker.ts`) and constructs no indexer at all.

**One implementation, made structural.** Everything a tab can observe — the case dispatch, the row projection, the derived progress figures, the clone refusals, the subscribe/unsubscribe bookkeeping and the "an unchanged report is not posted" rule — moved out of `serve.ts` into a new `src/host/cases.ts` behind a nine-question `HostBacking`. All three shapes reach it; `serve.ts` and `createIndexerState` each implement the backing.

**The suite.** `browser/hostingShapes.ts` holds 16 behaviour cases as DATA (the shape `@etherfold/state-store-conformance` and `readWorkload.ts` use). `browser/threeHostingShapes.spec.ts` runs that one list against a dedicated worker, a SharedWorker and the main thread in one real page, on three engines; `test/theThreeHostingShapesRunOneImplementation.test.ts` runs the same list against the main-thread shape under vitest + `fake-indexeddb` on every commit, plus seven main-thread-specific claims.

**Docs.** The README now leads with the dedicated worker, tables the three shapes and says the main-thread cost plainly. `CONTEXT.md`'s **indexer host**, **tx inclusion** and **processor kind** entries were rewritten in place. ADR-0082's frontmatter is gone. Changeset added. Two observation notes captured (a browser run dirties committed spike results; a `failure` outliving a restarted driver on both hosts).

## Decisions

**`createIndexerState` keeps its exact name and signature, and the shape is reached as `indexer.mainThreadHost()`.** Alternatives considered: (a) a top-level `mainThreadHost(spec)` symmetric with `dedicatedWorkerHost` / `sharedWorkerHost` — rejected, it is precisely the second main-thread constructor ADR-0082 closes, and an app that used both would open two containers over one store and meet a writer refusal three layers from the line that caused it; (b) a hosting argument on `createIndexerState` (the ADR's other option) — rejected, there is no second value it could take, because the hook's three reactive stores publish a live read HANDLE that cannot cross a port, so a worker-backed `createIndexerState` is unbuildable without deleting `state`. The consequence to weigh: the third shape's construction is ASYMMETRIC with the other two (they take a factory, this is a method). That is the honest shape of "the host here already exists", and the port code after it is identical, which is the criterion that matters. Touches every later task that adds a shape or a port case.

**The three shapes do NOT share one implementation as cleanly as ADR-0082 assumes, and this is the finding worth reading.** They share the whole BOUNDARY (now one module, `src/host/cases.ts`) and they do not share the DRIVER: the worker shapes run `serveIndexerHost`'s loop, the main-thread host runs the hook's auto-index loop with its four verbs. I considered rebuilding `createIndexerState` on top of `serveIndexerHost` so there were literally two files. Rejected, and the reasons are structural rather than effort: `serveIndexerHost` takes its provider/source at CONSTRUCTION and starts folding immediately, while the hook defers them to `init` (which every existing caller, example and test depends on); and the hook carries seven surfaces the worker host does not have and does not want (the stream-seed install, the scheduled prune, demotion, `reset`, `promote`, `updateProcessor`, `updateIndexer`, the `createIndexer` injection point, `trackNumRequests`) — folding them into the worker driver would move a large amount of main-thread-only machinery into every worker bundle. So "one implementation" here precisely means: one implementation of everything that crosses the port, over two drivers. `cases.ts`'s module docstring and the **indexer host** glossary entry both say this in those words rather than letting the ADR's sentence stand unqualified. Touches any later task that adds a port case (it needs one edit) versus one that changes driver behaviour (it needs two).

**How the two runners were handled, since the shapes cannot all be driven by one.** The cases are data in `browser/hostingShapes.ts`, run three times in the Playwright page (the only place a `Worker` and a `SharedWorker` exist) and once under vitest against the main-thread shape. I deliberately did not simulate the worker shapes in node: `test/utils/port.ts` already serves the worker driver over a local `MessageChannel`, and running the suite over that would have looked like three shapes while being two. Alternative considered: assert only in Playwright — rejected, the acceptance gate has no browser binaries, so the list would be exercised by nothing on a normal commit.

**The main-thread wire is a real `MessageChannel`, not direct calls.** A hosting shape is how a port is obtained, and the port's contract includes what may cross it. Direct calls would make the shape an app develops against more permissive than the shape it ships (`assertClonable` would never fire, a class instance would sail through). The cost is one structured clone per call on a path that has no thread boundary to justify it; that is the right trade for a shape whose main job is tests and small backfills. Touches nothing else.

**`stopAutoIndexing()` now also stops an already-running cycle from re-arming the loop.** It previously cleared only the pending timer, so a stop landing mid-cycle was undone by that cycle's own re-arm. This is a real behaviour change to an existing verb and it was forced: `IndexerPort.stopIndexing()` promises no chain request is made after it answers, and the shared suite's stop case fails on the main thread without it. Alternative considered: keep the bug and weaken the suite's stop case — rejected, that is the "three test files that agree" failure mode. Called out in the changeset. Touches anything relying on `stopAutoIndexing` being timer-only; nothing in the repo did.

**`serveIndexerHost` stays exported, and `test/utils/port.ts` keeps labelling itself `main-thread`.** A reviewer grepping exports for "a second way to build a main-thread indexer" will see `serveIndexerHost`, so it now says in its own docstring that it is the WORKER hosts' driver, that it is reached from the two entry helpers (both of which refuse to run in a document), that it is exported so a deployment can write a shape this package does not ship, and that calling it on the UI thread would be the second container. `test/utils/port.ts`'s `host: 'main-thread'` label is still accurate about the CONTEXT the driver runs in, but it no longer names the main-thread SHAPE, so its docstring now says which is which and points at the new test. Alternative considered: unexport `serveIndexerHost` — rejected, it forecloses a custom hosting shape (an iframe, a `MessagePort` somebody else owns) for a naming worry the docstring answers.

**`mainThreadHost()` may be called more than once, its access carries no `reopen`, and `{watch: false}` is documented rather than forced.** Several wires to one host is what a SharedWorker already does for several tabs, so it needed no new rule. No `reopen` because a host on this thread cannot die independently of the tab holding the port — the restart task had already flagged `watch: false` as "right only for a host that cannot die independently (the coming main-thread shape)". I did not make `connectToIndexerHost` default `watch` off when an access has no `reopen`: that would silently change the SharedWorker path's future behaviour, and the port options are the app's to pass.

**`dispose()` does not close the wires it handed out.** A port belongs to whoever obtained it, exactly as it does on the other two shapes where `close()` is the sole owner of the wire. So a dispose resets the port's view instead: the phase goes back to `waiting`, a read that was waiting for a store this container will now never build is REJECTED rather than left hanging, and a later `init` gives the same wire a fresh container to answer from. Alternative considered: close the wires on dispose — rejected, it would break the documented "after `dispose()`, `init(...)` may be called again" contract for an app that also holds a port, and it would surface an ordinary teardown to the app as an `onHostDeath`.

**`packages/browser`'s `package.json` description was updated.** It described only the main-thread hook, which is now one of three shapes; it is published metadata, so it is a user-visible string and the changeset covers it.


wighawag:
APPROVE ✅

Gate-3 (conductor's diff-vs-criteria review). Gate-1 green; Gate-2 off for this drive by configuration. This is the TERMINAL task of the spec, so it carries three obligations beyond its code, and I checked each of them explicitly.

**The browser criterion, verified rather than assumed** — re-ran `test:browser` off this branch: **51 passed** across chromium, firefox and webkit (48 before this task, plus the new three-shapes case on each engine).

**The three terminal obligations:**

1. **Exactly ONE way to construct a main-thread browser indexer** — met, and verifiable the way the criterion asks (by grepping exports). `createIndexerState` is the only exported main-thread constructor; the third shape is reached as `indexer.mainThreadHost()`, a METHOD on the thing you already have. `host/mainThread.ts` and `host/cases.ts` are deliberately not re-exported, with a comment at the export site saying why: "a second exported way to serve a container is exactly what ADR-0082 closes". This is the criterion's real intent rather than its letter, and the temptation it resists (a tidy top-level `mainThreadHost(spec)` factory beside the other two) is named and rejected in the code with the right reason: it would be a second way to build a main-thread indexer with no rule for choosing between them.

2. **`CONTEXT.md` updated** — met on all three counts the task names. The `indexer host` entry loses `NOT YET BUILT`, and the `tx inclusion` and `processor kind` entries now describe the browser surface as the three shapes rather than as one hook. Those are exactly the entries that would have started lying the moment this merged.

3. **ADR-0082's `status: accepted, not yet implemented` removed** — met, and removed correctly: the whole frontmatter block goes, matching the convention confirmed earlier in this drive on ADR-0081 (an implemented ADR carries no frontmatter; only a non-default status gets one).

**Per remaining criterion:**

- **The main thread is a named, selectable shape and `createIndexerState` IS it** — met, adapted rather than wrapped, which is what the task and ADR-0082 both required.
- **Existing callers keep working, or the break is deliberate and documented** — met on the easier branch: nothing breaks. The name, the arguments and every returned verb are unchanged; the returned object GAINED `mainThreadHost()`. An app that never wants a port sees none of this.
- **One test exercises the SAME behaviour across all three shapes and passes on each** — met, and this is the best-constructed test in the drive. One shared `hostingShapeCases` list is consumed by the node suite and by `threeHostingShapes.spec.ts`, which runs all three shapes in a real browser, asserts `passed === hostingShapeCases.length` per shape, and reports failures PER SHAPE (`expect(\`${shape}: ${JSON.stringify(failures)}\`)`) so a divergence names itself instead of surfacing as one opaque red suite. It then pins the one thing the shapes legitimately differ in: `DedicatedWorkerGlobalScope`, `SharedWorkerGlobalScope`, `Window`. The main-thread shape honestly reporting `Window` is the whole cost of that shape, stated as an assertion.
- **App code against the port is unchanged across the shapes** — met, asserted rather than claimed.
- **The guide documents the three shapes, leads with the dedicated worker, and says what the main thread costs** — met: `## Where the indexer runs: three hosting shapes`, then `### The default: a dedicated worker`, with the main-thread section stating plainly that the fold is on the UI thread.
- **Changeset** — met. Updating the published package DESCRIPTION as well is a good catch: it described the old single-shape reality and would have shipped stale to npm.

**Two off-path findings, correctly captured and not fixed:**

- `a-host-failure-outlives-a-restarted-driver`: a driver that stopped on a non-retryable refusal records `failure`, and a later `startIndexing()` moves the phase on without clearing it, so a tab can be handed `phase: 'at-tip'` with a stale failure attached. The judgement to leave BOTH drivers matching rather than fix one side is right and is stated as the reason: diverging the two would be worse than the wart.
- `a-browser-run-dirties-committed-spike-results`: a full `test:browser` run rewrites the nine result JSONs committed by the previous task, because `ranAt` and the random `instance` ids change every run, so any browser run leaves an unrelated dirty tree that a later `git add -A` would sweep into an unrelated commit. This one is notable because the drive CAUSED it: it is a direct consequence of PR #140's otherwise-excellent decision to commit per-engine results. This PR correctly reverted the churn rather than committing it (the result files are absent from this diff). I am carrying it to the end-of-drive triage as task-worthy.

**Non-blocking nit for triage:** these two notes omit the YAML frontmatter (`title`/`slug`/`observed`) that the drive's four earlier observation notes carry. The work contract requires frontmatter `source:` only for FINDINGS, so this breaks no rule, but it is an inconsistency inside one folder and worth normalising.

**Vocabulary:** clean; no `body`/`shell`.

This closes ADR-0082 and the eight-task worker spec. Merging.

PR #142, opened 2026-09-13
feat(the-fold-publishes-what-it-just-changed): The fold publishes what…

wighawag:
Task: `work/tasks/done/the-fold-publishes-what-it-just-changed.md`

## What I built

The chain-facing producer of ADR-0083's signal: the side that applied a block now tells the sides that are reading, in one shape every later transport carries unchanged.

**Drift check first.** The task's premises hold against the code today: `EventProcessor.process` still returns an opaque `ProcessResultType`, the word *mutation* still does not occur in `@etherfold/core`, `Indexer` still filters its per-generation callbacks on `entry === this.current`, `VersionedStateEventProcessor` still builds its inner fold lazily on first use, and `updateProcessor` still swaps in place. ADR-0083 is unsuperseded. Nothing was stale, so I built it. (One non-blocking logistics note: the task body is at `work/tasks/backlog/the-fold-publishes-what-it-just-changed.md`, not `work/tasks/ready/` as the launch prompt says — the runner's done-move should account for that.)

**Production changes**

- `packages/core/src/stateMoved.ts` (new): `StateMoved` (`{block, coherence, entities, generation}`), `StateMovedHandler`/`StateMovedDetach` (the app-facing plain callback), `coherenceToken()` and `StateMovedPublisher` — the one assembly both containers will publish from: subscribe/detach, a throwing handler caught and logged, a `rotate(reason)` hook for the two later rotations, and nothing held per subscriber.
- `packages/core/src/types.ts`: `AppliedBlock` / `AppliedBlockReporter` plus ONE optional additive member, `EventProcessor.setAppliedBlockReporter`. `process` is untouched; no existing signature changed.
- `packages/core/src/container.ts`: `Indexer.onStateMoved(handler)` returning the detach; the relay attached to every generation and filtered at publication; re-attached across `updateProcessor`; the signal assembled with `generationDigestOf(entry.record)`.
- `packages/processor-entities`: `applyEventStream` takes an optional reporter and reports each block *after* `applyBlock` returned, with the entity names its mutations carried (deduplicated, sorted); `EntityEventProcessor.setAppliedBlockReporter` wires it.
- `packages/processor-sqlite`: the wrapper holds the reporter and re-applies it in `folded()`, so an attach before the inner fold exists still lands.
- `CONTEXT.md` gained a glossary entry for the signal and its coherence token; ADR-0083's status line now says what is built and what is not.

**Tests** (written at the agreed seam): `packages/conformance-workload-stratagems/test/theFoldPublishesWhatItJustChanged.test.ts` drives the container/indexer publication boundary with an in-process subscriber over the real stratagems capture — 9 applied blocks, 4 empty sets, `["commitment"]` on four, and `["cell","cellOwner","commitment","computedPoints","placement","placementPlayer"]` on the last, each compared against a second run of the same handlers. `packages/core/test/theFoldPublishesWhatItJustChanged.test.ts` covers the container claims (one per applied block, one unchanged token, opacity, follower silence, reconfigure survival, subscribe/unsubscribe symmetry and producer shape, a throwing handler, silence for a fold that reports nothing). `packages/processor-entities/test/applied-block-report.test.ts` and `packages/processor-sqlite/test/applied-block-relay.test.ts` pin the producing layer and the wrapper forwarding. A changeset accompanies the change.

Gate: `format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test` all green (`changeset status --since=main` only sees a changeset once it is tracked, which the runner's commit does).

## Decisions

**1. The shape of the cross-package channel: one optional method on `EventProcessor`, carrying `AppliedBlock` per block.** `setAppliedBlockReporter(reporter | undefined)` is a single owner-set slot on the processor, and `AppliedBlock` is `{block, entities}` — per BLOCK, not per batch, because one notification per applied block is the contract and only the layer below knows which blocks it applied. Alternatives considered: widening `process` (ruled out by the task and ADR-0083); a separate structurally-detected interface plus a type guard (more machinery, no extra safety, and it hides the channel from the seam it belongs to); a settable callback FIELD rather than a method (a wrapper would have needed a getter/setter pair anyway, and a method makes the forwarding obligation legible). Precedent for the optional-method form is `ExistingStream.setStreamConfig?`. It touches `@etherfold/core` (seam + relay), `@etherfold/processor-entities` (producer), `@etherfold/processor-sqlite` (forwarding), and every downstream task that consumes the signal: `the-receiving-container-publishes-what-it-applied` reuses `StateMovedPublisher` and this same channel, and the three transport tasks adapt `StateMoved`/`StateMovedHandler`. Its known cost, stated at the seam: being optional, a wrapper that forgets to forward it fails silently — which is why the SQL wrapper's forwarding has a test whose only job is that.

**2. A processor that implements no channel publishes NOTHING; a processor with no entity DECLARATIONS publishes an EMPTY set.** ADR-0083's "an empty set is the honest answer there" is implemented as the second of those: an entity fold declaring no entities reports `entities: []` per applied block. The first case (a custom `EventProcessor` outside the entity path) gets silence, because core cannot know which blocks such a fold applied and inventing them from the delivered stream would be exactly the fabrication the ADR rejects — and it would put a second publication site in core, which is how two producers drift. Alternative considered: a core-side fallback publishing one empty-set notification per delivered block for non-reporting processors. Rejected for those two reasons; the entity path is the only authoring path (ADR-0037), so the reachable case is covered. It touches any third party writing a bespoke `EventProcessor`.

**3. A block that touched nothing PUBLISHES, with an empty entity set.** The rule is "one notification per APPLIED block", not "per block that changed something": the block was recorded and the cursor moved with it, and a reader's two-line rule stays one rule. A batch that applied NO block (a scanned range with no logs) publishes nothing — there is no block to name. Documented on `StateMoved.entities`, on `AppliedBlock`, and as rule 5 of `applyEventStream`.

**4. The publication fires AS THE BLOCK LANDS — inside `process()`, per block — therefore BEFORE the batch's `onStateUpdated`.** The block is durable when it is reported (block + cursor are one atomic unit behind the storage seam, ADR-0027), so a reader that re-reads the instant it is told sees that block's effects. The alternative — holding reports until the batch's state notification — would BUFFER, which is the one thing the producer must not do, and would collapse a batch's blocks into a single moment for no reader's benefit. Consequence to know: a `process` that throws mid-batch has already published the blocks that landed, which is true rather than optimistic.

**5. The relay is attached BEFORE `updateProcessor` delegates, so the rebuild inside the verb publishes.** A swap is followed by a `load` that replays the cached stream through the new fold; a relay attached afterwards would go quiet for exactly those blocks. Attaching to a processor whose swap is then declined is harmless (it folds nothing) and it is detached on that branch. Consequence: a reconfigure that rebuilds over a cached stream publishes one notification per replayed block. That is the canonical fold genuinely re-applying them, and with no rotation on a discard yet, narrow invalidation per block is the correct advice for a reader. Touches `the-promotion-policy`-adjacent work and the retraction/promotion rotation tasks, which may want to rotate on a discard too.

**6. `Indexer.onStateMoved(handler): () => void` — an `on*` name that SUBSCRIBES rather than an assignable callback slot.** The container's other `on*` members are single assignable fields, so this is the same prefix with a different shape; I chose it anyway because the acceptance criteria require many subscribers and symmetric detach, and because `@etherfold/browser`'s `IndexerPort.onHostDeath(listener): () => void` already establishes exactly this form in the repo. The alternative (`subscribeToStateMoved`) was rejected as a second convention for one idea. Touches all three transport tasks, which will adapt this name.

**7. The token is a per-producer random value, not a counter, and rotation is `StateMovedPublisher.rotate(reason)`.** A counter repeats across a restart, which is the one case where an unchanged token would be a lie (a rebuilt fold telling a reconnecting reader nothing is stale). Same construction as `@etherfold/state-store`'s writer token, deliberately duplicated rather than shared because core cannot depend on the storage seam (ADR-0016). `reason` is logged and never published — the token names nothing. `rotate` is protected on the container, so the retraction and promotion tasks each add one line at the point their reason occurs; there is deliberately no public getter for the current token yet, since no shipped caller needs one and a transport can carry the token it last relayed.

**8. `entities` is deduplicated and SORTED.** It is a set rendered as an array; sorting makes two runs of one block produce one payload and lets a transport or a test compare them as values. Touches every consumer that compares payloads.


wighawag:

Reviewed the diff against all 15 acceptance criteria. Gate 1 was green in a fresh worktree rebased on `main`; I did not re-run it. My pass was the judgement the gate cannot do, weighted onto the two places this task could have passed green while being silently wrong.

**The two silent-failure traps, both honoured:**

1. **Only the CANONICAL fold publishes.** `publishStateMoved` returns early unless `entry === this.current`, and it deliberately reuses the filter the per-generation callbacks in `add` already apply rather than inventing a second rule that could disagree. The relay is attached to every generation and filtered *at the publication*, which is the right choice: attaching only to the canonical one would need re-attaching at every promotion. The test is non-vacuous, which is what makes it worth having: it asserts `expect(follower.applied.length).toBeGreaterThan(0)` *before* asserting silence, so a follower that quietly applied nothing cannot make the test pass by accident.

2. **The relay survives the SQL wrapper's lazy inner fold.** `VersionedStateEventProcessor` holds the reporter in its own slot and re-applies it in `folded()` unconditionally (including `undefined`), following the same hold-and-re-apply shape `config` already uses. Since `folded()` runs on every `process()`, an attach *after* first use lands on the next call rather than being lost. Pinned by `packages/processor-sqlite/test/applied-block-relay.test.ts`, whose only job is that forwarding — the right response to a seam whose optionality means a forgetful wrapper fails silently.

**The seam:** the entity set is produced in `@etherfold/processor-entities` where the mutations are and relayed up; `process` is NOT widened. The only removed production lines in the whole diff are the ADR status line, one re-indented `applyEventStream` call and one import, so no signature reaching processor-sqlite, the CLI or the browser changed. `setAppliedBlockReporter?:` is optional and additive.

**Criteria spot-checked to named tests:** one notification per applied block with block/generation/entities; empty set for a block that touched nothing (and the documented rule that it *is* published, because "one per applied block" is one rule not two); token stable across N appends; token opaque and only compared; subscribe/unsubscribe symmetry with no per-subscriber growth; a throwing handler contained as `onStateUpdated` already contains one; reconfigure survival asserting both detach and re-attach; silence for a fold that reports nothing. The conformance test drives a real fold over the stratagems capture and compares entity names against a second run of the same handlers, so the names asserted are the ones a real processor produces.

**Decisions block:** present and substantive. The channel shape (one optional `EventProcessor.setAppliedBlockReporter` carrying a per-BLOCK `AppliedBlock`) is recorded with alternatives and their costs, including the honest statement that optionality means silent failure for a forgetful wrapper. That is exactly what the next three tasks need to build on, and it states its own known cost rather than hiding it.

Nothing here contradicts ADR-0083, and the publication-order note (fires inside `process()`, before the batch's `onStateUpdated`, safe because a block and its cursor are one atomic unit per ADR-0027) is reasoned rather than accidental.

Merging.

PR #143, opened 2026-09-13
feat(a-retraction-names-the-fork-point-it-withdrew): A retraction names…

wighawag:
Task: `work/tasks/done/a-retraction-names-the-fork-point-it-withdrew.md`

## What I did

Made a reorg a first-class case of the state-moved signal, and made a MISSED reorg self-correcting.

**Drift check first (no drift found).** `the-fold-publishes-what-it-just-changed` landed as described (commit `d5f10393`): `StateMoved`/`StateMovedPublisher`/`Indexer.onStateMoved` exist, `StateMovedPublisher.rotate(reason)` is there waiting for its first caller, the relay runs up from `@etherfold/processor-entities`, and ADR-0083's status line still says both rotations are unbuilt. `revertTo(keepUpTo)` still returns `void` at the seam, and `applyEventStream` still derives the fork point one line above it, exactly as the task assumes. (The task body was at `work/tasks/backlog/`, not `ready/` — a launcher path detail; I left every `work/` file alone apart from one observation note.)

**The seam, test first.** The failing test went in at the existing reorg machinery in core (`packages/core/test/aRetractionNamesTheForkPoint.test.ts`): the fake chain serves branch A, then branch B, and the engine derives the `removed` markers itself. All 7 cases were red before any production code moved.

**Production change:**
- `@etherfold/core`: `StateMoved` is now a discriminated union on `kind` — `StateApplied {kind:'applied', block, coherence, entities, generation}` and `StateRetracted {kind:'retracted', forkPoint, coherence, generation}`. `StateMovedPublisher.publishRetraction()` rotates the token *as it publishes*, so a retraction under the old token is unexpressible; the container filters both the publication and the rotation on `entry === this.current`, so a follower replaying a stored stream's reorg rotates nothing.
- The processor-seam channel now carries both cases: `FoldReport = AppliedBlock | Retraction`, `FoldReporter`, `EventProcessor.setFoldReporter?` (renamed from `setAppliedBlockReporter`; see Decisions).
- `@etherfold/processor-entities`: `applyEventStream` reports `{kind:'retracted', forkPoint}` after `revertTo` returned and before the replacements are applied. `@etherfold/processor-sqlite` forwards it, pinned by its own case.
- `StateStore`, `WritableStateStore` and `revertTo` are untouched; no backend and no `@etherfold/state-store-conformance` case changed (`git diff --stat` confirms). The conformance **workload** package gained a test, which is what the prompt asked for.

**Tests:** core (container semantics, the reader that drops the retraction, the control reader that ignores the token and keeps the dead branch, the non-canonical fold that rotates nothing); processor-entities (fork point derived once at the lowest removed block, a retraction with no replacement, nothing reported if `revertTo` throws); processor-sqlite (the wrapper forwards retractions); and `@etherfold/conformance-workload-stratagems`, which causes a real absence reorg over the committed capture through the real `EntityEventProcessor`, drops the retraction, and asserts a reader converges while the same reader without the token rule is still rendering rows from the abandoned branch — every stale row belonging to an entity the next append never named.

**Also:** changeset added; `CONTEXT.md`'s signal entry and ADR-0083's status/consequences updated to match what is built.

Gate: `format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test` all green. (`changeset status --since=main` cannot see an untracked changeset; it resolves at the runner's `git add -A`.)

One out-of-scope signal captured, not fixed: `work/notes/observations/check-refs-crashes-on-an-unstaged-rename.md`.

## Decisions

**1. The retraction is distinguished by a `kind` tag on BOTH cases, and `StateMoved` becomes a discriminated union.** `StateApplied {kind:'applied', ...}` | `StateRetracted {kind:'retracted', ...}`. The criterion forbids inferring the case from a field being absent, and `'block' in moved` is exactly that inference, so the tag goes on both arms rather than only on the new one: a reader that reads `block` off a retraction then fails to compile instead of rendering a number that means the opposite of what it thinks. Alternatives considered: a separate `onStateRetracted` channel (two attach conventions for the three transport tasks to adapt, against ADR-0083's "one notification model"); an optional `forkPoint` beside an optional `block` (the forbidden shape). Cost: the append payload gained a fifth key, so the previous task's `Object.keys(...)` shape assertions were updated. Touches every downstream consumer of `StateMoved` — the three transport tasks and `the-receiving-container-publishes-what-it-applied`.

**2. A retraction carries `{forkPoint, coherence, generation}` and deliberately NO entity set and NO block.** `forkPoint` rather than `block` because the number means "the highest block that still stands", not "a block that was applied", and reusing `block` would let a reader's `if (block <= rendered) ignore` silently drop a retraction. No entity set because the honest answer would have to come back out of `revertTo`, which returns `void` on all four backends (the constraint this task is fenced by), and because a rotated token already says invalidate everything — a narrower retraction would also invite a reader to repair its cache by hand, which is what goes wrong at the next reorg. `generation` is kept so a retraction says whose answers moved, symmetrically with an append.

**3. The rotation lives INSIDE `publishRetraction`, not at the container call site, and the retraction carries the NEW token.** ADR-0083 said "one line at the point where the reason occurs"; I put that line in the publisher so the reason and the rotation cannot be separated by a future caller — the receiving container (next task) reuses this assembly, and a second call site that forgot to rotate would silently reintroduce the exact bug this task exists to close. Carrying the new token (rather than rotating after publishing) means a reader that RECEIVED the retraction invalidates everything once and then holds precisely the token the following appends carry, so it goes straight back to narrow invalidation. `rotate(reason)` stays public for the promotion task. Touches `the-promotion-policy`'s rotation task and the receiving-container task.

**4. The seam's channel is RENAMED: `setAppliedBlockReporter` → `setFoldReporter`, `AppliedBlockReporter` → `FoldReporter`, payload `AppliedBlock` → `FoldReport = AppliedBlock | Retraction`.** The task says the fork point should ride the channel that exists; once it does, a member named "applied block reporter" means two things, which is the coherence failure CONTEXT.md's conventions tell me to fix rather than build around ("nothing is published, so a breaking change costs a changeset"). Alternative considered: keep the name and widen only the payload (rejected: the name would lie at the one seam third parties implement); a second optional method `setRetractionReporter` (rejected: two slots for one channel, two things a wrapper can silently forget to forward, and the task explicitly says to ride the existing one). It touches `@etherfold/core`, `@etherfold/processor-entities`, `@etherfold/processor-sqlite` and the previous task's done-record/changeset prose, which I did not rewrite: my changeset names the rename instead.

**5. A state DISCARD still does NOT rotate the token, and I did not make it.** The previous task's decision 5 flagged that a reconfigure's rebuild republishes every replayed block and suggested the rotation tasks might want to rotate on a discard too. ADR-0083 names exactly two rotation reasons (a retraction, a promotion); a discard is neither, it is user-visible, and it belongs to whoever owns the promotion/reconfigure rotation. Flagging rather than doing it, so the omission is a decision and not an oversight.

**6. The core world helper was extracted to `test/utils/stateMovedWorld.ts` and shared with the previous task's test.** Two files now drive the same container with the same fold, and `streamCacheWorld.ts` sets the precedent that a second copy of the fake is a second definition of what the layer below reports. The shared fold gained the revert-then-report behaviour so it mirrors `applyEventStream`; `appendsIn()` narrows the union in the old assertions and REFUSES a retraction rather than filtering it, so a fold that never reverts publishing one is a failure and not a skip.

**7. I did NOT rename the two test FILES (`applied-block-report.test.ts`, `applied-block-relay.test.ts`) although their contents now cover both halves.** I tried, and `pnpm check:refs` died with an unhandled `ENOENT`: it enumerates `git ls-files` (the index) and reads each path, so a working-tree rename the agent cannot stage crashes the gate. Reverting the renames costs only a slightly stale file name; risking a gate crash costs a bounce. The latent script bug is captured in `work/notes/observations/check-refs-crashes-on-an-unstaged-rename.md` rather than fixed here.


wighawag:

Reviewed against all 7 acceptance criteria, plus the seam constraint this task was explicitly fenced by.

**The fenced constraint holds.** `StateStore`, `WritableStateStore` and `revertTo` are untouched: no file under `packages/state-store*` and no `state-store-conformance` case appears in the diff. (The two `conformance-workload-stratagems` files are the stratagems *workload* test package, not the store conformance suite.) The temptation to widen `revertTo` to learn which rows moved was resisted, and decision 2 explains why the entity-level detail was not needed: a rotated token already says invalidate everything.

**The two halves, both real:**

1. **Type-level distinction.** `StateMoved` is now a discriminated union `StateApplied | StateRetracted` with `kind` on *both* arms, not just the new one — so a reader that reads `block` off a retraction fails to compile rather than rendering a number meaning the opposite of what it assumes. `forkPoint` is deliberately not called `block`, which forecloses a reader's `if (block <= rendered) ignore` silently dropping a retraction. That is the criterion's "rather than inferred from a field being absent" honoured in the strong form.

2. **The rotation is structural, not conventional.** Rotation lives *inside* `publishRetraction`, so "a retraction which forgot to rotate is unexpressible". This is better than the one-line-at-the-call-site the ADR suggested: the receiving container reuses this same assembly next task, and a second call site that forgot to rotate would silently reintroduce precisely the bug this task closes. Carrying the new token on the retraction (rather than rotating afterwards) also means a reader that *did* receive it returns straight to narrow invalidation.

**The convergence criterion is asserted the hard way.** The test drops the retraction in transit (`if (notification.kind === 'retracted') return; // LOST IN TRANSIT`) against a real rendering reader, and `ABANDONED_ENTITY` is chosen as an entity written *only* by the block the chain abandons — so "the reader converged" is an assertion about the token and not about luck. Critically, the next test is a negative control: the same reader with token-comparison disabled *does* keep the dead-branch row. That pair makes the property falsifiable rather than merely observed, and it also proves the first test is not vacuous.

**The reorg is caused, not simulated.** `world.chain.serve([...BRANCH_B], BRANCH_B_TIP)` over the existing `streamCacheWorld` machinery, with the engine deriving the `removed` markers itself. Nothing hand-builds a retraction message.

**Unprompted but correct:** a non-canonical fold replaying the same retraction does *not* rotate the token, with a test. That extends the previous task's canonical-only discipline to the new case, which is exactly where it would otherwise have leaked.

**On the seam rename** (`setAppliedBlockReporter` → `setFoldReporter`, `AppliedBlock` → `FoldReport = AppliedBlock | Retraction`): I checked this deliberately, since it renames a seam that landed only minutes ago. It is sound. Once the fork point rides the channel, a member named "applied block reporter" names two things, and the repo's own convention is to fix that rather than build around it while nothing is released. It stays ONE optional member, `process` is still not widened, and the SQL wrapper still holds-and-re-applies under the new name (`foldReporter` slot, re-applied in `folded()`), with its forwarding test intact — so the silent-empty-set trap remains closed.

**Decisions block:** thorough, and two entries deserve credit. Decision 5 *declines* to rotate on a state discard and says why (ADR-0083 names exactly two rotation reasons), turning a possible oversight into a recorded decision and leaving it to the promotion task. Decision 7 hit a genuine latent bug — `pnpm check:refs` dies with an unhandled `ENOENT` on a working-tree rename it cannot stage — and captured it in `work/notes/observations/check-refs-crashes-on-an-unstaged-rename.md` instead of fixing it off-path. Both are the right instinct.

Merging.

PR #144, opened 2026-09-13
feat(a-promotion-rotates-the-coherence-token): A promotion rotates the…

wighawag:
Task: `work/tasks/done/a-promotion-rotates-the-coherence-token.md`

Done. Here is what I built and verified.

## Drift check (first, before building)

The task's premises all hold against current reality. `a-retraction-names-the-fork-point-it-withdrew` landed (`work/tasks/done/`), `StateMovedPublisher.rotate(reason)` exists and is public precisely "for the promotion task", `StateMoved` already carries `generation`, and `container.ts` still fires `onPromoted` **before** `applyAtNotification` with the comment explaining why. ADR-0083's status still listed "the token's OTHER rotation (a promotion)" as not yet built. No drift, so I built it.

## What changed

**Test first** (`packages/core/test/aPromotionRotatesTheCoherenceToken.test.ts`, new, 7 cases). It drives two real generations over one stream through the existing machinery (`utils/stateMovedWorld.ts` + a successor added beside the live one, which follows the incumbent's stream and re-folds it), with a reader subscribed across the pointer move. I confirmed the four rotation-dependent cases go RED without the production line and green with it.

**Production** (`packages/core/src/container.ts`): one call in `movePointerTo`, inside the existing `superseded !== entry` guard, **after** the registry write and **before** both `onPromoted` and the state notification that applies the move:

```ts
this.stateMoved.rotate(`a pointer move: reads are answered by the generation {...} from here on`);
```

**Docs**: ADR-0083's status now records both rotations as built, plus an "As BUILT" paragraph on where it sits; `CONTEXT.md`'s state-moved-signal entry replaces "(a promotion's rotation is not built yet)" with the built description; `promote`, `movePointerTo` and `StateMovedPublisher.rotate` JSDoc updated.

**Changeset**: `.changeset/a-promotion-rotates-the-coherence-token.md` (`@etherfold/core: minor`).

Acceptance: token rotates on a promotion ✓; `generation` names the retired lineage before and the successor from the first notification after ✓; a reader re-reading on the changed token gets the named generation's answer (rows only the successor's fold produces) ✓; ordinary appends carry one token on both sides of the move ✓; one mechanism, no second event kind, nothing published at the move ✓; caused by a real promotion between two real generations ✓; changeset ✓.

Gate: `format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test` (full monorepo) all green. `pnpm changeset status --since=main` cannot see an untracked changeset (it reads `git diff`), so it resolves once the runner stages; I did no git.

## Decisions

**1. A promotion PUBLISHES nothing; it rotates the token the next notification carries.** The acceptance criterion is phrased as "the next notification after a pointer move carries a different one", and a pointer move has no block to name and no fold that applied anything, so an `'applied'` notification manufactured at the move would claim a block application that did not happen (and a `'promoted'` kind is the second kind the task explicitly forbids). Alternatives considered: publishing an `'applied'` naming the new canonical generation's cursor (rejected: `StateApplied.block` is documented as "the block that was just applied", and a reader's `if (block <= rendered) ignore` would drop it); a third union case (rejected by the task and ADR-0083). **Residual, stated rather than discovered**: on a chain that then goes quiet, an out-of-heap reader is not told until the next block moves. That is the same best-effort limit ADR-0083 already accepts, and in-heap readers are still told at once by `onStateUpdated`, which `applyAtNotification` fires. Touches every future transport task and `the-receiving-container-publishes-what-it-applied`, which inherit "a promotion is not an event".

**2. The rotation fires BEFORE `onPromoted` and before the state notification, and AFTER the registry write.** Both of those callbacks are a reader being told to re-read, so a rotation after either would let one notification's worth of questions about the new generation be answered under the retired one's token, which is the exact failure the task's prompt names. After the registry write, so a `moveCanonicalTo` that throws does not invalidate every reader's cache for a move that did not happen. Alternative considered: rotating at the top of `movePointerTo` (rejected: rotates on a failed move).

**3. It sits at the POINTER MOVE (`movePointerTo`), not at the `promote` verb, so a REVERT and a policy-driven promotion rotate too.** Coherence check: `promote` is the human/operator verb, while `movePointerTo` is the one place the *canonical pointer* moves (`immediate` at `add`, `on-catch-up` at `settlePromotion`, `manual` at `promote`, and the backwards move CONTEXT.md calls a *revert*). A reader can see nothing of a move except that a different fold answers, so gating the rotation on the direction or on which caller asked would make the token mean two different things depending on who moved the pointer. A case pins the automatic `on-catch-up` path so the claim is tested and not just asserted. Touches `theCanonicalPointerMovesBack` behaviour (a revert now rotates) and the promotion policy paths.

**4. A `promote` naming the generation that is already canonical rotates NOTHING.** It is inside the existing `superseded !== entry` guard, which is also what `onPromoted` uses, so the two agree. Nothing a reader holds became suspect, and rotating would charge every reader a full re-read for a lineage change that did not happen. It is a user-visible default (`promote` is a public verb), hence recorded rather than buried.

**5. The RECEIVING container is deliberately NOT given this rotation, and I did not add one.** It moves a pointer too (`ReceivingIndexer.movePointer`) but holds no `StateMovedPublisher` at all: publishing from it is `the-receiving-container-publishes-what-it-applied`'s task, per ADR-0083's status. Flagging it so the omission reads as a decision rather than an oversight: whoever builds that publisher owes the same one-line rotation at that pointer move.

**6. Two additive changes to the shared test world (`test/utils/stateMovedWorld.ts`), used by two other tasks' test files.** `reportingFold` gained a `rows` getter (two folds over one stream otherwise produce identical rows, so a read could not say which generation answered — the same problem `promotion.test.ts` solves by MARKING what a fold produces), and `openWorld` gained an optional `promotion` option defaulting to the existing `{policy: 'manual'}`. Both are additive; no existing case changed, and the whole core suite (1133 tests) is green.


wighawag:

Reviewed against all 8 acceptance criteria. The design trap on this task was criterion 5 (one mechanism, not a second event kind), and it was avoided in the strong form: no new `kind` literal appears anywhere in source, the union is unchanged, and a test asserts the pointer move publishes *nothing* while every notification stays `'applied'`. A promotion rotates the token the *next* notification carries, which is exactly what the criterion asked for.

**Criterion 3 is tested for real, not by proxy.** The reader is shown to be answered by the successor's fold using rows the retired generation could not have produced (`successorOnlyRows`), while the incumbent goes on folding beside it holding different rows. That is a genuine lineage assertion, which is the whole point of the criterion: a re-read after a promotion must not be silently served by the retired lineage.

**Criterion 4 is asserted on both sides of the move**, with the comment naming it as "the change most likely to break it" — one token across the appends before, one token across 108 and 110 after, and the two sets disjoint. The stability property the append case bought is not regressed.

**Placement is well reasoned.** The rotation sits in `movePointerTo`, the single place the canonical pointer moves, and fires before `onPromoted` and before the state notification (both of which tell a reader to re-read, so rotating after either would let one notification's worth of questions about the new generation be answered under the retired token) but *after* the registry write, so a `moveCanonicalTo` that throws does not invalidate every reader's cache for a move that never happened. Both halves of that ordering are the right call and the rejected alternative (rotating at the top of the verb) is recorded.

**Two judgements I checked and agree with:**

- *Rotating on every pointer move, including a revert and a policy-driven promotion, not on the `promote` verb* (decision 3). A reader can see nothing of a move except that a different fold answers, so gating on direction or on which caller asked would make the token mean different things depending on who moved the pointer. The automatic `on-catch-up` path has its own case, so the broader claim is tested rather than merely asserted.
- *No rotation when `promote` names the generation already canonical* (decision 4), inside the existing `superseded !== entry` guard so it agrees with `onPromoted`. Nothing a reader holds became suspect, and rotating would charge every reader a full re-read for a lineage change that did not happen.

**Honest residual, stated not discovered** (decision 1): on a chain that then goes quiet, an out-of-heap reader is not told until the next block moves. That is the best-effort limit ADR-0083 already accepts, and in-heap readers are told at once via `onStateUpdated`. Recording it beats discovering it later.

**Decision 5 is the kind of flag that earns its place:** the receiving container moves a pointer too (`ReceivingIndexer.movePointer`) but holds no `StateMovedPublisher` yet, so the rotation was deliberately not added there and the omission is recorded as a decision. I am carrying that forward as a review point for `the-receiving-container-publishes-what-it-applied`, which owes the same one-line rotation at its own pointer move.

Test-world changes are additive only (a `rows` getter so two folds over one stream can be told apart, and an optional `promotion` option defaulting to the existing `{policy: 'manual'}`), no existing case altered, full core suite green.

Merging.

PR #145, opened 2026-09-13
feat(a-tab-learns-the-state-moved-across-the-port): A tab learns the st…

wighawag:
Task: `work/tasks/done/a-tab-learns-the-state-moved-across-the-port.md`

All checks are green. Summary of what I did.

## What I built

**One more push on the existing port envelope, not a channel** (`@etherfold/browser`, `src/host/`):

- `envelope.ts`: `PortPushes` gains `stateMoved: StateMoved` (core's own type, carried unchanged), with `subscribeToStateMoved` / `unsubscribeFromStateMoved` beside the progress pair. Added `PORT_PUSH_SUBSCRIPTIONS`, which declares once which pair of cases turns which push on/off.
- `cases.ts` (the one place a case is served, reached by all three shapes): `HostBacking.onStateMoved`, a second subscription count kept apart from progress, and a post that composes/filters/coalesces nothing. The host's own subscription to the fold is taken on the first subscribe and released with the last.
- `serve.ts` (worker hosts) and `IndexerState.ts` (the main-thread host): a forwarder set plus one `container.onStateMoved`, attached as the container is built, so a tab that subscribes before the container exists hears the fold's first block; detached on dispose.
- `sharedWorker.ts`: the per-client push filter is now per push NAME, read from `PORT_PUSH_SUBSCRIPTIONS`, so a tab watching progress is not posted the signal and vice versa.
- `port.ts`: `IndexerPort.onStateMoved(listener)` returning the detach, restored across a host restart, released on `close`.
- The `progress` push, its helper and its cadence are untouched.

**Tests.** New `test/aTabLearnsTheStateMovedAcrossThePort.test.ts` (real `MessagePort`, gated chain, six cases: one notification per applied block with core's exact payload and the canonical generation's digest; silence at the tip while the driver keeps fetching; a real reorg producing a retraction with the rotated token the replacement blocks wear; unsubscribe going quiet on the wire; progress asserted unchanged; a foreign message still ignored). Two cases added to the parameterised hosting-shapes suite, which now runs green in a real browser against a real dedicated worker, a real SharedWorker and the main thread on Chromium, Firefox and WebKit, plus under vitest on every commit.

**Standing conventions.** Changeset added; ADR-0083's status line records that the first transport landed; the `CONTEXT.md` port entry now says two things are pushed and how they differ; `packages/browser/README.md` shows the two-line reader rule.

One byproduct to flag: running the browser suite regenerated the committed evidence JSONs under `docs/spikes/a-sharedworker-serves-several-tabs-from-one-host/results/`. I re-ran that spec on all three engines so all nine files are consistent; the entire diff there is the port surface gaining `onStateMoved`.

Note on the task file: it is at `work/tasks/backlog/a-tab-learns-the-state-moved-across-the-port.md`, not `ready/` as the prompt stated, and its spec is in `specs/tasked/`. Content matched current reality (dependency `the-fold-publishes-what-it-just-changed` is in `done/`, ADR-0082 and ADR-0083 both hold), so I built rather than stopping.

## Decisions

**The push is named `stateMoved`, with `subscribeToStateMoved` / `unsubscribeFromStateMoved` and `IndexerPort.onStateMoved`.** It takes the vocabulary that already exists end to end (`StateMoved`, `StateMovedPublisher`, `Indexer.onStateMoved`, and `CONTEXT.md`'s *state-moved signal*), so one name spans core, the port and the app. Alternatives considered: `signal` (too generic, and *signal* is already used loosely for the progress push in ADR-0082's prose), `changed`/`invalidate` (names what a READER does with it, which would bake one reader's policy into the producer's boundary). Touches every later transport task, which should reuse this name rather than coin one: `a-reader-tab-learns-from-the-indexing-tab` and `one-handler-for-every-transport`.

**A tab attaching part way through is told NOTHING until the fold next moves, so `subscribeToStateMoved` answers `undefined`.** This is the one place the pair differs from `subscribeToProgress`, whose response IS the current progress: progress is a state, the signal is an event, and replaying the last notification would tell a fresh subscriber a block just landed when it landed some time ago, under a token it has never held. A late tab reads through the four reads it already has. Alternatives: answer with the last notification (rejected as above); answer with the current coherence token so a tab can compare (rejected: it invents a "current token" question ADR-0083 does not give the reader, and a token is only ever compared against one the reader was PUSHED). This is the residual ADR-0083 already records for every transport, so no new limit is introduced. Touches the remaining transports, which should answer this the same way, and it is asserted as a case on all three hosting shapes.

**The hosting-shapes case signature gained a second argument: the port plus what that port has been told since it was opened (`StateMovedWatch`, `watchStateMoved`).** The suite's stated criterion was "handed a PORT and nothing else", and this widens it. The reason is WHEN the behaviour happens: notifications fire as the fold APPLIES blocks, and by the time any case runs the fold is at the fixture's tip with nothing left to apply, so a case holding only a port could never observe one and "it works on every shape" would have been asserted somewhere else, in three places that agree. Each shape's setup now does what an app does (subscribe the moment it has a port, which for the workers is before the worker boots and on the main thread is before `init`), and the cases assert on that recording. Alternatives: reconfigure from inside a case to cause fresh folding (rejected: the suite forbids it, and it would change the answers every later case asserts on); a chain that grows on a timer (rejected: an elapsed-time test, which the task explicitly rules out); assert per-block delivery only in the node test (rejected: the criterion names the shared suite). `openOnTheMainThread` now joins its port before `init` for the same ordering reason. Touches every later case added to that suite and both of its runners (`browser/cut.ts`, `test/theThreeHostingShapesRunOneImplementation.test.ts`).

**`PORT_PUSH_SUBSCRIPTIONS` is a new named thing on the envelope (coherence check).** It does not re-mean anything: it states as data the pairing the case names already imply, and it sits on the envelope where the case map and push map live. It exists because a SECOND place needs that pairing: the SharedWorker fan-in filters pushes per client, and the previous task recorded that this filter is the one bit of protocol knowledge in a shape file and that "any later PUSH will need its case names added to this filter". A hand-written list there would drift one push behind the map; the `satisfies` on this one makes forgetting a line a compile error. Alternative: keep the two case names inline in `sharedWorker.ts` and add two more (rejected on that recorded hazard). Touches `sharedWorker.ts` and any later push.


wighawag:

Reviewed against all 9 acceptance criteria. One of them needed work the acceptance gate structurally cannot do, so I did it myself — see below.

**Criterion 7 verified in real browsers, because the gate does not.** `packages/browser` has `"test": "vitest run"` and a *separate* `"test:browser": "playwright test"`, and the root `pnpm test` only runs the former. So the real-browser suite this criterion names is outside the gate, and green alone would not have evidenced it. I cloned the branch to a throwaway tree (never the working checkout), installed, built, and ran the playwright suite: **18/18 on chromium, and 36/36 across firefox and webkit**, including `threeHostingShapes.spec.ts` ("runs one behaviour suite against all three hosting shapes, and passes on each") and `progressPushedFromTheWorker.spec.ts`. Criteria 3 and 7 are therefore evidenced, not assumed.

**Criterion 5 (`progress` push unchanged) holds structurally:** `packages/browser/src/host/progress.ts` is not in the diff at all, the progress playwright spec passes unmodified in substance, and there is a dedicated case asserting progress is still subscribed to separately and pushed on its own cadence. The two pushes were not merged, as the task insisted.

**One module, not three that agree** (criterion 3): the subscription is taken once in `serve.ts` as the container opens and fanned out to port listeners, with the reasoning stated — "a shared host serving several tabs holds one handler on its fold and not one per client." That is the right shape, and it carries task 1's no-per-client-state property across the transport instead of quietly reintroducing per-client bookkeeping at the boundary.

**The value crosses unchanged** (criterion 2), and notably the builder tracked the seam as it moved: there is a case asserting the **retraction** arm crosses whole, carrying the rotated token that the appends after it wear. `StateMoved` became a discriminated union only one task ago, and covering the new arm across the port rather than only the `'applied'` shape the task's prose describes is exactly right. Both arms are plain data, so the structured-clone constraint is satisfied by the landed design and no bespoke serialisation was invented — and `clone.ts`'s existing walk covers the new case for free.

Remaining criteria each have a named case: told once per applied block; silence while the driver rests at the tip however many times it advances; stops on unsubscribe and resumes when a listener returns; a message that is not ours still ignored.

**Decisions block** — four, all substantive, and two worth calling out:

- *A late subscriber is told nothing until the fold next moves, so `subscribeToStateMoved` answers `undefined`.* The right call, and the reasoning is the important part: progress is a **state** so its subscribe answers the current value, while this is an **event**, and replaying the last notification would tell a fresh subscriber a block just landed when it landed some time ago, under a token it never held. Rejecting "answer with the current token" is also correct, since that invents a question ADR-0083 never gives a reader.
- *`PORT_PUSH_SUBSCRIPTIONS` states the case/push pairing as data with a `satisfies`.* This responds to a hazard the *previous* port task recorded — that the SharedWorker fan-in filter is the one bit of protocol knowledge in a shape file and any later push must be added to it. A hand-written list would drift one push behind the map; now forgetting a line is a compile error. That is the good kind of small abstraction: it exists because a second place needed the pairing, not on speculation.

**The one judgement I weighed** is decision 3, which widens the hosting-shapes case signature from "handed a PORT and nothing else" to also receive what the port has been told. It is a documented widening of a stated criterion of that suite, so I checked it rather than waving it through. It is justified: notifications fire as the fold *applies* blocks, and by the time any case runs the fold is already at the fixture's tip, so a case holding only a port could never observe one, and "it works on every shape" would have ended up asserted in three places that agree. The rejected alternatives are each genuinely worse (the suite forbids reconfigure from inside a case; a timer-based chain is the elapsed-time test the task rules out; asserting only in the node test abandons the criterion's point). Accepted.

The `docs/spikes/.../results/*.json` churn is just the port API-surface snapshots gaining `onStateMoved`, which is expected rather than scope creep.

Merging.

PR #146, opened 2026-09-13
feat(the-receiving-container-publishes-what-it-applied): The receiving…

wighawag:
Task: `work/tasks/done/the-receiving-container-publishes-what-it-applied.md`

Done. The gate is green (`pnpm build && pnpm typecheck && pnpm test`, plus `format:check`, `check:adr`, `check:refs`, `check:changesets`).

The task's premises all hold. `the-fold-publishes-what-it-just-changed` landed as described, with one naming difference from its own changeset: the cross-package channel shipped as `EventProcessor.setFoldReporter` carrying a `FoldReport` (`AppliedBlock | Retraction`), not `setAppliedBlockReporter`/`AppliedBlock` — widened by the retraction task. ADR-0083 and `CONTEXT.md` are both already written against the landed name, so there is nothing stale to build on. `ReceivingIndexer` did indeed publish nothing, `@etherfold/server` still applies no blocks, and `singleContextEntry`/`indexerEntryOn` are still the two entry shapes. No STOP.

## What I built

**`@etherfold/core` (`receivingContainer.ts`)** — the receiving container now holds a `StateMovedPublisher` (the same class, not a second implementation), exposes `ReceivingIndexer.onStateMoved(handler): StateMovedDetach`, attaches the fold-report relay to **every** fold it holds in `add` (and detaches it in `dropSuperseded`), filters publication to the canonical fold, and rotates the token in `movePointer` — which covers `promote`, the policy's own move, and a revert, because that is the one pointer-move site on this runtime.

**`@etherfold/server`** — `IndexerRegistryEntry.onStateMoved?`, forwarded by `indexerEntryOn`; absent on `singleContextEntry`, on the same capability-statement rule as `generations`/`promote`.

**`etherfold` (CLI)** — the `index` command forwards it on the entry it writes out longhand.

**Docs** — ADR-0083's `status` line and its "both containers publish" section updated to record what is now built (and the one place the two containers differ); the `CONTEXT.md` glossary entry now names `ReceivingIndexer.onStateMoved` and the host-side reach.

**Tests** — the seam test is `packages/conformance-workload-stratagems/test/theReceivingContainerPublishesWhatItApplied.test.ts`: a real fold over the committed capture driven through `ReceivingIndexer` by wire batches, with a subscriber, asserting the same properties the chain-facing file asserts (per-block notifications, real entity names compared against an independent `applyEventStream` run, empty set for a no-op block, one token, a **caused** reorg publishing the real fork point with the token rotated, payload shape carrying no handle). `packages/core/test/theReceivingContainerPublishesWhatItApplied.test.ts` covers the container half (canonical filter over a real follower rebuild, including that the filter covers the *token*; the pointer-move rotation publishing nothing; a contained throwing handler). `packages/server/test/aHostSubscribesToTheStateMoved.test.ts` posts a batch at `/{indexer}/ingest` on a real `createServer` + SQL registry assembly and asserts the subscriber attached to the **entry** is told, with real entity names through `VersionedStateEventProcessor`.

I verified the tests genuinely depend on the production change: disabling the relay turns 5 of the 7 conformance cases red.

## Decisions

**1. How a host reaches the subscription: an OPTIONAL `onStateMoved?` on `IndexerRegistryEntry`, forwarded by `indexerEntryOn`, plus one line on the CLI's `index` entry.** The publication lives in core, as the task says, but a server ROUTE holds an `IndexerRegistryEntry` and never the container, so `ReceivingIndexer.onStateMoved` alone would leave the transport task with nothing attachable inside `@etherfold/server`. I followed the existing precedent exactly rather than inventing one: `generations?`/`promote?` are already optional-and-forwarded-when-present, and absence is already read as a capability statement there. That makes `singleContextEntry` (a host with a bare receiver and no container) report no subscription, which is what lets the transport task REFUSE rather than attach to silence — the failure mode ADR-0083 names for Cloudflare Workers. Alternatives considered: (a) leave the server untouched and let the transport task add the member — rejected, because "a host can subscribe, so the transport task has something to attach to" is this task's acceptance criterion and the shape is the decision, not the wiring; (b) a separate `subscribeToStateMoved` name — rejected as a second convention for one idea (`@etherfold/browser`'s `IndexerPort.onStateMoved` already fixed the name). It touches `@etherfold/server`, the CLI's `index` command, and the remote-transport task, which now has a named attachment point.

**2. Which fold is canonical is answered from an IN-MEMORY `canonicalFold`, re-read wherever the pointer is already consulted AND on the two paths that precede a fold.** "Only the canonical fold publishes" must be decidable synchronously, because a fold reports from inside `process()`. The chain-facing container compares against `Indexer.current`, an in-memory read pointer it owns; the receiving container deliberately has none and derives every answer from the durable registry. I added `noteCanonical`, called wherever this container already reads or writes the pointer (`add`, `canonical`, `canonicalGeneration`, `settlePromotion`, `movePointer`) and — this is the part that costs something — added one `registry.canonical()` read to `liveIngestions()` and `rebuildMore()`, the two paths a batch and a rebuild chunk go through before folding. Without those two, a pointer moved by another process (or by an admin route on a sibling isolate) would leave a retired generation publishing to readers. Alternatives considered: caching only at open (wrong across processes and across an admin promotion), and publishing without a filter (rejected outright: a follower here re-folds a whole stored stream, so an upgrade would fire thousands of notifications about past blocks). The residual is stated on the field, in the ADR and in the changeset: a cross-process move is not seen until the next read, so this container can briefly publish from a fold that stopped being canonical elsewhere — the same in-process staleness the chain-facing twin has, bounded by the same best-effort decision. It touches every deployment's ingest hot path by one small SELECT per batch.

**3. Nothing was extracted from the previous task's implementation; the shared assembly was already the right seam.** `StateMovedPublisher` already owned the subscription, the containment and both rotations, so "reuse rather than mirror" is satisfied by each container holding an instance. What is duplicated is the ~40-line container half (`relayFoldReports` / `publishFoldReport` / `onStateMoved`), and I deliberately did NOT hoist it into a shared base class or mixin: the two containers share no hierarchy today (`ReceivingIndexer` is not an `Indexer` and could not be — one holds chain-facing engines, the other `StreamBuilder`s), and the half that differs is precisely the load-bearing one (`entry === this.current` versus `fold === this.canonicalFold`, decision 2). A base class would have had to abstract exactly the line the two genuinely disagree about, which buys nothing and hides the disagreement. Alternative considered: a free function taking `(publisher, generationRecord, report)` — rejected because the filter, not the assembly, is what would still have been written twice. The difference is recorded in ADR-0083 so it is a stated asymmetry rather than a discovered one.

**4. A test-util change shared across three existing suites.** `receivingWorld.ts`'s `foldingProcessor` now implements `setFoldReporter` and reports a retraction plus one applied block per block, from inside `process()` in the same order the shipped entity fold reports (its reported entity name is a constant, `REPORTED_ENTITY`, because names are asserted where they are produced). To avoid a second definition of "which blocks did this stream apply" and "what fork point did it revert to", it imports `appliedBlocksOf`/`forkPointOf` from `stateMovedWorld.ts` and I made those two generic over `ABI` rather than casting. This touches `rebuild.test.ts` and `theCanonicalPointerMovesBack.test.ts`, which share that world; both stay green (reporting with no reporter attached is a no-op).


wighawag:

Reviewed against all 7 acceptance criteria, with particular attention to the two things this task could have got wrong in good faith: putting the publication in the wrong package, and conflating two different silences.

**The publication is in core; the server only gets a way to REACH it.** `packages/server` was touched, so I checked it closely rather than assuming. What it gained is an *optional* `onStateMoved?` on `IndexerRegistryEntry` that forwards to `ReceivingIndexer.onStateMoved`, with the reason stated at the seam: this package applies no blocks, an ingest route delegates to a receiver the host constructed, so the signal is produced in core and what an entry adds is the only handle a route has on a name. No fold and no publication was added to `@etherfold/server`. That satisfies criterion 4 without the error the task warned about.

The optionality is the right shape, and it follows existing precedent (`generations?`/`promote?` are already optional-and-forwarded, with absence read as a capability statement) rather than inventing a convention. It also pays off immediately: a host holding a bare receiver and no container reports *no* subscription, with a test for exactly that, which is what will let the remote-transport task REFUSE rather than attach to silence — the Cloudflare Workers failure mode ADR-0083 names.

**The two silences were separated correctly** (criterion 5). The receiving container's documented rule 2 read "It publishes no state handle **and notifies nobody**", and that conjunction was the trap: keeping it whole leaves the task undone, dropping it whole reopens a decision made on purpose. The builder split it, removing only the notification half and preserving the state-handle half. No `onStateUpdated` and no state getter appears in the diff, and a case asserts the exact key set of every notification (`block, coherence, entities, generation, kind`) so no rows, mutations or handle can ride along.

**Criterion 2 is discharged, and it is the one I was watching for.** The previous task (`a-promotion-rotates-the-coherence-token`) recorded in its decisions that the receiving container moves a pointer too and that whoever built this publisher owed the same one-line rotation there. It is here and asserted on this path rather than inherited by assumption: a retraction rotates as it publishes, a pointer move rotates while publishing nothing and the next notification names the fold that now answers, and a `promote` naming the generation already canonical rotates nothing. The follower-silence rule is mirrored too, which matters more here than on the twin, since a follower on this path re-folds a whole stored stream.

**Criterion 1 and 3:** each container holds an instance of the same `StateMovedPublisher`, not a second implementation, and the entity set arrives through the same `setFoldReporter` relay. The conformance test proves the "real names on a server" claim with a real processor over the stratagems capture (`['commitment']` on the append after the fork, declared-but-untouched entities absent, an empty set for a block whose handlers changed nothing).

**Two decisions I weighed and accept:**

- *Decision 2* is the substantive one. "Only the canonical fold publishes" must be decidable synchronously because a fold reports from inside `process()`, and this container deliberately has no in-memory read pointer — so it needed `noteCanonical` plus one extra `registry.canonical()` read on `liveIngestions()` and `rebuildMore()`, the two paths a batch and a rebuild chunk pass through before folding. That is a real cost on every deployment's ingest hot path, and the builder states it plainly as one small SELECT per batch, along with the residual that a cross-process pointer move is not seen until the next read. I accept it: without those reads a retired generation would go on publishing to readers after an admin promotion on a sibling isolate, and the staleness window is the same best-effort bound the chain-facing twin already lives with. Worth the human knowing it is there.
- *Decision 3* declines to hoist the ~40-line duplicated container half into a base class or mixin, because the two containers share no hierarchy and the half that differs is exactly the load-bearing filter (`entry === this.current` versus `fold === this.canonicalFold`). A base class would have had to abstract precisely the line the two genuinely disagree about. That is the correct call, and recording the asymmetry in ADR-0083 makes it stated rather than discovered.

Test-util changes are additive and the two suites sharing that world stay green, since reporting with no reporter attached is a no-op.

Merging.

PR #147, opened 2026-09-13
feat(a-reader-tab-learns-from-the-indexing-tab): A reader tab learns fr…

wighawag:
Task: `work/tasks/done/a-reader-tab-learns-from-the-indexing-tab.md`

Done. Everything is green and the tree holds only this task's changes.

## What I did

**Drift check first.** ADR-0083's status still says the cross-tab `BroadcastChannel` is not built; the blocking task (`a-tab-learns-the-state-moved-across-the-port`) landed as this task assumes (`IndexerPort.onStateMoved`, `PORT_PUSH_SUBSCRIPTIONS`, one push beside `progress`); nothing in the repo uses a `BroadcastChannel` yet. No stale premise, so I built it.

**The adapter** — `packages/browser/src/stateMovedAcrossTabs.ts`, exported from the package index:

- `stateMovedChannelName({databaseName})` composes `etherfold/state-moved/<databaseName>` and nothing else, resolving the default through the store's own `DEFAULT_DATABASE_NAME` so the two cannot answer differently.
- `openStateMovedAcrossTabs({databaseName})` returns one object with `publish` / `onStateMoved` / `close`, over ONE `BroadcastChannel` used for both directions. What is posted is core's `StateMoved` inside a protocol envelope (so a stranger's traffic on the channel is ignored, exactly as `isPortPush` ignores it on the port); what a listener is handed is the value the fold published, unchanged. A runtime without `BroadcastChannel` is told in a sentence that names what is missing and falls back to nothing.
- No election, lease, heartbeat or publisher identity; no buffering, no replay, no per-receiver state.

**Tests, failing first at the agreed seam.** `packages/browser/test/aReaderTabLearnsFromTheIndexingTab.test.ts` (6 cases over real `BroadcastChannel`s and a real host on a `MessagePort`) went red before the module existed. The real-browser case is a NEW one: `packages/browser/browser/readerTabLearnsFromTheIndexingTab.spec.ts` plus seven `cut.ts` cases — three tabs, three of their own dedicated-worker hosts, two databases: a reader tab whose own host claims the store and folds nothing, an app NEXT DOOR folding into another store, and an indexing tab held at block 103 so a listener can attach half way through. Green on chromium, firefox and webkit; byte-stable results committed to `docs/spikes/a-reader-tab-learns-from-the-indexing-tab/` with a README.

**Docs:** `packages/browser/README.md` (a "second window is not a stale window" section), the `CONTEXT.md` state-moved entry, ADR-0083's status line, and a changeset.

## Decisions

**The channel name is composed from the STORAGE IDENTITY alone, spelled as the store's `databaseName`, and passed as the same config value the store was built with.** The name is `etherfold/state-moved/<databaseName>`, defaulted through `DEFAULT_DATABASE_NAME` so the channel and the store cannot resolve the default differently. Alternatives considered: (a) reading the identity off the store instance — impossible on the publishing side, because the store lives in the worker and `openForWriting`'s `ClaimedStateStore` wrapper does not expose `databaseName` anyway; (b) widening the `StateStore` seam with a `storageIdentity` — a four-backend, conformance-suite-wide change to move a fact the browser caller already holds, which ADR-0078's precedent argues against; (c) an independent channel-name option — rejected outright, it is the "a string an app passes" scoping the spec names as the silent failure. What this touches: apps on a NON-IndexedDB backend (`BrowserStateStoreConfig`'s `{backend: fn}` arm) have a storage identity this parameter cannot spell, and passing that config would silently scope to the default name; I did not add a refusal for it, because the remedy would have to name an identity concept the type does not yet have. If `one-tab-indexes-and-the-others-read` or a SQL-backed browser store needs it, the identity should come from the store rather than from a second parameter here.

**A tab hosting its own worker does not hear its own publication, because publish and listen share ONE channel object.** `BroadcastChannel` never delivers to the object that posted, so the duplicate (once over the port, once off the channel) cannot arise, and nothing has to name the publisher in order to filter it out. The alternative — stamping each publication with a publisher id and dropping your own — was rejected as exactly the half-election the spec fences off: it would put "who is publishing" on the wire, which `one-tab-indexes-and-the-others-read` would then have to adopt or unpick. The cost, stated: two separate `openStateMovedAcrossTabs` calls in ONE document DO hear each other (right, since silently dropping by document would be a scope nobody asked for), and this is why the API is one object rather than a publish function and a subscribe function.

**A publication on a CLOSED adapter is dropped and logged at `info`, not raised.** The ordinary cause is a teardown releasing the channel before the port subscription feeding it, which is a tab going away rather than a failure, and `publish` is wired directly onto a fold's notification path where a throw would reach code that has already applied its block. A genuine `postMessage` failure is still logged as an error. Touches: any app wiring `indexer.onStateMoved(tabs.publish)` and unmounting in either order.


wighawag:

Reviewed against all 9 acceptance criteria, including the real-browser one the acceptance gate cannot reach.

**Criterion 7 verified in real browsers.** As with the port task, `pnpm test:browser` (playwright) is outside the gate, so green alone would not evidence this. I fetched the branch into a throwaway tree, installed, built and ran the full suite: **57/57 across chromium, firefox and webkit**, including the new `readerTabLearnsFromTheIndexingTab.spec.ts` on each engine. The case is genuinely the one the criterion asked for and not the easy one it warned against: each tab holds its **own dedicated worker** over a shared database, with a third next-door tab on its own host and own database, rather than the shared-worker model where both tabs already hold a port to one host and would be pushed to anyway.

Crucially, the case guards its own non-vacuity: it asserts the reader tab's own host never got anywhere (`ownHostLastToBlock < START_BLOCK`), which is what proves the reader learned from the **channel** and not from a fold of its own. Without that assertion the test would have passed while demonstrating nothing, which is precisely the failure the criterion was written to prevent.

**Criterion 3, both directions, in one case:** a tab of the same store heard every applied block, and the app next door on the same origin heard nothing at all, with the channel names asserted equal for one store and unequal across two.

**The scoping rule follows the writer token's reasoning rather than re-deriving it.** The name is `etherfold/state-moved/<databaseName>` — storage identity alone, per `CONTEXT.md` — and it is defaulted through the same `DEFAULT_DATABASE_NAME` the store uses, so the channel and the store cannot resolve the default differently. That last detail is the kind of thing that silently breaks later, and closing it deliberately is the right instinct. The rejected alternatives are the right ones to reject, especially an independent channel-name option, which is exactly the "a string an app passes" scoping the task names as the silent failure.

**No election was built** (criterion 6). The only occurrences of leader/lease/heartbeat in the new module are disclaimers, and there is a case asserting a second publishing tab is noise rather than an error and that nothing on the wire names who published. Decision 2 earns particular credit here: rather than stamping publications with a publisher id and filtering your own out, publish and listen share ONE channel object, so `BroadcastChannel`'s own semantics (it never delivers to the object that posted) make the duplicate unarisable. That avoids putting "who is publishing" on the wire, which `one-tab-indexes-and-the-others-read` would then have had to adopt or unpick. The cost is stated plainly: two separate `openStateMovedAcrossTabs` calls in one document do hear each other, which is why the API is one object rather than a publish function beside a subscribe function.

Remaining criteria have cases: the value handed across is the one the fold published and identical to what crosses the port; nothing is buffered for a tab that was not listening and that tab converges on the next notification; a message on the channel that is not ours is ignored.

**Decision 3** is consistent with the model established at the root: a publication on a closed adapter is dropped and logged rather than raised, because `publish` is wired directly onto a fold's notification path and a throw there would reach code that has already applied its block. That matches "a subscriber that throws does not break the fold", and a genuine `postMessage` failure is still logged as an error.

**One residual I am recording rather than blocking on.** Decision 1 states that an app on the non-IndexedDB `{backend: fn}` arm of `BrowserStateStoreConfig` has a storage identity this parameter cannot spell, so passing that config would silently scope to the default channel name — and no refusal was added, because the remedy would have to name an identity concept the type does not yet have. I agree with not inventing it here (widening the `StateStore` seam for it is the four-backend change ADR-0078's precedent argues against, and this task is fenced to a transport). It is a genuine gap for a custom-backend app, it is surfaced honestly rather than hidden, and the suggested home — the identity coming from the store, when `one-tab-indexes-and-the-others-read` or a SQL-backed browser store needs it — is the right one. I am carrying it into the drive report as a follow-up worth tasking, not a defect in what this task was asked to deliver.

Merging.

PR #148, opened 2026-09-13
feat(a-remote-client-learns-the-state-moved): A remote client learns th…

wighawag:
Task: `work/tasks/done/a-remote-client-learns-the-state-moved.md`

Everything is green. Here is what I built.

## What landed

**The third transport of ADR-0083: `GET /{indexer}/state-moved`, server-sent events** (`packages/server/src/api/stateMoved.ts`, wired in `src/index.ts`).

- It is a **transport and never a producer**: it subscribes at `IndexerRegistryEntry.onStateMoved` (the seam the previous task landed) and adds no fold, no publication and no storage watch. The tests attach an **in-process subscriber to that same seam beside the HTTP client** and assert both are handed the same value, which is how "a second transport needs no producer change" is demonstrated rather than asserted.
- Two frame kinds, mirroring the split ADR-0082 already makes on a tab's port: `state-moved` carries `@etherfold/core`'s `StateMoved` as JSON, untouched and never replayed; `progress` carries `{lastToBlock, latestBlock, blocksBehindTip, coherence, generation}` on connect and whenever the fold moves.
- A connecting client is told the **position and the coherence token at once** (new `ReceivingIndexer.coherenceNow()` / `Indexer.coherenceNow()`, forwarded as `IndexerRegistryEntry.coherenceNow`), which is how a remote reader converges with no state query surface.
- Nothing is held per client: one handler reference per stream, a disconnect detaches (stream `cancel` plus the request's abort signal), nothing buffered or replayed. A test attaches three clients, closes them and asserts the producer's attachment count is back to zero.
- Two `501` refusals: `state-moved-unsupported-runtime` (the host has not declared `ServerOptions.holdsStreamsAcrossRequests`) and `state-moved-not-published` (a bare receiver with no container). `platforms/nodejs` declares the capability; `platforms/cf-worker` deliberately does not, and says so in prose.
- No GraphQL anything; the package dependency list is unchanged and the existing `platformAgnostic.test.ts` still passes, with a new focused scan asserting the refusal is a **declaration, not a runtime sniff**.

Tests: `packages/server/test/aRemoteClientLearnsTheStateMoved.test.ts` (11 cases: same value/same serialisation, progress, mid-fold connect with no replay, a real reorg's retraction, a real promotion's rotation, detach, and five refusals) plus `test/utils/signalStream.ts`. Docs: `packages/server/README.md` gained the route row and a section; ADR-0083's status line now records this transport as built. Changeset added. Gate run locally: `format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test` (full monorepo) all green; `changeset status --since=main` only sees committed files, so it will pass once the runner commits.

I also dropped one out-of-scope note: `work/notes/observations/the-combined-run-cannot-serve-the-state-moved-stream.md` (`etherfold run` registers no named indexer on purpose, so it answers `501` here).

## Decisions

**The route is `GET /{indexer}/state-moved`, public like the feed, carrying TWO frame kinds.** Path from `CONTEXT.md`'s own term (*the state-moved signal*), segment-per-tenant like every other named route. Anonymous, because `INGEST_TOKEN` guards the routes that can move the cursor and this one moves nothing and reads no rows; what crosses (a block number, entity NAMES, two opaque digests) is strictly less than the feed already serves anonymously. Alternative considered: putting it behind the ingest token, which would hand every reader the credential that can WRITE. Touches: any host that wants it private (its own edge, same as the feed).

**Progress rides the stream as its OWN frame (`progress`) and not as fields on the signal.** ADR-0083 fixes the payload, and the task requires the browser and the network to receive the same value, so widening `StateMoved` was not available. Naming reuses the browser's existing words (`lastToBlock`, `latestBlock`, `blocksBehindTip`) rather than coining a second vocabulary; `blocksBehindTip` because bare `blocksBehind` already means how far a non-canonical generation is behind the canonical one. Touches: `sync-progress-rides-the-signal-to-a-reader` (the cross-tab half), which should reuse these names.

**The progress figures come from the stored stream's COVERAGE CLAIM, not from any cursor.** `readStreamCoverage` gives `lastToBlock` and `latestBlock` from ONE row that the fold writes on every batch (including one that carried no logs), so the pair cannot disagree and no opaque cursor is deserialised (ADR-0027). Alternatives: `expectedFromBlock` (wrong — it is deliberately pulled back inside the finality window, so a level fold would render as "finality blocks behind"), and the host's `/status` cursor reporter (opaque and host-shaped). Stated cost, documented at the site: the claim is the STREAM's, so a canonical generation that were mid-rebuild would be behind it; that cannot arise today because the pointer moves at the END of a rebuild.

**`progress` carries the coherence token, unlike the port's progress push.** A tab attaching late READS the store it shares; a remote reader has neither that store nor a query surface (deferred to `the-same-query-runs-against-a-worker-and-a-server`), so ADR-0083 makes its convergence "be told the position and the token on connect". Touches: the query spec, which will add per-operation block pinning and re-query convergence beside this.

**A new READ was added to core: `coherenceNow()` on BOTH containers, surfaced as the optional `IndexerRegistryEntry.coherenceNow`.** The token in force was reachable only by receiving a notification, so the connect-time answer was unbuildable without it. It rotates nothing and publishes nothing, and it is deliberately not a replayed notification (which would have a reader invalidate for a block it may already have read). Put on both containers so the two keep differing in exactly ONE place (which fold is canonical). Alternative considered and rejected: changing `onStateMoved`'s shape, which is the one shape all three transports adapt to. Touches: `@etherfold/core`'s public API, `@etherfold/server`'s registry, the CLI's hand-written entry in `indexCommand.ts`, and `one-handler-for-every-transport`.

**`onStateMoved` and `coherenceNow` are PAIRED: an entry with one but not the other is refused.** They are one publisher's two answers (`indexerEntryOn` forwards both, `singleContextEntry` neither). Serving the stream without the token would silently drop the convergence promise a reconnecting client depends on. Alternative: degrade to a token-less `progress` frame, which is the "appears to work" failure this endpoint exists to avoid.

**The unsupported-runtime refusal is a HOST-DECLARED capability, `ServerOptions.holdsStreamsAcrossRequests`, and ABSENT means NO.** Fail-closed, on the same rule the credential guard follows and the same absent-is-a-capability-statement convention as `generations` / `promote` / `onStateMoved`. This is a user-visible default: every existing host gets `501` until it declares, which is why `platforms/nodejs` now declares it (so `etherfold index` serves the stream) and `platforms/cf-worker` deliberately does not. Declared at the platform layer rather than by the CLI, because `platforms/*` is the only place a runtime may be named at all. Alternative rejected: defaulting to supported, which is exactly the Worker failure ADR-0083 forbids (compiles, passes on Node, never fires).

**NO heartbeat is sent to an idle stream.** An interval invented here is the polling interval the signal exists to replace, and no number fits a Node process, a reverse proxy and a CDN at once. What makes that cheap rather than negligent is the connect-time answer: a dropped idle stream costs a reconnect, and a reconnect is answered at once with the position and the token, so a client cut off during a quiet chain learns it missed nothing. `X-Accel-Buffering: no` IS sent, for the neighbouring problem (a proxy withholding frames that were written). Touches: any deployment needing connections held open configures its own edge.

**Progress refresh cadence: on connect, then after each notification, with at most one read in flight per client and value-deduped.** Notifications fire per BLOCK while the figures move per BATCH, so an uncoalesced refresh would cost a pair of reads per block during catch-up. The coalescing is back-pressure from the database rather than an invented interval.

**ADR-0083's status line was updated to record this transport as built.** That is keeping a claim about the code true in the same change, and is NOT the reserved removal of an `accepted, not yet implemented` line, which `ADR-FORMAT.md` assigns to `one-handler-for-every-transport`. `CONTEXT.md`'s glossary was deliberately left alone for the same reason: that task owns the vocabulary entry.


wighawag:

Reviewed against all 11 acceptance criteria. This was the largest task in the chain and the one with the most ways to be quietly wrong; it is right on each of them.

**The package-boundary trap is avoided, and provably.** The task's central warning was that `@etherfold/server` applies no blocks, so a fold or publication added there would be rebuilding what exists one layer down. I grepped the whole of `packages/server/src` for `new StateMovedPublisher`, `.publish(`, `publishRetraction`, `applyBlock`, `setFoldReporter` and `.process(`: **no matches**. The route attaches to `IndexerRegistryEntry.onStateMoved` — the attachment point the previous task deliberately left — and the module says so at the top: "This package APPLIES NO BLOCKS, so this is a TRANSPORT and never a producer."

**Criterion 7 is implemented the way it was specified, not the way that would have been easier.** The refusal is a host-declared capability (`ServerOptions.holdsStreamsAcrossRequests`) where **absent means no**, so it fails closed; `platforms/nodejs` declares it (naming a runtime is only permitted there) and `platforms/cf-worker` deliberately does not, documenting that an absence rather than a `false` is the safe answer. The server package sniffs nothing, `platformAgnostic.test.ts` was not weakened or modified, and there is a case asserting the refusal is decided on **what the host declared and never on a runtime it sniffed for**. Two further refusal cases cover a bare-receiver entry and a host with no registry, so the endpoint refuses rather than attaching to silence — which is exactly the capability-absence the previous task left in place for this.

**Criterion 3 without a cursor:** the figures come from the stored stream's coverage claim, one row the fold writes on every batch including an empty one, so the pair cannot disagree and nothing opaque is deserialised. The rejected alternatives are rejected for good reasons (`expectedFromBlock` is pulled back inside the finality window and would render a level fold as "finality blocks behind"). The cost is documented at the site.

**Criterion 6, and the one new core read.** `coherenceNow()` was added to both containers, and I checked this deliberately because the root task recorded that there was *no* public token getter and that a transport could carry the token it last relayed. That reasoning does not extend to a **connecting** client, which has no prior notification to have carried one, so the connect-time answer was genuinely unbuildable without it — the root task's wording was "no shipped caller needs one yet", and this is that caller. It rotates nothing, publishes nothing, and is deliberately not a replayed notification (which would make a reader invalidate for a block it may already have read). Rejecting a change to `onStateMoved`'s shape was right: that shape is what all three transports adapt to.

**The pairing decision is the best thing in this diff.** `onStateMoved` and `coherenceNow` are refused unless an entry has **both**, because they are one publisher's two answers and serving the stream without the token would silently drop the convergence promise a reconnecting client depends on. Degrading to a token-less progress frame would have been precisely the "appears to work" failure this endpoint exists to avoid. Fail-closed, again.

**Progress as its own frame** rather than fields on `StateMoved` is correct: ADR-0083 fixes that payload and criterion 2 requires the browser and the network to receive the same value, so widening it was not available. Reusing the browser's existing words (`lastToBlock`, `latestBlock`, `blocksBehindTip`, the last disambiguated from the existing `blocksBehind`) avoids a second vocabulary.

**No heartbeat, argued rather than omitted:** an interval invented here is the polling interval the signal exists to replace, and no single number fits a Node process, a reverse proxy and a CDN. What makes that cheap is the connect-time answer — a dropped idle stream costs a reconnect, and a reconnect is answered at once with position and token, so a client cut off during a quiet chain learns it missed nothing. Sending `X-Accel-Buffering: no` for the neighbouring proxy-buffering problem is the right distinction.

Remaining criteria have cases: an HTTP client is handed the same value an in-process subscriber gets, serialised identically (which is also how criterion 8's transport-agnostic claim is demonstrated rather than asserted); a retraction and a promotion both cross with their rotations; nothing is held about a client that went away and many can attach without growth; a mid-fold connect is told position and token at once and replays nothing. Criterion 9 holds — every GraphQL mention in the diff is prose describing what was deliberately not built, and no dependency was added.

**The observation note is a real contribution and I am carrying it forward.** `the-combined-run-cannot-serve-the-state-moved-stream` records that `etherfold run` — the combined deployment `CONTEXT.md` calls the milestone, and the shape most apps would point at — answers `501` on this endpoint, because it deliberately registers no named indexer to avoid opening `/{indexer}/ingest` to a second writer. That was correctly captured rather than fixed in place, since whether a read-only registration belongs there is a design decision about the write path. It is the most consequential open question this drive surfaced, and it lands in my report.

Leaving `CONTEXT.md`'s glossary to `one-handler-for-every-transport` per `ADR-FORMAT.md`, while still updating ADR-0083's status line, is the correct split.

Merging.

PR #149, opened 2026-09-13
feat(sync-progress-rides-the-signal-to-a-reader): Sync progress rides t…

wighawag:
Task: `work/tasks/done/sync-progress-rides-the-signal-to-a-reader.md`

Everything is green. Here is my report.

## What I built

"Syncing, N blocks behind" is now renderable in a tab that is not the one folding, over the one cross-tab channel a reader already listens to.

**Drift check first (no drift found).** `a-reader-tab-learns-from-the-indexing-tab` landed as this task assumes: `openStateMovedAcrossTabs` / `stateMovedChannelName`, one channel object per tab, scoped to the **storage identity**. The port's progress push is as ADR-0082 shipped it (`HostProgress`, `blocksBehindTip`, the subscribe that answers with the current value). ADR-0083's cross-tab paragraph is unchanged. `a-remote-client-learns-the-state-moved` landed the cross-network half with `lastToBlock` / `latestBlock` / `blocksBehindTip` and explicitly asked this task to reuse those names, which it does.

**Production code** (`packages/browser/src/stateMovedAcrossTabs.ts`): two verbs beside the two the channel had, `publishProgress` and `onProgress`, carrying the host's own `HostProgress` unchanged as a second `kind` on the existing envelope (plus an internal `progressAsk` kind). `createProgressReadable` binds to the cross-tab end unchanged, because it takes the push and not a port — so a reader tab's progress bar is the same line a hosting tab's is. `sameProgress` moved from `host/cases.ts` to `host/envelope.ts` (beside the type it is about) so a reader tab that imports only the channel does not pull the case server in; both transports now suppress a repeat with one definition of "the same report". The port's `progress` push is untouched in shape, cadence and subscription behaviour.

**Tests, failing first, at the agreed seam and its node twin:**
- `packages/browser/browser/syncProgressRidesTheSignalToAReader.spec.ts` — four real tabs, one database, one channel: an indexing tab with its own worker (held at block 103, then released), a reader tab with its OWN held host, a reader tab with no host, and a newcomer tab opened after the fold reached the tip. Both readers' full sequences equal the indexing tab's port pushes element for element; the hosted reader renders the indexing tab's numbers while its own host sits at block 0 with `blocksBehindTip` absent; the newcomer is handed exactly one report. Green on chromium, firefox and webkit; the whole playwright suite (60 tests) is green.
- `packages/browser/test/syncProgressRidesTheSignalToAReader.test.ts` — 11 cases on real `BroadcastChannel`s in node, run on every commit.
- Evidence: `docs/spikes/sync-progress-rides-the-signal-to-a-reader/` (README + per-engine results), referenced from the spec.

**Standing conventions:** changeset added (`.changeset/sync-progress-rides-the-signal-to-a-reader.md`); ADR-0083's status line amended to record what is now built (not the reserved `accepted, not yet implemented` removal, which `one-handler-for-every-transport` owns). `CONTEXT.md` deliberately untouched, on the same ground the remote-transport task left it: that task owns the glossary entry.

Gate run locally: `format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test` all green (`pnpm test`: 0 failures across every package). `changeset status --since=main` cannot see the new changeset only because it is still untracked; it resolves once the runner commits.

## Decisions

**A newly-attached reader is handed the last report this tab HEARD, and if it holds none it ASKS on the channel.** The port's rule is that progress is a STATE whose subscribe answers with the current value, so the channel end keeps the last report another tab published and hands it to a listener attaching later. That alone is not enough for a tab that just OPENED: a host at the tip pushes nothing, so a window opened into a quiet chain would be blank until the chain moved (hours). So attaching with nothing held posts a `progressAsk`, and a tab holding a report of its own re-posts it. Alternatives considered: (a) tell a late listener nothing until the next push, which fails the criterion in exactly the quiet-chain case; (b) have the reader ask its own host, which it may not have, and whose answer is wrong when it does. It is an ask and not a request — nothing awaited, retried or correlated, no reply address, and an ask nobody can answer is silence. Touches: the analogue on the network transport is the server's connect-time `progress` frame, deliberately the same promise by a different mechanism (a remote client has a connection to answer on; a broadcast has none).

**Only what THIS tab published is re-posted in answer; a tab does not gossip on a report it merely heard.** Re-posting a heard report would keep a departed indexer's last number alive on the channel indefinitely. Cost, stated: if the only publisher's tab is gone, a newcomer renders nothing until a tab starts folding again, which is the honest answer.

**A report identical to the one a tab already holds is not delivered.** This is the rule the host already follows on the port (it posts nothing where the report would repeat), applied at the receiving end so several tabs asking at once cost one render rather than one per ask. It is suppression, never composition: what is held is one report, by reference, replaced wholesale — the wrapper-is-a-VIEW rule. It reuses `sameProgress` rather than a second notion of sameness, which is why that function moved to `envelope.ts`.

**Cadence is INHERITED from the port and no timer is introduced.** A report is posted when the host pushes one at the tab holding the port (batch applied or phase changed), not at the cadence of the applied blocks this channel otherwise carries, and not on an interval. The app wiring is one line beside the one it already wrote: `indexer.onProgress(tabs.publishProgress)`.

**The value crosses WHOLE, including `host` and `scope`, which describe the publishing tab's host.** The criterion is "identical in shape and meaning to what a hosting tab receives over its port", so narrowing it to a reader-flavoured subset would have been a second vocabulary. A hosting SHAPE is not a publisher identity (nothing can be addressed, counted or elected from `'dedicated-worker'`), so the previous task's "nothing on the wire names the publisher" property is not weakened. Touches: `one-handler-for-every-transport`, which compares payloads across the three transports — the cross-tab progress frame is `HostProgress`, while the server's is `StateMovedProgress` (three shared block-figure names plus a coherence token a tab does not need, because a tab can re-read).

**Before an election exists, several tabs holding hosts may publish progress, and a receiver renders the LAST report rather than a merge.** This is the same residue the notification has ("every indexing tab publishes, every tab listens"), but with a sharper edge: two reports are contradictory where two notifications were merely redundant. I did not add a filter, because any filter here (prefer the higher block, prefer the writer) would be the receiver inventing policy and a second source of truth, and picking the writer is precisely `one-tab-indexes-and-the-others-read`'s job. It is documented at the site, tested as noise rather than an error, and is why the browser fixture publishes progress only from the tab whose fold is moving.

**`sameProgress` moved from `host/cases.ts` to `host/envelope.ts`.** One definition of "the same report" for both transports, and it keeps a reader tab that imports only `openStateMovedAcrossTabs` from being tied to the module that serves every port case. No behaviour change; `cases.ts` imports it from its new home.

**ADR-0083's status line records this transport half; `CONTEXT.md` is left alone.** Keeping a claim about the code true in the same change, on the precedent the remote-transport task set. The glossary entry for the cross-tab channel belongs to `one-handler-for-every-transport`, which has it as an acceptance criterion.


wighawag:

Reviewed against all 9 acceptance criteria.

**Criterion 4 checked properly, because two files that could have broken it were touched.** `host/progress.ts` and `host/envelope.ts` appear in the diff, so I did not take "untouched" on trust: `envelope.ts` has **zero** removed lines, and `progress.ts` loses only two doc-comment lines. The additions are a reworded comment and a new pure `sameProgress(a, b)` value-equality function. The port's `progress` push is genuinely unchanged in shape, cadence and subscription behaviour, and the two mechanisms were not unified — which was the trap this task was written to avoid.

**Criterion 7 verified in real browsers,** since playwright sits outside the acceptance gate. Full suite in a throwaway tree: **60/60 across chromium, firefox and webkit**, including the new `syncProgressRidesTheSignalToAReader.spec.ts` ("every tab renders how far behind the fold is, sourced from the one tab that knows") on each engine, and reusing the own-host-per-tab case the previous task added rather than the shared-worker case where the result would prove nothing.

**Criterion 3 holds structurally:** one `BroadcastChannel` in the package, now carrying three frame kinds (`stateMoved`, `progress`, `progressAsk`) rather than a second channel with its own lifetime.

**Criterion 5 is the interesting one and it is well solved.** The port's rule is that progress is a *state* whose subscribe answers with the current value, so the channel end holds the last report and hands it to a late listener. The builder correctly spotted that this is insufficient for a tab that just *opened* into a quiet chain — a host at the tip pushes nothing, so the window would be blank for hours — and added a `progressAsk`. The design restraint here is what makes it acceptable: it is explicitly "an ask and not a request", with nothing awaited, retried or correlated and no reply address; an ask nobody can answer is silence rather than a failure (with a case for exactly that). The answer is broadcast and deduped through `sameProgress` at the receiving end, so several tabs asking at once cost one render, not one per ask — which is also how criterion 6 stays true (a case asserts one report is held and not a backlog, however many tabs ask).

**Two restraint decisions I particularly agree with:**

- *A tab re-posts only what it published itself, never a report it merely heard.* Gossiping a heard report would keep a departed indexer's last number alive on the channel indefinitely. The cost is stated honestly: if the only publisher's tab is gone, a newcomer renders nothing until some tab starts folding again, which is the truthful answer rather than a stale one.
- *Several publishing tabs render as the LAST report, with no filter added.* This is sharper than the notification case (two reports contradict where two notifications were merely redundant), and the reasoning for leaving it is right: any filter — prefer the higher block, prefer the writer — would be the receiver inventing policy and becoming a second source of truth, and choosing the writer is `one-tab-indexes-and-the-others-read`'s job. Documented at the site and tested as noise rather than an error.

**Criterion 2 is met in the strong form:** the value crosses whole, including `host` and `scope`, rather than being narrowed to a reader-flavoured subset that would have become a second vocabulary. I checked this against the previous task's "nothing on the wire names the publisher" property and it is not weakened — a hosting *shape* is not a publisher identity, since nothing can be addressed, counted or elected from `'dedicated-worker'`. Cadence is inherited from the port with no timer introduced, and the app wiring stays one line beside the one already written.

Moving `sameProgress` to `envelope.ts` is justified (one definition of "the same report" for both transports, and it stops a reader tab that imports only `openStateMovedAcrossTabs` being tied to the module serving every port case) and is behaviour-neutral. Binding through the existing `createProgressReadable` keeps the wrapper-is-a-VIEW rule intact, with a case for it.

**One asymmetry I am carrying into the drive report** rather than treating as a defect: the cross-tab progress frame is `HostProgress` while the server's is `StateMovedProgress` — three shared block-figure names plus a coherence token a tab does not need, because a tab can re-read. That is a defensible difference and the builder flagged it deliberately for `one-handler-for-every-transport`, which has cross-transport payload comparison as an acceptance criterion. It is the next task's business, and it is now on the record for it.

Leaving `CONTEXT.md`'s glossary entry to that same task, while updating ADR-0083's status line here, follows the precedent the remote-transport task set.

Merging.

PR #150, opened 2026-09-13
feat(one-handler-for-every-transport): One handler for every transport

wighawag:
Task: `work/tasks/done/one-handler-for-every-transport.md`

## What landed

**A new package, `@etherfold/state-moved-conformance`** — the suite a state-moved *transport* must pass to be an adapter rather than a second semantics. Cases are data (`{group, name, run}`), parameterised by a `StateMovedTransport` adapter, with `describeStateMovedConformance` (vitest) and `runStateMovedConformance` (runner-free) as thin adapters — the `@etherfold/state-store-conformance` shape, not a third convention. Four chapters: `one handler`, `the coherence token`, `a dropped notification`, `coherent with what a reader reads`.

**One handler, literally.** `readerRule()` in the suite is the ADR's two lines written once above every transport; the cases attach that same `StateMovedHandler` to all three and assert on what it *decided* (`{invalidate: 'everything'} | {invalidate: [...names]}`), not on bytes. **No transport forced a variation.**

**Three runners over real substrates.** `packages/browser/test/oneHandlerForEveryTransport.test.ts` runs it over a tab's port and the cross-tab channel (real `MessageChannel`, real `BroadcastChannel`, real fold over a forkable fake chain, real IndexedDB). `packages/server/test/oneHandlerForEveryTransport.test.ts` runs it over `GET /{indexer}/state-moved` (real HTTP + SSE, real ingest). Every verb moves a real fold; the retraction is *caused* by serving a forked branch / re-sending a block under a new hash. 28 browser cases + 11 server cases, green.

**The suite is asserted to be able to fail**: four cases feed `runStateMovedConformance` transports broken in the ways a real adapter goes wrong (replaying to a late joiner, re-stamping the token, dropping the retraction, answering neither convergence question) and check *which* cases went red.

**ADR-0083**: status frontmatter removed (absence means accepted and current); the as-built facts it carried moved into a new body section, *The three transports, as built, and the ONE SUITE that keeps them one model*.

**`CONTEXT.md`**: names the network transport as built, and adds **transport conformance suite** with an explicit "do not call it *the* conformance suite" line.

**Guide** (`docs/guide/indexing-in-a-browser-app/`): new *How your app learns the state moved* section — the union, the two-line rule, three transports one handler, the late-joiner rule, cross-tab wiring, and the cache wiring against TanStack Query's `invalidateQueries` with no dependency added.

**Two notes**: a finding on what the payload costs a normalised cache, and an observation on a rebuild/follow stalling within `finality` of a stream's start block.

Gate: `format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test` all green. `changeset status --since=main` cannot see the changeset while it is untracked (changesets reads changeset files *via git* since the ref); it resolves once the runner commits, as it has for every prior task.

## Decisions

**A new published package rather than a module inside one package.** The three transports live in two packages (`@etherfold/browser` × 2, `@etherfold/server` × 1) and nothing depends on both. `packages/browser/browser/hostingShapes.ts` (the other model the task named) is an in-package module because its three substrates are in one package; that shape cannot span two without inventing a browser↔server dependency edge. So I took the `@etherfold/state-store-conformance` shape instead: a package depending only on `@etherfold/core`, dev-depended by the two packages that own transports. Alternative considered: a private package dev-depending on browser *and* server so all three run in one file, which would make "identical outcomes" one run rather than two. Rejected because it forks the convention the repo already has for exactly this, and because a published suite is what lets the anticipated GraphQL subscription adapter — explicitly out of this repo's scope in ADR-0083 — check itself. Touches: the workspace, the release plan, `CONTEXT.md`'s glossary.

**The glossary name is "transport conformance suite", not "conformance suite".** `CONTEXT.md` already reserves **conformance suite** for the storage one and **conformance workload** for its subject. This is the same *shape* at a different *seam*, so it reuses the concept under a qualified name rather than re-meaning the bare term, and its entry says in as many words not to call it *the* conformance suite. Touches: every future artifact that says "the conformance suite".

**The adapter drives a REAL world; it is not a pipe over a synthetic publisher.** A pipe adapter (hand the transport a bare `StateMovedPublisher`) would be far cheaper and would give exact control over payloads, but acceptance criterion 5 needs a real read path on the browser transports, and a suite whose producer is synthetic cannot assert "a read does not answer from below the block you were told about". The cost is that each adapter carries a small world (chain/fold/port, or ingest/receiving container/SSE). Stated in `StateMovedTransport`'s docstring: an adapter that posts a value of its own passes every case while demonstrating nothing.

**The last chapter is claim-driven, and answering NEITHER question is a FAILURE, not a skip.** A transport whose reader has a state surface implements `readsUpTo` and is asked the coherence question; one with none (the server, whose query layer is deferred to `the-same-query-runs-against-a-worker-and-a-server`) implements `positionOnConnect` and is asked how a connecting reader converges. A transport offering neither fails a case saying so. This mirrors the store suite's rule for a backend that claims `singleWriter` and hands the suite no way to contend for it; a capability-driven selection that can select *nothing* is how a suite becomes decoration. Touches: whoever adds a fourth transport.

**The ADR's status line was not the literal string the task names, and I removed the whole frontmatter rather than editing it.** ADR-0083 no longer said `accepted, not yet implemented`: each task in the chain had amended it into a long "as built" narrative ending "The GraphQL subscription adapter is still not built". `ADR-FORMAT.md` says absence means accepted and current and that a status is added only when the plain reading would mislead, so the correct act is deletion. Because that line carried real as-built facts that were *not* in the body (the cross-tab naming rule, the SSE endpoint, `holdsStreamsAcrossRequests`, `coherenceNow`), I moved them into a body section rather than deleting them. Alternative considered: leave a `> **Built.**` marker as ADR-0073 does — rejected, since that ADR says itself it marks the exception, not the norm. Touches: anyone reading ADR-0083 for the built state.

**The guide attaches the handler to a PORT, and says the main-thread hook deliberately has none.** While writing it I found that `createIndexerState(...)` does not expose `onStateMoved` at all — the signal is on `IndexerPort`. That is correct rather than a gap (an app holding the hook has the indexer in its own heap and already has `state`), but it is not obvious, so the guide states who needs the signal and who does not, and shows `connectToIndexerHost(indexer.mainThreadHost(), {watch: false})` for the main-thread case. Touches: any future doc or example that reaches for `indexer.onStateMoved`.

**The cache wiring surfaced a real limit, and I wrote it down instead of smoothing it.** The two-line rule composes: the coarse line (`token changed → invalidate everything`) is one call in TanStack Query, Apollo and urql as they stand. The narrow line does **not** come free: `entities` carries entity *names*, which is the processor's vocabulary, and no cache library knows it — TanStack matches only if the app keyed queries by entity name (and silently no-ops if not), Apollo has no type-level invalidation primitive, urql needs the operations named. So the narrow half costs an app-declared mapping from entity name to that library's invalidation unit. That is a finding about the model, recorded in `work/notes/findings/what-the-state-moved-payload-costs-a-normalised-cache.md` and stated in the guide, with the reasons it is *not* an argument for shipping ids. What I checked is the libraries' documented invalidation surfaces; I did not build an app against any of them, and the note says so.

**Divergences found between the three transports.** No semantic divergence: the same handler produced the same decisions everywhere, and every case passes on all three. Three *legitimate* differences are now pinned as such rather than being allowed to become a second handler shape: (1) the server sends a `progress` frame on connect where the browser transports send nothing — that is ADR-0082's state-versus-notification split, and it is expressed as the claim-driven last chapter, not as a variation in `onStateMoved`; (2) only the browser transports have a read surface, same treatment; (3) *attaching* is synchronous in-heap and asynchronous over the network, so `StateMovedTransport.onStateMoved` admits a `Promise<Detach>` while `StateMovedHandler` itself is untouched — the handler is identical, the attach is not. One difference outside the signal, worth knowing: after a promotion the receiving container needs `rebuildMore` scheduled by the host before the fold that answers applies anything, where the browser container advances followers inside `indexMore`; the two adapters differ there and the notifications do not.

**A deliberate gap: no case for a block that is APPLIED but touches NO entity.** The prompt names "what an empty block does" as one of the four drift points. I did not build it. Both worlds' folds mutate on every event they carry, so producing one needs an event the processor has no handler for, which means a second ABI event — cheap in the browser fixture (mine, isolated) and not cheap on the server, where it means adding an event to `packages/server/test/utils/feedHarness.ts`'s shared ABI, used by 24 test files. Making it a case only two of three transports answer is exactly the capability fiction I refused one decision above, so I left it out entirely rather than half in. What is covered instead: the exact key set on both cases of the union, `entities` asserted as an array of strings (empty is legal), and the retraction — which carries no entity set at all — asserted whole. What is *not* covered by a test is a transport that swallowed a notification because it judged it empty; I verified by reading that none of the three does (the port forwards the value by reference, the channel posts it verbatim, the route `JSON.stringify`s it). A reviewer who wants this closed should say so: the honest fix is a second event on the server's shared harness ABI.


wighawag:

Reviewed against all 10 acceptance criteria. This is the fan-in that judges the other eight, and it does the adversarial job it was asked to do rather than declaring victory.

**The suite can fail, which is the only thing that makes it worth having.** The best part of this diff is not the suite but the meta-tests beside it: three deliberately-broken transports, each modelling an "improvement" somebody might genuinely make — replaying the last notification to a late joiner, re-stamping the coherence token per connection, dropping the retraction as an internal detail — each asserting *which named case* goes red. That proves every case is load-bearing and mirrors `state-store-conformance`'s lying-backend precedent instead of inventing a convention. A conformance suite that cannot fail is decoration, and this one is tested against that.

**Criteria 1 and 2:** the cases are data in a package parameterised by a transport factory, with the browser package running two transports and the server one. All four required outcomes are present and identical across all three: an append leaves the token alone, a retraction is carried whole with its fork point and no entity set, a promotion publishes nothing while the next notification wears an unheld token, and a reader that missed a notification (including a retraction specifically) converges on the next one alone. Criterion 3 holds in the strong form — `readerRule` is the same function attached on every transport.

**Two design decisions I want to credit:**

- *The adapter drives a real world rather than piping a synthetic publisher.* That is more expensive and it is correct: criterion 5 needs a real read path, and a suite whose producer is synthetic cannot assert "a read does not answer from below the block you were told about". The docstring names the failure it forecloses — an adapter that posts a value of its own would pass every case while demonstrating nothing.
- *The last chapter is claim-driven, and answering NEITHER question is a failure rather than a skip.* A transport with a read surface implements `readsUpTo`; one without (the server, whose query layer is deferred) implements `positionOnConnect`; one offering neither fails a case saying so. "A capability-driven selection that can select nothing is how a suite becomes decoration" is exactly right, and it matches the store suite's rule for a backend claiming `singleWriter` with no way to contend for it.

**On the ADR status line, the reasoning is right and matches mine.** I had independently found that the literal `accepted, not yet implemented` string was gone before this task started — the root task replaced it and each later one refined it — so criterion 6 as written was unsatisfiable. Deletion of the whole frontmatter is the correct act: `ADR-FORMAT.md` says absence means accepted and current, and that a status is added only where the plain reading would mislead. Importantly, the line had accreted real as-built facts not present in the body (the cross-tab naming rule, the SSE endpoint, `holdsStreamsAcrossRequests`, `coherenceNow`), and those were moved into a body section rather than deleted with it. Rejecting a `> **Built.**` marker was right too, since the ADR that uses one says itself it is the exception.

**The divergence hunt produced a real answer, not a shrug.** No semantic divergence: the same handler yields the same decisions everywhere. Three differences are pinned as legitimate rather than allowed to become a second handler shape — the server's connect-time `progress` frame (ADR-0082's state-versus-notification split, expressed as the claim-driven chapter), the presence of a read surface, and synchronous in-heap versus asynchronous network *attach* (so `StateMovedTransport.onStateMoved` admits a `Promise<Detach>` while `StateMovedHandler` itself is untouched — the handler is identical, the attach is not). Noting that the receiving container needs `rebuildMore` scheduled by the host where the browser container advances followers inside `indexMore`, and that the notifications do *not* differ there, is the kind of detail only this fan-in could see.

**The cache finding is exemplary.** Criterion 4 said that if the payload does not compose cleanly with a normalised cache, that is a finding to surface rather than smooth over. It was surfaced: the coarse line is genuinely one call in TanStack Query, Apollo and urql, while the narrow line costs an app-declared mapping from entity name to each library's invalidation unit, because `entities` is the processor's vocabulary and no cache knows it. TanStack's silent no-op when queries were not keyed by entity name is called out as the dangerous case precisely because it looks like a working integration. The note scopes itself honestly (documented surfaces read, no app built, nothing measured), argues why this is *not* a case for shipping ids, and the guide says the narrow half costs something rather than implying it is free. Finding the guide gap along the way — that `createIndexerState(...)` deliberately exposes no `onStateMoved`, because an app holding the hook already has `state` — is a good catch.

**The one declared gap, and why I am not blocking on it.** There is no case for a block that is applied but touches no entity, which the prompt named among its four drift points. I weighed this carefully. Criterion 1 enumerates exactly four required outcomes (append, retraction, promotion, dropped-notification convergence) and all four are delivered; the empty block is investigative guidance in the prompt rather than an acceptance criterion. The decision is declared rather than hidden, the reasoning holds (both worlds' folds mutate on every event they carry, so the case needs a second ABI event — cheap in the builder's own browser fixture, and on the server a change to a shared harness ABI used by 24 test files), and making it a case only two of three transports answer would contradict the capability-fiction rule this same task established one decision earlier. What is covered instead is stated (exact key sets on both union arms, `entities` asserted as an array of strings with empty legal, the retraction asserted whole), what is not covered is stated (a transport that swallowed a notification it judged empty), and the residual was verified by reading all three paths with each mechanism named — the port forwards by reference, the channel posts verbatim, the route `JSON.stringify`s. The honest fix is named. I am recording it as a follow-up worth tasking rather than treating a reasoned, documented scope decision as an unmet criterion.

Criterion 7 is met (`CONTEXT.md` gains the signal, the coherence token, the retraction-as-notification and the cross-tab channel), and the qualified glossary name "transport conformance suite" correctly avoids re-meaning the bare term the storage suite already reserves. Criterion 8 is met with a new guide section carrying the union, the two-line rule and the compare-never-parse warning. The done-move is clean and the suite genuinely executes in the gate through the two thin runners.

Both observation/finding notes filed during this task are real and correctly captured rather than fixed in place.

Merging — and with this, ADR-0083 is fully implemented.

PR #151, opened 2026-09-14
feat(run-registers-a-read-only-named-indexer): The combined run registe…

wighawag:
Task: `work/tasks/done/run-registers-a-read-only-named-indexer.md`

## What I did

**Drift check first.** The task's premises still hold: `IndexerRegistryEntry.liveIngestions` was required, `api/resolve.ts` still draws the three-way `501` / `404` / serve distinction, `indexCommand.ts` still writes its registration longhand with the "must not pull hono into a folding process" reason, and `run` still injected no registry. One stale claim I found and corrected rather than built on: `packages/cli/test/run.test.ts` asserted in prose that `--indexer` is *refused* by `run`, which `config.ts` has not done for some time (it is optional there, defaulting to `default`). ADR-0057 contains a sentence this task deliberately overturns, so I amended it in place rather than silently contradicting it.

**The seam** (`@etherfold/server`): `IndexerRegistryEntry.liveIngestions` is now OPTIONAL, and its absence is a capability statement — "this name accepts no ingestion" — documented against the empty-list meaning it must never be confused with. `indexerEntryOn` forwards it only where what it was handed answers it, exactly as it already does for `generations` / `promote` / `onStateMoved` / `coherenceNow`. Both ingest routes refuse on it with `501 ingestion-not-accepted`, decided before the body is read.

**The command** (`etherfold run`): registers the one name it folds under as a read-only entry (longhand, per `indexCommand.ts`'s reasoning about the lazy server import), forwarding `canonicalGeneration`, `generations`, `promote`, `onStateMoved` and `coherenceNow`, and omitting `liveIngestions`. Its feed, canonical view, state-moved stream and promote/revert route now answer; a push with a valid `INGEST_TOKEN` gets `501 ingestion-not-accepted`; an unknown name gets `404`; an anonymous caller still gets `401` first. A `feed:` line joins `status:` on startup.

**Tests** (written before the production code, both red first): `packages/server/test/aNamedIndexerThatAcceptsNoIngestion.test.ts` (reads answer, signal delivers, pointer surface answers, ingestion refused with a credential present, the three refusals are distinct, an empty live list still behaves exactly as it did, and `indexerEntryOn` forwards conditionally) and `packages/cli/test/runServesTheIndexerItFolds.test.ts` (the same claims through a real `run` process over real HTTP, including an SSE client watching a block this process fetched and applied itself). `index` is untouched; the whole suite (all packages) is green, as are `format:check`, `check:adr`, `check:refs`, `check:changesets` and `changeset status --since=main`.

**Docs kept true:** `CONTEXT.md`, both READMEs, the root README, `api/resolve.ts` / `api/ingest.ts` prose, and a dated amendment on ADR-0057. Changeset added naming `etherfold`, `@etherfold/server` and `@etherfold/platform-nodejs`.

## Decisions

**1. "Accepts no ingestion" is expressed by OMITTING `liveIngestions` from the entry, not by a new flag.** The codebase already reads an absent optional member as a capability statement (`generations?`, `promote?`, `onStateMoved?`, `coherenceNow?`), so this reuses that convention instead of forking a second one (`acceptsIngestion: false`, `readOnly: true`), which would have meant two ways to say a capability is missing on one type. Alternatives considered and rejected: a boolean field on the entry (a second vocabulary for absence); a host-level `ServerOptions` flag on the model of `holdsStreamsAcrossRequests` (wrong layer — it would be a statement about the deployment as a whole, and a later host wanting to accept pushes for one name and refuse them for another could not express it; per-entry absence makes that case free); and `liveIngestions()` answering `[]` (explicitly forbidden by the task and rightly: an empty list is a legitimate transient state on a host that does accept pushes). **Touches:** `IndexerRegistryEntry` (every host that builds one — `platforms/nodejs`'s passthrough type in its test needed a `NonNullable`), `indexerEntryOn`, both ingest routes, and `etherfold run`. **Residual risk, stated rather than hidden:** a host that simply *forgets* the member now silently becomes read-only instead of failing to compile. That is the same trade the four existing optional members already make, and the refusal is loud (`501` naming the reason and the command that does receive pushes) rather than silent at runtime.

**2. The ingest route answers `501 ingestion-not-accepted`, a THIRD refusal beside `ingestion-not-configured` and `unknown-indexer`.** `501` because it is the status every other capability refusal on this server already uses (`ingestion-not-configured`, `generations-not-held`, `state-moved-not-published`) and because nothing about the request is wrong and no retry helps. A NEW error code rather than reusing `ingestion-not-configured`, because the acceptance criterion is precisely that an operator can tell "this deployment does not accept pushes" from "this host has no registry at all" from "no such indexer here" — and existing tests assert the old code for the read-tier case, which is unchanged. Alternatives considered: `400 context-mismatch` with an empty `expected` (rejected — it is a payload refusal, and it would tell a sender its batch was wrong when nothing about it was), and `403` (rejected — this is not about who is calling; the caller is correctly authenticated). It is decided before the body is buffered, since no payload could change it. **Touches:** `packages/core`'s `createHttpIngestion` classification only in that this lands in the non-retryable 4xx/5xx handling it already has for `501`; no sender change was needed.

**3. `run` prints a `feed:` line at startup beside `status:`.** User-visible output change, small but not invisible: the route segment is the one thing an app pointed at a combined process must know and cannot guess (the name may have been defaulted to `default`). `index` already prints its ingest URL for the same reason. Alternative considered: print nothing and leave the name discoverable only through `--indexer` / the config docs — rejected as the worse default for the shape the milestone calls the one to reach for.

**4. No dedicated `readOnlyEntryOn` helper was added.** The statement at the seam is the ABSENCE, and `run` cannot import from `@etherfold/server` at assembly time anyway (the lazy-import reason `indexCommand.ts` records), so a third entry constructor would have had test-only callers while adding a new named concept ("read-only entry") to the package's API. A host holding a container that wants to state this passes `indexerEntryOn` what it wants answered, without that question; that path is asserted. **Touches:** anyone later adding a second read-only host (e.g. if `serve` is ever given a registry) — they get the seam, not a helper, and adding one then is cheap.


wighawag:

Reviewed against all 9 acceptance criteria, with both named traps checked first because either would have produced a green build that was wrong.

**Trap 1 avoided: the refusal is not faked with an empty list.** I grepped for `liveIngestions: () => []` across the CLI and server sources and found nothing. The statement is a genuine ABSENCE: `liveIngestions?()` is now optional, and both ingest routes check for the member itself rather than for an empty answer. That matters exactly as the task argued, because an empty list is a legitimate transient state on a host that does accept pushes, so overloading it would have made a permanent refusal indistinguishable from a momentary one.

**Trap 2 avoided, and asserted the hard way.** The refusal does not lean on `INGEST_TOKEN` being unset. The test environment is built with `INGEST_TOKEN: TOKEN` present, and there is a case named "refuses a push WITH A VALID CREDENTIAL, on the capability and never on a missing token". That is the assertion the task asked for, and it forecloses a door held shut by a missing environment variable.

Two cases beyond what I asked for, both good: an anonymous caller is still refused FIRST, so the capability is not something to probe (preserving the token-guard-ahead-of-capability-lookup ordering the package already had), and the deployment "folds nothing of the batch it refused", which proves the refusal is not accept-and-drop.

**Criterion 5 is met in the strong form.** There are now three distinguishable answers: `501 ingestion-not-accepted` (this deployment takes no pushes), `501 ingestion-not-configured` (no registry at all), and `404 unknown-indexer` (no such name here). A new code rather than reusing the existing one is the right call, since telling those apart IS the criterion, and existing tests still assert the old code for the read-tier case.

**Criteria 1 to 3** have cases: both feed views, the state-moved stream with the payload a split deployment serves, and the operator's pointer surface, which is the one that most obviously should have worked already given `run` is the shape that holds generations and may add one.

**Criterion 7 holds:** `indexCommand.ts` does not appear in the diff at all, so `index` registers exactly as it did.

**The ADR-0057 amendment is the best judgement in this diff.** Rather than quietly changing behaviour the ADR describes, it names the conflation: "registers no named indexer at all" was how the refusal happened to be EXPRESSED, not the reason for it. The reason (this command withholds INGESTION, because a remote sender pushing into a process already fetching for itself is a second writer) is restated as still true and unchanged. It also records that `serve` still answers `501`, since a read tier registers nothing, which stops the next reader assuming this generalises.

**The residual I weighed and accept.** Making a required member optional means a host that simply forgets it now silently becomes read-only instead of failing to compile. That is a real loss of type safety and the decisions block states it plainly rather than burying it. I accept it on the grounds given: it is the same trade the four existing optional members already make, the failure is loud at runtime with a `501` naming both the reason and the command that does receive pushes, and the alternative (keeping the member required and adding a boolean) would have meant two vocabularies for absence on one type. The rejected host-level flag is rejected for the right reason too, since it could not express a host that accepts pushes for one name and not another.

Declining to add a `readOnlyEntryOn` helper is also correct: the statement at the seam is the absence, so a third constructor would have added a named concept to the package's API with test-only callers.

The `feed:` startup line is a user-visible output change, small and justified (the route segment is the one thing an app pointed at a combined process cannot guess, especially when the name defaulted), and it is flagged as such rather than slipped in.

Merging.

PR #152, opened 2026-09-14
feat(a-replay-never-asks-below-a-streams-start-block): A replay never a…

wighawag:
Task: `work/tasks/done/a-replay-never-asks-below-a-streams-start-block.md`

Done. Working tree holds only this task's changes, and the full gate (`format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test`) is green.

## What I built

**The floor.** `getFromBlock` (`packages/core/src/internal/engine/utils.ts`) now floors at `defaultFromBlock` on the level branch, which is the same number its `latestBlock === 0` branch already returned. Both terms of the `min` are untouched: a level fold still deliberately reaches back by `finality`.

**Verification of the premise the task asked me to check, rather than assume.** The floor's correctness rests on three things I checked in the code, not on the observation's summary:

- `installStreamSeed` *refuses* a seed whose `coverage.fromBlock` is above `reachBackTo`, which defaults to `defaultFromBlockOf(source)`. So a seed-installed stream opens at or *below* the floor, and lower is always served. The divergence the task warned about exists but only downward, where it is harmless.
- `getFromBlock` is computed on **both halves of the wire** from the source alone (`StreamBuilder.expectedFromBlock`, and `generateStreamToAppend`'s `UnexpectedFromBlockError`, which is ADR-0004's resumption protocol). A floor derived from the stream is not available to the side holding no keeper, so it is not a candidate.
- The stall is *permanent* for a follower and a rebuild, not momentary as the observation supposed: their `latestBlock` comes from the read that was refused, so the chain moving on never changes the comparison. Confirmed by making the two fixtures reproduce it.

**The window property, asserted rather than argued** (`test/utils.test.ts`): the floor never clamps past `lastToBlock + 1` over a grid of cursors, every window block is inside the clamped range, and a contradiction at a window block in a level fold still concludes the retraction through `generateStreamToAppend`.

**The two seams.** `test/follower.test.ts` — a follower level with its own stream's start block still follows what the writer appends (and `keyedStream` now models `startBlock`, which is why nothing in that suite could meet this before). `test/rebuild.test.ts` — a rebuild's *second* chunk, resuming from the checkpoint the first committed, is served rather than stopped; it sits beside the existing "opens too high" case, which still reports `does-not-reach-back` with `retryCanAdvance === false`.

**The two fixture leads (criterion 6).** Both were only working around this defect, and I proved it: with each lead removed and the old floor in `dist/`, the server fixture fails with `a generation never finished rebuilding` and the browser one with `the fold did not reach block 101`; with the new floor both pass. Both leads are removed, and the comments now say why running *without* one is what keeps the fix asserted from outside core.

**Docs.** `StreamRead`, `ReplayRead` and `RebuildStop` now say what `does-not-reach-back` means after the change; `CONTEXT.md`'s **follower** entry no longer says a seeded generation is the shape that produces it. Changeset added.

## Decisions

**Which floor: the SOURCE's earliest block (`defaultFromBlock`), not the stream's own start.** The task said the stream's start is the right floor *if the two can differ in a way that matters*. They can differ, but not in a way this fix can use, and not in a way that matters: `installStreamSeed` refuses a seed opening above `reachBackTo` (defaulting to `defaultFromBlockOf`), so an installed stream opens at or below the floor; and decisively, `getFromBlock` is the number *both halves of the wire* derive from the source alone (`StreamBuilder.expectedFromBlock`; `generateStreamToAppend`'s `UnexpectedFromBlockError`), so a floor derived from a stored stream would make the sender and the receiver disagree about one cursor and break ADR-0004's resumption protocol. The alternative considered and rejected was passing the stream's `startBlock` into `getFromBlock` (or re-reading from `startBlock` after a refusal): both require relaxing that from-block check, which is the one refusal that keeps a batch from applying into a hole. **Touches:** every fold's read start, `StreamBuilder.expectedFromBlock`, `installStreamSeed`'s `reachBackTo`, and both stream keepers' `startBlock` check.

**How the two kinds of "does not reach back" are distinguished: by making the young one stop producing the verdict, not by adding a variant to it.** I deliberately did not add a new `RebuildStop` reason or a new field. After the floor, the read start is never below `defaultFromBlock`, which is the lowest block *any* fold over this source ever asks from — so a stream that served the first load serves every later one, and `does-not-reach-back` is left meaning exactly one thing: a stored stream no fold over this source can be whole on. The two answers a caller branches on are therefore the existing union members: `absent`/`nothing-stored` (the writer has not appended yet, retry) and `does-not-reach-back` (intervene), and `retryCanAdvance` gives the right answer for both without change. I considered splitting the verdict on `startBlock <= foldedThrough + 1` and rejected it: **both** halves recur for ever (a refused read never moves `latestBlock`), so a second reason would buy no scheduling difference while adding a concept overlapping one ADR-0069/ADR-0070 already own. **Touches:** `StreamRead`, `ReplayRead`, `RebuildStop`, `retryCanAdvance`, and the CLI's `newlyStalledFollowers` (all unchanged in shape).

**The residual I did not build, stated so it is not mistaken for coverage.** One shape still reports `does-not-reach-back` where this particular fold has no hole: a stream that opens *mid-history* (above `defaultFromBlock`) read by a fold that has already folded past its start. That is reported as needing a human, which is right — a mid-history subtree is the damage `startBlock` exists to catch — but the fold in question is individually whole. Serving it would mean reading from `max(fromBlock, startBlock)`, which needs the from-block refusal relaxed, and that is a different task. No ADR: the change is one expression, reversible, and carries no residual trade-off; what it *did* narrow is a claim ADR-0070 makes in passing ("a seeded generation is the shape that produces it"), which I corrected in the type docs and `CONTEXT.md` rather than by amending the ADR.

**Test-double fidelity, changed as part of the fix.** `keyedStream` in `follower.test.ts` now records `startBlock` and answers `does-not-reach-back`, as both shipped keepers do. A double that served whatever it held regardless of the block asked for could not express the refusal at all, which is why the core suite never met this. **Touches:** every existing case in `follower.test.ts` (all still green).


wighawag:

Reviewed against all 8 acceptance criteria. This changes the read-start rule for every fold in the system, so I weighted the two things that could have gone quietly wrong: the choice of floor, and whether clamping can skip a block a reorg reaches.

**The floor was verified rather than assumed, and the argument is better than the one the task supplied.** I asked for the caveat to be checked (`defaultFromBlock` is the SOURCE's earliest block, while the defect is about the STREAM's start, and the two can diverge for a seed-installed stream). The answer establishes that they can differ but not usably, for a reason I had not seen: `getFromBlock` is the number **both halves of the wire** derive from the source alone, so a floor read off a stored stream would make the sender and the receiver disagree about one cursor and break ADR-0004's resumption protocol. The seed path is separately safe, since `installStreamSeed` refuses a seed opening above `reachBackTo`. The rejected alternative is rejected for the right reason too: passing the stream's `startBlock` in would require relaxing the from-block refusal, which is the one check that keeps a batch from applying into a hole.

**Criterion 3, the property this change could have broken, is properly pinned.** Three cases carry it: the read start still never goes past `latestBlock - finality`, it still reaches back over the WHOLE unconfirmed window once the fold is clear of the start, and, decisively, a block inside the window that a reorg reaches is still re-read **with the floor in force**. That last one is the case I would have written, and it is the difference between arguing that clamping upwards cannot skip anything and demonstrating it.

**Criterion 2 is covered on both paths, which matters because they fail differently:** a follower "KEEPS FOLLOWING while it is level with its own stream START BLOCK", and a rebuild "KEEPS ADVANCING" in the same state. A fold reading its own cached stream could always self-heal by clearing; a follower holds a read-only view whose clear is a no-op by design, so it was the one that could stall for ever.

**Criterion 4 solved by elimination rather than by a new variant, and I prefer it.** Instead of adding a `RebuildStop` reason, the floor means the young-stream case stops producing the verdict at all, leaving `does-not-reach-back` meaning exactly one thing: a stored stream no fold over this source can be whole on. So a caller branches on the union members that already exist, and `retryCanAdvance` is correct for both without change. The rejected split is rejected on a real observation, that both halves recur for ever because a refused read never moves `latestBlock`, so a second reason would have bought no scheduling difference while adding a concept ADR-0069 and ADR-0070 already own.

**Criterion 6 produced the outcome I hoped for.** The fixture lead was a workaround, and it is gone. The comment previously insisted the tip running `FINALITY` blocks ahead "is load-bearing rather than decoration"; it now says "The lead was not a property of this world, it was a workaround." A fixture that had to be shaped around a defect is a good signal that the defect was real, and removing the shaping is how you prove it is fixed.

**The best catch in the diff is not in the fix at all.** `keyedStream` in `follower.test.ts` never recorded `startBlock` and served whatever it held regardless of the block asked for, so it could not express the refusal that both shipped keepers make. That is why the core suite never met this bug: the double was more permissive than the real thing. Tightening it to match is properly part of the fix, and every existing case in that file stays green.

**The residual is stated rather than left to be discovered:** a stream opening mid-history, read by a fold already past its start, still reports `does-not-reach-back` although that fold individually has no hole. Reporting it is defensible, since a mid-history subtree is exactly the damage `startBlock` exists to catch, and serving it would need the from-block refusal relaxed, which is a different task. Correcting ADR-0070's passing claim in the type docs and `CONTEXT.md` rather than amending the ADR is the right weight for a claim made in passing.

Merging. This closes the last of the observations that had a task.

PR #153, opened 2026-09-14
feat(a-successor-that-was-never-canonical-is-superseded): A successor t…

wighawag:
Task: `work/tasks/done/a-successor-that-was-never-canonical-is-superseded.md`

All green. Here is the report.

## What I did

**Drift check first.** The launch prompt's premises hold against the tree *as re-scoped*: the task body at `work/tasks/backlog/a-successor-that-was-never-canonical-is-superseded.md` matches the prompt I was given (option B, the narrowed in-memory predicate), and I verified its claims in the code rather than trusting them: `GenerationRecord` really is `GenerationId & {createdAt}` (`packages/core/src/generation/registry.ts:58`), the durable row really is `{stream, processor, createdAt}` (`packages/server/src/generations.ts`), `everCanonical` really is in memory (`receivingContainer.ts`, ADR-0057 consequences), `maxStreams` really is counted over the distinct streams among registered generations, and a cap really refuses and never evicts. Nothing had superseded it. Two staging notes for the runner/human, neither blocking: the item is at `work/tasks/backlog/` (not `ready/`) with `needsAnswers: true` and an unanswered `work/questions/task-a-successor-…md` sidecar whose Q1 the re-scope commit has in fact answered.

**The change** (`packages/core/src/receivingContainer.ts`): `add` now drops every **abandoned successor** before it registers the new one.

- The predicate is `successorsAddedHere` minus `everCanonical`: a generation *this* container registered, after `open`, that was not canonical then and that the pointer has not named since. A generation this process only *resolved*, or never saw at all, is never touched.
- `everCanonical` became a set of generation identities (not fold objects) and is now filled in one place, `noteCanonical`, i.e. at every point the container reads or writes the pointer, so a move another process made counts the moment this one sees it.
- Cross-stream by design; the drop is **declined**, and said out loud, while the abandoned fold writes a stream another held fold follows (`dropSuperseded`'s ADR-0044 rule, with the fold about to be added counted as a follower). The pass walks newest-first so an abandoned follower + its abandoned writer both go in one call.
- Each drop deletes the registry row, drops the state namespace (ADR-0053) and reaps the stream when no registered generation is left on it, and is reported through `named-logs`.

**Tests.** `packages/cli/test/anAbandonedSuccessorIsDropped.test.ts` — the seam the task named: one container over one real libSQL database with real per-generation table namespaces, the stored emission stream and the SQL registry. Seven cases covering all criteria (registry contents after a run of changes, the disk coming back, the predecessor + revert surviving, both caps never reached, the decline, and the restart residual). I confirmed it fails without the production change (6 of 7 red; the 7th is the never-touch-the-canonical safety assertion, which must pass either way).

**Standing conventions.** Changeset added (`@etherfold/core` minor, `@etherfold/server` + `etherfold` patch for the touched test files); `CONTEXT.md` gains an **abandoned successor** glossary entry. `pnpm format:check && check:adr && check:refs && check:changesets && build && typecheck && test` are all green. (`changeset status --since=main` reads changesets from *git*, so it only sees the new file once the runner commits; the packages it names match exactly the packages this change touches.)

## Decisions

**The concept is named an ABANDONED successor, not a "retired" or "superseded" one.** Both of the task's own words are already taken in this system and mean the *opposite* case: `supersededRecord` / `dropSuperseded` / the `CONTEXT.md` "drop-on-promotion" entry use *superseded* for the generation a PROMOTION moved the pointer off, and "the retired generation" is used in the same sense in ~8 prose sites (`container.ts:59`, `browser/IndexerState.ts`, `browser/host/serve.ts`). Both of those are RETAINED because a revert moves back to them; this one is DELETED because nothing can. Reusing either word would make one term mean a thing and its opposite. I kept the verb the system already has for this (`DROP`, as in ADR-0053 and `dropSuperseded`) and changed only the adjective, and wrote the distinction into the `CONTEXT.md` entry. Touches: the glossary, the changeset, the new test file name, and any later artifact that reuses the term.

**The stream IS reaped when nothing is left folding it, which departs from one sentence of the Prompt** ("leave the STORED STREAM alone: the cap is reclaimed even though the raw logs are not"). I used `registry.deleteGeneration` unchanged, which reaps exactly when the dropped generation was the last registered on its stream. Leaving an unclaimed subtree alive breaks the invariant the sweep exists to maintain ("a stream subtree no registered generation claims is what the sweep collects"), and it is reachable in the very loop this task is for: the developer flips the source back, a fresh generation registers on that stream, is the only one on it, is therefore its writer, has an empty cursor, and appends the whole stored range a second time — the silent second history ADR-0052/ADR-0055 exist to prevent. Alternatives considered: a `reapStream: false` option on `deleteGeneration` (rejected: it buys only "the raw logs survive until the next open sweeps them" and pays for it with that corruption), and keeping the subtree plus a guard against re-registration (rejected as a new mechanism for a case the reap already answers). Touches: `maxStreams` relief is unaffected (the cap counts registry rows either way); what changes is that a flip back to a previous source re-fetches that stream instead of re-folding it.

**The retirement is reported through `named-logs` only, not through a new return type or signal.** `add` still returns `HeldFold`, `onStateMoved` still carries nothing about it. Alternatives: returning a drop report from `add` (a breaking signature change for a fact no caller acts on), or publishing it as a state-moved notification (wrong: nothing a reader can see moved, and rotating the coherence token would have every reader throw its cache away because a *second* generation was tidied up). The log line names the dropped generation, the successor that took its role, why it is safe, and whether the stream was reaped — which matches how `dropSuperseded` and writer succession already report. Touches: `/status`'s `generations` array is the existing place an operator sees the shorter list.

**`everCanonical` is keyed on the generation identity and filled on every pointer READ, not on the moves this process makes.** Previously it was a `Set<HeldFold>` written only by `movePointer`/`add`. Keyed on objects it would read a *second fold for a generation already registered* (a reconfigure back to a still-registered fold resolves rather than creates) as one the pointer had never named — which is precisely how the predicate would have deleted a predecessor kept for a revert. Filling it in `noteCanonical` means a pointer move made by another process counts the moment this container reads it. Side effect, deliberate and in the safe direction: `movePointer` now classifies a move back to a generation this container merely *observed* as canonical as a REVERT, so drop-on-promotion declines there too. Touches: `dropSuperseded`'s `wasRevert`, and ADR-0057's recorded consequence (unchanged in direction, only more observant).

**`successorsAddedHere` records only a generation this container actually REGISTERED, which costs one extra `registry.list()` per `add`.** `create` resolves-or-creates, so without the pre-read a reconfigure onto an identity another process registered would be marked as this container's own and become droppable. `add` is called at open and at reconfigure, so the read is not on any hot path.

**A same-stream churn settles at three generations, not two, and that is the residual.** When the abandoned fold writes the stream the newcomer is about to follow, the drop is declined, so `incumbent + that writer + the newest` coexist. With `SERVER_GENERATION_CAPS` (four) the loop never meets the bound, which is what the criteria ask; a host that states `maxGenerations: 2` would still wall on that shape. The alternative — dropping the writer first and letting the newcomer become the writer of a stored stream — is the double-append above. Touches: any host choosing its own caps, and the browser's tighter numbers if the chain-facing container ever adopts this rule.

**Scope is the RECEIVING container only.** `Indexer` (`container.ts`, chain-facing, what a browser tab runs) is unchanged. The task names the receiving container, its caps and its `add`, and the seam it asks to be tested at is the receiving one; porting the rule to the chain-facing twin is a separate change with its own `everCanonical` shape (a flag on the entry) and its own cap arithmetic.

**Two existing tests encoded the old behaviour and were adapted rather than deleted.** `packages/core/test/rebuild.test.ts` ("orders across STREAMS too") counted three registry rows after three adds; it now registers its third generation through `resolveGeneration` so all three coexist, since its subject is how `create` stamps `createdAt` while the others are present. `packages/server/test/twoNamedIndexers.test.ts` ("a cap refuses in that one only") reached `maxStreams` with a successor the pointer had never named; it now promotes that successor first, so the two generations it counts are ones nothing may reclaim, and its actual subject (whose registry the bound is counted over) is untouched. Related, worth knowing: deleting the newest generation lets the next `create` re-issue its `createdAt`; ADR-0072's invariant (no two *surviving* records tie, so `writerOf` is never ambiguous) still holds, because `create` takes `max(now, newest surviving + 1)`.

**`scripts/check-work-refs.mjs` now exempts `work/questions/`, which is a repo-wide gate change I did not set out to make.** The `verify` gate includes `check:refs`, and it was already RED on `main` for four dead citations inside two question sidecars — verbatim bounce reports quoting the paths items were at when they were raised, one of which exists *precisely* to report that the launch snapshot and the item's real folder disagreed. Left alone it would bounce this task (and every other) for an unrelated reason. Rewording the quotes would falsify the record, which is the script's own stated argument for exempting `.changeset/` and `work/notes/observations/`, and the script's failure message prescribes exactly this fix for a historical surface. Alternative considered: fixing the four citations in place (rejected: it edits another task's sidecar and destroys the observation one of them makes). Touches: every future item with a question sidecar — dead `work/` paths inside one are no longer reported.


wighawag:

Reviewed the diff against the acceptance criteria and against the three hazards this task was flagged for. I verified the load-bearing claims in the code rather than trusting the report.

**The predicate (the whole safety argument).** Correct, and defended three ways rather than one. `successorsAddedHere` minus `everCanonical` is "never been canonical" and not "not canonical right now": a predecessor kept for a revert is unreachable because this container never registered it (`registeredBefore.some(...)` excludes a `create` that merely RESOLVED an existing row), and `dropAbandonedSuccessors` additionally asserts `sameGeneration(fold.record, canonicalNow)` directly rather than deriving safety from the set. Re-keying `everCanonical` from fold objects to identities, and filling it in `noteCanonical` at every pointer READ, is a genuine strengthening: a pointer move another process made now counts the moment this one sees it.

**Not cap-pressure eviction.** Confirmed. `dropAbandonedSuccessors` runs unconditionally before `registry.create`, gated only on `opened` and on there being an incumbent, never on how full the registry is.

**Caps unchanged.** Confirmed mechanically: the diff of `receivingContainer.ts` contains no `GENERATION_CAPS` line at all.

**Cross-stream rule.** Implemented as the re-scope specified, and the test asserts the `maxStreams` half separately, which is what makes it a property rather than a claim.

**On the one departure from the prompt, which I accept.** The prompt said to leave the stored stream alone; `deleteGeneration` reaps it when no registered generation is left on it. The departure is right and the reasoning given is sound: an unclaimed subtree is exactly what the sweep exists to collect, and it is reachable in this task's own loop (flip the source back, the fresh generation is the only one on that stream, so it is the writer, with an empty cursor, and appends the whole stored range a second time). It is also consistent with the task's own criterion, which says the stream is untouched *if anything else still needs it*, and that condition is enforced by `wouldStrandAFollower`. My prompt sentence was stricter than the criterion and would have introduced the double-append ADR-0052/ADR-0055 exist to prevent. Flagging loudly rather than silently complying was the right call.

**On the `check:refs` exemption, which is a gate change.** I verified independently that `check:refs` is RED on current `main`, for dead citations inside the two `work/questions/` sidecars the runner's own surface step wrote (they quote the launch-snapshot paths, which have since moved). The script's failure message prescribes precisely this remedy for a historical surface, so this is the prescribed fix rather than a weakened gate. Recorded as a finding: the surface mechanism breaks `check:refs` on main every time it fires.

**Tests** are at the seam the task named (one container over a real libSQL database with real per-generation namespaces), and cover the criteria individually, including the two that matter most: the predecessor surviving with the revert still reaching its state, and the restart residual. Changeset present and correctly scoped.

The `needsAnswers: true` still on the done record and the now-answered question sidecar are protocol housekeeping rather than anything about this diff; I am handling both post-merge.

PR #154, opened 2026-09-14
feat(an-endpoint-triggers-a-reconfigure-in-a-running-process): An endpo…

wighawag:
Task: `work/tasks/done/an-endpoint-triggers-a-reconfigure-in-a-running-process.md`

All checks green. Final summary:

## What I built

**The trigger a reconfigure never had: `POST /{indexer}/admin/reconfigure`.**

`@etherfold/server`
- `IndexerRegistryEntry.reconfigure?()` — a new OPTIONAL capability, and `ReconfigureReport`, the three answers it gives. Absent is a CAPABILITY statement (`501 reconfigure-not-held`), independent of `generations`/`promote`, forwarded by `indexerEntryOn`.
- The route, under the existing `/:indexer/admin/*` guard, so `ADMIN_TOKEN` is inherited rather than invented. It decides only the transport's two questions (who may call, which status each answer is) and resolves no module: the package still names no runtime.
- `packages/server/test/aReconfigureReachesARunningDeployment.test.ts` — the three answers, the two 501 shapes, the credential, and "the incumbent still answers across every outcome", over a real container and a real server.

`etherfold` (CLI)
- `src/reconfigure.ts` — the re-read: re-resolve config, re-import the module through a cache-busting importer, re-resolve the source, compute the identity, and `container.add` only when it differs. Calls are serialised. Nothing restarts, re-opens or tears down.
- `src/folding.ts` — `foldPartsFor`/`FoldParts` and `openIndexingSource` factored out of `openFolding`/`prepareIndexing`, so a re-read builds its successor the one way this deployment builds folds.
- `run` registers the capability; `build`, `index`, `serve` do not.
- `packages/cli/test/anEndpointReconfiguresARunningRun.test.ts` — a real `run`, a real processor module **on disk**, edited between calls. Mutation-checked: removing the cache-busting makes the edit invisible (`unchanged`) and four of five cases fail.

Also: changeset, `CONTEXT.md` glossary, and the now-false "no route adds a generation" comment in `driveCycles` corrected.

Two observation notes (signals spotted, not fixed): a premise error in the source note, and a load-sensitive browser test flake.

## Decisions

**The task's hazard 2 is factually wrong, and I built on the corrected premise rather than the stated one.** The task and `a-reconfigure-cannot-reach-a-running-run.md` say the processor half of a generation identity is a hash of the handlers' SOURCE TEXT, and conclude that "an ordinary edit does change source text, so this is not the common case". It is not: the identity is `getVersionHash()` = `${version}-${simple_hash({entities, config})}`, and `getCodeFingerprint()` is a separate ADVISORY value that `packages/core/src/utils/fingerprint.ts` deliberately keeps out of the identity (a recorded deviation from ADR-0008). `CONTEXT.md`'s own `generation` entry already has it right. I did **not** stop, because the correction does not change what to build — it strengthens the reason the endpoint must report `unchanged` distinctly, which the spec's story 4 already demanded — and folding the fingerprint into the identity to "fix" it is the one thing `fingerprint.ts` argues against by name. What it changes is the FREQUENCY: `unchanged` is what an ordinary handler edit gets, not the rare case, so the `unchanged` message names `version` as the thing to bump. Alternative considered and rejected: making the identity fingerprint-sensitive (invalidates every deployment's state on a minifier change). Touches: the source observation note and the task body, both of which now mislead (observation note filed).

**A failed re-read is `409`, not `400`/`500`/`503`.** The request is well formed (`400` is wrong), the capability is present (`501` is taken), and this is not a host fault (`500` is alarming and wrong). What is true is that the deployment's current state conflicts and an identical re-send after the next build succeeds, which is RFC 9110's `409` and is exactly what this repo already spends `409` on as ADR-0004's one resumable refusal. Touches: what a watcher branches on.

**The three outcomes ride in the BODY (`outcome: registered | unchanged | failed`), not in status-code subtleties.** `200` for both successes rather than `201`/`204`, because `201` implies a Location and a watcher distinguishing outcomes by status nuance is a worse contract than a named discriminant. `registered` and `unchanged` both NAME the generation in the pointer move's own shape (`{stream, processor, digest}`), so the value answered is the value an operator matches or promotes.

**The type is `ReconfigureReport`, not `ReconfigureOutcome`.** Coherence check: `ReconfigureOutcome` is already taken in `@etherfold/core` for the IN-PLACE verb's discard verdict, and `@etherfold/browser` already met this exact collision and sidestepped it (`HostReconfigure`, with a JSDoc saying why). I initially wrote `ReconfigureOutcome`, caught it against the glossary, and renamed. The wire field stays `outcome` because that is the caller-facing word. Touches: `@etherfold/core`'s type of the same name, `@etherfold/browser`'s `HostReconfigure`.

**The module cache is defeated by a cache-busting file-URL query, not a fresh worker.** Cost, stated rather than discovered: every reload adds a module instance to the ESM registry and none are ever collected, since the registry is keyed by URL for the life of the process. That is nothing for a dev loop and a slow unbounded growth for a long-lived process poked frequently, so a production caller should trigger per BUILD, not per minute. The alternative (a worker per reload) costs a process boundary the fold would then have to cross. An INJECTED `importModule` wins untouched, because a test double is stating what comes back. Touches: `LoadProcessorModuleOptions.importModule`, and any future decision to run the fold out-of-process.

**An identity this process already holds a FOLD for is answered `unchanged` WITHOUT calling `add`.** `add` does not deduplicate: the registry resolves an identity it already holds, but the container still builds a second fold over the same state and starts re-folding the stream into it beside the first. So the identity is computed first and compared against `container.held()`. Without this, a watcher calling on every save would quietly accumulate duplicate in-memory folds writing into one namespace. Touches: `ReceivingIndexer.add`'s resolve-rather-than-create behaviour.

**A re-read whose config names a different INDEXER or a different DATABASE is `failed`, not silently ignored.** Those two decide where rows go and what key they carry, and neither is re-appliable on a running process. Every other re-read input (node URL, rps, port) is deliberately NOT refused and is documented as belonging to the fetcher and server this call does not rebuild — refusing on those would be noise. Touches: `resolveCommandConfig`, and the repo's "nothing is accepted and ignored" rule.

**Scoped to `run`; `index` does not get it.** `index` holds a processor path and a container and could answer the capability nearly free, but its source is explicit and its fold is driven by a remote writer, so it is a different story with its own refusals. The capability is an entry method, so `index` can gain it later with no change to the route. It currently answers `501 reconfigure-not-held`, which is honest. Touches: a possible follow-up task.

**A filter/source change registers a generation this process's single fetcher will not feed.** Pre-existing and out of scope (`a-deliberate-freeze-is-not-visible-on-a-filter-change`): `run` builds one fetcher over one scalar source. The processor-only case (the dev loop this exists for) is a follower on the same stream, advanced by the scheduled rebuild and auto-promoted — asserted end to end.


wighawag:

Checked the diff against the acceptance criteria and against the three hazards this task was flagged for, verifying the load-bearing claims in the code rather than trusting the report.

**No teardown.** Confirmed mechanically: `packages/cli/src/reconfigure.ts` contains no `dispose`, `close`, `stop`, `destroy`, `restart` or `reopen` call at all; the only matches are the comments explaining why it registers beside and never re-opens. The server test asserts the incumbent answers across *every* outcome, including the failure path, which is the property the whole feature exists to preserve.

**The module cache is genuinely defeated.** A unique query on the module's file URL, supplied through the existing `LoadProcessorModuleOptions.importModule` seam so an injected importer still wins. The cost is stated rather than discovered: one uncollected module instance per reload, so a production caller should trigger per build rather than per minute. The CLI test proves the cache-bust is actually exercised rather than merely present, since a cached module would return the old identity and the assertion would fail.

**A failed import changes nothing.** Ordering is right: the import is caught and returns `failed` before `container.add` is ever reached, so nothing partial can be registered, and both failure shapes are covered (a module that throws and one that cannot be parsed). Calls are also serialised through a queue, so a watcher firing twice cannot race.

**On the task's hazard 2, which the build found to be false, and which I verified.** `fingerprint.ts:14` states that the code fingerprint "is **advisory** and stays out of `getVersionHash()`", so the identity is the declared version plus a hash of entities and config, not the handlers' source text. The task and its source observation both had this backwards. Not stopping was the right call: the correction strengthens rather than undermines the reason `unchanged` must be a distinct outcome, and folding the fingerprint into the identity is the one remedy the codebase argues against by name.

**The residual this creates, stated plainly because it is the one thing a reader of the task will not expect.** Since a bare handler-body edit does not change the identity, the common dev-loop edit answers `unchanged` and the developer must bump `version` for the reload to register a generation. That is a property of the pre-existing identity scheme rather than anything this endpoint does, it is reported by the endpoint itself (the `unchanged` message names `version` as the remedy), and an observation note records it. It does leave the spec's headline story needing one extra step, which is worth a follow-up decision but not a reason to hold this PR.

Everything else lands as asked: three distinguishable outcomes carried as a named discriminant in the body, the admin credential with the ingest token refused, the honest `501` for a host that cannot serve it, a burst of six saves leaving one successor beside the incumbent (the end-to-end assertion of the supersession rule that landed in #153), and a correctly scoped changeset.

PR #155, opened 2026-09-14
feat(the-cli-selects-its-promotion-policy): The CLI selects its promoti…

wighawag:
Task: `work/tasks/done/the-cli-selects-its-promotion-policy.md`

## What I built

An operator can now select the promotion policy on `etherfold run`, and the CLI reports back which one is in force.

**The input** (`packages/cli/src/config.ts`): `--promotion <on-catch-up|immediate|manual>` with `PROMOTION_POLICY` behind it, plus `--drop-on-promotion` (flag-only boolean). Both go through the existing `INPUTS` / `OWNERSHIP` / `REFUSALS` tables, so they get the same flag-beats-environment rule, the same blank-variable-is-unset rule, and the same refusal shape as every other input. A value nobody recognises is refused naming the three that exist and the default; nothing given resolves to `undefined` rather than to `'on-catch-up'`, so the default stays written only in `@etherfold/core`'s `resolvePromotionConfig`.

**The refusal** (`resolvePromotion`): `--promotion immediate` with `--drop-on-promotion` is refused by the pure resolver, before a module is imported, a database opened or the chain dialled, naming what to use instead. The container's own `refuseImmediateDrop` stays where it is.

**The plumb**: `RunConfig.promotion` → `openFolding`'s context → `openReceivingIndexer({promotion})`. One path, no second read of the container's configuration.

**The visibility**: `/status` gained a `promotion: {reported, policy, dropOnPromotion}` field, built exactly like the existing `fetcher` field — a new `ServerOptions.getPromotionPolicy` reporter, a `StatusPromotion` envelope that degrades to `{reported: false, reason}`, a passthrough on `platforms/nodejs`, and `run` injecting `container.promotion` (the *resolved* value, so the default is reported as a value rather than as an absence). Absent entirely on hosts that decide no promotion.

**Tests** (written before the production code): resolution, precedence, refusals and ownership in `packages/cli/test/configuration.test.ts`; `packages/server/test/theDeploymentReportsItsPromotionPolicy.test.ts` for the envelope; and `packages/cli/test/theDeploymentSelectsItsPromotionPolicy.test.ts`, which stands a real `run` up under each policy, introduces a successor by editing the processor module on disk and calling `POST /{indexer}/admin/reconfigure`, and asserts *when the pointer actually moves* (immediate: already moved when the call returns, with the successor having folded nothing; default: incumbent answers complete until level, then moves; manual: still incumbent once level, and moves when an operator asks). I checked these are not vacuous: reverting the one-line plumb fails 5 of the 8.

Docs: `packages/cli/README.md` (run flag table, the `/status` field, the variables table) and `packages/server/README.md` (the new reporter). Changeset: `.changeset/the-cli-selects-its-promotion-policy.md`.

`pnpm build && pnpm typecheck && pnpm test` green (3,769 tests), plus `format:check`, `check:adr`, `check:refs`, `check:changesets`.

I also dropped one out-of-scope observation at `work/notes/observations/an-index-process-can-never-finish-a-processor-upgrade.md`: an `index` process restarted with a changed processor registers the successor and then nothing ever arms, advances or promotes it, and `indexCommand.ts`'s header claims an affordance (a batch creating a generation) that no code path implements.

## Decisions

**1. `--drop-on-promotion` is exposed too, not just the policy.** Acceptance criterion 4 requires the `immediate`+dropping combination to be refused as a configuration error, and with only the policy exposed `dropOnPromotion` is always `false`, so that combination is unreachable and the criterion is vacuous. The alternative was to expose only the policy and drop criterion 4; I rejected it because the container's own refusal message says "use `on-catch-up` with `dropOnPromotion`", which is unreadable advice to an operator who cannot set it. Touches: a new user-visible flag that can destroy the revert path (story 4), default `false` and argued by ADR-0046, so the unsafe direction stays opt-in.

**2. Only `run` owns the input; `build`, `index`, `fetch` and `serve` REFUSE it by name.** The task says "every command that holds generations accepts it", but the policy only ever speaks when a successor is added to a *running* container, and `container.add` is reached from exactly one place in the tree: `reconfigure.ts`, wired only by `run` (`build` stops at the tip and never calls `rebuildMore`; `index` wires no reconfigure route and schedules no rebuild; `serve` folds nothing). Accepting it elsewhere would ship the accepted-and-ignored flag into the one module whose header forbids it. Alternative considered: accept on `index` and `build` anyway, for symmetry. Rejected on ADR-0048's own recorded asymmetry — refusal→optional is additive later, removing an accepted flag is breaking. Touches: four commands' ownership rows and refusal messages, and whichever future task gives `index` a reconfigure (it will need to flip one cell).

**3. A new environment variable, `PROMOTION_POLICY`.** Criterion 1 requires selection "by a flag or by the environment". ADR-0048's *rule* puts it there (the environment carries what varies between deployments of one image; the same processor image iterated locally versus shipped is exactly that), though the ADR's prose enumerates a list of six that was already stale before this change (`PRUNE_INTERVAL`). I updated the stale count/claim in `config.ts`'s own doc comment for `INPUTS`, and did not edit the ADR. Touches: the deployment contract of the `etherfold` image.

**4. `--drop-on-promotion` gets no environment variable.** It is a boolean, and this module has never had a truthiness spelling for the environment — the only other boolean, `--no-auto-setup`, is flag-only for that reason. Inventing `DROP_ON_PROMOTION=true|1|yes` would add a parsing convention for one input. Alternative: give it one and pick a convention. Reversible if a containerised deployment needs it.

**5. The policy is reported on `/status` as a top-level `promotion` field, not inside the `cursor` envelope.** The `cursor` envelope is about progress, and a promotion policy is not progress; putting it there would re-mean that field. Instead it mirrors `getFetcherLimits` exactly — an injected reporter for a deployment-level fact only the process holding the thing can answer, typed rather than opaque (it is `@etherfold/core`'s `UsedPromotionConfig`, not a host's private vocabulary), degrading to a reason rather than an omission. Alternative considered: `GET /{indexer}/admin/canonical-generation`, which is the operator's pointer-move route; rejected because the policy is a process-level configuration rather than a fact about one named indexer, and `/status` is the page an operator already watches a rebuild on. Touches: `@etherfold/server`'s `ServerOptions` and `platforms/nodejs`'s `StartOptions` (both additive and optional).


wighawag:

Checked the diff against the acceptance criteria and against the two hazards this task was flagged for.

**No development default, which was the hazard most likely to be got wrong.** Verified mechanically rather than by reading the prose: the only occurrences of `NODE_ENV`, `import.meta.env` or a dev sniff anywhere in the diff are documentation sentences stating that no such inference exists and will not be added. Better than that, `resolvePromotionConfig` returns `undefined` when nothing was configured, so a deployment that selected nothing passes no promotion config at all and the default stays written in exactly one place (`generation/promotion.ts`). The CLI does not restate it, which is the strongest possible form of "the default is unchanged".

**The refused combination is refused at start-up, naming what to use instead.** `config.ts:842` raises during resolution, before anything is opened, and the message names both escapes (`on-catch-up` with the drop, or `immediate` while retaining) and the reason (ADR-0046's deferral is not built on this runtime). Tested both ways: the pair refused, and either half accepted on its own.

**The tests are behavioural, which is what makes this task honest now and would not have been possible a week ago.** Each policy is stood up on a real deployment and the POINTER is asserted: `immediate` has already moved it by the time the reconfigure answers, the default keeps the incumbent answering until the rebuild is level, and `manual` leaves it alone even once the successor is level and moves it when asked. That is exactly the seam the original bounce said could not be written honestly, and it is only writable because the reconfigure endpoint landed first. The resolution rules are covered separately, including the subtle ones: a blank variable reads as unset, an ambient variable is ignored by a command that does not own the input, and an unrecognised value is refused naming the three that exist.

**On "every command that holds generations accepts it", where the build made a judgement I agree with.** `index` holds generations but REFUSES the flag, alongside `build`, `fetch` and `serve`. That is not a criterion missed but the repo's stronger rule applied: an `index` process can never apply a policy, so accepting the input there would be the accepted-and-ignored flag this task exists to avoid. The build did not just assert that, it found out WHY and filed the observation: `applyPolicyTo` returns while `opened` is false and `indexCommand.ts` schedules no `rebuildMore`, so a restarted `index` never even arms its successor as a candidate and can never finish a processor upgrade at all. That is a real gap worth its own task, and recording it rather than papering over it with a flag was the right call.

Reporting, docs, both READMEs, `/status` (including degrading to a reason rather than failing the page), and a correctly scoped changeset are all present.

PR #156, opened 2026-09-16
feat(a-reload-that-changed-nothing-reports-processor-drift): A reload t…

wighawag:
Task: `work/tasks/done/a-reload-that-changed-nothing-reports-processor-drift.md`

## What I built

**The task's premises held against current reality** (I checked `fingerprint.ts`, `reportProcessorDriftIfAny`, the receiving container, the reconfigure module, ADR-0008 and the two unaccepted ADRs 0084/0085): the receiving side really did write `processorFingerprint` into every cursor and compare none of them, and `unchanged` really was one reading with two causes. Nothing had drifted, so I built it.

**One comparison, shared by four surfaces** — `packages/core/src/processorDrift.ts` (new). `processorDriftReport(...)` turns a pair of fingerprints into a `ProcessorDriftReport` or `undefined`, and holds every caller to the three rules: either side missing is *unknown* and never *drifted*; the message is a question for the author (it names minification / transpiler / comment edits and says the fingerprint is advisory); nothing here acts. `announceProcessorDrift` logs at `error` and contains a throwing listener. The chain-facing `IndexerGeneration` now builds its report through it instead of its inline copy, so there is exactly one `PROCESSOR DRIFT` phrasing.

**The receiving container reports in its own right** — `StreamBuilder` and `GenerationRebuild` compare on the branch where a cursor is ADOPTED (both engine shapes, because which one a fold is in is registration order, not anything an author did), report **once per engine** rather than per batch (a receiver re-reads its cursor on every call), refresh nothing stored, and relay through the new `ReceivingIndexer.onProcessorDrift`, the same field name the chain-facing container publishes. There is deliberately no receiving twin of `strictProcessorDrift`: a receiver that threw would refuse a sender for a condition the sender cannot see or fix.

**The endpoint's `unchanged` splits into two readings** — `ReconfigureReport`'s `unchanged` arm carries `drift?: ProcessorDriftReport`, the admin route returns it verbatim beside the message, and the CLI's reconfigure builds it from the two things it already holds. Three outcomes stay three; nothing is registered or discarded because of drift (asserted in both suites, and the drifted reload returning the *same* generation digest is itself the assertion that no identity contains the fingerprint).

**Tests** (written before the production code, at the two seams the task named): 7 new cases in `packages/core/test/receivingContainer.test.ts` (drift on a restart with an edited handler, advisory end to end, the stored fingerprint not refreshed so the next boot reports again, once-per-receiver, genuine no-op silent, both *unknown* cases silent, a bumped version silent, a throwing listener contained), 3 in `processorDrift.test.ts` pinning the shared builder including the reload wording, and 2 in `packages/cli/test/anEndpointReconfiguresARunningRun.test.ts` driving a real deployment with a real module edited on disk (`version` static, handler body changed → `unchanged` + `PROCESSOR DRIFT` + nothing registered + the feed still served; and an un-edited reload reporting no drift).

Gate: `pnpm build && pnpm typecheck && pnpm test && pnpm format:check` all green (core 1176 passed, cli 243 passed), `check:adr` / `check:refs` / `check:changesets` green, changeset added, working tree carries only this change. `CONTEXT.md`'s `generation` entry was amended so the glossary still describes the endpoint truthfully. Note that `work/notes/observations/the-reconfigure-note-conflates-the-version-hash-with-the-code-fingerprint.md` already records this gap's "what stays OPEN" half, so I added no new note; this closes the reporting half of it (bumping `version` is still the author's step, by design).

## Decisions

**Which two fingerprints each surface compares, and it is said in the type.** The receiving container asks the BOOT question: the STORED cursor's fingerprint against the code loaded now ("the state I am about to serve was computed by different logic"), matching `IndexerGeneration`. The reconfigure endpoint asks the RELOAD question: the RUNNING fold's loaded fingerprint against the module just re-imported ("the code I have just read differs from the code I am running"), because that is what a developer who pressed save is asking, and it is the only one true of an edit made after a process came up with no drift. They can disagree, so I made the answer explicit rather than implied by provenance: `ProcessorDriftReport` gains `compared: 'persisted-state' | 'reloaded-module'`. Alternative considered: use the stored fingerprint on the endpoint too (one comparison, but it answers the wrong question and would stay silent for exactly the developer this task is for); or leave the question implicit in the prose (a host routing both into one alert channel could not tell them apart). Touches: `ProcessorDriftReport` consumers (core only, plus the new `ReconfigureReport.drift`).

**`storedFingerprint` renamed to `previousFingerprint`.** On the reload question nothing about that value is stored — the running fold's cursor may carry a third value — so keeping the old name would make one field mean the cursor in one report and a loaded module in another, which is the coherence failure the brief warns about. Alternative considered: a discriminated union with different field names per arm (honest, but duplicates the shape and forces every consumer to switch). This is a breaking type change to a published `0.x` package with no external consumers; it is in the changeset. Touches: `packages/processor-sqlite/test/version.test.ts` and `packages/core/test/processorDrift.test.ts` (both read the report).

**Drift is carried ON `unchanged`, not made a fourth outcome, and it replaces the message rather than appending to it.** Nothing was registered and nothing changed, which is exactly what `unchanged` means; a fourth outcome would be the first step towards acting on an advisory signal. The drift message already states that the identity did not move, that nothing was registered and what to do about it, so appending the generic explanation would repeat the only sentence that matters. Absence of `drift` covers both "the code is identical" and "one side cannot be fingerprinted", deliberately — I did not invent a third reading for *unknown*. Touches: `ReconfigureReport` (`@etherfold/server`), the admin route's JSON body, any watcher reading it.

**`processorDriftReport` is exported from `@etherfold/core`; `announceProcessorDrift` is not.** The reload comparison can only be made where a module is re-imported, which is the CLI, and a message assembled there would be a second `PROCESSOR DRIFT` phrasing for an operator to grep. Where a report GOES, though, is the caller's: the CLI logs through its own logger and hands the report back on its response.

**The receiving container relays drift from EVERY held fold, unlike the chain-facing one, which filters to the canonical entry.** A generation that answers no read yet is precisely the one being built to answer them next, and a report suppressed until promotion would arrive after the upgrade it was about; `processorHash` names which fold it is about. Touches: `ReceivingIndexer.onProcessorDrift` consumers (none today; the log is the deployment-facing surface).

**`FoldParts` gains `codeFingerprint`** in `packages/cli/src/folding.ts`, beside `versionHash`, taken from the author's own declared object exactly as `EntityEventProcessor.getCodeFingerprint()` takes it — one formula, one spelling — so the re-read compares like with like. It enters no identity and nothing branches on it. Touches: `openFolding` and `reconfigure.ts`, the two callers of `foldPartsFor`.


wighawag:

Reviewed the diff against the nine acceptance criteria and against the three traps the prompt named. Gate 2 is off in this repo (`dorfl.json` `review: false`), so this is the only review layer above the acceptance gate; I read the code rather than the report.

## Per criterion

1. **A drifted no-op reads differently from a genuine one.** ✅ `ReconfigureReport`'s `unchanged` arm carries `drift?: ProcessorDriftReport`, the admin route returns it verbatim, and `anEndpointReconfiguresARunningRun.test.ts` asserts both readings against a real module edited on disk: the handler-body edit answers `unchanged` **with** `drift.compared === 'reloaded-module'`, the un-edited reload answers `unchanged` with `drift` undefined and no `PROCESSOR DRIFT` in the message.
2. **The answer names the remedy.** ✅ Both arms of `whatDiffers` end in "Bump the processor's `version`", and the tests assert `message` contains `version` rather than only that two hashes differ.
3. **The receiving container reports in its own right.** ✅ `StreamBuilder.reportProcessorDriftIfAny` and `GenerationRebuild.reportProcessorDriftIfAny` both compare on the branch where a cursor is **adopted**, and `announceProcessorDrift` logs at `error` unconditionally, so a deployment that never calls the endpoint still learns from its logs. Covering **both** engine shapes is the right call and is justified in the code: which shape a fold is in is registration order, not anything an author did.
4. **The fingerprint stays ADVISORY, asserted directly.** ✅ This is the criterion the whole shape rests on and it is asserted, not argued: `generations()` still `['v1']` after a drifted boot, `world.cleared` is `[]`, `rowsIn(namespace)` is unchanged, and at the endpoint the drifted reload returns the **same generation digest** as the incumbent, which is itself the assertion that no identity contains the fingerprint. Nothing in `processorDrift.ts` acts.
5. **Either side absent reports NOTHING and is never drift.** ✅ Enforced in one place (`processorDriftReport` returns `undefined` on a falsy side) and asserted in both directions: a cursor written before the field existed, and a processor that cannot fingerprint itself.
6. **A genuine no-op is silent.** ✅ Asserted, plus the separate case that a *bumped* version is silent (an upgrade is never a drift).
7. **One vocabulary, reused.** ✅ Verified by reading the deletion, not the claim: `indexer.ts` loses its entire inline message construction, its own `namedLogger.error` and its own listener try/catch, and now calls `processorDriftReport` + `announceProcessorDrift`. `strictProcessorDrift` correctly throws `report.message`. There is exactly one `PROCESSOR DRIFT` string in the tree for an operator to grep.
8. **Tests.** ✅ 7 cases in `receivingContainer.test.ts`, 3 in `processorDrift.test.ts`, 2 end-to-end in the CLI suite.
9. **Changeset.** ✅ Present, and it declares the `previousFingerprint` rename rather than burying it.

## The prompt's three traps

- **Which two fingerprints.** Decided explicitly and, better, made legible in the type: `compared: 'persisted-state' | 'reloaded-module'`. The boot question stays the containers'; the reload question is the endpoint's, which is the one a developer who pressed save is actually asking. Making the distinction a field rather than leaving it implicit in provenance is the right resolution of "they can disagree" — a host routing both into one alert channel can now tell them apart.
- **Do not refresh the stored fingerprint.** ✅ Honoured, and asserted across two boots (`fp-A` still on the cursor, both boots report). This is the property that keeps the report repeating until the author acts.
- **Resist making drift do anything.** ✅ Nothing registers, nothing discards, no re-fold. `GenerationRebuild`'s docstring names the temptation and rejects it.

## Notes, non-blocking

- `storedFingerprint` → `previousFingerprint` is a breaking type rename. It is right (on the reload question nothing about that value is stored, so the old name would mean the cursor in one report and a loaded module in another), it is confined to `@etherfold/core` consumers inside this repo, and it is declared in the changeset as a `minor` on a `0.x` package. No external consumer exists yet.
- Reporting from **every** held fold rather than only the canonical one is a deliberate divergence from the chain-facing twin and is argued convincingly: the generation being built to answer reads next is exactly the one whose drift matters, and `processorHash` names which fold it is about.
- `CONTEXT.md`'s `generation` entry was amended in the same change, so the glossary does not go stale on the day the behaviour lands. Good.

No criterion unmet, no drift note violated, nothing advisory turned load-bearing. Merging.

PR #157, opened 2026-09-16
feat(a-successor-lands-in-a-durable-slot-that-holds-one): A successor l…

wighawag:
Task: `work/tasks/done/a-successor-lands-in-a-durable-slot-that-holds-one.md`

Done. Everything is green: `format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, and the full test suite (18 packages).

## What I built

**Drift check first:** the task still matches reality. ADR-0084 is accurate against the code, `a-successor-that-was-never-canonical-is-superseded` landed exactly as described (in-memory `everCanonical` + `successorsAddedHere`, narrowed predicate, stated restart residual), and no later ADR moved the ground. ADR-0084's `status: accepted, not yet implemented` is left in place (arming, collection and the chain-facing twin follow).

**The registry now holds three durable named slots** (`packages/core/src/generation/registry.ts`). `GenerationRegistryState.canonical` becomes `slots`, `GenerationRegistryWrite.canonical` becomes `slots` with three cases (absent leaves, `null` clears, an identity assigns). New surface: `SLOT_NAMES`, `SlotName`, `GenerationSlots`, `SlottedGenerations`, `SlotAssignment`, `slotHolding`, `registry.slots()` and `create(id, {slot})`. `create` keeps the first-generation-is-canonical rule, leaves a generation some slot already names where it is, and otherwise assigns the named slot, replacing what it held. `moveCanonicalTo` assigns `predecessor` and empties `successor` in the same commit as the pointer write. Deletions clear the slots that named what they removed.

**All three substrates carry it**: memory, IndexedDB (one small record per slot beside the entries, outside the entry range) and SQL (`_generation_pointer` renamed to `_generation_slots`, a column pair per slot on the one guarded row, `SCHEMA_VERSION` still 1, no migration).

**`ReceivingIndexer` registers into `successor`** and replaces what it finds there, with the drop reusing `deleteGeneration` unchanged (row, state namespace, stream reaped, decline while a held fold still follows the stream). **Both in-memory sets are deleted**: `everCanonical` and `successorsAddedHere` are gone, `dropAbandonedSuccessors` is now `replaceTheSuccessor`, and promotion-vs-revert is read from the slot. Nothing of the ever-canonical set is still earning its keep here; the chain-facing `Indexer` keeps its own flag until its own task.

**Tests**: `packages/cli/test/aSuccessorLandsInADurableSlot.test.ts` at the asked-for seam (a container over a real libSQL database, plus a SECOND CONTAINER over the same substrate as the restart), asserting the slot ROW, the restart replacement, eight restarts holding at one successor, `predecessor` assigned on promotion and unreachable from a replacement, the assignment-not-identity case, the caps, and the reported replacement. It replaces `anAbandonedSuccessorIsDropped.test.ts`, whose last case asserted the residual this removes. Registry rules are covered substrate-neutrally in `packages/core/test/generationRegistry.test.ts`, and the IndexedDB key layout in `packages/browser/test/generationRegistry.test.ts`.

**Docs**: changeset added; `CONTEXT.md`'s "abandoned successor" glossary entry becomes **generation slot** (recording what it replaced and why); ADR-0057's `everCanonical`-is-in-memory consequence carries a dated amendment.

Note: `pnpm changeset status --since=main` reports "no changesets found" only because it reads `git diff` and the new changeset is still untracked; it resolves once the runner commits.

## Decisions

**A slot is represented as a `slots` MAP on the registry state, replacing the single `canonical` field, rather than two fields added beside it.** ADR-0084's claim is that `canonical` is *merely the first slot*, and two representations of one kind of fact drift. The write side needs three cases, not two, so `SlotAssignment` distinguishes absent (leave), `null` (clear) and an identity (assign) — `COALESCE`-style "absent means leave" cannot express a clear, and a deletion must be able to clear the slot that named what it removed. Alternative considered: keep `canonical` at the top level and add `successor`/`predecessor` beside it (rejected: it makes `canonical` a special case forever, which is exactly what the ADR argues against). Touches: every `GenerationRegistryPort` implementation (the three in this repo are updated) and anything constructing a port in a test.

**The SQL table is RENAMED `_generation_pointer` → `_generation_slots`, with a column pair per slot on ONE row, and `GENERATION_SLOT_TABLE` replaces the exported `GENERATION_POINTER_TABLE`.** A table called "pointer" holding three pointers would make the name say the opposite of the decision. One row rather than a slot table keeps the promotion shuffle a single guarded write (ADR-0054's reasoning for putting `revision` there), and keeps the commit at the same statement count. The three-case write is expressed as a bound TOUCHED FLAG per slot inside the upsert's `CASE`, because `COALESCE` cannot tell "leave" from "clear". `SCHEMA_VERSION` stays 1 and no migration is written, per `CONTEXT.md` (nothing published, no persisted state), and because `predecessor` is unreconstructible for a pre-slot registry anyway. Alternatives: keep the name and widen it (rejected: the name would lie), or a child table keyed on slot (rejected: two statements and a second guard for no gain). Touches: `@etherfold/server`'s exports, the wrangler D1 path (same static `db.sql`), and four tests that named the table.

**The promotion shuffle lives INSIDE `registry.moveCanonicalTo`, and `predecessor` ALWAYS names the generation the pointer moved off — with no direction inference anywhere.** Atomicity is the argument: a crash between a pointer write and a separate `predecessor` write would leave a moved pointer with no record of what it moved off, which is the missing fact all over again. The uniform rule means a revert is itself undoable (after reverting to A, `predecessor` names B) rather than needing an exception clause for "a move back". Alternative considered: CLEAR `predecessor` on a move back to it (rejected: it adds a direction test to the one place that should have none, and it makes the generation just reverted away from collectable immediately, which is a surprising amount of destruction for an undo). Touches: the chain-facing `Indexer`, which shares this registry and now gets `predecessor` assigned by its own pointer moves — harmless, and a head start for its own task; and the next task's arming rule, which must read `successor` and never `predecessor`.

**Drop-on-promotion now fires only when the target is what `successor` names; every other move drops nothing.** That includes an operator promoting a generation no slot names, which the old in-memory rule could classify as a forward move (and therefore drop the incumbent) whenever this process had not seen the pointer on the target. The new rule is strictly more conservative in the destructive direction and strictly more accurate in the retaining one (a restarted process no longer misreads a genuine promotion as a revert). Touches: ADR-0057's recorded consequence, which I amended in place rather than leaving false.

**What a registration may DISPLACE is bounded to two things: the generation the `successor` slot named (whether or not a fold for it is held — this is the restart case and the whole point), and held folds that no slot names (which is what a declined drop leaves behind).** An unslotted row this container holds no fold for is deliberately NOT dropped: reclaiming those is the operator verb `a-generation-no-slot-names-is-reclaimed-on-request`, and a registration silently deleting rows it never touched would pre-empt that task's decision about where the blast radius lives. The second clause is load-bearing rather than cosmetic: without it the decline case would strand a generation until the reclaim verb existed, and a source-change loop would still reach `maxStreams`.

**A FAILED drop does not refuse the registration.** The arriving generation takes `successor` regardless, and what failed to go is left named by no slot — collectable, retried by the next registration or by the reclaim verb. The alternative (refuse the `add`) would turn a stuck deletion into an outage for the deployment trying to move forward, and the slot is what makes the lenient path safe: the failed generation is no longer claimed by anything. It is logged at error level naming both facts.

**A DECLINED drop (it writes a stream another held fold follows) also leaves the generation unslotted rather than keeping it in the slot.** The slot holds at most one and the newcomer takes it; keeping the declined one there would mean two pending successors in one slot, which is the property this task exists to establish. The consequence is that it is collected by the next registration that moves off its stream (asserted), or by the operator verb.

**The superseded test file was DELETED rather than kept beside the new one.** `packages/cli/test/anAbandonedSuccessorIsDropped.test.ts` asserted, as its last case, that a restart drops nothing — the exact residual ADR-0084 removes — and its other cases are this task's acceptance criteria under the old vocabulary. All of them are carried over into `aSuccessorLandsInADurableSlot.test.ts` with slot-row assertions added. Touches: the old changeset and the earlier done record cite that path; `check:refs` skips a tracked path absent from the working tree, and both citations remain historically accurate about what was built then.


wighawag:

This is the foundation of ADR-0084 and four tasks rest on it, so I reviewed it against the three traps the prompt singles out before the criteria, and I checked the durability claim by following it to the schema rather than by finding the word "durable" in a type.

## The three traps

**1. The slot is beside the record and NOT inside `GenerationId`.** ✅ Verified by reading the type, not the prose: `GenerationId` is still exactly `{stream, processor}` and `GenerationRecord` is still identity + `createdAt`. The slots live on `GenerationRegistryState.slots` as a `GenerationSlots` map of assignments. This is the trap that would have turned the same content under two slots into two generations, two namespaces and two folds of one stream, and it is avoided structurally. The SQL confirms it: the slot columns are on `_generation_slots`, keyed by indexer, and nothing was added to `_generations`.

**2. `predecessor` is ASSIGNED by the promotion, never inferred.** ✅ The assignment is `slots.predecessor = identityOf(movedOff)` INSIDE `registry.moveCanonicalTo`, in the same guarded commit as the pointer write. That is the right place and the reasoning given for it is the correct one: a crash between a pointer write and a separate predecessor write would leave a moved pointer with no record of what it moved off, which is the non-derivable fact all over again. Two edges handled well: a move onto where the pointer already sits assigns nothing (no generation becomes its own predecessor), and a promotion clears the `successor` slot. Nothing anywhere derives which generation a revert would want.

**3. The in-memory machinery is DELETED, not left agreeing with the slot.** ✅ Checked by grep, not by report: `everCanonical` and `successorsAddedHere` no longer exist as fields in `receivingContainer.ts`. What stands in their place is a `WHAT USED TO BE HERE` comment block naming both and saying what reads them now, which is the right residue. `movePointer` now decides promotion-vs-revert by reading the slots before the move (`wasPromotion = successor names this id`), and the comment makes the strictly-better argument explicit: the old set was empty after a restart, so a restarted process read every move as a revert and a genuine promotion dropped nothing. `candidates` correctly REMAINS — deleting it is the next task's criterion, not this one's.

## Per criterion

1. **A run of N changes leaves incumbent + ONE.** ✅ Asserted, including over a run of differing shapes.
2. **It holds ACROSS A RESTART.** ✅ This is the criterion that distinguishes the task from the in-memory rule already shipped, and it is asserted the right way: a genuinely fresh container over the same rows, with the test's own comment stating that this one registered nothing and remembered nothing. It asserts the slot contents read back out of the DATABASE, that the replaced generation's namespace tables are actually gone, and that the incumbent still answers. There is also a run-of-restarts case for the redeploy-per-commit loop, and a restart on the canonical generation that displaces nothing.
3. **The incumbent answers throughout and `canonical` is never touched.** ✅ Asserted separately, stream included.
4. **A generation a revert needs is never replaced.** ✅ Asserted directly, and the test is the best one in the file: it promotes, checks `predecessor` is the superseded incumbent, then changes its mind twice more and shows the predecessor surviving both — with the comment making the point that "not canonical right now" is precisely the wrong predicate. It then drives a real revert back and asserts the STATE (`ownerOf(...)` is still ALICE), so the way back is shown to be real rather than merely named. A restart registering into the slot is covered as its own case.
5. **A slot is an assignment, not an identity.** ✅ Asserted: naming what a slot already holds makes no second generation.
6. **A replacement is REPORTED.** ✅ Asserted, naming what went, what took its place and why it was safe.
7. **The redundant machinery is deleted.** ✅ See trap 3.
8. **Caps UNCHANGED, and churn no longer approaches them.** ✅ `BROWSER_GENERATION_CAPS` is untouched (the diff to `OnIndexedDB.ts` contains no cap change at all), and both halves are asserted: registering through a run of processor changes no longer refuses at `maxGenerations`, and a source change frees the STREAM slot too, which is the half that is easy to miss.
9. **Tests / changeset.** ✅ Both.

## Durability, checked properly

The lesson I applied here deliberately: a task that says "durable" has to be followed to the schema, the row type and every substrate, because an identifier existing proves nothing. It holds. `_generation_pointer` becomes `_generation_slots` with six real nullable columns (a pair per slot) plus the existing `revision` guard; `OnIndexedDB.ts` and `memory.ts` both carry the same three-case assignment; and the `SlotAssignment` type distinguishes absent (leave), `null` (clear) and an identity (assign), which is what makes "a deletion clears the slots that named what it removed" expressible at all. Keeping all six columns on ONE row so the promotion shuffle is a single guarded write is the correct call and is argued from ADR-0054.

## Notes, non-blocking

- **On ADR-0084 being "a net deletion of machinery":** it is a net deletion of INFERENCE and a net addition of LINES, and I think that is correct rather than a miss. The two in-memory sets are gone and the container no longer guesses a move's direction; what was added is a schema, three substrate implementations and a three-case write type. ADR-0084 itself makes no literal line-count claim (its argument is against a durable `everCanonical` COLUMN, which it rejects for answering a harder question than the one that needs answering), and that argument is respected here: the slot records what a generation is FOR, not how it arrived.
- **ADR-0057 was amended in place.** Correct call and correctly executed: an `AMENDED 2026-09-16` blockquote appended per `ADR-FORMAT.md` with the original bullet left intact, and it is precise about scope — the receiving container's flag is gone, the chain-facing one survives until `the-chain-facing-container-holds-its-generations-in-slots`. Leaving that bullet reading as though `everCanonical` were still in memory everywhere would have been the silent falsehood.
- **ADR-0084's `status: accepted, not yet implemented` is correctly LEFT IN PLACE**, per this task's explicit instruction. It is the chain-facing twin's to remove, and I have planted that pointer in `the-chain-facing-container-holds-its-generations-in-slots` (commit `f50b1e90`), since nothing in that task's body said so.
- **The superseded test file was deleted rather than kept.** Right: its last case asserted that a restart drops nothing, which is the exact residual this task removes, so keeping it would have meant asserting the old behaviour and the new one in one suite.
- Bounding what a registration may displace to "what `successor` named" plus "held folds no slot names", and explicitly NOT reaping unslotted rows this container holds no fold for, is a good boundary: it leaves the blast-radius decision to `a-generation-no-slot-names-is-reclaimed-on-request` instead of pre-empting it.

No criterion unmet, no trap triggered. Merging.

PR #158, opened 2026-09-16
feat(a-generation-no-slot-names-is-reclaimed-on-request): A generation…

wighawag:
Task: `work/tasks/done/a-generation-no-slot-names-is-reclaimed-on-request.md`

Everything is green: `pnpm build`, `pnpm typecheck`, `pnpm test`, `pnpm format:check`, `pnpm check:adr`, `pnpm check:changesets`, `pnpm check:refs`.

The task's premises hold against the code today. `a-successor-lands-in-a-durable-slot-that-holds-one` landed (commit `b647fb84`) with exactly the shape this task assumes: three durable slots in the registry (`SLOT_NAMES`, `GenerationRegistry.slots()`, `slotHolding`), `predecessor` assigned by the pointer move, and the drop path in `ReceivingIndexer` (`deleteGeneration` → row + state namespace + stream reaped, declined where a held fold follows the stream). Its done record even reserves the unslotted rows for *this* task ("An unslotted row this container holds no fold for is deliberately NOT dropped: reclaiming those is the operator verb"). ADR-0084's cap consequence still names the missing verb. Nothing superseded. Built.

## What I did

**Core (`@etherfold/core`)**
- `unslottedGenerations(generations, slots)` in `generation/registry.ts`: ADR-0084's collection rule in one place (a refcount over the slot rows; `canonical` needs no extra clause because it *is* a slot).
- `GenerationDeletion.records`: how many substrate records a reaped stream subtree held, so a report can say what actually came back (`StreamDeletion` already reported this; `deleteGeneration` was throwing it away).
- `ReceivingIndexer.slots()` (the SEE half) and `ReceivingIndexer.reclaim()` (the verb), plus `ReclaimReport` / `ReclaimedGeneration` / `DeclinedReclaim`. `reclaim` reuses `deleteGeneration` and the existing `wouldStrandAFollower` decline; it is newest-first so a replaced follower goes before the writer it would strand. Extracted `stopDriving(record)` (shared with `dropReplaced`) so a reclaimed fold stops being driven and stops reporting.

**Server (`@etherfold/server`)**
- `IndexerRegistryEntry.slots?()` / `reclaim?()`, optional and paired, forwarded by `indexerEntryOn`.
- `POST /{indexer}/admin/reclaim-generations` — no body, on `ADMIN_TOKEN`, `501 reclaim-not-held` where a host holds no registry, 200 with `outcome` / `reclaimed` / `declined` / `slots` / `message`.
- `GET /{indexer}/admin/canonical-generation` widened: `slots`, `unslotted`, and a `slot` per listed generation.

**CLI** — `run` and `index` forward `slots` / `reclaim` on their registry entry.

**Tests** — `packages/cli/test/aGenerationNoSlotNamesIsReclaimed.test.ts` (9 cases) at the seam the task named: one container over a real libSQL database holding canonical + pending successor + predecessor + a generation no slot names, asserting which survive, that the namespace is really gone and the stream really reaped, that the revert target still answers from its own state afterwards, the decline, the caps, and `nothing-to-reclaim`. `packages/server/test/theOperatorReclaimsWhatNoSlotNames.test.ts` (7 cases) asserts the HTTP affordance: the listing, the report, the guard (anonymous and ingest-token both `401`), the `404`, and the `501`.

**Docs** — changeset; ADR-0084's cap consequence amended in place (the verb it named as missing now exists, and where); `CONTEXT.md`'s `generation slot` entry, the two package READMEs and the root README.

## Decisions

**The verb is an HTTP admin route (`POST /{indexer}/admin/reclaim-generations`) and NOT a sixth CLI verb.** ADR-0057 already decided this class of question and I followed it rather than re-opening it: a Cloudflare Worker is reachable only over HTTP, so a CLI verb serves no serverless deployment at all, and the command set is pinned at five names that denote *deployment intents* — a one-shot control action is not one. The blast-radius argument the task raises is real and is answered where ADR-0057 put the answer: the credential is `ADMIN_TOKEN`, which fails closed and is deliberately never the ingest token a log shipper holds, so the surface that DELETES state is the last one a fetcher can reach. Alternative considered: a CLI verb (cannot be reached remotely, cannot be misauthorised) — rejected because an affordance that exists on one deployment shape is not an affordance, and because the CLI still inherits the route by hosting the same app, so the operator of `run`/`index` gets it anyway. Touches: `IndexerRegistryEntry` (two new optional capabilities), `run` and `index` wiring, and the Worker host, which answers `501 reclaim-not-held` because it registers a bare receiver.

**Acceptance criterion 7 ("follows the rules every other command input obeys, ADR-0048") is satisfied by having NO input rather than by ADR-0048 literally.** ADR-0048 governs CLI flags and their environment fallbacks; with the verb on HTTP it has no flag to name. The verb takes no body at all (the rule decides which generations go, so there is nothing for a caller to name and nothing to get wrong), and it follows the refusal shapes this surface already established — `401` on the path guard, `404` for an unknown name, `501 <capability>-not-held`, in the `reconfigure-not-held` / `generations-not-held` family. The task prompt explicitly rebinds the criterion this way ("if you choose the route, it belongs on the admin credential and the refusal shapes there are already established").

**The SEE half WIDENS the existing `GET /{indexer}/admin/canonical-generation` instead of adding a second listing route.** That route already enumerates every generation; under ADR-0084 `canonical` is merely the first slot, so reporting the other two beside it is a widening of the same answer rather than a new resource. Alternative considered: a new `GET /{indexer}/admin/generations` pair with `POST .../reclaim` under it — rejected because it would leave two enumerations of one thing that can drift, which is the failure this repo consistently designs against; the cost is that the route's NAME now under-describes its body, which it arguably already did. The reclaim verb is a flat kebab segment like `reconfigure`, matching the established admin path shape.

**Reclaiming nothing answers `nothing-to-reclaim`, and a reclaim that was entirely held back answers `declined` — three outcomes, not an empty array.** "Nothing happened" had three causes an operator must tell apart, exactly as `ReconfigureReport`'s `unchanged` exists so a save that changed nothing reads as itself: reporting a full decline as "nothing to reclaim" would tell an operator whose disk is full that there was nothing to free, which is the false answer the verb exists to end. All three are `200` successes, because the verb ran. A per-generation deletion FAILURE is a `declined` entry with `reason: 'deletion-failed'` rather than a thrown call, on the same ground `dropReplaced` uses: the generation is still named by no slot, so nothing reads it and the next call tries again. Touches: the response shape any operator tooling reads, and the `ReclaimReport` type consumers of `@etherfold/core` see.

**The report NAMES rather than counts, and the message is built once in core.** Each reclaimed generation carries its identity, the stream reaped with it and how many records that subtree held (which is why `GenerationDeletion` gained `records`); each declined one carries the reason and what to do about it; `slots` carries what survives, from the same read the rule was decided on. The sentence is composed in the container and carried on the report, so a log line and the HTTP response say the same thing rather than two renderings that drift.

**It is not a collector, and I did not make it one.** Nothing calls `reclaim` on a timer, at `open`, or from any drive path; every call in the tree is a test or an operator route. The caps are untouched. If an automatic sweep is wanted, that is a separate decision (it deletes with nobody present) and ADR-0084 does not make it.


wighawag:

This verb DELETES state, so I reviewed the two properties that make it safe by reading the selection rule and grepping the call sites, not by reading the report.

## The two things that make it safe

**1. Nothing any slot names is ever reclaimed, `predecessor` included.** ✅ Structural rather than enumerated, which is the right shape: `unslottedGenerations` is `generations.filter(record => !slotHolding(slots, record))`, and `slotHolding` iterates `SLOT_NAMES`, so a fourth slot added later is protected by construction rather than by someone remembering to extend a list. The selection takes ONE read of the records and ONE of the slots before anything is dropped, with the reason stated (reading twice could pair a listing with slot assignments from either side of another process's write) — that is the concurrency hazard a delete verb has and it is handled.

   Asserted directly, and better than the criterion asks. The first test keeps all three slotted generations and checks their **namespace tables still exist**, so the state is verified present rather than just the row. The second drives the revert for real: promote back to the predecessor and assert `ownerOf(...)` is still BOB. The comment on it makes the point explicitly — "not canonical right now" is exactly the wrong predicate — which is the same trap `a-successor-lands-in-a-durable-slot-that-holds-one` guarded against.

**2. It stays a VERB an operator runs, and has not become a sweep.** ✅ Verified two ways. No `setInterval`, `setTimeout` or open-hook appears anywhere in the diff. And the only two `container.reclaim()` call sites in `packages/*/src` are the registry-entry wirings in `run.ts` and `indexCommand.ts` that back the HTTP route — no drive path, no cycle, no `open`. The decision block says so and the tree agrees. The ADR amendment records it as a decision ("deliberately NOT a collector: nothing fires it on a timer or at `open`, because an automatic reclaim deletes with nobody present and this ADR does not make that decision"), which is the right place for it.

## Per criterion

1. **An operator can SEE what a deployment holds.** ✅ Each slot, what it names, and everything no slot names, without matching digests by eye. Widening `GET .../admin/canonical-generation` rather than adding a second enumeration is the right call: under ADR-0084 `canonical` is merely the first slot, so this is the same answer reported fully. Two enumerations of one thing is exactly what drifts.
2. **RECLAIM everything no slot names, in one action, and get back what was freed.** ✅ And it NAMES rather than counts: each reclaimed generation carries its identity, the stream reaped with it and how many records that subtree held (`GenerationDeletion` gained `records` for this). The prompt was right that this matters more than it looks, and it is honoured.
3. **Never reclaims a slotted generation, `predecessor` included, asserted directly.** ✅ See above.
4. **DECLINED where dropping would strand a follower, and says so.** ✅ Reuses the existing ADR-0044 decline, with a message saying it is retained and will go on the next reclaim once nothing follows its stream.
5. **Caps UNCHANGED.** ✅ Asserted as its own case, and no cap constant is touched anywhere in the diff.
6. **Reclaiming nothing is a success that says so.** ✅ Better than asked: THREE outcomes, because "nothing happened" had three causes. `nothing-to-reclaim` vs `declined` is a real distinction — telling an operator whose disk is full that there was nothing to free is precisely the false answer this verb exists to end. All `200`, because the verb ran.
7. **ADR-0048 / one input shape.** ✅ Satisfied by having NO input: the rule decides which generations go, so there is nothing for a caller to name. Refusals follow the established family (`401`, `404`, `501 reclaim-not-held`). The rebinding from ADR-0048-literally to the admin surface's shapes is the one the task prompt itself authorises.
8–9. **Tests / changeset.** ✅ Both, 458 + 336 lines of new test across CLI and server.

## Notes, non-blocking

- **Route vs CLI verb** was the decision the task flagged, and it is decided explicitly with the reasoning the task asked for. The argument that settles it is the one I would have wanted: a Cloudflare Worker is reachable only over HTTP, so a CLI verb is an affordance that does not exist on one deployment shape, while the CLI still inherits the route by hosting the same app. The blast-radius concern the task raises is answered where ADR-0057 already put the answer (`ADMIN_TOKEN`, fails closed, never the ingest token a log shipper holds). The Worker host answering `501 reclaim-not-held` for a bare receiver is the correct completion of that.
- **ADR-0084 was amended, not closed.** Correct: an `AMENDED 2026-09-16` blockquote under the cap consequence, and the `status: accepted, not yet implemented` line is LEFT IN PLACE. That line belongs to `the-chain-facing-container-holds-its-generations-in-slots`, the last of this ADR's family, and I have planted that pointer in its body (`f50b1e90`). The amendment is also precise that the chain-facing container has no such verb yet because it has no slots yet.
- A per-generation deletion failure becoming a `declined` entry rather than a thrown call is right and consistent with `dropReplaced`: the generation is named by no slot, so nothing reads it and the next call retries.
- `stopDriving` on a reclaimed fold is a good catch — folding into state that has been dropped would be writing into nothing.

No criterion unmet, nothing slotted at risk, no sweep introduced. Merging.

PR #159, opened 2026-09-16
feat(the-chain-facing-container-holds-its-generations-in-slots): The CH…

wighawag:
Task: `work/tasks/done/the-chain-facing-container-holds-its-generations-in-slots.md`

## What I found first (drift check)

The branch already carried the prior attempt's full slot port; I reviewed `git diff origin/main...` before touching anything and built on it rather than restarting. Two premises did not match reality:

- The task body is already at `work/tasks/done/the-chain-facing-container-holds-its-generations-in-slots.md` on this branch (the prior attempt's commit did the done-move), not at `work/tasks/ready/`. I left it exactly where it is and did no git.
- **The forward-pointer's premise for removing ADR-0084's status line is FALSE.** It says "the receiving slot, the arming and the reclaim verb have all landed". Checked rather than assumed: `a-successor-lands-in-a-durable-slot-that-holds-one` and `a-generation-no-slot-names-is-reclaimed-on-request` are in `work/tasks/done/`, but `promotion-arms-from-the-slot-so-a-restart-can-finish-an-upgrade` is still in `work/tasks/ready/` (re-scoped after a measured build STOPPED on it). So ADR-0084's `opened`-gate consequence is genuinely unimplemented and I did **not** remove the line. This is not a stale premise the build rests on, so I proceeded and recorded it below.

## The defect, fixed (three parts, as directed)

**1. `follows` is derived from `writerOf`** (`packages/core/src/container.ts`, `Indexer.add`). The old comment claimed the two forms "now give the same answer here, always" because a record being added sorts LAST. That is true only while the record is NEW; `create` resolves an existing record, which keeps its earliest `createdAt` and sorts FIRST, and that is what a reload is. The comment is replaced with one that says why the set question was chosen, why both of ADR-0071's preconditions have since been met, and exactly which case broke the equivalence. I did not widen into reconciliation: the comment states explicitly that `readOnlyStream` is still frozen at construction and that this only matters if the writer changes mid-fold.

**2. The assertion that would have caught it** (`packages/browser/test/aTabHoldsItsGenerationsInSlots.test.ts`, the reload test). The reload now keeps its own `fakeChain`, drives `afterReload.indexMore()` to the tip, and asserts ranges were actually asked of the node and that the canonical state reaches `EXPECTED_A`, with `canonical.follows === false` beside it to name the mechanism. Verified red-before-green both ways: against the pre-fix core the `follows` assertion fails, and with that line removed the purely behavioural one still fails with `expected 0 to be greater than 0` — zero ranges fetched, which is the silent stall itself.

**3. ADR-0071 amended** — its rejection of the `writerOf` form is marked superseded, with the tie gone (ADR-0072) and the reconciliation shown to be a different, still-unbuilt case. Its headline claim (the source is the durable registry, never this process's array order) is called out as unchanged.

Everything else from the prior attempt is untouched: the slot port, the replacement rule, the reload replacement case, canonical/predecessor unreachable, the deleted IndexedDB database, `everCanonical` gone, the cap arithmetic and its ADR-0084 amendment, the observation note.

## Gate

`pnpm format:check`, `pnpm build`, `pnpm typecheck`, `pnpm test` (4542 tests across 18 packages), `check:adr`, `check:refs`, `check:changesets` all green. Working tree holds only the seven intended files, nothing untracked.

## Decisions

The first four are **carried forward verbatim from the prior attempt** (they are the rationale for the slot port itself and must survive the requeue); the last three are new to this continuation.

**The rule is SHARED as one pure function; the ACTION is REPEATED in each container.** `displacedBySuccessor(arriving, generations, slots, heldHere)` lives in `generation/registry.ts` beside `slotHolding` and `unslottedGenerations`, which are already exactly this kind of shared pure rule, and both containers call it. I shared it rather than repeating it because it carries the safety clause — a generation any OTHER slot names is untouchable, which covers the incumbent and the revert target together — and two copies of a safety predicate is how a revert target gets deleted by the twin nobody was reading (ADR-0071's "one rule has one home"). I did NOT hoist the slot LOGIC: whether a drop strands a follower, and what stopping a fold means, are repeated and worded alike, because a fold is an engine on one side and a receiver on the other, and a shared base parameterised over storage, caps and what a fold IS is the options-growing abstraction ADR-0071's closing section already refused. Alternatives considered: repeat everything (rejected — the safety clause would have two homes), or share a `SlotReplacer` collaborator (rejected — five callbacks to save sixty lines). Touches: `ReceivingIndexer.replaceTheSuccessor` was refactored onto the shared function, and `displacedBySuccessor` is a new public export of `@etherfold/core`.

**What a cap of TWO means under THREE slots: the tab holds `canonical` + `successor` OR `canonical` + `predecessor`, never all three, and the third registration is REFUSED rather than dropping the revert target.** Three slots occupied at once is three generations and `BROWSER_GENERATION_CAPS` is two, so after a promotion (which assigns `predecessor` and clears `successor`) a further save meets `maxGenerations`. I did not soften that, and specifically did not let a replacement reach `predecessor` to make room: that would be an eviction, and no policy can know which generation was being kept, while the re-index it costs may not be available at all from a public node. This is NOT a new wall — `promotion.test.ts`'s existing "a superseded generation is never EVICTED by a cap" pins the identical refusal against the pre-slot code, and the acceptance criteria say the caps are unchanged — and what slots DO fix is the churn case, which is what a cap was always the wrong instrument against: repeated reconfiguration without an intervening promotion is now bounded at two forever. It is asserted as a test ("REFUSES a third generation rather than dropping the revert target") so the trade is visible rather than silent, and recorded as an amendment in ADR-0084 beside the consequence that argues the caps should stay. Alternative considered and deliberately NOT taken: raise `BROWSER_GENERATION_CAPS.maxGenerations` to three so all three slots are expressible at once — that is a storage decision of its own and the acceptance criteria forbid touching the caps. Touches: `an-hmr-update-reconfigures-the-tab-it-is-running-in` (an HMR loop that promotes between saves will meet this), and any future task that revisits the browser caps.

**A move onto a generation NO slot names now drops nothing, where `everCanonical` could have classified it as a promotion.** Deleting the flag means "was this a promotion" is `successor` naming the target, so an operator promoting some other registered generation is treated as a move that supersedes nothing. Strictly more conservative in the destructive direction, and it makes the chain-facing container agree with the receiving one, which recorded the same choice when it landed.

**A host opened with several generation specs now puts every spec after the first into `successor`, so the ones between are replaced and dropped as they are added.** That follows from registering with a slot at all and is the correct reading of "successor holds at most one"; the only shape that behaves differently from before is a container opened with three or more specs, which nothing in this repo does outside tests. Called out in the changeset because it is the one caller-visible behaviour change.

**ADR-0084's `status: accepted, not yet implemented` line STAYS, and the forward-pointer's condition is why.** The pointer said to remove it once the receiving slot, the arming and the reclaim verb had all landed, and told me to check rather than assume. Two of the three are in `work/tasks/done/`; the arming (`promotion-arms-from-the-slot-so-a-restart-can-finish-an-upgrade`) is still in `work/tasks/ready/`, re-scoped after a measured build stopped on it, so ADR-0084's `opened`-gate consequence is genuinely unbuilt and the line is still a true claim about one consequence. Alternatives considered: remove it anyway (rejected — it would assert in the ADR record that a correctness cliff is closed when it is open, which is the exact failure mode the pointer exists to prevent), or rewrite it to a fourth status phrasing such as "implemented except the `opened` gate" (rejected — the status vocabulary is free prose today with five phrasings, and coining a sixth is a decision the conductor did not sanction, for a line that comes off entirely one task later). Instead I pinned a dated note on the `opened`-gate consequence itself saying it is the last one open, that it is why the front-matter line stays, that every other consequence has landed, and that this task checked and found the premise unmet. Touches: `promotion-arms-from-the-slot-so-a-restart-can-finish-an-upgrade` now owns the removal, and the note tells its agent so.

**ADR-0072 is amended too, not only ADR-0071.** The handoff named ADR-0071. But ADR-0072's "What this does NOT change" section closes with "that container asks the registry a set question instead, and that is unchanged and still correct", which this change makes false in the same sentence that grants the permission it relies on. Leaving it would put two ADRs in contradiction about the same line of code, so it gets a short amendment pointing at ADR-0071 for the reasoning. Alternative considered: amend ADR-0071 only and let ADR-0072's sentence rot (rejected — a reader arriving from ADR-0072 would conclude the code was wrong).

**The observation note gets an appended RESOLVED paragraph rather than being left as-is or deleted.** `work/notes/observations/` is append-only capture, so appending is the sanctioned edit and deleting is not; but a note asserting a live silent stall, in code that no longer has one, is a false signal for whoever reads the bucket next. The appended paragraph is dated, says the cause was narrower than the note assumed (no reconciliation needed, only the initial derivation), and points at the ADR-0071 amendment and the new test. It records no build-time rationale — that is all here, per the work contract.


wighawag:
BLOCK — Gate-3 (conductor's diff-vs-criteria pass)

The slot port itself is good work, and most of this diff I would approve without hesitation. I am blocking on one thing, and it is a thing this PR's own observation note states plainly: **this change converts a loud refusal into a silent stall in the container a browser tab runs, in exactly the scenario the task is named after.**

## The blocking finding

`work/notes/observations/a-reloaded-container-makes-its-canonical-generation-a-follower.md`, added by this PR, says it better than I can:

> Before slots this path was mostly unreachable at `BROWSER_GENERATION_CAPS`: a reload with a changed processor registered a third generation and the cap REFUSED, so the container did not open at all. With slots the reload replaces what `successor` holds, so the container now opens — and lands in this case instead. **A loud refusal became a silent stall, which is the direction that matters.**

I verified the mechanism rather than taking the note on trust. `Indexer.add` computes `follows` from the DURABLE registry (`container.ts:777`): `alreadyOnThisStream = registry.list().filter(other => other.stream === record.stream && !sameGeneration(other, record))`, and `follows = alreadyOnThisStream.length > 0`. On a reload over a durable registry that already holds canonical A and successor B on one stream, re-adding A finds B still registered, so **A — the canonical generation — is built as a FOLLOWER** with a `readOnlyStream` and `followMore()`. Every generation the reloaded container holds then follows a stream nothing writes. The tab stops fetching, and `/status`-shaped reporting and the state both still look fine. `readOnlyStream` is baked into the engine config at construction (`container.ts:781-784`), so nothing reconciles it afterwards.

Why that is this PR's problem and not merely a pre-existing one it happened to notice:

1. **The reachability change is caused by this diff.** The bug in `follows` predates it; the fact that a browser tab now *reaches* it does not. Trading a refusal the developer can see for a stall they cannot is a regression, and it is in the direction the task explicitly cares about.
2. **It falsifies the task's own title.** "So a browser tab survives an afternoon of reloads" is the headline claim, and after this change a tab that reloads over durable storage comes up not fetching. Surviving a reload is the deliverable, not a side effect.
3. **The task prompt rules on precisely this.** "This container is what a user's app holds. A reload that briefly answers nothing is worse here than anywhere, because there is a UI attached to it." A tab that *permanently* fetches nothing is strictly worse than one that briefly answers nothing.
4. **The gate cannot catch it.** The reload suite (`aTabHoldsItsGenerationsInSlots.test.ts`, "a RELOAD replaces what it finds in the slot") asserts the slot contents after the reload and never asserts that the reloaded container still FETCHES. That is why this is green. The missing assertion is the one that matters: after the reload, does the tab still make progress?

Note the scope precisely, because it bears on the fix: the DEFAULT browser registry is `openMemoryGenerationRegistry(...)`, where a reload starts empty and `alreadyOnThisStream` is empty, so this does not bite. It bites on the DURABLE (`OnIndexedDB`) registry — which is the configuration criterion 2 is written against ("a fresh container over the same durable storage") and the one this whole task exists to make work.

## What I am NOT blocking on

**The cap arithmetic is handled correctly and I want to say so**, since it was the thing I expected to block on. Three slots against `maxGenerations: 2` means a tab holds `canonical` + `successor` OR `canonical` + `predecessor` and never all three, and a registration needing a third meets the cap and is REFUSED. That is asserted as a test (`REFUSES a third generation rather than dropping the revert target`, with `dropped` asserted empty), recorded as an ADR-0084 amendment, and the cap raise is deliberately deferred as a separate decision with its own storage argument. The revert promise is preserved rather than quietly traded away, which is the right end of the trade: an eviction cannot know which generation was being kept, and on a public node the re-index it costs may not be available at all. That is a needs-attention signal delivered properly, not a silent trade.

For the record, everything else ticks: the replacement rule and its restart case, `canonical` and `predecessor` unreachable from a replacement, the retired generation's own IndexedDB database actually deleted (not merely unregistered), caps untouched, `everCanonical` deleted from the chain-facing entry, the two ADR-0084 amendments, and ADR-0084's `status: accepted, not yet implemented` line correctly removed now that this is the last of its family (thank you for picking up the forward-pointer). Sharing exactly one thing between the twins — `displacedBySuccessor`, the RULE — and deliberately repeating the ACTION because a fold is an engine on one side and a receiver on the other is the right call, and ADR-0071's refusal of an options-growing base class is the right citation for it.

## What would unblock this

Any one of these, and I do not want to prejudge which, because the third may be a task of its own:

1. Make the reloaded container NOT build its canonical generation as a follower — `writerOf` would name A correctly here, and ADR-0072 has since delivered the stable registration order that ADR-0071 lacked when it rejected that form. This is the real fix and it may well exceed this task.
2. Failing that, **refuse loudly rather than open into the stalled state**, so the property this change removes (the developer finds out) is preserved until the real fix lands.
3. At minimum, a reload test that asserts the reloaded container still FETCHES, so the stall is a red test rather than a note — and if that test cannot be made green in this task's scope, that is the needs-attention signal, and the honest one.

Routing to needs-attention with this as the reason. The branch is preserved; the slot port is sound and should be recoverable on top of whichever resolution is chosen.

wighawag:
Follow-up to the block above: I found the precise cause, and the fix is much narrower than the three options I listed. Requeuing with this as the handoff.

## The false premise is in the code's own comment

`Indexer.add` explains why it derives `follows` from "is anyone else already registered on this stream" rather than from `writerOf`, and ends:

> ADR-0072 removed that: `createdAt` is strictly increasing within a registry, so **a record being added always sorts LAST** and `writerOf` can never name it while others exist. **The two questions now give the same answer here, always.** This one is kept anyway because it is the one that does not DEPEND on that.

The claim in bold is false, and the reload is the counterexample. "A record being added always sorts LAST" holds only while the record being added is NEW. On a reload the container re-adds a generation that is **already registered**, so `create` finds the existing record and it keeps its original, earliest `createdAt`. It sorts FIRST.

With canonical A and a leftover successor B on one stream, re-adding A gives:

- `writerOf(stream)` → the oldest surviving generation → **A**. A writes. Correct.
- `alreadyOnThisStream` → `[B]`, non-empty → **A follows**. Wrong, and it is what makes the tab stop fetching.

So the two forms diverge in exactly one case: when the record being added already exists. That is what a reload does, and it is what slots made reachable, which is why this surfaced now rather than earlier.

Worth stressing that the reachable case is the ORDINARY dev loop, not an edge. It does not need a changed processor: save once (creating successor B), then reload the page unchanged. B is durable now, so A comes back as a follower and the tab silently stops fetching while its status and state both still look healthy.

## The fix

1. **Derive `follows` from `writerOf` in `Indexer.add`**, and correct that comment, which is now provably wrong rather than merely conservative.
2. **Add the reload test that would have caught it**: the existing reload case asserts the slot contents and never asserts that the reloaded container still FETCHES. That missing assertion is why this is green.
3. **Amend ADR-0071** to record that its rejection of the `writerOf` form is superseded: it was rejected for depending on an ordering the registry did not then guarantee, and ADR-0072 supplies exactly that ordering.

## Why this rather than the bigger options

- **The comment already concedes `writerOf` is correct now.** It kept the other form defensively, against an ordering ADR-0072 guarantees. That defensive choice is the thing that broke, so removing it is a simplification and not a risk.
- **It converges the twins**, which is this task's own acceptance criterion 6. `ReceivingIndexer` already uses `writerOf` in five places; the chain-facing container is the outlier, so this makes the two say the same thing in the same words rather than adding a third dialect.
- **The reconciliation half is NOT needed.** The observation added in this PR correctly notes that `readOnlyStream` is baked into the engine config at construction and so cannot be recomputed later the way `reconcileWriters` does on the receiving side. True, and irrelevant here: the reload case needs only the INITIAL derivation to be right. Recomputation matters when the writer changes while a fold is held, which is a different scenario and stays out of scope.

## What stays as it is

Everything else in this PR is good and should be kept: the slot port, the replacement rule and its reload case, `canonical` and `predecessor` unreachable from a replacement, the retired generation's own IndexedDB database actually deleted, `everCanonical` removed from the entry, and the cap arithmetic worked out honestly (a third generation REFUSED rather than the revert target evicted, asserted with `dropped` empty and recorded as an ADR-0084 amendment). ADR-0084's status line is correctly removed. The observation note you added is the reason this was findable at all, and it should stay.

Continuing from this branch rather than starting over.

wighawag:
APPROVE ✅ — Gate-3, second pass

The blocking defect is fixed, fixed at the right level, and fixed without the scope growth I was worried about. Merging.

## The fix

`Indexer.add` now derives `follows` from `writerOf` rather than from the set question, and the comment that carried the false premise has been replaced with one that names it exactly:

> That claim is FALSE, and exactly one case breaks it: a record that ALREADY EXISTS. `create` RESOLVES rather than duplicating, so a generation re-registered keeps its original `createdAt` and sorts FIRST, and re-registering what is already there is precisely what a page RELOAD is.

That is the defect stated better than I stated it, and having it sit next to the code is worth more than the fix itself, because the next person to reach for the "robust" set question will read why it is not.

The scoping paragraph is exactly right and I want to call it out, since it was the thing most likely to sprawl:

> This is still the INITIAL derivation and nothing recomputes it [...] That asymmetry is untouched here and only matters if the writer CHANGES while a fold is held; what a reload needs is this one derivation being right at the start.

`reconcileWriters` is left alone, the receiving container is untouched, and the ADR-0071 asymmetry is documented rather than half-closed.

## The test that would have caught it

```
expect(afterReload.canonical.follows).toBe(false);
expect(reloadedChain.ranges.length).toBeGreaterThan(0);
expect(await readState(afterReload.state)).toEqual(EXPECTED_A);
```

Asserting on RANGES ASKED OF THE CHAIN is the right assertion, because it is the one the slot contents cannot fake. The comment above it makes the point that the previous suite was green while the tab fetched nothing, which is the lesson worth keeping.

## You corrected my forward-note, and you were right

I planted a pointer in this task's body telling it to remove ADR-0084's `status: accepted, not yet implemented` line on the grounds that it was the last of that ADR's family. It is not, and this PR restores the line with a note saying why: `promotion-arms-from-the-slot-so-a-restart-can-finish-an-upgrade` is still open, having STOPPED on a measured finding that the `opened` gate is one of three blockers rather than the only one.

My note was true when I planted it and stopped being true when the arming task stuck and was re-scoped. Checking rather than obeying is the correct response to a forward-pointer, and it is exactly the failure mode `ADR-FORMAT.md` warns about, caught from the other direction.

## Amendments

All three are properly scoped. **ADR-0071** records that its rejection of the `writerOf` form is superseded while its headline claim (the source is the durable registry, never process array order) survives untouched. **ADR-0072** records that it is what made the change possible, its own last sentence having expired. **ADR-0084** keeps its status line with the reason.

## Everything from the first pass still stands

The slot port, the replacement rule and its reload case, `canonical` and `predecessor` unreachable from a replacement, the retired generation's IndexedDB database actually deleted, `everCanonical` gone from the entry, caps untouched, and the cap arithmetic settled honestly (a third generation REFUSED rather than the revert target evicted, `dropped` asserted empty). The observation note that made this findable stays.

Both criteria that were at risk are now met: a browser-shaped deployment no longer reaches its bound through repeated reconfiguration, AND the tab it reaches that state in still fetches.

PR #160, opened 2026-09-17
feat(a-tab-can-or-cannot-instantiate-a-processor-from-bytes-under-a-csp…

wighawag:
Task: `work/tasks/done/a-tab-can-or-cannot-instantiate-a-processor-from-bytes-under-a-csp.md`

The gate is green and the tree carries only the intended changes.

## What I did

This was a **continuation**. The branch already carried a complete, measured spike (commits `3e2bc390` + `ed608782`): the server, page, fixture bundle, driver, raw results for three engines, and the finding note. I reviewed it rather than restarting it, and found exactly one thing wrong: the **acceptance gate had failed on a dead `work/` reference**, not on the measurement.

**The fix** (`docs/spikes/a-tab-can-or-cannot-instantiate-a-processor-from-bytes-under-a-csp/README.md`, one line): the README cited the task at `work/tasks/ready/…`, which had since moved to `work/tasks/done/…` as part of the done-move in the same branch. `check:refs` flagged it verbatim. Updated the citation to `work/tasks/done/…`, which is also the house precedent (`docs/spikes/sqlite-in-the-browser/README.md` cites its task in `done/`).

**I verified the prior measurement rather than trusting it.** Re-ran the spike against nixpkgs' playwright browsers: Chromium and Firefox reproduced the committed results **exactly**, 110 verdicts per engine, zero differing. I then restored `results/` to the committed files so the three engine files stay one coherent session, matching the finding's `source:` line. WebKit would not relaunch (see below).

**Acceptance criteria**, all met by the finding at `work/notes/findings/a-tab-instantiates-retained-bytes-only-through-a-same-origin-url.md`: a mechanism-by-policy matrix over 8 policies x 2 deliveries x 3 engines (§1), the boundary rather than a verdict (`blob:` and `data:` are separate source expressions, neither a superset; `'strict-dynamic'` is *more* permissive than `script-src 'self'`), what a refusal looks like from inside the page including the corrupt-bytes control and the `securitypolicyviolation` correlation that is the only way to tell refusal from damage (§3), the reference deployment's real gateway headers fetched live (§4), the three alternatives named and costed a paragraph each (§6), and what was not measured (§7). Nothing outside `docs/spikes/` and `work/` is touched; no shipped package changed, so no changeset applies.

**Gate:** `format:check`, `check:adr`, `check:refs`, `check:changesets`, `changeset status --since=main`, `build`, `typecheck`, `test` all green.

**Observation captured** (out of my task's scope, not fixed): `work/notes/observations/webkit-did-not-relaunch-when-the-csp-spike-was-re-run.md` — WebKit died at `newPage` on the re-run under the same documented recipe that had produced its rows, so that column is not reliably re-runnable on this host. The finding is not in doubt, since the two engines that did run agreed verdict-for-verdict.

## Decisions

- **The dead `work/` citation was REWORDED, not exempted.** `check-work-refs.mjs` offers two fixes: update the citation, or add the holding surface to `EXEMPT` because the reference is historical. I updated it, because a spike README's pointer at its originating task is *navigational* (a reader follows it to find out what was asked), not a frozen quote like a changeset or a question sidecar. The alternative, exempting `docs/spikes/`, would have blinded the gate across every spike README to keep one link alive. What it touches: nothing beyond this file; `docs/spikes/sqlite-in-the-browser/README.md` already cites its task in `done/`, so this is the existing convention rather than a new one.
- **The mechanism list was EXTENDED past the two the task names** (`blob:` and `data:` import) to include `new Function`, a `blob:` worker, a same-origin worker under two different response policies, and a service-worker-synthesised same-origin URL. Why: the task's own bar is the BOUNDARY, and with only the two named mechanisms the answer would have been a flat "no under every realistic policy", which is false and would have changed the retention design wrongly. The alternative considered was staying literal and reporting the no. What it touches: the retention spec's browser half, and potentially a new task, because the finding's recommendation ("a browser instantiation path implies a service worker in `@etherfold/browser`") is a component that does not exist and is app-visible.
- **The spike keeps its own `esbuild` + `playwright` devDependencies, and `playwright` was re-pinned from the prior attempt's `1.62.1` down to exactly `1.61.1`.** The task says "do not add a dependency"; I read that as the shipped packages, since `docs/spikes/` is outside the workspace globs (`packages/*`, `examples/*`, `platforms/*`) and every other browser spike here carries its own `package.json`. The downgrade is because npm's playwright and `nixpkgs#playwright-driver.browsers` must agree on browser revisions, and nixpkgs is at 1.61.1; without it only chromium launches on this host and the finding would have been single-engine. Alternative considered: keep 1.62.1 and report chromium only. What it touches: nothing outside this folder, but a re-runner on another host should read the README's NixOS section before assuming the pin is arbitrary.
- **The finding recommends but does not decide.** It says what the retention spec should do (do not build on `blob:`/`data:`; decide explicitly whether a browser revert is worth a service worker) and names/costs the three alternatives, without editing the spec or opening a task. The spec is a launch snapshot and the runner owns work-item transitions, so the answer lives in the finding and the routing is a human's.


wighawag:

Reviewed the diff against the task's seven acceptance criteria. Gate 1 (acceptance) is green; Gate 2 is off per this repo's `dorfl.json`, so this is the only review layer beyond the gate.

**The question is answered, and the boundary is the answer.**

- [x] **Answered for the mechanisms that could work, against at least no CSP / `script-src 'self'` / a policy allowing `blob:`.** Exceeded: eight policies × two delivery mechanisms (header and `<meta http-equiv>`) × three engines, 16 runs per engine, and the mechanism set widened past `blob:`/`data:` import to `new Function`, a `blob:` worker and a service-worker URL. A verdict counts as `ran` only when the processor actually FOLDED an event to the expected value, which is the right bar: a missing error is not a result.
- [x] **The answer states which mechanism works under which policy rather than yes-or-no.** Section 1's matrix. Three findings in it are load-bearing and non-obvious: `blob:` and `data:` are separate source expressions so neither degrades to the other; `script-src 'nonce-…' 'strict-dynamic'` (the hardened policy) is *more* permissive here than the naive `script-src 'self'`; and `new Function` is doubly wrong because it needs `'unsafe-eval'` AND cannot accept an ESM bundle at all, so it would force a second artifact and therefore a second identity under ADR-0086.
- [x] **What a blocked instantiation looks like from inside the page is recorded.** Section 3, and this is the criterion the task said would lead to different designs. It distinguishes the two catchable rejections from the `worker.onerror` case whose message is empty in all three engines, and then does the thing that was not asked for and matters most: a truncated-bundle CONTROL, standing in for a partially written retained artifact. The result is that under a blocking policy a refusal and a corrupt artifact are *indistinguishable* on the error, and separable only out-of-band via `securitypolicyviolation`. That is a real constraint on the eventual error surface rather than a reassurance.
- [x] **The reference deployment's own CSP is checked, so this is about a real app.** Section 4, with the five hosts fetched live the same day. The finding that the policy belongs to whichever IPFS gateway a user happens to reach (so the same build meets different policies at different users, and the mechanism cannot be chosen at build time) is a stronger and more useful result than the task asked for. Pinata's header being quoted verbatim, including the note that it reads as permissive because of `img-src`'s `blob:` while refusing every script mechanism, is exactly the "a real app rather than a hypothetical one" the criterion wanted.
- [x] **Alternatives named and costed.** The answer is a qualified yes rather than a no, so this criterion was arguably not triggered; section 6 delivers it anyway for all three alternatives the task named, one paragraph each, costed and not designed. Each one is honest about its own defeat condition (the in-memory factory does not survive a reload; build-time registration reintroduces an author obligation of the kind ADR-0086 deleted; read-only revert makes `predecessor` mean different things in a tab and on a server).
- [x] **The finding is written to `work/notes/findings/`** as the surviving deliverable, in house style, with a `source:` line that names the playwright and engine versions, the host, the run counts and the date.
- [x] **The spike code is NOT merged: no production path, no new export, no dependency added.** Verified by path: the diff touches nothing under `packages/`, `platforms/` or `examples/`, and does not move the root `package.json`, `pnpm-workspace.yaml` or `pnpm-lock.yaml`. The harness lands in `docs/spikes/<slug>/` with its own isolated `package.json` + `package-lock.json`, which is this repo's established convention and precisely what the precedent the task itself cites (`browser-storage-headroom-for-generations.md`) did with its own harness, results JSONs and manifest. I checked that rather than assuming it, because 15k added lines looks at first glance like the thing this criterion forbids; it is not, and the criterion's real content (nothing in the shipped surface, no workspace dependency) holds exactly.

**Two things beyond the criteria, both to the good.** The agent re-ran its own committed results before landing and reports Chromium and Firefox reproducing 110 verdicts each, zero differing, which makes the stored rows evidence rather than an assertion. Where that re-run failed (WebKit died at `newPage` on the same host and recipe that had just produced its column) it did not quietly drop it or re-characterise the result: it filed `work/notes/observations/webkit-did-not-relaunch-when-the-csp-spike-was-re-run.md` labelling it a harness signal, noting the two engines that did run agreed verdict-for-verdict so nothing in the finding is in doubt, and warning the next re-verifier not to read a red run as a changed answer. That is the right disposition.

**Section 7 is not padding.** Naming private browsing (Firefox has historically disabled service workers there, so the recommended mechanism may have no coverage at all for those users), a service worker actually reading from IndexedDB with its cold-start ordering, and `require-trusted-types-for 'script'` as UNMEASURED keeps the finding from being over-read. Given section 5 concludes that a browser instantiation path implies a service worker, those three gaps are exactly what the retention spec has to settle before it commits.

**The one blocking thing found earlier is fixed.** The first gate run failed `check:refs` because the spike README cited `work/tasks/ready/<slug>.md`, which the done-move relocates. The README now cites `work/tasks/done/<slug>.md`, the reference is live, and the gate is green on the rebased tip.

No nits worth carrying: Gate 2 is off in this repo, so there is no `review-nits-*` note for this task, and I found nothing in the diff to route. The only judgement call was the 15k-line `docs/spikes/` addition against "the spike code is NOT merged", and the repo's own precedent settles it.

Merging. The finding does not unblock anything in the ADR-0086 family (it gates the RETENTION spec, `a-generation-retains-the-code-that-folds-it`), so the family's order is unchanged; what it changes is that the browser half of retention now has a costed answer instead of an assumption.

PR #161, opened 2026-09-17
feat(a-processor-artifact-is-bytes-a-hash-and-a-loader): A processor AR…

wighawag:
Task: `work/tasks/done/a-processor-artifact-is-bytes-a-hash-and-a-loader.md`

Done. The task is built, the gate is green, and nothing existing changed behaviour.

## What I did

**The unit** — `packages/utils/src/processorArtifact.ts`, three capabilities and nothing consuming them:

- `processorArtifactIdentity(bytes)` — SHA-256 over the octets, rendered `sha256:<64 lowercase hex>`, reusing the convention `streamSeedContentHash` established so a pasted literal says which function produced it.
- `unresolvedImportsOf(bytes)` — what SELF-CONTAINED means, checked against the bundle: every module it still expects somebody else to resolve, bare (`viem`) and relative (`./abi.js`) alike, statically, including a specifier only a dynamic `import()` carries. A builtin (`node:crypto` or bare `crypto`) is admitted, because a `data:` URL really does resolve one.
- `loadProcessorArtifact(bytes, {processorConfig})` — hash, admit, then instantiate by importing a `data:text/javascript;base64,` URL: no path, no temp file, no cache breaker. Everything checkable runs before EVALUATION, which is the irreversible act here (the module joins the process registry for the life of the process and its top-level code runs). Refusals are DATA (`not-self-contained` naming the specifiers, `unreadable-module`, `not-a-processor`), each carrying the artifact's identity, because bytes have a name whether or not they load. The module-shape rules stay `instantiateProcessor`'s (including ADR-0037's kind-tag refusal), caught and converted here exactly as the seed loader converts `parseStreamSeed`'s throw.

**The fixture** — `packages/utils/test/fixtures/processor-artifact/`: a real processor (one declaration, one handler, one sibling module) bundled once by the documented command (`esbuild 0.28.1 --bundle --format=esm --minify`) and committed, with a README carrying the command and the reason `--minify` is mandatory. Listed in `.prettierignore`, since reformatting an artifact would rename it.

**The evidence** — `docs/spikes/a-processor-artifact-is-bytes-a-hash-and-a-loader/`: a script measuring what a `data:` URL module can resolve in Node, and its output. It is what decided the builtin exception, and it is the case that proves the check must be STATIC: a bundle whose only unresolved import is dynamic imports cleanly and would fail at the first event it folds.

**Tests** — 28 in `packages/utils/test/processorArtifact.test.ts` (utils: 54 total). The round trip is asserted against the committed bundle: same declarations and the same mutation as the processor imported straight from the source. Identity is asserted as a property (stable, and moves on one byte) plus the published SHA-256 of the empty input, so a rebuilt fixture is not a test failure. One test proves the refusal precedes evaluation via a global marker that never lands. Verified the tests bite by disabling the check (2 fail).

Untouched: `getVersionHash()`, the declared `version`, `getCodeFingerprint()`, and every caller. The only other edits are the changeset, the utils README/description, `.prettierignore`, and the lockfile line for the new `esbuild` devDependency.

Two notes: I left `work/questions/task-…md` alone (its question is the orchestration fault that destroyed the earlier run, not a defect in the task); its recorded design notes match what I built, and where I diverged is below. I also dropped one observation (`work/notes/observations/a-relative-processor-modules-own-error-is-swallowed-by-the-package-fallback.md`) about `loadProcessorModule` discarding a relative module's own import error — noticed, not fixed.

One gate step cannot run locally: `pnpm changeset status --since=main` uses `git diff`, which is blind to untracked files, so it reports my changeset as missing until the runner commits. Everything else (`format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test`) is green.

## Decisions

**Home is `@etherfold/utils`, not `@etherfold/core`.** Chosen because utils is the package that already owns "a host was handed a processor by an operator" (`loadProcessorModule`, `instantiateProcessor`), so the artifact is a second ARRIVAL beside the path one and can REUSE the module-shape rule rather than forking a second answer to what a processor module is. Core was the alternative (it holds the stream-seed prior art and the identity vocabulary), and rejected: core is what a browser bundles, and a `data:` URL import shipped there is dead weight pointing at a path that is measured dead under a real CSP. What it touches: `a-deployment-runs-from-a-bundle-identified-by-its-hash` and `a-path-naming-an-unbundled-entry-point-is-refused` import from here; the CLI already depends on this package. The lost run's notes reached the same placement independently.

**A node BUILTIN is self-contained; everything else is refused.** Measured, not assumed: a `data:` URL resolves `node:crypto` AND bare `crypto`, so refusing them would refuse an artifact that runs (a `--platform=node` bundle keeps those imports). An absolute `file:`/`https:` URL resolves too and is still refused, because a fold depending on the instantiating machine's disk or on a network fetch is not covered by the identity. Alternative considered: refuse every import without exception, simpler but a false refusal an author cannot act on. What it touches: the CLI refusal task reuses this judgement, so `import 'node:fs'` in an entry point will not be what makes it "unbundled".

**The self-containment check is a bounded two-part regex, not the string/comment-aware scanner the lost run's notes recommend.** The statement half requires a statement position (start, or after `;`, `}`, a line break); the dynamic-call half stays loose. The asymmetry is chosen on which way each fails: a missed STATIC import costs only a less precise refusal, because Node refuses to link such a module and the loader reports `unreadable-module` anyway, whereas a missed DYNAMIC one would admit an artifact that dies at its first fold. A tokenizer (strings, templates, comments, regex literals) is ~100 lines with its own ambiguity, for a false positive that requires build output to embed JavaScript source in a string; and `--minify` strips comments, which removes the largest class. The residual fails CLOSED with the matched specifier named. What it touches: the same function is the CLI's "this is not a bundle" judgement.

**Three refusal reasons, and no logging.** `not-self-contained` / `unreadable-module` (bytes that do not parse, or a module body that threw — collapsed like the seed loader's `unreadable-format`) / `not-a-processor`. Each carries a `why`, so unlike the seed install this unit logs nothing: the caller owns the context worth logging (the path or the request the bytes came from) and would otherwise say half of it twice. What it touches: `a-path-naming-an-unbundled-entry-point-is-refused` owns the operator-facing wording.

**No `isProcessorArtifactIdentity`-style predicate.** I wrote one, then deleted it and left the omission documented at the site. ADR-0086 rests on nothing parsing `GenerationId.processor`, and a helper asking "does this look like a hash" is the first parser; the sibling task names exactly that hazard. `pinnedStreamSeedContentHash` is not a precedent, because nothing pins a processor identity.

**The loader hands back the AUTHORING object plus the module, not an `EventProcessor`.** A bundle's `createProcessor` yields declarations plus handlers and names no backend (ADR-0037), and the host builds the runtime around it, so satisfying the `EventProcessor` seam here would mean this unit picking a store. The module rides along because `contractsData` / `contractsDataPerChain` live on it and resolving a source is the caller's step. The committed fixture keeps `version: '1.0.0'` with a comment saying it is not what identifies the artifact, because the entity runtimes still require one until the contract task deletes the field — which is what lets the next task reuse the fixture end to end.


wighawag:

Reviewed the diff against the task's eight acceptance criteria, with particular attention to the two ways an EXPAND step can pass a green gate while being silently wrong: deleting something a later batch needs, and reaching for a bundler to obtain an identity in a test. Neither happened.

- [x] **Bytes hash to a stable `sha256:<hex>`.** `processorArtifactIdentity` is SHA-256 over the octets with the algorithm in front of the digits, reusing `streamSeedContentHash`' convention rather than re-deciding it, and the JSDoc says why (the value is pasted into builds and generation records, where a bare hex string cannot say which function produced it). The byte domain is stated as the whole of the contract and enforced by the signature taking `Uint8Array`, so a caller cannot introduce a text-encoding disagreement. Asserted as a known digest, not just self-consistently.
- [x] **Instantiated from bytes with no filesystem, and the result satisfies the processor seam.** A `data:text/javascript;base64,` import: no temp file, no path, no cache-busting query. The module-shape rule is not reimplemented, it delegates to `instantiateProcessor`, so the retired `{kind, processor}` tag (ADR-0037) is refused by the one rule that already owns it. There is a test asserting specifically that the bytes were never on a filesystem.
- [x] **A bundle that is not self-contained is REFUSED, naming the unresolved specifier.** Wider than asked and correctly so: bare specifiers, relative ones (a `data:` URL is not a directory), side-effect imports, re-exports, and the DYNAMIC `import()` that would otherwise survive to the first fold. Builtins are admitted, prefixed and bare, and that was MEASURED against Node rather than assumed (the measurement is kept in `docs/spikes/<slug>/`). The detector's asymmetry is reasoned on which way each half fails: a missed statement costs only a less precise refusal because Node will not link the module anyway, while a missed call would admit an artifact that dies at the first event, so the call half stays loose. That is the right axis to choose on.
- [x] **The refusal is DATA in the manner the seed-install path refuses, not a bare throw.** Three reasons with three different remedies, `why` carrying the distinguishing detail, and `unresolvedImports` on the arm that has them. The identity is on BOTH arms, which the module argues is the point of the unit rather than a convenience: bytes have a name whether or not they turn out to be a processor, so a refusal can say which artifact it refused. The ordering is the borrowed half and it is structural: everything checkable happens before EVALUATION, which is the irreversible act here because a module joins the process registry for the life of the process and its top-level code runs. There is a test asserting the top-level code never runs on a refusal.
- [x] **A small REAL bundle fixture, built once by the documented command and committed.** 258 bytes at `packages/utils/test/fixtures/processor-artifact/`, built with `esbuild --bundle --format=esm --minify`, with its source committed beside it and a README carrying the exact rebuild command. Two judgements in that README are better than the criterion required. It argues against building in the test path not only on toolchain dependence but on CIRCULARITY: an artifact built on the fly by the process that checks it can only tell you the two agree with each other. And it pins no hash: identity is asserted as a PROPERTY (identical bytes, identical name; one changed byte, a different name), so a newer esbuild is not a test failure, while the one pinned value is the digest of a known input, which pins the RENDERING rather than this artifact. The fixture is also a genuine bundle (a sibling module flattened in) rather than a single file, so it exercises what the criterion meant.
- [x] **NOTHING existing changes behaviour.** Verified mechanically, not by reading the summary: the entire diff contains exactly two `-` lines and both are prose (the package description and a README sentence). `getVersionHash()`, the declared `version`, `assertProcessorVersion` and the code fingerprint are untouched, appearing only inside new explanatory prose, and no caller is migrated. The changeset says so in as many words ("Nothing consumes this yet, deliberately"), and the module header repeats it with the reason. This is the criterion that keeps the four migrate batches buildable, so I checked it by diff rather than by claim.
- [x] **No test runs a bundler to obtain an identity.** `esbuild` is added as a DEV dependency of `@etherfold/utils` only so the documented rebuild command runs via `pnpm exec`; nothing in `src/` or `test/` invokes it, and the test's only filesystem use is reading the committed fixture, which is the caller's job by design. This was the failure mode I was most watching for and it is clean.
- [x] **Tests mirror the repo's style,** and the 30 cases are chosen rather than enumerated. The detector's false-positive surface is tested where it actually bites: a bundle that merely MENTIONS a package name, an import statement quoted inside a string (not at a statement position), an export clause, a default export, `import.meta`, and statements at every position a bundler can emit one. That matters beyond this task, because `a-path-naming-an-unbundled-entry-point-is-refused` is told to reuse this judgement rather than invent a second heuristic.
- [x] **A changeset accompanies it,** and it is a good one: it states the sequencing, names the three capabilities, and records that there is no browser path AS A DECISION rather than an omission.

**Placement was the open question and it was answered and recorded.** An earlier run of this task stopped short of implementing, having reached exactly one undecided point: `@etherfold/utils` versus `@etherfold/core` beside `stream/seed.ts`. This build chose `@etherfold/utils` and justified it in the module header and the barrel export: it is a different ARRIVAL rather than a variation of one (that package takes a PATH and resolves it through the module system; this takes BYTES that carry their own name and resolve nothing), the node-side package can use `node:crypto` where `@etherfold/core` needs a hash a browser can bundle, and the browser exclusion this task makes deliberately matches a package documented as not being a browser bundle. That is the right call and it is now a decision on the record rather than an accident of where someone typed.

**It consumed the finding that landed an hour earlier.** The reason there is no browser path is no longer an assertion: it cites the CSP measurement from `a-tab-can-or-cannot-instantiate-a-processor-from-bytes-under-a-csp` (#160), concludes that a browser arrival implies a service worker, and declines to make that decision here. The absence of `isProcessorArtifactIdentity` is likewise argued rather than omitted: `GenerationId.processor` is a string nothing in the tree parses, so a predicate sniffing "is this a hash" would be the first parser and would break the two-arrivals-two-derivations property the family depends on. That is precisely ADR-0086's invariant, held by refusing to add something.

**One off-path finding filed, correctly not fixed here:** `work/notes/observations/a-relative-processor-modules-own-error-is-swallowed-by-the-package-fallback.md` records that `loadProcessorModule` discards a relative module's own import error and reports the node_modules resolution failure instead, so an operator whose module throws is told their PATH is wrong. Out of scope, captured, with a suggested shape. I am carrying it forward: it lands squarely on `a-path-naming-an-unbundled-entry-point-is-refused`, whose whole deliverable is that an author meets a kind refusal rather than a stack trace from inside a loader.

Nits: none blocking. The `esbuild` devDependency is benign (dev only, never shipped, and it keeps the fixture's rebuild toolchain named next to the fixture). Gate 2 is off in this repo, so there is no `review-nits-*` note for this task.

Merging. This unlocks `a-deployment-runs-from-a-bundle-identified-by-its-hash`, which is the tracer bullet the four migrate batches then move onto.

PR #162, opened 2026-09-17
feat(a-deployment-runs-from-a-bundle-identified-by-its-hash): A deploym…

wighawag:
Task: `work/tasks/done/a-deployment-runs-from-a-bundle-identified-by-its-hash.md`

Everything is green. Final summary:

## What I did

**The road, in `@etherfold/utils`:** new `openProcessorArrival(path, options)` (`packages/utils/src/processorArrival.ts`) — one operator-supplied path landing on whichever of the two arrivals it describes, answering `{processor, processorModule, identity?}`. A path whose bytes are self-contained is read, hashed (`sha256:<hex>`) and instantiated through `loadProcessorArtifact`; anything else goes through `loadProcessorModule` + `instantiateProcessor` exactly as before and comes back with no identity. Covered by `packages/utils/test/processorArrival.test.ts` (8 cases).

**The seam, in `@etherfold/processor-entities`:** `EntityEventProcessorOptions.identity`, the identity a host HANDS a fold. `getVersionHash()` answers with it when given, else the declared hash unchanged. This is what keeps ONE answer to "which generation is this" through the registry, the receiver, the feed and the stored cursor — `@etherfold/core` is untouched.

**The CLI:** `prepareIndexing` (`run`/`build`), `indexCommand` and `reconfigure` all resolve the processor through the arrival; `foldPartsFor`/`openFolding` take the arrival's identity and pass it to both the state namespace and the fold. `FoldParts.versionHash` → `processorIdentity`. `reconfigure`'s `unchanged` message now says the right thing for each arrival.

**Fixtures + tests:** two committed real bundles (`packages/cli/test/fixtures/processor-bundle/`, esbuild 0.28.1, `--bundle --format=esm --minify`, both prettier-ignored, README documents the rebuild command) that differ in ONE handler line and in nothing a declared identity is computed from. `packages/cli/test/aDeploymentRunsFromABundle.test.ts` (10 cases) drives a deployment the way the CLI tests already do: bundle folds end to end; identity IS the bundle hash; byte-identical bundles at different paths give the same generation; one edited handler gives a different one where `getVersionHash()` cannot tell them apart; the declared path still registers `entityProcessorVersionHash`; a bundle that throws leaves the deployment exactly as it was; and a rebuilt bundle reaching a running `run` over `/admin/reconfigure`.

Changeset added. `pnpm build && pnpm typecheck && pnpm test && pnpm format:check` all green; `check:adr`, `check:refs`, `check:changesets` pass. Working tree holds only this task's files.

## Decisions

**A BUNDLE is defined as "a module that expects nobody else to resolve anything" (`unresolvedImportsOf`), and that is what discriminates the two arrivals.** Chosen because it is the repository's ONLY definition of self-contained: it is what the artifact loader refuses on, and `a-path-naming-an-unbundled-entry-point-is-refused` is explicitly told to reuse that same judgement. The alternative considered and rejected was a second, transitional INPUT ("this path names a bundle") to select the derivation explicitly: cleaner during expand, but it invents a user-visible concept that four later tasks would have to delete without knowing it exists, and it contradicts "a PATH is still how a deployment names its processor". The consequence I am flagging rather than burying: **a hand-written entry point that happens to import NOTHING is a bundle by this definition and is identified by its bytes**, not by its `version`. That is the same answer the end state gives it and it is honest about what the file is, but it is a behaviour change for that class of configuration during the expand step. What it touches: the refusal task inherits this judgement wholesale; and two existing CLI suites (`anEndpointReconfiguresARunningRun`, `theDeploymentSelectsItsPromotionPolicy`) wrote import-free modules to disk and would have silently moved onto the artifact route, so their fixtures now import their ABI from a sibling file — which is what a real unbundled entry point does — with a comment at each saying why not to inline it back. One case in the drift suite genuinely flipped (`registered` instead of `unchanged`) before that fix, which is how the hazard was found.

**The identity is handed to the FOLD (`EntityEventProcessorOptions.identity`), not to the container.** Chosen because the container derives `wanted` from `processor.getVersionHash()`, and `StreamBuilder` derives its advertised `generation` and its `lastSync.context.processor` the same way — so telling only the container would have left the registry saying one thing and the receiver, the feed and the stored cursor saying another. Handing it to the engine gives ONE identity everywhere with no change to `@etherfold/core` at all. Alternative considered: `ReceivedGenerationSpec.processorIdentity` in core, which would have needed the engine change anyway plus edits inside `core-takes-its-identity-from-the-arrival`'s batch fence. It is NOT the caller-declared version hash ADR-0043 rejected: it is derived from something the class cannot see and REPLACES the computation rather than sitting beside it. What it touches: `the-processor-packages-take-their-identity-from-the-arrival` (this is the option its batch migrates onto) and `core-takes-its-identity-from-the-arrival` (which stays free of this task's edits).

**`FoldParts.versionHash` is renamed to `processorIdentity`.** A field whose name says DECLARED VERSION while it may hold a hash of bytes is exactly the silent re-meaning the coherence check warns about. Alternative: leave the name and widen the doc — rejected, because every later reader inherits the wrong noun. What it touches: `the-cli-server-and-examples-take-their-identity-from-the-arrival` will see the new name; nothing outside `packages/cli` referenced the old one.

**`openProcessorArrival` RAISES on an artifact refusal, where `loadProcessorArtifact` answers with data.** An operator-supplied path is configuration, and this repo's configuration layer refuses by raising; the module arm beside it raises too, so one input has one failure shape. The artifact's `identity`, `reason` and `why` are carried into the message so the refusal still names WHICH bytes it refused. What it touches: `a-path-naming-an-unbundled-entry-point-is-refused` owns the operator-facing wording and may want to reshape this message; nothing here claims that wording.

**`reconfigure`'s `unchanged` message branches on WHICH ARRIVAL supplied the identity.** Telling an author who cannot declare a `version` to bump one is a wrong instruction at the exact moment they are confused. It branches on the arrival answering "did I derive an identity", never on the identity string — nothing parses `GenerationId.processor` and this does not start. What it touches: `the-declared-version-and-the-drift-report-are-deleted` removes the declared half of this branch.

**`--processor`'s help text and the CLI README now say a path may name a bundle.** An accepted input that the documentation does not describe is a deployment believing something untrue. Kept to one paragraph in the README and one flag description; the full build-command documentation is `the-build-command-and-its-pinning-rule-are-documented`'s and I did not pre-empt it.


wighawag:

Reviewed the diff against the seven acceptance criteria. This is the tracer bullet, so the two things I weighted hardest were whether the DECLARED path really survives (four migrate batches and a contract task depend on it) and whether the identity property is demonstrated rather than asserted.

- [x] **A deployment configured with a BUNDLE runs end to end: read, hashed, instantiated, registered, folding blocks.** The arrival is threaded through every CLI layer the ordinary path already goes through (`config.ts`, `folding.ts`, `index.ts`, `indexCommand.ts`, `reconfigure.ts`) rather than bolted onto one entry point, which is what the task asked for and what makes the later batches a move onto an existing road.
- [x] **The registered generation's `processor` identity IS the bundle's `sha256:<hex>` and nothing else.** Asserted directly, and separately from the end-to-end fold, so a pass cannot come from the fold alone.
- [x] **Byte-identical bundles give the SAME generation; one changed handler gives a DIFFERENT one, with no author action either way.** This is the property the whole family exists for and the way it is tested is better than the criterion required. The same-identity case copies the bundle to a DIFFERENT PATH first, so it proves the identity is the bytes and not the location, which is user story 3 (a laptop and CI checking out into different directories). And then there is the test that earns the PR: `entityProcessorVersionHash(edited())` is asserted EQUAL to `entityProcessorVersionHash(base())`, using two fixtures that differ in one handler line and in nothing a declared identity is computed from. So the pair is provably indistinguishable to the old identity and distinguishable to the new one. That is ADR-0008's silent wrong-state condition exhibited and then made impossible, in one test, rather than argued in a comment.
- [x] **The DECLARED path is untouched.** Checked by diff and not by claim, because this is the criterion the rest of the family rests on. `getVersionHash()` still exists on `EntityEventProcessor` and still computes `entityProcessorVersionHash` — it gained a `??` fallback, not a replacement. `assertProcessorVersion`, `getCodeFingerprint()`, `packages/core/src/utils/fingerprint.ts`, `packages/core/src/processorDrift.ts` and the `PROCESSOR DRIFT` report are all still on the branch, tests included. Two tests assert the declared path positively (a module that is not a bundle keeps `getVersionHash()`; a path naming a module that still expects somebody else to resolve an import still resolves). Nothing is deleted anywhere in `packages/` except lines that moved.
- [x] **A bundle that fails to instantiate leaves the deployment exactly as it was, with nothing partially registered,** and the refusal names both the path it read and the identity of the bytes it refused. That second part is the artifact unit's "a refusal can say WHICH artifact it refused" paying off at the layer above it.
- [x] **Tests mirror the repo's existing test style,** and the fixture pair (`nfts.bundle.js` + `nfts-edited.bundle.js`) is built once by the documented `esbuild --bundle --format=esm --minify` command and committed, with sources beside it and a README carrying the exact rebuild line. No test runs a bundler: the README explicitly declines to add `esbuild` to the CLI, pointing at the workspace's copy instead, on the grounds that nothing in the CLI bundles anything. Neither hash is pinned; identity is asserted as a property, so a newer esbuild is not a test failure.
- [x] **A changeset accompanies it,** correctly scoped to the three packages that changed, and it states that the declared path and the fingerprint are untouched and that the batches depend on that.

**`EntityEventProcessorOptions.identity` is the right shape and it anticipated the obvious objection.** The header distinguishes it from the caller-declared version hash ADR-0043 rejected, and the distinction is real rather than rhetorical: that one was a caller restating a value the class also computes, so the two could silently disagree, whereas this is derived from something the class cannot see (bytes it was never given) and REPLACES the computation instead of sitting beside it. So there are still never two live answers, and ADR-0053's before-the-processor-exists namespace naming stays honest because the host passes the same value to both. It also notes that `configure()` cannot move an arrival-supplied identity, and why that is correct rather than a limitation (the config a bundle was built with is in the bundle).

**One behaviour change, disclosed rather than hidden, and I judge it correct.** "A bundle is a module that expects nobody else to resolve anything" reuses `unresolvedImportsOf`' existing judgement, so a hand-written entry point that imports NOTHING counts as a bundle and is identified by its bytes rather than by its `version`. That is a real edge of the DECLARED-path-untouched criterion, and the changeset says so in as many words instead of letting someone discover it. I am not blocking on it, for three reasons: the task explicitly instructed reuse of that judgement rather than a second heuristic, and a "is this REALLY meant to be a bundle" predicate would be that second heuristic; the shape is test-shaped rather than deployment-shaped, since a real processor imports its ABI; and the contract task makes bytes identity universal two steps later, so this is early arrival at the destination rather than a divergence from it. Nothing is published, so no persisted state is at stake either way.

**A rebuilt bundle reaching a running deployment is covered too** ("unchanged while the bytes are, and registers the new hash once they move"), which is the reconfigure path picking up hash identity without the drift report being touched — exactly the split this step needs, since retiring that report belongs to `the-declared-version-and-the-drift-report-are-deleted`.

Nits: none blocking. Reaching into `../utils/node_modules/.bin/esbuild` in the fixture README is slightly fragile as a path, but it is documentation for a manual rebuild and it buys the right thing (no bundler in the CLI's dependency tree), so it stands. Gate 2 is off in this repo, so there is no `review-nits-*` note for this task.

Merging. This unlocks all four `*-takes-its-identity-from-the-arrival` migrate batches at once: `core`, `the-processor-packages`, `the-browser`, and `the-cli-server-and-examples`. I am landing them one at a time rather than in parallel, because two concurrent `dorfl do` jobs on this repo were measured today deleting each other's worktree (`work/notes/observations/two-concurrent-dorfl-do-jobs-on-one-repo-delete-each-others-worktree.md`), so the batches' deliberate file-orthogonality is unusable through `do` for now.

PR #163, opened 2026-09-17
feat(core-takes-its-identity-from-the-arrival): `@etherfold/core` takes…

wighawag:
Task: `work/tasks/done/core-takes-its-identity-from-the-arrival.md`

All checks green. Here is my report.

## What I did

One MIGRATE batch of ADR-0086's wide refactor, scoped to `packages/core`. Ground truth before starting: 60 `getVersionHash` mentions in the package, of which **11 were real call sites in source** (`container.ts` 1, `receivingContainer.ts` 2, `indexer.ts` 4, `streamBuilder.ts` 2, `generation/rebuild.ts` 2) and the rest were fakes, doc comments and assertions in tests.

**The seam.** `packages/core/src/internal/processorIdentity.ts` (new, internal — not on the published surface) holds the one expression every engine asks: `processorIdentityOf(processor, supplied)` = the arrival's identity, or the processor's own declared hash where no arrival derived one. Every one of the 11 call sites now asks it, so exactly **one** `getVersionHash()` call remains in core source: the fallback inside that function, which the contract task deletes.

**Where a host hands one over** (all optional, all documented at the site):
- `GenerationSpec.processorIdentity`, inherited by `ReceivedGenerationSpec` through its existing `Pick`.
- `IndexerGenerationOptions.processorIdentity` — a fifth constructor argument on `IndexerGeneration`, held *beside* `processor` and replaced *with* it by `updateProcessor`, which also gained `processorIdentity` in its options.
- `StreamBuilderOptions.processorIdentity` and `GenerationRebuildOptions.processorIdentity`.
- `IndexerOptions.createGeneration` gained a fifth parameter carrying the **resolved** identity; a four-parameter factory still compiles.

Both containers resolve the identity once, register under it, and hand `record.processor` down to whichever engine shape the fold turns out to be, so the record, the receiver, the rebuild and the engine cannot name one generation four ways. `ReceivingIndexer.cursorOf` now reads the identity off the record rather than re-asking the processor.

**Tests migrated with their source.** A new `packages/core/test/utils/processorIdentity.ts` gives a fold an identity by supplying **bytes**: `bundleBytes(marker)` → `identityOfBytes` → `sha256:<hex>`, plus `markerOf` so slot assertions still read as sentences. No bundler anywhere. 27 test files moved onto it; all 67 core suites pass (1193 tests).

**Nothing deleted.** `version`, `getVersionHash()`, `getCodeFingerprint()`, `utils/fingerprint.ts` and the `PROCESSOR DRIFT` report all still exist and still work; `types.ts` now says in a docstring what `getVersionHash()` is *for* during the migration and which task removes it.

**Gate:** `format:check && check:adr && check:refs && check:changesets && build && typecheck && test` — green across the whole tree. `changeset status --since=main` passes with the changeset staged (verified through a scratch index; the real index was not touched). Only `packages/core/**` and one new `.changeset/*.md` are modified; no sibling package was edited.

Two notes on things outside this batch, fixed by nobody here: `CONTEXT.md`'s glossary still describes a generation as `{stream, processor version hash}` — that is explicitly the contract task's acceptance criterion, so I left it. And `packages/browser/src/host/serve.ts` still names its state namespace from `built.getVersionHash()`, which is the browser batch's own call site.

## Decisions

**The name is `processorIdentity`, and it is a field on the options bag that accompanies the processor at every entry point.** Chosen because `packages/cli` already spells exactly this concept `FoldParts.processorIdentity` (landed by `a-deployment-runs-from-a-bundle-identified-by-its-hash`), so the batch that migrates the CLI meets a name it already uses. Plain `identity` was rejected as a re-meaning: a generation's identity IS `{stream, processor}`, so `GenerationSpec.identity` would name the whole `GenerationId` and mean two things one line apart. What it touches: the three sibling batches all see this name, and the contract task deletes the fallback behind it.

**It is deliberately NOT a key on `ProvidedIndexerConfig`, even though that is the bag already threaded from the container to the engine and would have needed no signature change.** Rejected because `IndexerOptions.config` is ONE value shared by every generation a container builds, so a `processorIdentity` there would silently give two generations one name — one registry record, one state namespace, and two specs folding a stream that is only folded once. The cost of the alternative I took is a fifth parameter on `IndexerOptions.createGeneration`. What it touches: `@etherfold/browser`'s `createIndexerState({createIndexer})` mirrors that hook. An existing four-parameter factory still type-checks and still compiles (TypeScript accepts fewer parameters), but it would silently DROP the identity, so the browser batch must widen its own factory when it migrates. I flag it rather than bury it because a compiler will not.

**`updateProcessor`'s "the swap was skipped" warning now branches on WHICH ARRIVAL supplied the identity.** Telling an author who cannot declare an identity to "bump the processor's version hash" is a wrong instruction at the exact moment they are confused. This mirrors the precedent the blocking task set for the CLI's `reconfigure` `unchanged` message, so the two read the same. It branches on whether an arrival answered — `options.processorIdentity === undefined` — and never on the identity string, because nothing in the tree parses `GenerationId.processor` and this does not start. What it touches: it is user-visible prose on a core log line; `the-declared-version-and-the-drift-report-are-deleted` removes the declared half of the branch.

**The identity is resolved LAZILY at each site, not captured at construction.** An arrival-supplied identity is a constant, but the DECLARED fallback is not: `getVersionHash()` covers a processor's config as well as its version and `configure()` can move it, which is precisely why `StreamBuilder.generation` reads it on every call today. Capturing at construction would have been tidier and would have changed un-migrated behaviour. The alternative is available to the contract task for free, once the fallback is gone.

**One test is deliberately left on the declared fallback**: `streamBuilder.test.ts`'s `reads the DECLARED fallback live, so a processor reconfigured after construction is not misreported`. Criterion 3 says the old form must still *work*, and a migrate batch that left no assertion of that inside this package would be trusting the sibling packages to notice. It is labelled at the site as the case the contract task retires with the path it describes. Everything else in the package takes its identity from the arrival.

**The test helper spells the `sha256:<hex>` rendering itself instead of importing `processorArtifactIdentity`.** `@etherfold/utils` depends on `@etherfold/core` and not the other way round, so core cannot import it. Recorded because it looks like a forked derivation and is not one: what core is under test for is that it takes WHATEVER it is handed, so the exact derivation is the test's business. The docstring says so and points at both prior arts (`stream/seed.ts`'s `streamSeedContentHash` and the artifact loader).


wighawag:

This is the first and largest of four MIGRATE batches, so I checked the three ways a batch can pass a green gate while wrecking the batches that follow: deleting the old form, editing a sibling package's files, and reaching for a bundler in tests. All three are clean.

- [x] **Every `getVersionHash()` call site in `packages/core` takes its identity from the arrival instead, source and tests together.** Verified by grep rather than by claim: exactly three live `.getVersionHash()` invocations remain in the package, and each is correct. One is `processorIdentityOf`'s single fallback expression, the funnel that 62 scattered sites collapsed into. The other two are NEGATIVE test assertions (`expect(...).not.toBe(fold.processor.getVersionHash())`) that prove the registered identity is no longer the declared hash when an arrival supplied one, which is the migration demonstrating itself. Everything else that still mentions the name is prose explaining the old form, or the seam declaration.
- [x] **Nothing is deleted.** `getVersionHash(): string` is still declared on the seam in `types.ts`; `packages/core/src/utils/fingerprint.ts` and `packages/core/src/processorDrift.ts` are both still on the branch with their suites; `version`, `getCodeFingerprint()` and the `PROCESSOR DRIFT` report all still exist and still work. Every new option is OPTIONAL and absent falls back to the processor's own hash, so the three batches that have not run are unaffected. This is the criterion the whole sequence rests on and it holds.
- [x] **No file outside this batch's package is edited.** The diff is `packages/core/**` plus its own changeset plus its own task done-move, and nothing else. Zero sibling-package edits.
- [x] **Tests that declared a `version` now supply BYTES, and none of them runs a bundler.** `test/utils/processorIdentity.ts` is the new helper and it does exactly what the task prescribed: synthetic bytes per marker, SHA-256 over the octets, rendered `sha256:<hex>`. Its header argues the case rather than just doing it, including the sentence that shows the design was understood and not merely followed: "A test that shelled out to `esbuild` to get one would have misread the design." It also explains why the rendering is spelled out locally instead of imported (`@etherfold/utils` depends on `@etherfold/core`, not the reverse) and why that is fine (what core is under test for is that it takes whatever it is handed, so the derivation is the test's business and never core's). That is the right reason, not a convenience.
- [x] **The batch is GREEN on its own** with the rest of the tree unchanged, and the gate runs the WHOLE tree (`pnpm build && pnpm typecheck && pnpm test`) on the branch rebased onto `main`, so the claim that the unmigrated siblings still compile against the new optional options is verified rather than asserted.
- [x] **A changeset accompanies it,** and it enumerates each seam that gained an identity, says plainly that nothing is removed and no caller has to move yet, and names the contract task that will delete the fallback.

**`processorIdentityOf` is the pattern the other three batches will follow, and it is a good one to inherit.** Two details in its header are load-bearing. It refuses to add a predicate asking whether an identity "looks like a hash", on the grounds that this would be the first parser and the engine is not allowed an opinion on the value — the same restraint the artifact unit showed in #161, so the family is consistent about ADR-0086's invariant in both directions. And it explains why the identity is asked at each SITE rather than captured once at construction: the declared half is not a constant, because `getVersionHash()` covers the config and `configure()` can move it after construction, so a snapped value would advertise a fold that is no longer running. An arrival-supplied identity IS constant (the config a bundle was built with is in the bundle), so asking again simply answers the same thing. That reasoning is why this refactor did not quietly introduce a staleness bug.

**The drift suite was migrated rather than weakened.** `processorDrift.test.ts` now sources its identities from the arrival, but what it ASSERTS is unchanged: the fingerprint on a persisted cursor against the one the running processor reports, at an identity that did not move. Retiring that report belongs to `the-declared-version-and-the-drift-report-are-deleted`, and this batch correctly left it alive and passing.

**One disclosed behaviour change, and it is the right one.** `updateProcessor`'s "the swap was skipped" warning no longer tells an author to bump a version hash when the identity came from an arrival, because an author who cannot declare an identity cannot bump one. Critically it branches on WHETHER an arrival supplied a value, never on the value itself, so it does not become the parser the module just refused to add. `IndexerOptions.createGeneration` gaining a fifth parameter is backward compatible: an existing four-parameter factory still compiles.

Nits: none blocking. Gate 2 is off in this repo, so there is no `review-nits-*` note for this task.

Merging. Three migrate batches remain (`the-processor-packages`, `the-browser`, `the-cli-server-and-examples`); they are file-orthogonal to each other and to this one, and I am landing them one at a time because concurrent `dorfl do` jobs on this repo were measured deleting each other's worktree today.

PR #164, opened 2026-09-17
feat(the-processor-packages-take-their-identity-from-the-arrival): The…

wighawag:
Task: `work/tasks/done/the-processor-packages-take-their-identity-from-the-arrival.md`

Whole tree green, formatting and the three repo checks clean.

## What I did

Ground truth first: the measured 30 was **33 `getVersionHash` mentions** in the batch. Of those, most in `src/` were docstrings and two were the method definitions themselves (which must survive); the real identity-taking sites were in the tests plus one missing seam.

**The seam change (`packages/processor-sqlite`).** `VersionedStateEventProcessor` had no way to be handed an identity: `getVersionHash()` computed `entityProcessorVersionHash(processor, config)` unconditionally, while the `EntityEventProcessor` it delegates to already took `options.identity` (from the blocking task). Added `VersionedStateProcessorOptions.identity`, answered it from `getVersionHash()`, and **forwarded it** to the inner fold built in `folded()`, so the wrapper and the fold underneath give one answer rather than two. Test-first: three cases in `test/lifecycle.test.ts` went red, then green.

**Call sites moved onto the arrival (bytes, no bundler).** Added `test/utils/processorIdentity.ts` to each package: synthetic `bundleBytes(marker)` hashed to `sha256:<hex>` with `node:crypto` (the same rendering `processorArtifactIdentity` produces; no new dependency, no lockfile churn).
- `processor-entities/test/entity-event-processor.test.ts`: the stored-context-mismatch case no longer bumps `version: '2.0.0'`; it folds the same object under `identityOf('v2')`. The existing arrival case now uses the helper instead of a hand-spelled digest.
- `processor-sqlite/test/lifecycle.test.ts`: same migration, plus the three new arrival cases.
- `processor-sqlite/test/deployment-shapes.test.ts`: both deployment shapes now fold under **one** handed identity (`PROCESSOR_IDENTITY`), which restates that file's own claim: the fold's name cannot depend on the backend under it. The test-local `EventProcessor` answers what it was told instead of deriving from `authored.version`.

**Nothing deleted.** `version`, `assertProcessorVersion`, `getVersionHash()`, `getCodeFingerprint()` and the fingerprint all still exist and still work; `processor-sqlite/test/version.test.ts` and three cases in `lifecycle.test.ts` and one in `entity-event-processor.test.ts` are retained on the declared path and labelled at the site as belonging to `the-declared-version-and-the-drift-report-are-deleted`.

**Docs that instructed the old form** (in-package only): `BootstrapOptions.processor`, `createSnapshot`'s `processor`, `bootstrapFromSnapshot`'s example (which literally told a caller to call `getVersionHash()`), `entityProcessorVersionHash`, both `getVersionHash` docstrings, and `EntityProcessor.version`. The snapshot candidate *predicate* and the label's *value* are untouched, since `a-snapshot-is-labelled-with-the-identity-it-was-computed-under` owns those.

No file outside the two packages was edited (plus the required changeset).

One non-finding worth stating so a reviewer does not chase it: `packages/core`'s test count rose by one between runs. That is `core/test/pendingChangesets.test.ts`, which registers one case per file in `.changeset/` — my changeset. Not a flake, no note filed.

## Decisions

**The option is named `identity`, matching `EntityEventProcessorOptions.identity`, and NOT `processorIdentity` as `packages/core` chose.** Both spellings already exist in the tree and they sit on different layers: a *generation/engine* option says `processorIdentity` (because `identity` there would collide with `GenerationId`, which is exactly why core rejected the short name), while the thing *accompanying a processor object* says `identity` — `EntityEventProcessorOptions.identity`, and `ProcessorArrival.identity` in `@etherfold/utils`. `VersionedStateProcessorOptions.identity` forwards straight into the former, in the same package family, so matching it is what keeps one concept one word at this layer. The alternative, renaming `EntityEventProcessorOptions.identity` to `processorIdentity` for tree-wide uniformity, was rejected because its only caller is `packages/cli/src/folding.ts` (`{identity}`), so the rename would require editing a sibling package and break criterion 4 and every parallel batch. What it touches: `the-cli-server-and-examples-take-their-identity-from-the-arrival` meets `identity` here and `processorIdentity` in core/cli; if either batch or the contract task wants one word everywhere, the rename is theirs to make in one commit rather than mine to make across a fence.

**Criteria 1 and 3 conflict inside these packages, and I resolved it the way the sibling batch did: the DECLARED path's own test suites stay.** `processor-sqlite/test/version.test.ts` is entirely assertions *about* `getVersionHash()`, the fingerprint and `PROCESSOR DRIFT`, and `lifecycle.test.ts`'s `describe('getVersionHash')` pins the declared formula. Those are `getVersionHash()` call sites, so criterion 1 read literally says migrate them — but they are not sites that *take* an identity, they are the witness that the old form still *works*, which criterion 3 requires and which `the-declared-version-and-the-drift-report-are-deleted` is explicitly tasked to delete. Rewriting them onto an arrival identity would not migrate the declared path, it would silently delete its coverage while leaving the code, which is the failure mode the task's own prompt warns about. `core-takes-its-identity-from-the-arrival` recorded the same judgement for one `streamBuilder.test.ts` case. So every site that *sources* an identity now sources it from the arrival, and the declared path's own suites are retained with a comment at each naming the task that retires them. The alternative (migrate literally everything) was rejected on the above; the other alternative (delete them here) is forbidden outright. What it touches: the contract task will find ~14 retained declared-path assertions in these two packages, labelled, rather than none.

**The forwarding to the inner fold is asserted by reading a PRIVATE field (`innerFoldOf` in `lifecycle.test.ts`).** The inner `EntityEventProcessor` is never asked its identity by anything in production — core holds the wrapper — so the forwarding is unobservable through the public surface, and an assertion is the only thing that stops a later edit dropping it and reintroducing two live answers to "which fold is this". The alternative considered was not testing it at all (it changes no current behaviour) or exposing the inner fold, which would widen a published surface to serve a test. `version.test.ts` already reaches into a private field the same way (`(p as unknown as {version}).version = undefined`), so this is precedented rather than novel; the helper's docstring says why the reach is there.


wighawag:

Migrate batch 2 of 4. Same three failure modes checked as on #163 (deleting the old form, editing a sibling package, a bundler in tests), all clean, plus one reading of the criteria that this batch specifically needed to get right.

- [x] **No file outside this batch's packages is edited.** The diff is `packages/processor-entities/**` and `packages/processor-sqlite/**` plus its own changeset and task move. Nothing else, so the two batches still to run cannot collide with it.
- [x] **Nothing is deleted.** `entityProcessorVersionHash` is still exported, `getVersionHash()` is still implemented on both classes, and `version`, `getCodeFingerprint()` and the `PROCESSOR DRIFT` report all still exist and pass. The new `identity` is OPTIONAL on both sides, so absent behaves exactly as before. I re-grepped for removals of the old form specifically and found none.
- [x] **Every call site takes its identity from the arrival, read correctly for THIS batch.** These are the two packages that IMPLEMENT the seam, so a literal "no `getVersionHash()` anywhere" reading would have meant deleting the declared computation, which the same task forbids. The right reading, and the one the task's own framing supports, is that these packages must ACCEPT an arrival identity and answer with it, keeping the declared computation as the fallback. That is what landed: `VersionedStateProcessorOptions.identity` is what `getVersionHash()` returns when a host supplied one, and the remaining `getVersionHash()` calls in tests are assertions ABOUT that behaviour (`handed` returns the supplied identity; `declared` still returns `entityProcessorVersionHash`; the two differ), not unmigrated callers.
- [x] **The delegation is the part that mattered and it is right.** `VersionedStateEventProcessor` wraps `EntityEventProcessor`, so the identity is held on the wrapper AND forwarded to the inner fold it builds on first use, and the code comment names the hazard it is closing: two live answers to "which fold is this" is exactly what a wrapper would otherwise introduce. There is a test reaching into the inner fold to assert it received the identity, so the forwarding is verified rather than assumed. The type is no longer a pure pass-through to the store, and the diff says why (the store never sees an identity), which is the kind of thing that otherwise reads as an accident later.
- [x] **`configure()` cannot move an arrival-supplied identity,** asserted, with the declared case asserted to still move on configure. That pair is the right shape: the config a bundle was built with is in the bundle, so an arrival identity is a constant, while the declared one is a function of the config.
- [x] **Tests supply BYTES and none runs a bundler.** Consistent with the helper #163 established.
- [x] **The batch is GREEN on its own,** and the gate ran the whole tree on the rebased tip, so the two unmigrated batches still compile against this.
- [x] **A changeset accompanies it,** correctly `minor` for `processor-sqlite` (new option) and `patch` for `processor-entities` (documentation only), and it states that nothing is removed and names the contract step that will delete the declared path.

**A useful piece of groundwork, and it stopped at the right line.** The batch also corrected docstrings that told a caller to reach for the declared hash: `BootstrapOptions.processor` and `createSnapshot`'s `processor` now describe WHICH FOLD computed the rows (a value compared for equality and never parsed) instead of "the version hash", and `bootstrapFromSnapshot`'s example passes the identity the arrival handed the fold. That is preparatory for `a-snapshot-is-labelled-with-the-identity-it-was-computed-under`, and critically it left the CANDIDATE RULE untouched, which is exactly what that task is instructed not to change. `EntityProcessor.version` now says plainly that it is superseded and not for new code, which is the honest state during a migration.

**One nit, carried forward rather than blocking.** `packages/processor-entities/README.md` still shows `{processor: eventProcessor.getVersionHash(), …}` in its `openAndBootstrap` example, so the docstrings moved and that README block did not. It is not blocking: the declared path still works, the criterion covers source and tests rather than prose, and that example is squarely the snapshot task's subject matter. But it is a real spot that a green gate cannot catch, because a README is not compiled. I am carrying it into Gate 3 of `a-snapshot-is-labelled-with-the-identity-it-was-computed-under` (which should land the new value there) and re-checking it at the contract task, whose "no caller left behind" criterion a stale prose example could otherwise slip past.

Merging. Two migrate batches remain: `the-browser` and `the-cli-server-and-examples`.

PR #165, opened 2026-09-17
feat(the-browser-takes-its-identity-from-the-arrival): `@etherfold/brow…

wighawag:
Task: `work/tasks/done/the-browser-takes-its-identity-from-the-arrival.md`

## What I did

One MIGRATE batch scoped to `packages/browser`: every place the package **sourced** a generation's fold identity now takes the one the ARRIVAL supplied, with the author-declared path left intact underneath it.

**Source (3 sites + the threading):**
- `BrowserGenerationSpec.processorIdentity?: string` (`src/IndexerState.ts`) — handed straight to `GenerationSpec.processorIdentity` in `@etherfold/core`, and inherited by `HostedIndexerSpec`, so a worker host takes it too.
- `createIndexerState`'s per-generation store map (`generationSpecFor`) and the worker host's (`generationSpecOf`, `src/host/serve.ts`) now key on `processorIdentity ?? built.getVersionHash()` — the same expression the container resolves the registry record with, so the name filed and the name a store is looked up by cannot diverge.
- `addGeneration` takes a `processorIdentity` (a reconfigure is a new fold, so it arrives with its own), and the `createIndexer` factory option gained the fifth argument that core's `createGeneration` gained, documented because a 4-parameter factory still compiles and would silently drop it.
- `BrowserStateStore`'s bootstrap docstring no longer tells a caller to call `getVersionHash()`.

**Tests (7 sites), written test-first** at the seam the package already tests at: a new failing case in `test/generationContainer.test.ts` ("names the generation by the identity the ARRIVAL supplied, not by what the processor declares") — two generations built from the *same* definition, therefore the same declared version, registering as two generations because their bytes differ. New `test/utils/processorIdentity.ts` supplies synthetic BYTES (`sha256:<hex>`, `node:crypto`); no bundler anywhere. Migrated `setupIndexing`, `txInclusion`, `dispose`, `liveReload` (the fakes now answer a `declared-version-…` nothing reads), `entityBootstrap` (snapshot label and client identity are the arrival's, not a computed version hash), `promotion`, `generationProgress`, `everyGenerationIsLevelBeforeTheDriverRests`, `generationContainer`, `aTabHoldsItsGenerationsInSlots` and `utils/stateMovedTransports`.

**Nothing deleted:** `version`, `getVersionHash()`, `getCodeFingerprint()` and the fingerprint are untouched and still work; `processorIdentity` is optional everywhere and absent means the old fallback.

**Gate:** `format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test` all green across the whole tree (browser: 48 files / 358 tests). Changeset added at `.changeset/the-browser-takes-its-identity-from-the-arrival.md`. Only `packages/browser/**` plus that changeset are modified; no sibling package was edited.

Nothing wrong was found outside the batch, so no observation note was filed.

## Decisions

**`updateProcessor` is deliberately NOT given a `processorIdentity`, and its suites stay on the declared path.** What arrives at that call is a MODULE OBJECT a dev server handed the tab: there are no bytes to hash, and `a-module-handed-to-a-tab-is-identified-by-its-handler-sources` states in as many words that the app must not supply that arrival's identity ("do not let the app supply the identity … that is the author-declared identity re-entering through the one door left open"). Widening the options bag would have been the mechanical move (core already accepts `{force, processorIdentity}`) and would have let me migrate `reconfigure.test.ts` too, but it invents an input the HMR task would have to delete. So `reconfigure.test.ts`'s axis one and `liveReload.test.ts`'s four `updateProcessor` cases keep declaring versions, each labelled at the site with the task that owns them, and a `## WHY THERE IS NO processorIdentity HERE` section sits on `updateProcessor`'s JSDoc so the asymmetry with the generation spec is not a mystery. **Touches:** `a-module-handed-to-a-tab-is-identified-by-its-handler-sources` (which will add the derivation there) and `an-hmr-update-reconfigures-the-tab-it-is-running-in`.

**Criteria 1 and 3 conflict on two `getVersionHash()` sites, and I resolved it the way the sibling batches did: a site that does not SOURCE an identity stays.** `callShape.test.ts`'s free-form fixture is a value asserted to be REFUSED by `@ts-expect-error` (the retired JS-object shape, ADR-0037) and names no generation; `reconfigure.test.ts:68` is a docstring describing the declared formula on the HMR axis above. Migrating either would delete coverage rather than move a caller. Alternative considered: migrate literally every occurrence — rejected because criterion 3 requires the declared form to go on WORKING and something inside this package has to witness that. **Touches:** the contract task will find two labelled declared-path sites here, plus the fixtures below, rather than none.

**Generation specs in suites that hold exactly ONE generation and never observe an identity keep the declared fallback** (`utils/applied.ts`, `browser/readWorkload.ts`, `workload.ts`'s base definition and every `indexerFor` caller that passes nothing). Migrating them would be churn with nothing observable behind it, and they are the live witness for criterion 3. The cost I am flagging: after `version` is deleted those fixtures have no identity at all, so the contract task's sweep includes them — `indexerForProcessor`/`runWorkload` now take an optional identity parameter precisely so that sweep is a one-line change per call site. **Touches:** `the-declared-version-and-the-drift-report-are-deleted`.

**The name is `processorIdentity`, reused verbatim from `GenerationSpec.processorIdentity` and `FoldParts.processorIdentity`.** Coherence check: `identity` alone would re-mean the glossary's *generation* (identity IS `{stream, processor}`), and `versionHash` would say DECLARED VERSION while holding a hash of bytes. No new concept is introduced by this batch; the browser threads an existing one. **Touches:** every caller of `createIndexerState`, `addGeneration` and the worker hosts — documented in the package README and in the changeset, because an accepted input the docs do not describe is a deployment believing something untrue.

**The test identity helper is DUPLICATED into `packages/browser/test/utils/` rather than shared, and deliberately kept OUT of `browser/workload.ts`.** `@etherfold/core` and `@etherfold/processor-entities` each already carry their own copy for the same reason (a test folder is not a published surface, and what is under test is that the hook takes WHATEVER it is handed). It hashes with `node:crypto`, which is fine for the Vitest half but would break the Playwright specs that load `workload.ts` inside a real browser — so the harness only ever PASSES an identity through and never derives one. **Touches:** nothing outside this package; a later batch wanting one shared helper would have three copies to fold together.

**`aTabHoldsItsGenerationsInSlots.test.ts` stopped PARSING generation identities.** It read the declared version off the front of one (`record.processor.split('-')[0]`) to name slots and to key IndexedDB databases, which is exactly what ADR-0086 says nothing may do and which a hash identity makes meaningless. It now asks a test-only reverse map (`markerOf`) for which bytes it hashed, and the fixture's `dropState` looks a connection up by the whole identity, as a real host does. Alternative: keep the parse and make the synthetic bytes start with a version-shaped prefix — rejected as teaching the suite a lie it would pass on.


wighawag:

Migrate batch 3 of 4, and the one with an extra boundary to respect: this batch must move the ORDINARY browser paths and leave the HMR arrival alone, because that one has no bytes and belongs to `a-module-handed-to-a-tab-is-identified-by-its-handler-sources`.

- [x] **No file outside `packages/browser` is edited.** Diff is `packages/browser/**` plus its own changeset and task move.
- [x] **Nothing is deleted.** `processorIdentity` is optional everywhere it appears; `version`, `getVersionHash()`, `getCodeFingerprint()`, `utils/fingerprint.ts` and the drift report all still exist and pass. Grepped for removals of the old form: none.
- [x] **Every `getVersionHash()` call site takes its identity from the arrival.** Two live invocations remain and both are the fallback funnel (`processorIdentity ?? built.getVersionHash()`), one per hosting path, matching the shape #163 established in core.
- [x] **The HMR arrival is deliberately untouched, and the reason is recorded where a reader will meet it.** `updateProcessor` gained no `processorIdentity`, and the comment at that call site explains why rather than leaving a silent gap: what arrives there is a MODULE OBJECT a dev server handed the tab, so there are no bytes for an app to hash, and letting an app state one anyway would be "the author-declared identity coming back through the one door left open". It then names the task that owns the handler-sources derivation. That is the correct boundary AND a correct piece of reasoning: an optional `processorIdentity` on `updateProcessor` would have looked like a harmless symmetry and would in fact have reopened exactly what ADR-0086 closes.
- [x] **Tests supply BYTES and none runs a bundler for an identity.** `test/utils/processorIdentity.ts` mirrors core's helper (synthetic bytes per marker, SHA-256, `sha256:<hex>`), using `node:crypto` here. The one `esbuild` import in this package's tests is the pre-existing `bundlesForABrowser.test.ts`, which asserts the package is browser-bundleable and has nothing to do with identity; this batch did not touch it.
- [x] **The batch is GREEN on its own,** verified by the gate on the whole tree at the rebased tip.
- [x] **A changeset accompanies it,** and it flags the HMR exclusion explicitly so the exclusion is a documented decision rather than an omission someone later "fixes".

**One catch in here is specific to this package and is a real correctness find.** The resolved identity is now also what BOTH hosting paths key their own per-generation STORE record on (`createIndexerState` and `serveIndexerHost` each keep one, for the scheduled prune and for the read a tab makes across the port). So the name the registry files a generation under and the name its store is looked up by cannot drift apart. A batch that had only threaded the value into `GenerationSpec` would have passed its criteria and left two keys that agree today and stop agreeing the moment an arrival supplies an identity. Good catch, and it is the sort of thing only someone reading the package rather than grepping it would find.

**It was honest about a backward-compatibility sharp edge instead of hiding it.** `createIndexer`'s factory option gains a fifth argument carrying the resolved identity, mirroring the container's `createGeneration`; an existing four-parameter factory still compiles and would silently DROP the identity, and the option's documentation says so. Stating that is better than either breaking the signature or letting it fail quietly.

**It also fixed its own doc example,** which is worth noting because it makes the equivalent miss on #164 a genuine miss rather than a policy: `createBrowserStateStore`'s bootstrap example now passes the identity the app's arrival handed the hook instead of calling `eventProcessor.getVersionHash()`. I am still carrying the `packages/processor-entities/README.md` `openAndBootstrap` example forward to the snapshot task's review.

Merging. One migrate batch left: `the-cli-server-and-examples`, after which the contract task can delete the declared path.

PR #166, opened 2026-09-17
feat(the-cli-server-and-examples-take-their-identity-from-the-arrival):…

wighawag:
Task: `work/tasks/done/the-cli-server-and-examples-take-their-identity-from-the-arrival.md`

All checks are green. Here is the report.

## What I did

One MIGRATE batch of ADR-0086, scoped to `packages/cli`, `packages/server`, `packages/state-store` and `examples/`. Every call site in these packages that *sources* a generation's fold half now takes the identity its ARRIVAL supplied; every place that *describes* one says so in the new vocabulary. Nothing was deleted: `version`, `getVersionHash()`, `getCodeFingerprint()` and the `PROCESSOR DRIFT` report all still exist and still work, and the declared route is still asserted end to end inside `packages/cli`.

**Blast radius, measured rather than trusted.** The task estimated ~12 `getVersionHash` sites. The literal count in scope was ~13, but only two of those were real call sites (two hand-rolled `EventProcessor` fakes in `packages/server/test`); the rest were prose. The real migration surface was wider and different: ~14 `entityProcessorVersionHash(...)` identity-sourcing sites in `packages/cli/test` and ~12 `EntityProcessor.version`-sourced identities in `packages/server/test`.

**New, one per package:** `packages/server/test/utils/processorIdentity.ts` (synthetic bytes, `node:crypto`, matching the three existing copies in core / processor-entities / processor-sqlite) and `packages/cli/test/utils/processorIdentity.ts` (which *imports* `processorArtifactIdentity` from `@etherfold/utils` rather than re-spelling it, because the CLI already depends on it and already reads its own deployments' identities through it). No test runs a bundler; the one suite that genuinely instantiates from a path writes self-contained bytes itself or uses the committed fixture.

**`packages/server`** — `StreamBuilder` / `GenerationRebuild` / `IndexerGeneration` / `ReceivedGenerationSpec` all now receive `processorIdentity`, and `VersionedStateEventProcessor` receives `identity`, at every construction site in the suite (13 test files plus `utils/feedHarness.ts` and `utils/stateMovedTransport.ts`). `foldAt('v1')`/`entityProcessorAt(version)` became `foldAt(marker)` with a constant declared `version` labelled as naming nothing; `feedHarness.deploy`'s `processorVersion` option became `processorMarker`, and `PROCESSOR_VERSION` became `PROCESSOR_IDENTITY`. Source prose: `registry.ts`'s `ReconfigureReport` docstring, `generations.ts`, `api/admin.ts`'s refusal text, the `_generations.processor` column comment in `schema/sql/db.sql`, and the README's "what the advertised `generation` is made of" paragraph.

**`packages/cli`** — six suites moved off `entityProcessorVersionHash` onto arrival identities threaded into the namespace, the fold and the registry record together. `theDeploymentSelectsItsPromotionPolicy.test.ts` now ships a real self-contained BUNDLE on disk, so its successor is one edited handler line. Two source docstrings that had gone stale under the blocking task were corrected (`folding.ts`'s "why the namespace is computable up front", `reconfigure.ts`'s "why `unchanged` is common").

**`packages/state-store`** — `StateSnapshot.processor` and `SnapshotProcessorMismatchError` re-documented as WHICH FOLD, as the producing deployment's arrival derived it; the `openSnapshotAware` example no longer calls `eventProcessor.getVersionHash()`. The candidate predicate, the envelope and the format number are untouched (that value is `a-snapshot-is-labelled-with-the-identity-it-was-computed-under`'s).

**`examples/`** — both examples still compile and still run their tests. Neither has an arrival that derives an identity yet (see Decisions), so both keep their declared `version` with a label saying what it is, what replaces it, and which task retires it.

**Gate:** `format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test` all green (server 342, cli 267, state-store 171, examples 2, whole tree clean). `pnpm changeset status --since=main` cannot see the changeset until it is committed (it reads `git diff`, not the working tree), which is the runner's step.

## Found outside this batch (not touched)

- `packages/state-store-sqlite/src/identifiers.ts:63` and `src/store.ts:105` still spell a generation as `{stream digest, processor version hash}`. Stale vocabulary, sibling package, no `getVersionHash()` call site.
- `CONTEXT.md:16` still lists `getVersionHash` as part of the `EventProcessor` contract — that is `the-build-command-and-its-pinning-rule-are-documented`'s criterion ("`CONTEXT.md` describes how a processor is identified in the new vocabulary"), so I left it.

## Decisions

**`theDeploymentSelectsItsPromotionPolicy.test.ts` is moved onto the BUNDLE arrival, which reverses a "do not inline it back" comment the blocking task left.** `a-deployment-runs-from-a-bundle-identified-by-its-hash` pushed that suite's fixtures onto the module route (a sibling `abi.js` import) so the EXPAND step would not silently change their arrival, and said so at the site. Its stated reason was that the cases "express a successor as a `version` bump". This batch is the step that changes that: the successor is now one edited handler line in a self-contained module, so the sibling import no longer earns its keep and the fixture reads as what a real deployment ships. Alternative considered: leave it on the module route and migrate nothing there — rejected, because the suite's subject is WHEN THE POINTER MOVES and nothing in it is about the declared identity, so leaving it would have left a whole CLI suite naming generations by a field this family deletes. What it touches: the comment the blocking task wrote is gone from that file; `anEndpointReconfiguresARunningRun.test.ts` keeps its identical comment, deliberately (below).

**`anEndpointReconfiguresARunningRun.test.ts` is RETAINED whole on the declared arrival, labelled, and is this package's declared-path witness alongside `aDeploymentRunsFromABundle.test.ts`.** Criteria 1 and 3 conflict here and I resolved it the way all three sibling batches did: a site that does not SOURCE an identity, but WITNESSES that the old form still works, stays. That file's subject is the `PROCESSOR DRIFT` report and the `unchanged` message that names `version` — machinery that exists ONLY on the declared route, because it is the compensation for an identity an author had to remember. Rewriting it onto a bundle would not migrate the coverage, it would delete it while leaving the code, which is the failure the task's own prompt warns about. The same endpoint on the bundle arrival is already asserted in `aDeploymentRunsFromABundle.test.ts`. Alternative considered: split the non-drift cases onto bundles — rejected as duplicating the bundle suite while stranding the drift cases in a file whose fixture no longer produces them. What it touches: `the-declared-version-and-the-drift-report-are-deleted` will find one labelled CLI suite (plus `aDeploymentRunsFromABundle.test.ts`'s two declared cases and `utils/chain.ts`'s `nftProcessor.version`) to retire, rather than none.

**Both examples stay on the declared path, labelled, and I did NOT give either an identity.** `examples/browser-reference` hands a tab a module OBJECT (`createIndexerState`), and `examples/event-processor-nfts` runs `etherfold build -p ./dist/cli.js`, a `tsc` output that imports its ABI. Neither arrival derives an identity today. For the tab, `a-module-handed-to-a-tab-is-identified-by-its-handler-sources` says in as many words "do not let the app supply the identity … that is the author-declared identity re-entering through the one door left open", so passing `processorIdentity` from the app — the mechanical move, since `createIndexerState` now accepts it — is forbidden rather than merely premature. For the CLI example, making it ship a bundle means adding a bundler and a build command to an example, which is `the-build-command-and-its-pinning-rule-are-documented`'s (it owns the documented command, and its own criteria reference the example). So both keep `version`, each with a label naming ADR-0086, the arrival it is on, why the old rule still applies there, and the task that retires it — because an example that teaches a rule being deleted without saying so is the "claim that stopped being true" this batch includes examples to prevent. What it touches: `a-module-handed-to-a-tab-is-identified-by-its-handler-sources` (which will migrate `browser-reference`) and `the-build-command-and-its-pinning-rule-are-documented` (which will migrate `event-processor-nfts`'s build). Alternative considered and rejected: a Vite `define`d build id as the browser example's identity — that is an author-declared identity wearing a build tool's hat, and it would set a user-visible pattern the HMR task then has to un-teach.

**The CLI suites that stand a deployment up through an injected `importModule` were left alone** (`run.test.ts`, `indexCommand.test.ts`, `oneShot.test.ts`, `commands.test.ts`, `fixedTableNamespace.test.ts`, `generationNamespaceBesideFixedTables.test.ts` and the rest). They inject the MODULE arrival, which legitimately supplies no identity, and none of them sources or asserts one — they derive namespaces from what the deployment registered. Migrating them would have meant writing a bundle per case and deleting the injected-importer affordance, for no observable change. Alternative considered: convert them all to bundles — rejected as churn that removes a test seam. What it touches: the contract task's sweep is unaffected (they name no identity), but `a-path-naming-an-unbundled-entry-point-is-refused` will have to decide whether an injected `importModule` is still an accepted arrival, and these suites are where that lands.

**`SnapshotProcessorMismatchError`'s message now says "computed by processor `x`" rather than "by processor version `x`".** A user-visible string, so I am flagging it rather than burying it. It is a one-word honesty fix, not a predicate or format change: the value is compared for equality and never parsed, and `@etherfold/processor-entities`'s equivalent message already reads exactly this way, so this removes a disagreement between two messages about the same rule. What it touches: `a-snapshot-is-labelled-with-the-identity-it-was-computed-under` owns the VALUE in that field and may reshape the wording further; I claimed none of that, and left the envelope, the format number and the candidate predicate untouched.

**The CLI's test identity helper IMPORTS `processorArtifactIdentity` where the other three packages re-spell it.** Core, `processor-entities` and `processor-sqlite` each recorded that they spell `sha256:<hex>` themselves because they cannot depend on `@etherfold/utils`. `packages/cli` can and does — `aDeploymentRunsFromABundle.test.ts` already reads its own deployments' identities through that function — so a second spelling here could disagree with the one the CLI actually runs. Recorded because it looks like an inconsistency with the three siblings and is the opposite: the reason those three forked is a dependency direction that does not apply here.


wighawag:

Migrate batch 4 of 4, the last one, so after this the contract task can delete the declared path.

- [x] **No file outside `packages/cli`, `packages/server`, `packages/state-store` and `examples/` is edited.** Checked; nothing else.
- [x] **Every `getVersionHash()` call site in the batch's areas takes its identity from the arrival.** Zero live `.getVersionHash()` invocations remain across all four; the two surviving mentions are JSDoc prose in `packages/cli/src/folding.ts` describing the declared form, which is correct during a migration.
- [x] **Nothing is deleted.** `version`, `getVersionHash()`, `getCodeFingerprint()`, `utils/fingerprint.ts` and the `PROCESSOR DRIFT` report all still exist and pass. The only `-` lines matching the old form are two prose references.
- [x] **Tests supply BYTES and none runs a bundler for an identity.** Same helper shape as the earlier batches, in both `packages/cli/test/utils/` and `packages/server/test/utils/`. Every `esbuild` occurrence is documentation or a comment. `sql2ts.test.ts`'s `execFileSync` is the pre-existing SQL-to-TS codegen check, which is relevant here only because `db.sql` changed, and the gate is green so the generated types are consistent.
- [x] **The batch is GREEN on its own,** verified by the gate over the whole tree at the rebased tip.
- [x] **A changeset accompanies it,** correctly all-`patch` since this batch moves callers and vocabulary rather than surfaces.

**Two files needed checking because they belong to later tasks, and both stopped at the right line.**

`packages/server/src/schema/sql/db.sql` is a schema file, which is not what you expect a migrate batch to touch. The change is COMMENT-ONLY: no DDL, no column type, no migration. The `_generations.processor` comment now describes the value as which fold, as the arrival derived it, stored and compared for equality with nothing parsing it. That is the right thing to say in the one place a future reader will look to find out what that column holds.

`packages/state-store/src/snapshot.ts` is the subject of `a-snapshot-is-labelled-with-the-identity-it-was-computed-under`, a later leaf, so pre-empting it here would have been a real problem. It did not: the `processor` field keeps its name and type, `ENTITY_SNAPSHOT_FORMAT` is still `1`, and the candidate PREDICATE is untouched — which is exactly what that task is instructed not to let anyone change. What moved is documentation plus the mismatch message's wording ("computed by processor `x`" rather than "by processor version `x`"), aligning it with what `@etherfold/processor-entities` already says. The VALUE the producer writes is still the old derivation, correctly left for the task that owns it.

**The best judgement in this PR is the one about what NOT to migrate, and it would have been easy to get wrong.** The batch deliberately left `anEndpointReconfiguresARunningRun.test.ts` on the DECLARED route, and says why: the `PROCESSOR DRIFT` report and the `unchanged` message that names `version` exist only on that route, because they are the compensation for an identity an author had to remember. Rewriting those suites onto a bundle would have looked like thorough migration and would in fact have DELETED their coverage rather than moved it — and the contract task then needs that coverage present in order to remove it deliberately. Likewise `aDeploymentRunsFromABundle.test.ts` keeps its positive assertion that an unbundled module still takes its identity from `getVersionHash()`. A batch that migrated everything it could reach would have quietly dismantled the declared path's only witnesses while every criterion still read as met.

**The CLI's promotion-policy suite now ships a real self-contained bundle on disk,** so its successor is one edited handler line rather than a bumped `version`. That is the change a developer actually makes, which makes the suite a better witness than it was.

**The example's docstring is honest rather than aspirational.** `examples/browser-reference/src/processor.ts` now says plainly that this app is on the declared path, that a bytes arrival deletes `version`, that a tab is the one arrival with no bytes and will derive from handler sources, and names the task that will do it. It then states that the app deliberately does NOT supply an identity of its own, because that would be the author-declared identity re-entering through the one door left open — the same reasoning the browser batch used at `updateProcessor`. Three batches independently reaching the same conclusion in the same words is the family holding together.

Merging. All four migrate batches are now landed, which unblocks `the-declared-version-and-the-drift-report-are-deleted` — the fan-in where the wide refactor is finally verified as a whole.

PR #167, opened 2026-09-18
feat(a-module-handed-to-a-tab-is-identified-by-its-handler-sources): A…

wighawag:
Task: `work/tasks/done/a-module-handed-to-a-tab-is-identified-by-its-handler-sources.md`

All gates green (`format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test` — 4,000+ tests across the tree), and the working tree holds only this task's changes.

## What I did

**The derivation** (`packages/browser/src/moduleIdentity.ts`, new): `moduleProcessorIdentity(processor)` is the one expression that names a fold which arrived as a MODULE. It asks the processor for its handler-source fingerprint and takes nothing from the caller. Its doc carries the role (ADR-0086's "kept for this one arrival in a different role"), the limits (survives reformatting and handler re-ordering; does not survive minification, a change of transpiler, or a change the source text does not carry), and why a dev server is the only runtime those limits are acceptable in.

**Wired at every point the browser obtains a processor and no arrival named it**: `createIndexerState`'s generation spec (so `init` and `addGeneration`), the worker driver's own spec (`src/host/serve.ts`, so a fold moved off the UI thread is named identically), and `updateProcessor`, which is the direct module handover. Where an app supplied a `processorIdentity` (the bytes arrival) nothing changed: it is still used verbatim.

**Tests** (`packages/browser/test/aModuleIsIdentifiedByItsHandlerSources.test.ts`, written before the code): the same handlers under two declared versions are ONE fold and an edited handler under one declared version is TWO; an edit registers a successor at the container and handing the same module again resolves to the generation already held; `updateProcessor` discards for an edit and answers `stateDiscarded: false` for a save that changed nothing; a supplied identity is never overridden, so the same code is named differently as a module than as bytes; and the worker host names a generation the same way.

**The reconfigure suite's axis one was rewritten**, because the trap it asserted no longer exists: an edited handler now takes effect with no `version` bump, a no-op save says so, and `{force: true}` is re-pointed at the derivation's real blind spot. `browser/cut.ts` + `browser/indexing.spec.ts` (the Playwright half) moved with it.

**Docs**: `packages/browser/README.md`, the browser guide, `CONTEXT.md`'s `generation` entry (which claimed unqualified that the fingerprint "stays out of the identity"), `examples/browser-reference` (its processor doc told authors their edit would never run), and doc-only notes in `@etherfold/core` so a reader meeting "advisory" does not conclude the seam is free to delete. A changeset covers `@etherfold/browser` (minor) and `@etherfold/core` (patch, documentation only).

I did **not** remove ADR-0086's `status: accepted, not yet implemented`: `the-declared-version-and-the-drift-report-are-deleted` is still outstanding, so this is not the last task of that family.

## Decisions

**The derivation is ASKED of `EventProcessor.getCodeFingerprint()`, not computed by calling `processorCodeFingerprint` on the handed object — so the contract task must keep that seam.** The task's re-order note states that what this needs is `processorCodeFingerprint` and "is NOT the seam method `getCodeFingerprint()` that the contract task removes". That premise is false in detail: what a tab holds is a fold built OVER A STORE (`EntityEventProcessor`), and the handlers a developer edits live on the author's object INSIDE it, so `processorCodeFingerprint(thatObject)` hashes the library's own methods — identical for every processor ever built that way, a constant no edit could move, which is exactly the silent lie the derivation exists to remove. `getCodeFingerprint()` is the seam each implementation already answers from the author's object. Alternatives considered: reaching into the wrapper's internals (the browser would have to know about `EntityEventProcessor`'s shape), or a new seam method (a second name for a value that already exists — refused by the coherence check, since `getCodeFingerprint` already MEANS "a derivation over this processor's handler sources"; only its ROLE changes). **Touches:** `the-declared-version-and-the-drift-report-are-deleted`, which may delete the DRIFT REPORT but must leave `EventProcessor.getCodeFingerprint()` answering, or the one arrival with no bytes has no name at all. I said so in the JSDoc on both the seam and `processorCodeFingerprint`, and in the changeset.

**The derivation runs wherever the browser builds a fold with no supplied identity, not only at `updateProcessor`.** Criterion 1 reads literally as the direct handover, but naming only that one would leave the incumbent (named at `init` by the declared hash) and the newcomer (named by the derivation) incomparable, so the FIRST save after every page load would look like a change and discard a warm fold even when nothing was edited — the precise outcome this task exists to make honest. It is also the only way a dev app can name itself once the declared fallback is deleted. Alternative: `updateProcessor` alone, leaving `init` declared — rejected as making "a no-op save says so" false exactly once per page load, and leaving the contract task with a browser that cannot name anything. **Touches:** `an-hmr-update-reconfigures-the-tab-it-is-running-in` (whichever call its API uses is now named the same way) and the contract task's sweep of browser fixtures.

**A module whose handlers have no readable source keeps the declared fallback rather than being REFUSED.** `processorCodeFingerprint` answers `undefined` for a processor that is all bound or proxied functions, and "cannot tell" is a real answer. Inventing a refusal here would be a new user-visible error on a path nobody has reported, and it would be the contract task's problem to answer for anyway once the fallback goes. Alternatives: hash a constant (the silent lie), or throw (a new refusal). **Touches:** `the-declared-version-and-the-drift-report-are-deleted`, which has to decide what a non-derivable module arrival does once `getVersionHash()` is gone.

**A generation spec built by the browser must reach the container WHOLE, and two call sites were changed to stop spreading it.** The identity of a module arrival is only knowable after the processor object exists, and `Indexer.add` resolves `spec.processorIdentity` immediately after `createProcessor` returns — so the spec fills its own field in there. A spread (`{source, ...specFor(...)}`) would copy the field while still `undefined` and silently fall back to the declared hash, so `source` is now a parameter of the builder in both `IndexerState.ts` and `host/serve.ts`, with the reason stated at all four places. Alternatives: a getter (same hazard under spread, and more magical), or teaching `@etherfold/core` to accept a function-valued identity (a second way to say a thing the ADR deliberately keeps as an opaque string, and it would reach the bytes arrivals). The host-path half is pinned by a test that fails without it. **Touches:** any future caller that builds a browser generation spec.

**The identity covers handler SOURCE TEXT and nothing else, and the gap is documented rather than closed.** Entity declarations, imported helpers, and behaviour decided by a value the handler captured are all invisible to it, so a schema edit in a dev tab does not move the identity; `{force: true}` is the stated escape hatch. Closing it would need a new seam exposing the declarations, and folding the author's `version` back in is the thing ADR-0086 deletes. Stated in `moduleIdentity.ts`, the README, the guide, the example and the test that drives it (`processorVariant` versus `editedProcessorVariant` are that pair on purpose). **Touches:** nothing today; a later task that wants declarations in the name would add the seam, not reuse `version`.

**`@etherfold/core` and `examples/browser-reference` are edited, documentation only.** No behaviour outside the browser arrival changed: the bytes arrivals still hash bytes and core still only compares what it is handed. But `types.ts` said the fingerprint is "advisory ... never discards state because of it", `fingerprint.ts` said folding it into an identity is "unusable in practice", `CONTEXT.md` said it "deliberately stays out of the identity", and the reference app told authors their edit would never run — four statements this task makes false in one place each, and a muddled term is cheaper to fix now than after the next artifact inherits it. **Touches:** the contract task, which rewrites the server half of that `CONTEXT.md` entry; I added a scoped clause rather than rewriting the entry, so the two changes do not collide.


wighawag:

The one arrival with no bytes, given an identity of its own. This is also the task whose `blockedBy` edge I inverted to break the family's deadlock, so I paid particular attention to whether running it BEFORE the contract task caused any problem. It did not, and the build makes the case better than the re-order note did.

- [x] **A module object is identified by a derivation over its handler sources, with nothing declared and nothing supplied by the app.** `moduleProcessorIdentity` derives it inside the library and passes it down; `updateProcessor`'s public signature still takes `{force}` and nothing else. The changeset is explicit about why: an identity an application could state is the author-declared identity re-entering through the one door left open, and it would be silent whenever it was wrong. That is the third batch in a row to reach that conclusion independently.
- [x] **A handler EDIT produces a different identity, and the same module twice produces the same one,** asserted AT THE CONTAINER through `updateProcessor` rather than through an HMR API — which is what the criterion demanded, so that `an-hmr-update-reconfigures-the-tab-it-is-running-in` can build the three-outcome surface on top without the dependency being inverted.
- [x] **The limits are stated where a reader will meet them** — `src/moduleIdentity.ts`, the package README and the browser guide — and they are stated honestly rather than minimised: it survives reformatting and handler re-ordering, does NOT survive minification or a change of transpiler, and does not move for a change the source text does not carry (an edited helper the handler imports, a changed entity declaration, behaviour decided by a captured value). That last category is the one a developer will actually hit, and naming `{force: true}` as its answer is the right disposition.
- [x] **The same code as a MODULE and as a BUNDLE has different identities, documented as correct rather than papered over,** with a test asserting it directly ("never overrides a supplied identity, so the same code is named differently as bytes").
- [x] **Nothing reaches a non-browser runtime.** The bytes arrivals still hash bytes; `packages/core` is touched DOCUMENTATION-ONLY (I diffed it: no behaviour change in either file), and `cli`, `server`, `utils` and the processor packages are untouched.
- [x] **Tests and a changeset.** One extra test earns its place: a generation built through a worker host is named the same way as one built on the UI thread, so moving the fold off the main thread renames nothing. That is a real trap and nothing asked for it.
- [x] **Nothing is deleted.** `getVersionHash()` and the declared `version` still exist; what the browser stops doing is FALLING BACK to them, which is precisely what this task had to achieve for the contract step to become possible.

**The trap it removed is worth naming.** Before this, an edit under an unchanged `version` silently never executed in a tab. That was the family's motivating failure, in the runtime where it is most likely to happen, and it is now gone rather than documented.

**The core documentation edits are the right kind of forward-note.** `EventProcessor.getCodeFingerprint` and `processorCodeFingerprint` now say they have a second role, specifically so that a reader who meets the word "advisory" does not conclude the seam is free to delete. That is a note planted for the next builder, in the file that builder will open, which is exactly where it belongs.

**Two things I am carrying into the contract task, both surfaced by this PR rather than missed by it.**

First, and it contradicts a criterion I wrote yesterday: `moduleProcessorIdentity` calls the SEAM METHOD `processor.getCodeFingerprint()`, not the standalone `processorCodeFingerprint`. So `getCodeFingerprint()` must SURVIVE on the seam, where ADR-0086's consequence list and the contract task's criterion 1 both say it goes. I checked whether the builder should have used the standalone function instead so the seam could die, and it should not: the seam method is what delegates correctly through a wrapper (`VersionedStateEventProcessor` over `EntityEventProcessor`), while the standalone function applied to a wrapper would fingerprint the wrapper's own methods rather than the author's handlers — which is the exact hazard `types.ts` already documents. The build made the right call and said so in the changeset; the task and the ADR are what need correcting.

Second, the honest loose end the changeset flags rather than hides: a processor whose handlers have no readable source (all bound, or behind a proxy) answers `undefined`, and that fold currently keeps the declared fallback. Once the contract task removes it, that case has no identity at all. Correctly not decided here — it is the contract task's to settle.

I am fixing both in the contract task's body before dispatching it, so neither is rediscovered at build time the way the first false premise was.

Merging. This is the first of the two inverted edges paying off: the browser's module arrival now has a name of its own, which is one of the two things the contract task was missing.

PR #168, opened 2026-09-18
feat(no-suite-or-example-still-rests-on-the-declared-identity): The CLI…

wighawag:
Task: `work/tasks/done/no-suite-or-example-still-rests-on-the-declared-identity.md`

Done. Everything is green and the working tree holds only this task's changes.

## What I did

**Grounded the blast radius by measurement, not by the list.** I made both halves of the declared fallback throw in the built `dist/` (a scratch edit, reverted by rebuilding) and ran the suites. That found **12 files in `packages/cli`**, not the six named in passing: `run`, `indexCommand`, `oneShot`, `fixedTableNamespace`, `entityStore`, `equivalence`, `fetch` (via `test/utils/receiver.ts`), `fromTheEnvironment`, `runServesTheIndexerItFolds`, `scheduledPrune`, plus the two witness files. It also cleared two the task body implied were dirty (`commands.test.ts`, `generationNamespaceBesideFixedTables.test.ts` never take the fallback) and showed `packages/utils` was already clean, 62/62.

**Shape chosen, per suite family: keep the injection seam, name the arrival.** `packages/cli/src/index.ts` and `indexCommand.ts` gained `deps.processorIdentity`, the other half of the seam `importModule` already was. All ten command-driven suites keep their subject (namespaces, commands, one-shot, scheduling, the wire) and gain one line. `fixedTableNamespace` and `receiver.ts`/`entityStore` build their fold by hand, so they state it on the spec directly.

**`FoldParts.generation` now states its `processorIdentity`** instead of leaving `processorIdentityOf` to ask the fold. Same string, but the container no longer reaches for `getVersionHash()` — which is what actually goes dark.

**The example ships a bundle.** `pnpm build:bundle` runs `esbuild --bundle --format=esm --minify` (the fixtures' command), `build:db` and the suite point at `dist/cli.bundle.js` with no injected importer, and `build` produces it so the root README's sequence stays true.

**The three witnesses are untouched**, and a new test file, `anInjectedArrivalSuppliesItsIdentity.test.ts`, guards the seam (written red first) because its failure mode is silent regression to the declared identity.

**Demonstration (criterion 3), with both halves of the declared fallback throwing:** `packages/cli` 260/270, the only failures being the 7 in `anEndpointReconfiguresARunningRun.test.ts` and the 3 declared cases in `aDeploymentRunsFromABundle.test.ts`; `packages/utils` 62/62; the example 2/2. Under the stricter probe alone (`processorIdentityOf` refusing to fall back, the exact shape of the contract task's deletion) `packages/cli` is **28/28 green including the witnesses**.

Full tree: build, typecheck, `pnpm test`, `format:check`, `check:adr`, `check:refs`, `check:changesets` all green. Changeset added (`etherfold: patch`).

**Observation filed** (out of scope, not fixed): `work/notes/observations/four-more-packages-still-rest-on-the-declared-identity-fallback.md` — the same probe across the whole tree shows ~46 cases in `processor-sqlite`, `conformance-workload-stratagems`, `browser` and `platforms/nodejs-fetcher` still resting on the fallback, with no owner.

## Decisions

**`deps.processorIdentity` on `IndexingDependencies` / `IndexDependencies`, rather than converting the suites to bundle fixtures.** Both types are documented "what a test may substitute for the real world", and `importModule` already states WHAT comes back for a `--processor` path; this states what that thing is CALLED, which is what an arrival does. Alternatives considered: (a) point every suite at the committed `nfts.bundle.js` — rejected, its processor differs from `utils/chain.ts`'s (no burn branch), several suites need bespoke modules (reserved entity names, per-chain contract data, an `imported` spy), and a data-URL-evaluated module destroys the object identity three suites assert through, so it would rewrite subjects the prompt says to keep; (b) thread `readBundle` through — rejected, it takes the artifact route and evaluates the bytes, so it is (a) in disguise and cannot keep an injected module. **Precedence is `arrival.identity ?? deps.processorIdentity`**, so real bytes always win and the two can never need a refusal (criterion 5). **What it touches:** `a-path-naming-an-unbundled-entry-point-is-refused` still owns whether an injected importer is an accepted arrival — this neither answers nor forecloses that, and deletes no affordance. **Coherence:** `processorIdentity` is the name already used at this layer (`openFolding`'s context, `ReceivedGenerationSpec`, `FoldParts`) for exactly this thing; no new concept, no re-meaning.

**`FoldParts.generation` now carries `processorIdentity`, which is a source change beyond making the suites green.** Under a probe that deletes the `processorIdentityOf` fallback outright, *every* CLI deployment failed — including the bundle ones — because `foldPartsFor` returned only `createState`/`createProcessor`, so the container asked `processor.getVersionHash()`. That is a genuine "goes dark" site inside my packages, and the task's goal is that nothing does. The value is unchanged (it is the string the table namespace was already named from); only the container stops asking. Alternative: leave it to the contract task — rejected, it is inside this task's packages, it is value-preserving, and leaving it means the contract task discovers the CLI is not actually migrated. **What it touches:** `the-declared-version-and-the-drift-report-are-deleted` finds `packages/cli` fully green when it removes the fallback arm.

**The example borrows the workspace's esbuild binary instead of declaring a devDependency.** `examples/event-processor-nfts/package.json` runs `../../packages/utils/node_modules/.bin/esbuild`. Adding `esbuild` to the example would edit `pnpm-lock.yaml`, which is outside the three paths criterion 6 allows, and CI installs `--frozen-lockfile` so package.json and lock must move together. `packages/cli/test/fixtures/processor-bundle/README.md` already prescribes exactly this borrowing for this repo ("reach for the workspace's copy rather than adding a dependency a deployment would then carry"). The README shows a reader the plain `esbuild …` command instead. **Alternative:** add the dependency and accept the lockfile edit — rejected as a literal criterion-6 violation for a cosmetic gain. **What it touches:** `the-build-command-and-its-pinning-rule-are-documented` may prefer to declare the dependency once it is free to touch the lockfile; the command string it must match is unchanged.

**The example's bundle is built by `pretest`, and `build` now means `tsc && build:bundle`.** The suite drives the real bytes, so they must exist whether it is run alone or by the root `pnpm test`, which does not build examples. Folding the bundle into `build` also keeps the root `README.md`'s `build` → `build:db` sequence true without editing a file outside my scope. **Alternative:** build the bundle inside the test — rejected, it puts a bundler in the test path, which this repo's fixture README argues against.

**No `docs/spikes/` artifact for the probe.** Criterion 3 asks for a scratch edit explicitly *not* committed, and criterion 6 bars new files outside the three paths; the probe recipe and its numbers are recorded here and in the observation note instead, which is where they stay readable for the contract task.


wighawag:

I wrote this task yesterday to break the family's deadlock, so I reviewed it twice over: did the build meet the criteria, and did the task I wrote actually aim at the right thing. Both yes — and the build found something the task did not know to ask for.

- [x] **No suite in `packages/cli` or `packages/utils` falls through to `getVersionHash()`, except the three named witnesses.** The seam added is `processorIdentity` on `IndexingDependencies` / `IndexDependencies`, described exactly right: it is the other half of the seam `importModule` already was — one states WHAT comes back for a `--processor` path, the other states what that thing is CALLED. Two guards make it safe rather than a back door: BYTES ON DISK ALWAYS WIN (where the path names a real bundle its hash is the identity and this is not consulted, so it can supply a derivation where the injected arrival made none but can never overrule one), and no flag or environment variable reaches it, so it cannot become a way for a user to declare an identity.
- [x] **The example runs from a self-contained bundle,** built with the command this family already documents, and its `version` now names nothing. `pretest` builds the bundle so the suite drives the real arrival against real bytes rather than a stale artifact.
- [x] **Demonstrated rather than asserted, which was the criterion I cared most about.** With both halves of the declared fallback made to throw, `packages/cli` fails exactly the three witnesses and nothing else, and `packages/utils` and the example pass whole. That is the only check that can actually see a run-time fallback, and it was run.
- [x] **Nothing deleted, nothing refused.** `version`, `getVersionHash()`, `getCodeFingerprint()` and the drift report all still exist; an unbundled configuration still resolves; and the injected-importer affordance is untouched, with the changeset explicitly noting that whether it survives is `a-path-naming-an-unbundled-entry-point-is-refused`'s question and removing it here would answer it silently and out of order. That restraint is the right instinct.
- [x] **The three declared-path witnesses are untouched** — verified by path, not by claim.
- [x] **No file outside `packages/cli`, `packages/utils` and `examples/event-processor-nfts`.** Checked; clean.
- [x] **Green on its own, and a changeset.**

**`FoldParts.generation` now STATES its `processorIdentity`** instead of the container asking the fold for it. The value is unchanged, but the container no longer reaches for `EventProcessor.getVersionHash()` to learn it — which is, as the changeset puts it, a question a processor built from bytes cannot answer. That is the deeper fix and it is the one that makes the contract step possible in the CLI at all.

**The build did something the task did not ask for, and it is the most valuable thing in this PR.** Having built the probe (patch both halves of the fallback to throw, in the BUILT output so only CONSUMERS of a package are affected, then run each package's suite), it ran it across the WHOLE tree rather than only its own three paths — and found four more packages resting on the fallback: `processor-sqlite` (16), `conformance-workload-stratagems` (14), `browser` (10) and `platforms/nodejs-fetcher` (6). Roughly 46 cases that would go dark, owned by nobody. It filed `work/notes/observations/four-more-packages-still-rest-on-the-declared-identity-fallback.md`, was careful to say these are all CORRECT today, and was honest about what it did not do ("I did not read the failures to separate genuine witnesses from genuine remainders, because that is triage on somebody else's packages"). Staying inside its scope while making the next person's job possible is exactly right.

That finding means my fifth batch was necessary but not sufficient, and I am acting on it rather than letting the contract task discover it the hard way for a third time: a sixth and final batch follows, scoped to those four packages, reusing this probe. The probe covered every package, so the enumeration is now complete — there is no seventh.

Two nits, neither blocking:

- **A novel convention in a file readers copy.** The `"---- BUILD: … ----": ""` dummy keys in `examples/event-processor-nfts/package.json` are comments-in-JSON, and this repo used the pattern nowhere before this PR (I checked). The prose is genuinely useful and JSON has no comments, so I am not blocking, but an example is the worst place to introduce a house style of one, since copying it is the point.
- **`../../packages/utils/node_modules/.bin/esbuild`** reaches across packages into a pnpm layout. The reasoning is sound and stated (no devDependency for a reader to inherit, nothing here bundles at run time) and the comment tells a copier to run `esbuild` from wherever they have it, but the path itself only works inside this workspace.

Both are for the docs task to consider when it settles how the build command is presented; neither affects identity.

Merging.

PR #169, opened 2026-09-18
feat(the-last-four-packages-stop-resting-on-the-declared-fallback): The…

wighawag:
Task: `work/tasks/done/the-last-four-packages-stop-resting-on-the-declared-fallback.md`

Everything is green and the working tree holds only the intended changes.

## What I did

**Grounded the probe first.** Patched both halves of the declared fallback in the BUILT output (`packages/core/dist/internal/processorIdentity.js` throws when `supplied === undefined`; `packages/processor-entities/dist/EntityEventProcessor.js`'s `entityProcessorVersionHash` throws) and ran every package. The four-package table reproduced **exactly** — `processor-sqlite` 16, `conformance-workload-stratagems` 14, `browser` 10, `platforms/nodejs-fetcher` 6 — and turned up **one package the earlier measurement missed**: `packages/fetcher-host`, 2 cases. `dist/` was rebuilt afterwards; `grep PROBE` on both files is 0 and nothing under `dist/` is tracked.

**Migrated the remainders** (all four packages are now off the fallback):

| package | what moved |
| --- | --- |
| `processor-sqlite` | `deployment-shapes.test.ts`: all three `IndexerGeneration`s now take `{processorIdentity: PROCESSOR_IDENTITY}` as the **fifth** argument (it is the `options`, not the `config` — passing it in the config object silently does nothing, which is how this site hid) |
| `conformance-workload-stratagems` | new `test/utils/processorIdentity.ts` + `processorIdentity` on the generation spec in all three suites (publication, retraction, receiving container). Nothing they assert was reshaped |
| `browser` | `snapshotOnlyMode` / `streamSeeding` label their snapshot with an `identityOf(...)` constant instead of `entityProcessorVersionHash(definition)`, copying `entityBootstrap.test.ts`'s `APP_IDENTITY`; `liveReload`'s fake now reports a `getCodeFingerprint()`, which **is** the module arrival's derivation |
| `platforms/nodejs-fetcher` | `test/deployment.ts` hands one identity to BOTH halves: the fold's `identity` and `StreamBuilderOptions.processorIdentity` |

## Witness inventory (the batch's most valuable output)

Five labelled witnesses survive, in two packages. Each carries a comment naming ADR-0086, stating it exists to prove the declared path still works, and naming `the-declared-version-and-the-drift-report-are-deleted` as what retires it.

**`packages/processor-sqlite/test/version.test.ts`** — whole-file label, 11 cases:
- `getVersionHash >` `cannot contain the 'unknown' fallback…` / `has no 'unknown' branch left even with the constructor guard defeated` / `carries no fallback constant at all, configured or not` / `still changes when the entity SCHEMA changes at a fixed version` / `gives an unconfigured processor and one configured with undefined the SAME hash`
- `the code fingerprint > changes when a handler changes at an unchanged version`
- `drift, end to end >` `reports when a handler changed but the version did not` / `does not report when the version WAS bumped` / `does not report for a byte-identical processor across a restart` / `does not report against a cursor written before fingerprints existed` / `refuses to start under strict mode, and starts without it` — **this group reaches the CORE half** (`processorIdentityOf` via a real `IndexerGeneration`), not `EntityEventProcessor`

**`packages/processor-sqlite/test/lifecycle.test.ts`** — two `describe` labels, 5 cases:
- `the identity this fold answers with >` `is the one the ARRIVAL handed it, where a host had one` / `is not moved by a later 'configure'…` — kept because each establishes the arrival identity by CONTRASTING it with a declared one
- `getVersionHash, the DECLARED fallback >` `changes when the processor version changes` / `changes when the entity SCHEMA changes, even at the same version` / `is stable across instances of the same processor`

**`packages/browser/test/aModuleIsIdentifiedByItsHandlerSources.test.ts`** — 2 cases:
- `names the generation by its handler sources, with nothing declared and nothing supplied` (its `declared()` helper)
- `falls back to the declared hash for a module whose handlers have no readable source` — **NEW**, see Decisions

`conformance-workload-stratagems` and `platforms/nodejs-fetcher` leave **no** witness: nothing in either is about identity.

With the probe applied, those five are exactly and only what goes red in the four packages. `packages/cli`'s 10 (its three batch-5 witnesses) are unchanged and out of scope.

## The fifth package, left alone

`packages/fetcher-host/test/harness.ts:199-200` is the same un-migrated shape as `platforms/nodejs-fetcher`'s — 2 cases in `classification.test.ts`. Per the prompt I did not expand into it. Recorded in `work/notes/observations/fetcher-host-is-a-fifth-package-resting-on-the-declared-identity-fallback.md`, including the fact that it hides badly: the throw is swallowed and surfaces as `expected undefined to match object {…}`, so a probe run skimmed for `PROBE:` reads it as green.

`pnpm format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test` are all green. `changeset status --since=main` still reports "no changesets were found" only because `@changesets/git` reads changed files with `git diff` and the new `.changeset/*.md` is untracked until the runner commits.

## Decisions

**Added a NEW labelled witness in `packages/browser` rather than leaving `liveReload.test.ts` as one.** `liveReload`'s four failures came from a fake whose `getCodeFingerprint()` returned `undefined`, so `moduleProcessorIdentity` declined and the core fell back. `test/utils/fakeGeneration.ts` called that omission deliberate — but its own words were "Until [`a-module-handed-to-a-tab-is-identified-by-its-handler-sources`] lands, those cases keep the author-declared fallback", and that task HAS landed, so the deferral had expired and the cases are remainders (their subjects are reconfigure serialization and `{force}` passthrough, not identity). Migrating them, however, would have deleted the only live exercise of a **production** dependency: `src/moduleIdentity.ts` documents that `undefined` is a real answer and that the caller then falls back, calling itself "the last place that fallback is reachable in this package, and one `the-declared-version-and-the-drift-report-are-deleted` has to answer for". So I migrated the four and wrote one focused case that pins that branch (an author object with every handler `bind`-ed, so `processorCodeFingerprint` sees only `[native code]`). **Alternatives:** (a) leave all four as witnesses — rejected, four subject-mismatched cases give the contract task a muddle instead of a decision, and the stale doc would have stayed stale; (b) migrate and add nothing — rejected, the contract task would then delete a fallback production still takes with no test going red, which is precisely the failure this family keeps repeating. **What it touches:** `the-declared-version-and-the-drift-report-are-deleted` inherits one explicit decision ("what is an unnameable module called once there is no declared hash?") instead of a silent branch. **Coherence:** no new concept; it reuses `moduleProcessorIdentity` and the existing witness vocabulary.

**`liveReload.test.ts`'s fake now names itself by a marker-derived `getCodeFingerprint()` instead of by a moving `getVersionHash()` string.** That IS the module arrival's derivation (`moduleProcessorIdentity` is `getCodeFingerprint()` and nothing else), so the fake got more realistic rather than less, and the swap semantics are identical: same marker = same fold = declined, different marker = swapped. Two consequences are visible in the file: the recording harness's `updateProcessor` wrapper now **forwards its options** (it dropped them, so the identity the hook derived never reached the core — a genuine fidelity bug the probe exposed, and the reason one case failed even after `init` was fixed), and the `#5` assertion is now `toEqual({force: true, processorIdentity: makeProcessor().getCodeFingerprint()})`. **Alternative:** supply `processorIdentity` through the spec — rejected, `IndexerState.updateProcessor` deliberately accepts no identity from an app, so it would have fixed `init` and not the swap. **What it touches:** nothing outside the suite; no production code changed.

**The two `lifecycle.test.ts > the identity this fold answers with` cases are witnesses, not remainders.** Their subject is the arrival's identity, but each proves it by contrast with a fold built with no identity (`expect(handed…).not.toBe(declared.getVersionHash())`, and `declared.configure(…)` MOVES the declared hash). Migrating would mean deleting the contrast, i.e. deleting coverage. Per the prompt's tie-break I labelled rather than migrated. **What it touches:** the contract task deletes the `declared` half of each case and keeps the arrival assertion above it, which stands alone.

**`@etherfold/conformance-workload-stratagems` is deliberately absent from the changeset** even though it was edited: it is `private: true` and `.changeset/config.json` sets `privatePackages: false`, so naming it is the exact mixed-changeset refusal `scripts/check-changesets.mjs` exists to catch. Its share of the change is described in the changeset body instead.

**No `docs/spikes/` artifact.** The probe is a scratch `dist/` edit the task explicitly says not to commit, and its recipe and numbers are already recorded in the two observation notes and in the changeset, which is where the contract task will look.


wighawag:

The sixth and final migrate batch, and the one whose real product is an inventory rather than a migration. I wrote this task after #168's probe found these four packages, so again I checked both the build and the task.

- [x] **Scope held exactly.** Nothing outside `processor-sqlite`, `conformance-workload-stratagems`, `browser` and `platforms/nodejs-fetcher`. No fifth package was quietly absorbed.
- [x] **Nothing deleted, nothing refused.** `version`, `getVersionHash()`, `getCodeFingerprint()` and the drift report all still exist; no configuration that resolves today stops resolving. Verified by diff.
- [x] **Triage, not bulk migration** — which was the way this task could most easily have gone wrong. Every disputed case has a recorded decision with the alternative considered and rejected. The `lifecycle.test.ts` pair is the clearest: their subject IS the arrival's identity, but each proves it BY CONTRAST with a fold built without one, so migrating them would have deleted the contrast rather than moved it. Labelled, per the tie-break, with a note that the contract task deletes the `declared` half and keeps the arrival assertion, which stands alone.
- [x] **Every surviving witness is labelled in place** (14 labels added, each naming ADR-0086 and the contract task).
- [x] **The inventory exists, by file and case** — five witnesses: the whole of `processor-sqlite`'s `version.test.ts` (11 cases), the two `describe`s in its `lifecycle.test.ts` that consult the declared hash (5 cases), `browser`'s declared-hash contrast in `aModuleIsIdentifiedByItsHandlerSources.test.ts`, and the new production-fallback case. The contract task now starts from a list instead of a search, which is the whole point of this batch.
- [x] **Demonstrated, not asserted:** with both halves of the fallback made to throw, these four packages fail ONLY those labelled witnesses, and the tree is green without the probe.
- [x] **Green, with a changeset** correctly scoped to the three publishable packages.

**The best decision in here closed the loose end I had just written into the contract task.** `liveReload`'s four failures came from a fake whose `getCodeFingerprint()` returned `undefined`, and `fakeGeneration.ts` called that omission deliberate — but its own words deferred to `a-module-handed-to-a-tab-is-identified-by-its-handler-sources`, which has now landed, so the deferral had EXPIRED and those cases were remainders (their subjects are reconfigure serialization and `{force}` passthrough, not identity). Rather than either leaving four subject-mismatched cases as a muddle or migrating them and silently dropping the branch, it migrated the four AND wrote one focused case pinning the `undefined` path — an author object with every handler `bind`-ed, so the derivation sees only `[native code]`. That is precisely the case I added to the contract task's criteria yesterday, and it now has a test that will go red if the contract task deletes the fallback without deciding what an unnameable module is called. The reasoning it gives for rejecting "migrate and add nothing" is the family's own lesson stated back: the contract task would otherwise delete a fallback production still takes, with no test going red.

**It also found a real bug while it was in there.** The `liveReload` recording harness's wrapper around `updateProcessor` DROPPED its options, so the identity the hook derived never reached the core — a fidelity bug that made the suite lie about what the production path does, exposed by the probe and fixed with the migration rather than filed and left.

**A small correctness catch on the repo's own rules:** `conformance-workload-stratagems` is edited but deliberately absent from the changeset, because it is `private: true` and `privatePackages: false`, so naming it is the exact mixed-changeset condition `scripts/check-changesets.mjs` exists to refuse. Its share is described in the body instead. Knowing that rule without being told is good.

Nits: none. Gate 2 is off in this repo, so there is no `review-nits-*` note.

Merging. The migrate phase is now genuinely complete and, for the first time in this family, that claim rests on a measurement rather than a grep. The contract task is unblocked with an inventory in hand.

PR #170, opened 2026-09-18
feat(the-declared-version-and-the-drift-report-are-deleted): The declar…

wighawag:
Task: `work/tasks/done/the-declared-version-and-the-drift-report-are-deleted.md`

The refactor's fan-in is done and the whole tree is green.

## What I did

**The declared identity is gone from the tree.** `EntityProcessor.version`, `EventProcessor.getVersionHash()`, `assertProcessorVersion` (`core/src/utils/version.ts`, deleted), `entityProcessorVersionHash` and the `identity` option that existed only to answer `getVersionHash()` on `EntityEventProcessor` / `VersionedStateEventProcessor` are all removed. `processorIdentity` is now **required** on `IndexerGenerationOptions`, `StreamBuilderOptions`, `GenerationRebuildOptions` and `updateProcessor`; `processorIdentityOf` (the fallback expression) is replaced by `requireProcessorIdentity`, which REFUSES rather than back-fills.

**The drift report is gone with it, everywhere it reached**: `core/src/processorDrift.ts` and its test (deleted), `ProcessorDriftReport` / `ProcessorDriftComparison` / `processorDriftReport` export / `ContextIdentifier.processorFingerprint` / `strictProcessorDrift` / `Indexer.onProcessorDrift` / `ReceivingIndexer.onProcessorDrift` / `StreamBuilder` + `GenerationRebuild` reporting / `ReconfigureReport.unchanged.drift` / the admin route's response body / the CLI reconfigure endpoint's two-armed `unchanged` message / `CONTEXT.md`'s **generation**, **EventProcessor** and **EntityEventProcessor** entries / `README.md` / the browser guide. The phrase `PROCESSOR DRIFT` appears in no source, test, doc or `CONTEXT.md` — only in ADR-0086 and in already-written changesets, which are history.

**The derivation the HMR arrival needs survives**, in its new role: `utils/fingerprint.ts` and `EventProcessor.getCodeFingerprint()` both stay, with their docs rewritten to record why (and naming `a-module-handed-to-a-tab-is-identified-by-its-handler-sources`, which is done, #167, and calls the seam method). `processorFingerprint.test.ts` and its fixtures stay untouched.

**Test churn**: every suite that rested on the run-time fallback now supplies an arrival identity (the inventory the sixth batch labelled was the starting list, and the typechecker found the rest). `processor-sqlite/test/version.test.ts` is deleted and replaced by `test/codeFingerprint.test.ts`, which keeps the one live claim inside it — the wrapper delegates the fingerprint to the AUTHOR's object, which is what the browser arrival depends on. `anEndpointReconfiguresARunningRun.test.ts` is migrated from an unbundled entry point onto a real self-contained bundle (its drift cases replaced by "`unchanged` means what it says").

## Decisions

**An unnameable MODULE arrival is REFUSED, in `@etherfold/browser`.** `moduleProcessorIdentity` now returns `string` and throws when `getCodeFingerprint()` declines (all handlers `bind`-ed or proxied). *Why:* the other three options are all lies — hashing `[native code]` into a constant no edit can move, taking a name from the application (the author-declared identity through the one door ADR-0086 leaves open), or letting `undefined` reach `GenerationId.processor`. The refusal is loud, lands before anything is registered or any state claimed, and names the two ways out. *Alternatives:* leave it `undefined` and let core's structural guard fire (rejected: core cannot say WHY, and the message is what an app author needs); accept `processorIdentity` on `updateProcessor` (rejected explicitly by ADR-0086 and by the task that built the arrival). *Touches:* `@etherfold/browser`'s `updateProcessor` and `createIndexerState`/`serveIndexerHost`; the derivation is computed before the auto-index loop is paused, so a tab meeting the refusal is left exactly as it was. Pinned by `aModuleIsIdentifiedByItsHandlerSources.test.ts`.

**A `--processor` path naming an UNBUNDLED entry point is REFUSED, minimally, at arrival resolution (`requireArrivalIdentity`, `etherfold`).** *Why:* deleting the declared identity leaves such a deployment with no name for its fold, and this task may not leave one running unnameable. *Touches `a-path-naming-an-unbundled-entry-point-is-refused` directly:* that task is still correct and still needed — it moves this refusal to CONFIGURATION RESOLUTION in ADR-0048's shape and gives it the build command an author needs. I deliberately did NOT name a build command (the wording is `the-build-command-and-its-pinning-rule-are-documented`'s), and I did not put the refusal in `openProcessorArrival` (`@etherfold/utils`), because that would also refuse the injected-arrival test seam and would put a configuration decision inside a loader. *Alternative rejected:* deriving a CLI identity from handler sources — that is the production identity `fingerprint.ts` argues against, and it would give the same code two names depending on how it was loaded.

**`GenerationSpec.processorIdentity` stays typed optional but is REFUSED when absent.** *Why:* `add` reads state → processor → THEN identity, which is exactly what makes a module arrival expressible (the browser fills the field in from inside `createProcessor`). A required field would be unstatable there. The refusal (`requireProcessorIdentity`, `@etherfold/core`) is the structural backstop under both arrival-level refusals above.

**The `identity` option on `EntityEventProcessor` / `VersionedStateEventProcessor` is DELETED, not kept.** It existed solely to answer `getVersionHash()`. A fold that computes no identity of its own has nothing for a caller to override, so the "two live answers" hazard it closed no longer exists. *Touches:* every host now passes ONE value to the table namespace (ADR-0053), the registry and the engine, and none to the processor.

**DEVIATION FROM ADR-0086's consequence list, recorded as the task asked.** The ADR says `getCodeFingerprint()` and `utils/fingerprint.ts` go. They do not: the browser's module arrival names its fold by the SEAM method (not the standalone function, because the seam delegates correctly through a wrapper), so deleting it would leave the one arrival with no bytes with nothing to be. `fingerprint.ts` now records this the way it already recorded its deviation from ADR-0008, and the changeset states it.

**The pending changeset `a-reload-that-changed-nothing-reports-processor-drift.md` is DELETED, not amended.** *Why:* nothing has been released, so it would otherwise ship a release note announcing a report that never existed outside this repository. A changeset is a release note, not a history log; the history lives in the work record and in ADR-0086. My changeset states the deletion and the reasoning, so the retirement is explained in the notes that DO ship. *Alternative:* amend it to "shipped and retired" — rejected as noise about a release nobody saw.

**ADR-0086's `status:` front matter is left at "accepted, not yet implemented"**, matching the repo's practice (ADR-0084 is implemented and still says the same); flipping it is not this task's convention to invent.

## Notes for whoever reads next

- `pnpm changeset status --since=main` fails **only** because the new changeset file is untracked in my worktree (changesets resolves "new changesets" with `git diff --name-only <merge-base>`, which cannot see an untracked file). Everything else in `verify` is green: `format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, `test` (all packages, platforms and examples).
- `openAndBootstrap`'s `processor` option (the snapshot's identity check) can no longer be computed by an app from its declaration; the browser guide now takes it from a build-supplied constant. Giving a snapshot a self-describing label is `a-snapshot-is-labelled-with-the-identity-it-was-computed-under`, already in `work/tasks/ready/`.


wighawag:

The CONTRACT step, and the task that STOPPED twice before this. It stopped correctly both times: its premise was false, because hashing bytes requires having bytes and the arrivals with none had been left behind by every batch that honestly reported itself clean. With the two inverted dependency edges corrected, the fifth and sixth batches landed, and an inventory of labelled witnesses in hand, it is finally the removal it always claimed to be. I checked the deletions by grep rather than by changeset.

**Gone from the code:** `EventProcessor.getVersionHash()`, `assertProcessorVersion`, `EntityProcessor.version`, `entityProcessorVersionHash`, `ContextIdentifier.processorFingerprint`, `ProcessorDriftReport`, `ProcessorDriftComparison`, `processorDriftReport`, `strictProcessorDrift`, and both containers' `onProcessorDrift`. `assertProcessorVersion`, `strictProcessorDrift` and `onProcessorDrift` now appear ONLY in `CHANGELOG.md` files, which is history and correctly untouched. `getVersionHash` survives in exactly two source files and both are PROSE explaining what used to be there.

**`PROCESSOR DRIFT` is gone from every live surface** — zero in `packages/`, `platforms/`, `examples/`, `docs/guide/` and `CONTEXT.md`. It remains only in changesets that describe its deletion, in ADR-0086, and in `work/` records. The criterion said "nowhere in the tree", and rewriting decision records to erase a phrase would be the wrong reading of it; this is the same judgement the spike made about its own history, and it is right.

- [x] **`CONTEXT.md`'s glossary no longer describes any of the three as live.** The `generation` entry is rewritten to say identity is derived and never declared, that ALL edits move it, and that `unchanged` now has ONE reading. Its only remaining mention is the sentence "**The condition a drift report used to name cannot occur**, which is why the report is gone" — which is the criterion being met, not missed.
- [x] **`ReconfigureReport`'s `unchanged` arm no longer carries `drift`,** the admin route no longer returns it, and the three outcomes are still three.
- [x] **The derivation the HMR arrival needs was NOT deleted,** and the task that owns it is named where a reader will find it.
- [x] **The changeset says what this retires and why it is an ending rather than a reversal,** in the terms the task demanded: the report was the right fix for an identity that could LIE, ADR-0086 deletes the lie instead of reporting it, and the compensating machinery goes with the thing it compensated for.
- [x] **The pending changeset is dealt with deliberately: DELETED, with the reasoning stated.** Nothing has been released, so amending it would ship a release note for a report that never existed outside this repository, and the history belongs in the work record and the ADR rather than in notes to users. That is the right call and it is argued rather than assumed.

**It recorded the ADR deviation instead of quietly diverging.** ADR-0086's consequence list says `getCodeFingerprint()` and `utils/fingerprint.ts` go with everything else. Both survive, and the changeset says so under its own heading, explaining that the list is wrong in two places and that both were settled after it was written. Crucially it gives the real reason for using the SEAM method over the standalone function: the seam delegates through a wrapper to the author's own handlers, while the standalone function applied to a wrapper would fingerprint the wrapper's methods and produce "a constant no edit could move". That is a sharper statement of the hazard than I made when I corrected the criterion, and it follows this repo's existing habit of recording deviations in the file rather than leaving the ADR silently wrong.

**Both holes I flagged are closed with refusals rather than fallbacks, and the reasoning is the same in each:** every alternative was a lie. `processorIdentityOf` became `requireProcessorIdentity`, which refuses a spec with no identity BEFORE anything is registered, naming the two alternatives it rejected (inventing a name, which is the declared identity respelled; or writing `undefined` into a registry record and a persisted cursor). And a module whose handlers are all `bind`-ed or proxied — the `undefined` fingerprint case the sixth batch pinned with a test — is refused by `@etherfold/browser` in its own words, naming the two ways out. The core refusal explicitly declines to say WHY an arrival produced no identity, because it cannot know and the arrival can: that keeps the good message at the layer that has the context, which is the correct split.

**It did not eat the blocked leaf's work, and said so.** A `--processor` path naming an unbundled entry point is now refused, because after this deletion nothing could name its fold — the tree would otherwise be broken. But the refusal is deliberately minimal: it names the path and what is wrong, and does NOT give the build command, and it fires where the arrival is resolved rather than at configuration resolution. The changeset states plainly that `a-path-naming-an-unbundled-entry-point-is-refused` is what moves it earlier and adds the command. I verified that leaf still has real work against its own criteria (refuse before any database is opened; name the build command; ADR-0048's shape), and it does. That is the honest split between "keep the tree correct" and "make the message kind".

**It also declined to put the refusal in the loader,** on the grounds that resolving a path is a separate capability from naming what came back, and refusing there would take the test-substitution case with it and put a configuration decision inside a loader. Right call.

Nits: none blocking. Gate 2 is off in this repo.

Merging. Drift is now UNREPRESENTABLE rather than unreported: there is no declared identity left for the code to disagree with. That is ADR-0086's central claim, landed.

PR #171, opened 2026-09-18
feat(a-path-naming-an-unbundled-entry-point-is-refused): A path naming…

wighawag:
Task: `work/tasks/done/a-path-naming-an-unbundled-entry-point-is-refused.md`

The work is complete and the whole gate is green (`build`, `typecheck`, `test` across every package, `format:check`, `check:adr`, `check:refs`, `check:changesets`).

## What I did

**The refusal now fires at configuration resolution, before anything is imported or opened.**

- `packages/cli/src/config.ts` gains `refuseUnbundledProcessor(command, processorPath, options)`, beside the other input refusals and in ADR-0048's shape: it names the flag (`-p, --processor`), quotes the value, says in one sentence what the file is, and gives the build command with the path already substituted in. Two arms, one remedy: an **entry point** (naming the specifiers it still expects somebody else to resolve) and a **path this process cannot read at all** (where the command's `--outfile` is the path they named, because writing that file is what is missing).
- Called from all three folding paths (`prepareIndexing` for `run`/`build`, `index`, and the `reconfigure` re-read) immediately after `resolveCommandConfig`, before the provider, the module, the database and the port.
- Detection reuses `unresolvedImportsOf` (the artifact unit's judgement) rather than a second heuristic, so a bundle that merely mentions a package name in a string is accepted, and a `--platform=node` bundle keeping `node:crypto` still runs.
- `@etherfold/utils` gains `readProcessorPath`, which answers `bundle` / `entry-point` (with specifiers) / `unreadable` (with the reason) without importing anything; `openProcessorArrival`'s private `bundleAt` is now a thin wrapper over it, so the check and the arrival cannot mean two different files.
- Docs that had gone stale with the declared `version` are corrected: the `--processor` help text, and `packages/cli/README.md` ("both shapes work today" → the refusal, quoted verbatim, plus the `--minify` reason).

**Tests** (seam: the CLI's configuration tests, as the task asks): nine cases in `packages/cli/test/configuration.test.ts` covering the real committed bundle being accepted, the entry-point refusal naming path + import + command, the unreadable-path refusal, the string-mention false positive, a node builtin, relative-path resolution against `cwd`, the substituted arrival, and every folding command refusing identically. `aDeploymentRunsFromABundle.test.ts`'s refusal case is strengthened into an **ordering** assertion: the entry point's sibling now throws on evaluation, so the test fails if the refusal ever moves back into the loader. Three cases for `readProcessorPath` in `packages/utils`.

## Decisions

**The refusal is a SEPARATE async function beside `resolveCommandConfig`, not inside it.** `resolveCommandConfig` is documented and tested as pure, total and synchronous ("it opens nothing, dials nothing and imports nothing"), and `--processor` is the one input whose value is a PATH, so what is wrong with a bad one is a fact about the FILE. Making the whole resolver async and disk-touching to carry one input would have cost every other refusal its property. Alternative considered: put the check inside `openProcessorArrival` (rejected: that IS the loader, which is the placement the task names as the likely mistake). What it touches: three call sites (`prepareIndexing`, `indexCommand`, `reconfigure`) now each call it explicitly, and the ordering claim in each of their docstrings.

**A path this process cannot READ AT ALL is refused by the same refusal, not only an entry point that still imports.** A bare package specifier, a directory, or the output of a build that has not run previously fell through to the module route and failed with `ERR_MODULE_NOT_FOUND` from inside a loader. Under ADR-0086 none of them can name a fold, and "the build has not run yet" is the commonest way to meet this at all, so it gets the same message with `--outfile=<the path you named>`. This is a NEW user-visible refusal shape for `-p @scope/some-package`, which is why I am flagging it rather than burying it. Alternative considered: refuse only the entry-point case and leave unreadable paths to the loader (rejected: two different errors for one configuration mistake, and the loader's is the one an author cannot act on). What it touches: nothing configures a bare specifier today; `the-build-command-and-its-pinning-rule-are-documented` should describe one input with one rule ("a path to a self-contained bundle").

**A SUBSTITUTED arrival (`IndexingDependencies.importModule`) is exempt, and this answers the question that task left open.** `processorIdentity`'s docstring says explicitly that "whether an injected importer remains an accepted arrival at all is `a-path-naming-an-unbundled-entry-point-is-refused`'s question". It does: a caller that stated what comes back for a path has stated that there is no file to read, so the disk check says nothing. No flag and no environment variable reaches it, so it is not a way for a deployment to get round the refusal. Alternative considered: refuse regardless (rejected: it would break ~20 CLI suites that name a path no file backs, for no deployment-visible gain, and it would delete the arrival-substitution seam as a side effect of a message change). What it touches: every CLI test that injects `importModule`, and `IndexingDependencies.processorIdentity`'s open question, which is now closed.

**`requireArrivalIdentity` is KEPT and narrowed rather than deleted.** Its own docstring said this task "moves this to CONFIGURATION RESOLUTION"; I moved the refusal an author meets, and left the structural guarantee that no fold is registered without a name. It is still reachable by a substituted arrival that named itself nothing, and by a path whose bytes MOVED between the configuration check and the loader's read (a half-landed rebuild), which now fails closed instead of folding under a name nothing derived. Alternative considered: delete it and require the identity in the type (rejected: it would take the substituted-arrival case with it, which is a test seam four suites rest on, and the TOCTOU window would become unhandled). What it touches: `folding.ts`'s docstring, and `aDeploymentRunsFromABundle.test.ts`'s injected-arrival case, which still asserts the old message.

**`@etherfold/utils` gains a public `readProcessorPath`, so "what is at this `--processor` path" is written once.** The CLI needed the bytes AND which case it met; the arrival needs only "bundle or not". Rather than the CLI re-implementing the `isAbsolute`/`join(cwd, …)`/read rule, both now go through one function. Alternative considered: three lines of path resolution duplicated in `config.ts` (rejected: two resolution sites for one path is exactly how a check refuses a deployment that runs). Note the residual cost I accepted: the file is read twice per start-up, once by the check and once by the loader, which is also what makes the TOCTOU case above fail closed. What it touches: `@etherfold/utils`' published API (covered by the changeset, `minor`).

**The refusal pins the build command, which `the-build-command-and-its-pinning-rule-are-documented` must match.** It emits `esbuild <entry> --bundle --format=esm --minify --outfile=<bundle>`, the command the two committed fixture READMEs and the example already use, written in one place (`bundleCommand`) with the `--minify` identity reason at the site. That task's criterion "the error and the docs say the same thing" is therefore satisfiable by quoting this.


wighawag:

The migration's front door, made kind. #170 deliberately left this work rather than eating it, so the first thing I checked was whether there was real work left to do here — there was, and it is done properly.

- [x] **Refused at CONFIGURATION RESOLUTION, before any database is opened or any generation is registered.** It now sits in `config.ts` with the other input refusals, before a module is imported, a database opened, a port bound. The comment explains why this one check reads a disk when every other refusal in that module is about a string, and why the ORDER is the reason it belongs there rather than in a loader: refusing where the bytes are already being imported is "too late twice over", because a database may be open and what the author gets is an error about module resolution rather than about their configuration. That is exactly the criterion's reasoning, arrived at independently.
- [x] **The refusal names the path, says what is wrong in a sentence, and gives the build command.** It also names the specific import that is still unresolved (`it still imports "./abi.js"`), which turns "this is not a bundle" into something actionable — the artifact unit's original point, paying off three tasks later at the surface an author actually meets.
- [x] **A genuine bundle resolves and runs,** asserted against a REAL bundle so nobody who has migrated can meet the refusal.
- [x] **It follows ADR-0048 rather than inventing a second shape,** and all three folding commands (`run`, `build`, `index`) refuse identically, as does the re-read behind `POST /{indexer}/admin/reconfigure`, which reports it as `failed` with the live fold untouched.

**It is not a second heuristic, which was the task's sharpest warning.** `unresolvedImportsOf` remains this repository's only definition of self-contained — what the artifact loader refuses on and what the arrival picks its route with — and it decides here too. Both false-positive cases are tested explicitly: a bundle that merely MENTIONS a package name in a string is not refused, and a `--platform=node` bundle that legitimately keeps `node:crypto` still runs. Those are the two ways a naive check would have hurt real users.

**The best judgement here is one the task did not ask for.** It noticed that the commonest failure is not a misconfigured path at all but THE BUILD THAT HAS NOT RUN, and refuses an unreadable path in the same shape — with `--outfile` set to the path the operator actually typed, because writing that file is precisely what is missing. That turns the most frequent case from "no such file" into a command that fixes it.

**`readProcessorPath` closes a divergence before it could open.** A caller that REFUSES a path and the arrival that OPENS one now read it through the same unit, so they cannot mean two different files or return two different verdicts about one. Given this family's history of two things quietly disagreeing about identity, unifying that seam is the right instinct.

**It set up the last task rather than leaving it a puzzle.** `bundleCommand()` is the ONE place the command is written, with a note saying an author meets the refusal first and the documentation second so the two must match down to the flags, and naming where the documentation lives. `the-build-command-and-its-pinning-rule-are-documented` has a criterion that the refusal and the docs name the same command; it now has a single function to point at instead of a string to keep in sync. I will verify that match at its Gate 3.

Nits: none. Gate 2 is off in this repo.

Merging.

PR #172, opened 2026-09-18
feat(a-snapshot-is-labelled-with-the-identity-it-was-computed-under): A…

wighawag:
Task: `work/tasks/done/a-snapshot-is-labelled-with-the-identity-it-was-computed-under.md`

## What I built

**The end-to-end assertion the round trip was missing** — `packages/cli/test/aSnapshotIsLabelledWithItsBundleIdentity.test.ts` (6 cases). The CLI is the only package where the whole sentence can be said in one breath: the artifact and its identity live in `@etherfold/utils`, the producer and the candidate rule in `@etherfold/processor-entities`, and a deployment actually runs a bundle here.

- A deployment folds real blocks through the committed bundle fixture (`nfts.bundle.js`) and publishes what that fold computed. The snapshot's `processor` label is read off the generation the deployment **registered**, so it is the identity of the artifact that actually ran, and the envelope travels through `JSON.stringify`/`parse` the way a mirror serves it.
- A client that loaded the **same bytes** through `loadProcessorArtifact` (deriving its identity itself, told nothing by the producer) bootstraps: `{status: 'bootstrapped'}`, the producer's rows are its state, its floor is the snapshot's block, and an `EntityEventProcessor` built from that bundle's own authoring object answers `load` with the snapshot's block rather than the start block.
- The negative half: a client running `nfts-edited.bundle.js` (one handler line different, identical entity declarations, nothing declared different at all) gets `{status: 'not-bootstrapped', reason: 'processor-mismatch'}` with an untouched store, and a host that installs it anyway is refused with `SnapshotProcessorMismatchError` naming both folds. Nothing translated.

I verified the assertions bite by mutating the producer's label to a literal: three cases go red, and I reverted the mutation.

**Documentation at the two choice sites**: `createSnapshot`'s `processor` field now says where a producer gets the value (from the fold that wrote the rows, never derived a second way beside it) and why a wrong label is unrecoverable at a leaf; `ENTITY_SNAPSHOT_FORMAT` carries the format-number reasoning. `CONTEXT.md`'s `bootstrap` entry drops the stale "another **processor version**" phrasing for the ADR-0086 vocabulary and records that the envelope number did not move.

**The candidate predicate is untouched.** No production behaviour changed: the four migrate batches already moved what a deployment hands `createSnapshot`, so a producer running a bundle labels with the hash today. What was missing was the proof and the format-number decision, which is exactly what the task's re-ordering note said was left.

Gate: `format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, full `test` all green (4,600+ tests). `pnpm changeset status --since=main` only sees a changeset once it is in the index; I confirmed with a throwaway index (the repo's own index untouched) that all three changed packages are named by `.changeset/a-snapshot-is-labelled-with-the-identity-it-was-computed-under.md`.

## Decisions

**The snapshot FORMAT number does NOT move: this is a VALUE change, not a FORMAT change.** `ENTITY_SNAPSHOT_FORMAT` stays at `1`. The envelope's shape is identical, `processor` is the same field in the same place meaning the same thing, and it is opaque on both sides (compared for equality, never parsed, which is the invariant ADR-0086 rests on), so a label derived the old way is not half-understood by a new reader. The alternative considered was bumping it, on this repo's habit of refusing a document it cannot read rather than half-parsing it (ADR-0040). Rejected because that habit protects against a shape a reader would misread, and here the mismatch is already answered precisely: `processor-mismatch` (not a candidate) or `SnapshotProcessorMismatchError` at install. A bump would convert that precise refusal into `unreadable-format`, which tells a user their app or the publisher is out of date when the truth is that the snapshot is for another processor: strictly less information, on the one path where a client has no stream to re-fold. And nothing is published (`CONTEXT.md`), so there is no document on either side of the distinction. What it touches: `SnapshotFormatError` / `isReadableHead` / the `unreadable-format` outcome keep their current meaning, and `a-generation-can-be-seeded-from-a-published-artifact`'s sibling `STREAM_SEED_FORMAT` is unaffected. Recorded beside the constant with the condition that WOULD justify a bump (a field appearing, disappearing or changing meaning). Not an ADR: it is cheaply reversible (no published documents exist) and the trade-off is local to one constant.

**No producer-side derivation helper was added, and `createSnapshot` still takes the label from its caller.** The criterion "computed by the producer from the same artifact the deployment runs" is satisfied by taking the label from the fold that wrote the rows (its registered `processor`), which is what the test does and what the field's docstring now instructs. The alternative was a helper in `@etherfold/processor-entities` that hashes bytes for a publisher. Rejected on two counts: it would be a second spelling of `processorArtifactIdentity`, and it would put a node-only hash into the package a browser bundles, while publishing snapshots is deliberately out of scope (ADR-0028; `work/notes/ideas/publishing-snapshots-of-versioned-state.md`). What it touches: whoever builds the publishing spec inherits the docstring's instruction rather than an API to call.


wighawag:

The second of the two `blockedBy` edges I inverted to break this family's deadlock, so beyond the criteria I checked whether the reorder was actually sound. It was, and this PR is the evidence: the VALUE was already correct because the migrate batches had moved what callers hand `createSnapshot`, and what was genuinely missing was the assertion and the format-number decision. Neither needed the declared version to be gone first.

- [x] **The label is the identity ADR-0086 derives, computed by the producer from the same artifact the deployment runs** — the fold that wrote the rows, never derived a second way beside it. The docstring now says WHERE a producer gets it, which is the part that could have drifted.
- [x] **The candidate rule is UNCHANGED in meaning.** `processor-mismatch` still means not-a-candidate, and a snapshot installed anyway is still REFUSED rather than translated. The predicate itself is untouched — which the task explicitly warned against changing, and which I verified by diff.
- [x] **Asserted end to end rather than by comparing two strings.** This is the criterion that mattered and it is met in the strongest available form: a deployment folds blocks through a REAL committed bundle and publishes what it computed; a client that loaded the same bytes through the artifact loader installs it and resumes at the snapshot's block; a client running the EDITED bundle (one handler line different, nothing declared different) gets `processor-mismatch` and an untouched store. As the changeset puts it, neither side is handed the other's value, so it asserts that two derivations AGREE rather than that two constants are spelled the same. The extra case — the reseeded client asks the chain from the snapshot's block rather than the start block — proves the install actually took effect rather than merely being accepted.
- [x] **The FORMAT number was considered deliberately, and the answer is recorded beside the constant** rather than only in a report. It does not move, because this is a VALUE change and not a FORMAT change: `processor` is the same field in the same place meaning the same thing, opaque on both sides, so a label derived the old way is not half-understood by a new reader — it is simply another fold, which the candidate rule already answers precisely. And the argument against bumping is the good one: it would convert a precise `processor-mismatch` into `unreadable-format`, telling a user their app is out of date when the truth is that the snapshot is for another processor. Strictly less information, on the path where a client can least afford confusion.

**It understood why this small change carries a large promise.** The changeset states the leaf property in its own terms: a snapshot-seeded generation has no stream to re-fold and no history below its own block, and a public node frequently refuses the historical `eth_getLogs` a backfill would need — so such a client cannot recover from a mislabelled snapshot by re-indexing. That is exactly the reasoning that makes this task load-bearing rather than cosmetic, and it is why I was careful about inverting its dependency rather than just doing it.

**A nit I had been carrying since #164 is closed.** `packages/processor-entities/README.md`'s `openAndBootstrap` example still showed `eventProcessor.getVersionHash()` after the batches; it is gone now, swept by the contract task. Nothing left to route.

Merging. One task left: `the-build-command-and-its-pinning-rule-are-documented`, which is now unblocked on all three of its leaves and owns ADR-0086's expiring status line.

PR #173, opened 2026-09-18
feat(the-build-command-and-its-pinning-rule-are-documented): The build…

wighawag:
Task: `work/tasks/done/the-build-command-and-its-pinning-rule-are-documented.md`

All gates green (`format:check`, `check:adr`, `check:refs`, `check:changesets`, `build`, `typecheck`, full `test` at exit 0; `changeset status --since=main` verified green in a simulated committed tree, since it reads tracked files only).

## What I did

**Test first** — `packages/cli/test/theDocsAndTheRefusalNameOneBuildCommand.test.ts` (red before the docs existed, green now). It takes the command `refuseUnbundledProcessor` really emits and holds three copies to it, flag for flag and in order: the CLI README's canonical statement, the example README's worked instance, and the example's `build:bundle` script that actually runs. A fourth case pins `--bundle --format=esm --minify --outfile=` on its own, so the three copies cannot agree with each other and be wrong together. The doc copies are located by a `<!-- bundle-command: -->` anchor, on the precedent of the provider-surface claim in `@etherfold/core`.

**Measured rather than asserted** — `docs/spikes/the-build-command-and-its-pinning-rule-are-documented/` (a re-runnable script plus the numbers). One source built from a checkout root and from its package directory: `f2d21373…` vs `bce77ea9…` un-minified (the banner is `// src/abi.js` vs `// abi.js`, relative to the bundler's working directory), `bfbbcd68…` both times minified. `--sourcemap=external` leaves the bundle byte-identical; plain `--sourcemap` appends a `//# sourceMappingURL=` comment and so names a different generation.

**Documentation** — `packages/cli/README.md` gains `## Producing the processor bundle`: the one command, `--minify` as a correctness rule for identity with the measured fact, the pinning rule over the bundler's VERSION AND FLAGS with the "state is never reused" consequence, `rollup` as the alternative and `tsup` not recommended in a clause, source maps per spelling, and a migration subsection quoting the refusal. Pointers from the top of that README and from the root README; the example README keeps its runnable instance and defers the rule.

**ADR-0086** — `status: accepted, not yet implemented` removed, after checking that all twelve other tasks naming `spec: a-processor-is-a-bundle-and-its-hash-is-its-identity` are in `work/tasks/done/` (this was the only one in `ready/`) and that the code matches: `version`, `assertProcessorVersion`, `getVersionHash()` and the drift report are gone, `requireProcessorIdentity` refuses an arrival that supplies none, and `getCodeFingerprint()` survives only in its new HMR role. No "Built." ceremony added: ADR-0073 records that marking it is the exception, not the norm.

**CONTEXT.md** — two new glossary entries, **processor identity** and **processor artifact**, beside `EventProcessor`. **Changeset** — `.changeset/the-build-command-and-its-pinning-rule-are-documented.md` (`etherfold: patch`).

**Out of scope, captured** — `work/notes/observations/adr-0084-may-still-carry-an-expired-status-line.md`: ADR-0084 still declares `accepted, not yet implemented` while tasks that read like its family sit in `done/`. Not verified, not touched. (ADR-0085's line looks correct: its spec is still in `work/specs/ready/`.)

## Decisions

**Source maps are documented PER SPELLING, not as "they do not affect the identity".** The acceptance criterion asks for source maps to be "stated not to affect the identity". Measured, that is true of `--sourcemap=external` (byte-identical bundle) and FALSE of plain `--sourcemap`, which appends a `//# sourceMappingURL=` comment and so produces a different hash, hence a different generation. I documented the measured truth and named `--sourcemap=external` as the spelling to use, rather than writing the flat sentence the criterion suggests. Alternative considered: state it flatly and let an author discover the re-fold; rejected, because a documentation claim that is false for the DEFAULT spelling of the flag is exactly the class of error this task exists to prevent. Touches: nothing in code; it does set a user-facing recommendation of one flag spelling over another, and it strengthens the pinning rule (a source-map flag is a flag to pin).

**The canonical home is `packages/cli/README.md`, not the docs site.** `docs/guide/getting-started/index.md` is a two-line stub ("Documentation in Progress"), so a new guide page would sit in an unfinished site and need a sidebar edit to be reachable. The CLI README is what an npm user of `etherfold` reads, is where `-p` is documented, and is the reference for the `etherfold <command>` the refusal names. The example README keeps only its concrete invocation and links here for the rule; the root README carries a one-paragraph pointer. Alternative: a `docs/guide/` page (rejected, above) or the root README (rejected: it is browser-first, and the block is a CLI-authoring concern). Touches: anyone later writing the getting-started guide should link this section rather than restate it, or the "stated once" property dies.

**The section is called "Producing the processor bundle", not "Building the bundle".** Coherence check against the existing language: `build` already means `etherfold build`, a command "named for what it PRODUCES: a database", and the section sits directly under it. A heading with `build` in it would have made one word mean two things one heading apart. Touches: the anchor `#producing-the-processor-bundle` is linked from the root README, the example README and the spike.

**Two new glossary terms, both taken from names the code already uses.** **processor identity** (`requireProcessorIdentity`, `GenerationId.processor`) and **processor artifact** (`loadProcessorArtifact`, `processorArtifactIdentity`, `unresolvedImportsOf`), rather than coining a word such as "bundle identity". They overlap the existing **generation** entry, which already says a generation is identified by its stream plus the processor's identity; that entry is the composite and these are the term it names, so I left it alone and did not duplicate its content. Touches: `CONTEXT.md` is the glossary later artifacts inherit, so the spellings matter.

**The `<!-- bundle-command: -->` anchor is a new convention in a documentation file.** It follows the existing `<!-- provider-surface: -->` anchor in the root README, which is checked by `packages/core/test/theEngineDeclaresItsMethodSet.test.ts`; I copied the shape rather than inventing a second way to mark prose held to the code. Touches: any future edit to `bundleCommand()` in `packages/cli/src/config.ts` must change both documents in the same commit, which is what the anchor comment says out loud.


wighawag:

The last task of the family, and the one deliberately fanned in behind every leaf so that removing ADR-0086's expiring status line has an unambiguous owner.

- [x] **`ADR-0086`'s `status: accepted, not yet implemented` line is REMOVED,** after confirming the family had landed. The file now opens on its heading. This is the failure `work/protocol/ADR-FORMAT.md` warns about — "every task in a chain can see that it is not the last one while the actual last one has no way to know that it is" — and it did not happen.
- [x] **The build command is documented once, in a place an author starting out will reach** (`packages/cli/README.md`, under `Building the bundle`), with the rest of the tree pointing at it rather than repeating it.
- [x] **`--minify` is documented as MANDATORY with the identity reason, including the measured fact.** Un-minified, esbuild opens each module with a `// <path>` banner naming it relative to the directory the bundler ran in, so the building machine's layout lands in the bytes. Measured rather than asserted: one source built from a checkout root and from its package directory hashes differently un-minified and identically minified, with a script committed that re-runs the measurement. And it leads with the consequence in the right terms — drop the flag and a laptop and CI do not build a bigger bundle, they disagree about which generation they are.
- [x] **Pinning covers the bundler's VERSION AND its FLAGS,** with the reason: the output is a function of both, and the output is the identity.
- [x] **`rollup` named as the alternative; `tsup` not recommended in a clause** ("esbuild with a wrapper that adds a version to pin and no determinism"), not a paragraph.
- [x] **The migration note tells an existing author what changed and the one command to run.**
- [x] **`CONTEXT.md` describes how a processor is identified in the new vocabulary.**
- [x] **A changeset.**

**Source maps were handled better than the criterion asked, by contradicting it.** The criterion said to state that source maps do not affect the identity. That is FALSE for the default spelling, and the build measured it rather than repeating it: `--sourcemap=external` leaves the bundle byte-identical, while plain `--sourcemap` appends a `//# sourceMappingURL=` comment and therefore names a different generation. Documenting it per spelling is correct where a single reassuring sentence would have been a trap — an author who followed it would have silently changed their generation on every build. This is the third time in this drive a builder has been right to contradict a claim in its own task, and it is the behaviour I most wanted to see.

**The docs/refusal agreement is ENFORCED rather than checked once.** `theDocsAndTheRefusalNameOneBuildCommand.test.ts` extracts the command the refusal actually emits and compares it flag-for-flag and in order against three copies: the CLI README's canonical statement, the example's worked instance, and the `build:bundle` script that really runs. So the documented command is an EXECUTED one, and the three can never drift from the error message an author meets first. The criterion only asked that they agree today; this makes them unable to disagree tomorrow, which is the right shape for a rule about reproducibility.

**It captured an off-path finding instead of fixing or ignoring it.** `work/notes/observations/adr-0084-may-still-carry-an-expired-status-line.md` records that ADR-0084 still declares `accepted, not yet implemented` while what look like its tasks sit in `work/tasks/done/` — explicitly unverified and out of scope — and it went further by checking ADR-0085 and finding its line CORRECT, since its spec is still in `work/specs/ready/`. Noticing the class of bug while fixing an instance, and distinguishing a real case from a false positive before filing, is exactly right. I am routing it in the drive's triage rather than acting on it here.

Merging. That completes ADR-0086: an author can no longer state a processor's identity, and the ADR no longer claims to be unimplemented.

PR #175, opened 2026-09-19
feat(promotion-arms-from-the-slot-so-a-restart-can-finish-an-upgrade):…

wighawag:
Task: `work/tasks/done/promotion-arms-from-the-slot-so-a-restart-can-finish-an-upgrade.md`

All three parts are built, measured end to end, and the whole gate is green.

## What I did

**Part one, the gate (`packages/core/src/receivingContainer.ts`).** `applyPolicyTo` now arms what the `successor` slot names and nothing else, read from the registry rather than from how a fold arrived. `opened` and the in-memory `candidates` set are deleted; `settlePromotion` reads the slot; `movePointer` lost the held-fold argument it only needed in order to disarm one.

**Part two, the trigger can now be evaluated with no held incumbent.** `GenerationRegistryPort` gains `readStateCursor(id)`, `dropState`'s symmetric sibling, forwarded by `GenerationRegistry` under the same name. `cursorOf` reads a generation's `lastToBlock` by identity (namespace), never through `processor.load`, so both sides of the comparison can be generations this process holds no fold for. Nothing is retained, re-imported or reconstructed; no `import` was added anywhere. Hosts: the CLI supplies it from the same `stateFor` that supplies `dropState` (built **unclaimed**, so reading a position cannot take the writer claim from the fold that is writing it); the server's SQL factory and core's memory port take it as an option beside `dropState`; the browser's IndexedDB port takes it optionally (its container never asks — see Decisions).

**Part three, the settle is reached on `run`.** `packages/cli/src/index.ts` no longer gates the between-cycles `rebuildMore` on `container.followers().length > 0`; it is unconditional on `run`, exactly like the prune beneath it. Nothing about `index`'s scheduling was touched.

**The deliverable:** `packages/cli/test/aRestartFinishesTheUpgrade.test.ts` — a `run` stood up the way the CLI tests do, **stopped**, and re-run over the same libSQL handle with an edited bundle, driven until the pointer moves; plus the revert case driven through a restart under the default and under `immediate`; plus `manual` still waiting while level, and `immediate` on an unchanged restart not moving its own pointer. Also ADR-0046 amended, ADR-0084's `not yet implemented` status removed, `CONTEXT.md`'s trigger/arming glossary entry corrected, and a changeset.

## Measured, not inferred

Each part switched off in turn, rebuilt, the same file re-run:

| disabled | result |
|---|---|
| part two (cursor only for held folds) | 4 of 7 fail — the pointer never moves |
| part three (follower gate restored) | the same 4 fail |
| arming at `open` (the `opened` gate restored) | `immediate` on a changed restart fails; `on-catch-up` still passes, because the loop's settle reads the slot |
| the `predecessor` clause of the arming condition | the revert-under-`immediate` restart fails: the pointer goes straight back |
| the `canonical` clause | **nothing fails** (see Decisions) |

Two out-of-scope signals captured, not fixed: `work/notes/observations/status-reports-no-entry-for-a-generation-this-process-holds-no-fold-for.md` and `.../a-rerun-build-registers-a-successor-and-exits-without-ever-settling-the-pointer.md`.

Gate: `pnpm build`, `pnpm typecheck`, `pnpm test` (all packages), `pnpm format:check`, `check:adr`, `check:refs`, `check:changesets` all pass. (`changeset status --since=main` reports "no changesets found" only because the new changeset is still untracked — `git diff` cannot see an untracked file; it resolves on the runner's commit.)

## Decisions

**Which hazard each clause of the arming condition rules out, and one that rules out nothing.** The condition is stated positively — `slotHolding(slots, fold.record) === 'successor'` — because `add` always leaves the arriving fold in exactly one slot, so it excludes exactly two things. **`predecessor` is load-bearing and measured**: removing it makes a restart after a revert under `immediate` re-promote what the operator reverted away from, which is ADR-0046's hazard. **`canonical` rules out nothing observable**: I removed it and all seven cases still passed, because `moveCanonicalTo` writes nothing when the target is already what the pointer names, so the "promote whatever the host was built with" hazard is already neutral one layer down. I kept it anyway and said so at the choice site: the alternative (`!== 'predecessor'`) would arm the fold the pointer already names, make the rule depend on a distant no-op for its safety, and produce log lines asserting something false about that generation. This touches nothing else; it is the whole of the `run`/`index`/browser promotion behaviour and is what ADR-0046's amendment now records.

**The cursor-reading seam is on `GenerationRegistryPort`, returns a NUMBER, and is required on the port type.** Alternatives considered: a `ReceivingIndexerOptions.readCursor` beside `port` (smaller blast radius, but it splits one fact — where a generation's state lives — across two injection points, when `dropState` already established the port as its home), and returning the decoded `LastSync` (rejected: the window holds whole decoded blocks, and the comparison needs one number — the same reasoning ADR-0047 gives for not handing the cursor to `/status`). It is **required on the port type** (like `dropState`) so no substrate can silently omit it, and **optional in each factory's options** exactly as `dropState` is. Hosts that supply it: the CLI's `openFolding` (production), the core/server/CLI test hosts that stand a receiving container up. The browser's `openGenerationRegistryOnIndexedDB` takes it **optionally** and supplies nothing today — a tab runs the chain-facing container, whose trigger reads its held engines' cursors, so requiring it there would be a required option nothing consults. That is a deliberate exception to "every host that constructs a registry port supplies it", documented at the option.

**The per-fold identity check moved from the container to the port's contract.** The old `cursorOf` answered `undefined` when the stored cursor named another processor, which mattered because it read through a possibly-shared store. Addressing by identity makes that structurally impossible on every host in this repository (the namespace *is* the pair), so the CLI's implementation does not re-check it; the port's JSDoc states the rule ("a cursor another fold wrote answers `undefined`") so a host that maps several identities to one store still owes it.

**`settlePromotion` now consults the policy itself** (`promotionOnAdd(...) !== 'arm'` returns), because arming is no longer something a container remembers. Behaviour is identical to before under all three values; what changed is that the policy check moved from arming-time to settle-time, which is what keeps `manual` waiting while a successor sits in a slot.

**`run`'s gate was deleted rather than replaced** with a cheaper "is anything promotable" check. Such a check needs a registry read to answer, which is most of what it would save; the unconditional call costs about three registry reads per cycle in a gap the loop already waits, and it matches the prune beside it, which is unconditional for the same reason. It touches `run` only — `an-index-process-advances-the-successor-it-registered` still owns where a rebuild gets its turn against the ingest path on `index`.

**ADR-0046 was amended, ADR-0057 deliberately was not.** ADR-0046 owns the arming rule and two of its statements became false for the receiving container ("`open` arms nothing", "a successor created in a previous session does not auto-promote after a reload"), so both carry amendments naming this task and the chain-facing twin's unchanged behaviour. ADR-0057 needed nothing: its only sentence on this subject points at ADR-0046 for the arming, and its own `everCanonical` bullet was already amended by the slots task; the route it decides is untouched. ADR-0084's front-matter `not yet implemented` came off, as that ADR itself instructed, and its STILL-OPEN block became a LANDED block naming the three parts. `CONTEXT.md`'s "promotion TRIGGER" entry was rewritten, because it stated as domain language that the arming is in memory — now true only of the chain-facing container.

**Test hosts got two small shared seam helpers** (`packages/cli/test/utils/generationState.ts`, `packages/server/test/utils/generationDatabases.ts`) rather than five copies of the same four lines; the server's reads the cursor row with plain SQL, because `@etherfold/state-store-sqlite` is not one of that package's dependencies and adding one for a test double would be the heavier change. `theDeploymentSelectsItsPromotionPolicy.test.ts` was **not** rewritten onto the now-shorter restart path: what it asserts is the reconfigure endpoint, which nothing else covers. It gained a header pointer to the new file instead.


wighawag:

The correctness cliff ADR-0084 exists to remove. This task previously routed to needs-attention when a measured build implemented the narrowing exactly as written and the pointer still did not move, so the thing I checked hardest is whether all THREE parts landed and whether the end-to-end criterion — the one left unreachable last time — is genuinely asserted.

- [x] **A successor registered at `open` is promoted when it catches up, asserted END TO END on a restarted deployment.** `packages/cli/test/aRestartFinishesTheUpgrade.test.ts`: a deployment stood up the way the CLI tests stand one up, STOPPED, and re-run over the same libSQL handle with an edited bundle, driven until the pointer moves. That is the criterion's own wording, met in its own terms, and it is the one a green gate could not have caught last time.
- [x] **A restart after a deliberate REVERT does not re-promote what was reverted away from, under any policy value** — tested under the default AND under `immediate`, which the changeset correctly calls "the value with the most to say". This and the criterion above are the pair that makes the narrowing safe, and neither is sufficient alone.
- [x] **`immediate` still does not move a host's own pointer for no reason** ("does NOT move its own pointer when the restart changed nothing"), and there is a further case asserting a restarted successor under `immediate` answers AT ONCE, which is the other side of the same rule.
- [x] **`manual` still WAITS, however level the successor gets.** The slot says what a generation is FOR; the policy still says WHEN.

**The gate was not "simply removed", which the task warned against — its purpose was preserved and its cost was not.** The `opened` flag and the in-memory `candidates` set are both deleted, and arming is now `slotHolding(slots, fold.record) !== 'successor'`. That keeps both hazards closed structurally rather than by silence: `immediate` cannot promote the fold a host was built with because that fold is what `canonical` names, and a revert cannot be undone because what it returned to is what `predecessor` names and is armed under no policy. Arming stopped being something the container remembers, which is precisely ADR-0084's insight applied — the durable fact replaces the inferred one rather than sitting beside it.

**Part two is the one that makes the restart shape evaluable at all, and it took the narrow road.** `cursorOf` no longer goes through `fold.processor.load(...)`; it reads `registry.readStateCursor({stream, processor})`. So the comparison needs a NUMBER and gets a row addressed by the generation identity, with no engine — the rule the read tier already follows (ADR-0053). Crucially it did NOT drift into retention, which the task named as the likeliest failure: nothing retains, re-imports or reconstructs past processor code.

**Part three was real and would have left the whole thing unreachable.** `run` gated its bounded rebuild on `container.followers().length > 0`, and a restart-shape successor is not a follower, so the trigger was never reached. The diff's reasoning for removing it is the right one: the gate asked whether anything is being REBUILT when the question is whether anything might be PROMOTED.

**The folded-in criterion is honest about being a no-op.** The cancelled `a-folds-cursor-is-read-with-the-source-it-folded` correction is delivered, and the changeset says plainly that it fixes nothing observable today because neither implementation of `load` uses `source` to choose a cursor. Better still, it is gone STRUCTURALLY rather than patched: the new read is addressed by the generation identity, and that identity IS the pair. That is the cleanest possible discharge of a latent-defect criterion.

**ADR-0084's expiring status line is now `status: accepted`.** This is the last task of that family, and it removed the line — which also closes the question I triaged two days ago, where I judged `adr-0084-may-still-carry-an-expired-status-line` a false positive because `promotion-arms…` was still open. It was; it is not now; the line expired exactly here and was taken with it.

**Both ADR amendments are the right shape.** ADR-0046 and ADR-0084 gain dated, task-attributed `AMENDED` blocks that preserve the original rule and record only what superseded WHERE the fact lives and WHEN it may be read. ADR-0046's note is measured rather than asserted: with the `predecessor` clause removed, a restart after a revert under `immediate` re-promotes what was reverted away from. It also scopes itself honestly — the chain-facing `Indexer` is unchanged and still arms in memory, and unifying the two is named as separate work rather than smuggled in.

**One subtle correctness point worth calling out:** the new seam is built UNCLAIMED, so reading a generation's position never takes the writer claim from the fold that is writing it. Nothing asked for that; getting it wrong would have been a writer-token bug (ADR-0075) discovered much later.

**Two off-path findings filed, correctly not fixed.** `a-rerun-build-registers-a-successor-and-exits-without-ever-settling-the-pointer` is a genuine defect reasoned out from the same drive-loop gate — a re-run `build` over its own database with changed bytes folds a successor to the tip and exits with the pointer still on the old generation, so the artifact serves the OLD fold. And `status-reports-no-entry-for-a-generation-this-process-holds-no-fold-for` — an operator watching an upgrade across a restart sees a number climbing toward a target the page never shows. Both are labelled unmeasured. I will triage them at the end of this drive.

Merging. This unlocks `an-index-process-advances-the-successor-it-registered`, whose end-to-end criterion depended on the arming landing first.

PR #176, opened 2026-09-19
feat(an-index-process-advances-the-successor-it-registered): An `index`…

wighawag:
Task: `work/tasks/done/an-index-process-advances-the-successor-it-registered.md`

## What landed

**`packages/cli/src/indexCommand.ts`** — the whole of the change.

1. **`index` now schedules the bounded rebuild.** A host-owned timer at a new `DEFAULT_REBUILD_INTERVAL_SECONDS` (4) calls `container.rebuildMore()` unconditionally, exactly the call `run` makes in the gap between its cycles, and reports permanent stalls through `run`'s own `newlyStalledFollowers` (`followers.ts`) rather than a second copy. It fails soft: a failed chunk logs and the next tick retries.
2. **The rebuild and the prune share one in-flight guard.** The scarce thing is the one database handle, so a tick of either kind is skipped while the other is in flight. `--prune-interval` keeps its exact meaning, including `0`.
3. **A successor this half cannot feed itself is said out loud** at start-up, in the operator's own startup lines: the digest it holds, the canonical one still answering, and the statement that nothing here fetches so the successor advances only from what a sender pushes at `/{indexer}/ingest` for its own `{source, config}`. The diagnostic cannot refuse a start-up. A process holding no successor says nothing.
4. **The stale header claim is corrected, not made true.** `ReceivingIndexer.add` is reached from `open()` and from `reconfigure.ts` (a `run`-only surface) and from nowhere else; a batch matching no live wire context is refused `400 context-mismatch`. The header now says what the code does, and item 7 documents the new schedule. The list also said "Four things" while listing six.

**`packages/cli/test/anIndexProcessAdvancesItsSuccessor.test.ts`** (new, 8 cases) — a real receiver on a real port with a real `etherfold fetch` pushing at it, stopped, re-run over the same libSQL handle with an edited bundle, driven until the pointer moves. Written before the production code; verified failing against it (forcing the schedule off makes exactly the two pointer-move cases fail, and nothing else).

**Collateral corrections** to statements this change falsified: the `--promotion` refusal reason on `index` (`config.ts`), the same sentence in `packages/cli/README.md`, and the identical stale claim in `CONTEXT.md`'s two-containers glossary entry. Plus a changeset.

`run` is untouched. `pnpm build && pnpm typecheck && pnpm test` green (304 CLI tests, all packages), `format:check`, `check:adr`, `check:refs`, `check:changesets` green.

Drift check first: the blocking task landed at `11481a0a`, `applyPolicyTo` does arm from the slot, `indexCommand.ts` did schedule no rebuild, and the header claim is false. No stale premise.

## Decisions

**Where the rebuild gets its turn: a host-owned TIMER at 4s, not the ingest path, and not the prune's clock.** `index` is driven by arrivals, and that is precisely why the turn is not taken there: ADR-0022 makes the rebuild a call the host schedules and never a side effect of a write, and "one chunk after each batch" would put an arbitrarily long catch-up between a sender and its acknowledgement for work that batch did not cause. That is the same argument the scheduled prune beside it already records, so the shape transfers. **Cost under load:** a tick with nothing to advance is a few registry reads (the bargain `run`'s unconditional call already states); a tick with something to advance is bounded at `DEFAULT_MAX_EMISSIONS_PER_CHUNK` (2000), so the two contend for the one handle for the length of one chunk and the ingest path wins the next tick outright, because an overlapping tick is skipped rather than queued. **Alternatives considered:** riding the ingest path (rejected, above); riding the existing prune timer (rejected — it would re-mean `--prune-interval`, so `--prune-interval 0`, which a deployment and several tests use to disable a disk reclaim, would silently disable processor upgrades, and 60s is the wrong cadence for a promotion); a new flag (rejected — no command in the set takes one, the number a deployment wants is the one the constant argues for). **4 seconds** is `run`'s own cadence rather than a new number: the gap a caught-up `run` waits is its poll interval, default 4000ms. **Touches:** the prune schedule in the same command (they now share one in-flight guard, so a tick of either kind can be skipped while the other runs), and `IndexDependencies` gains `rebuildIntervalSeconds` as a test-only instrument with no flag behind it.

**What an unfeedable successor reports: a start-up line, not a new refusal or a fetcher.** On this command a fold added at `open` is never a follower (`add` decides that from "do I already hold a fold on this stream", and at `open` the fold list is empty), so it holds no rebuild over the stored stream and is fed by the wire alone. The one way the upgrade silently never finishes is a sender pushing a different `{source, config}`, whose refusal is visible on the sender and nowhere here, so the honest place to say it is at start-up on this half. **Alternatives considered:** giving `index` a fetcher so it could feed its own successor (rejected — it deletes the property that defines the command); refusing to start when a successor is held (rejected — the ordinary restart upgrade is exactly this state and it is supposed to work); a new `/status` field (rejected — that reporter answers progress per fold, and this is a fact about the deployment's configuration, not a moving number). **Touches:** nothing else; it adds no flag, no route and no refusal.

**The module header is CORRECTED, and so are the other three places that repeated it.** Nothing creates a generation from an unseen batch, and inventing that path would be a feature — a receiver taking its deployment's shape from a sender's say-so — so correcting is the honest fix, as the task expected. I did not open a needs-attention signal for it, because I concluded the path should NOT exist rather than that it should. **Touches:** `CONTEXT.md`'s glossary entry for the two containers carried the identical sentence; correcting one and leaving the other is how a false claim survives being fixed, so both moved.

**The `--promotion` refusal on `index` keeps refusing, but its REASON changed, because my change falsified it.** It said the receiving half "schedules no rebuild, so nothing is ever added beside its live fold and the policy would have nothing to act on". After this change a successor its configuration named at start-up *is* carried to level and promoted here, under the default policy — so the second half is false. **I did not make `index` accept the flag:** that is a user-visible new input on another command's ownership row, the acceptance criterion asks only for the default policy, and ADR-0048's recorded direction is that refusal → optional is additive later while the reverse is breaking. The refusal now says it is about this command's INPUTS and that whether it should take the flag is a question nothing has answered. **Touches:** `config.ts`'s ownership/refusal table, `packages/cli/README.md`, and whichever future task gives `index` a reconfigure or a promotion input — the done record for `the-cli-selects-its-promotion-policy` already flagged that cell as the one to flip.


wighawag:

The other half of the gap whose first half landed in #175.

- [x] **An `index` process holding a successor ADVANCES it,** and **a restarted `index` FINISHES the upgrade** — asserted end to end on a split deployment stood up the way the CLI tests do it: a real receiver on a real port with a real `etherfold fetch` pushing at it, stopped, and re-run over the same libSQL handle with an edited bundle, driven until the pointer moves.
- [x] **The incumbent goes on answering every read until the move,** tested as "the upgrade is not an outage".
- [x] **The rebuild is BOUNDED the way `run`'s is** — `DEFAULT_MAX_EMISSIONS_PER_CHUNK` (2000) stored emissions per tick, with a test asserting it "does not starve the ingest path this command exists to serve".
- [x] **The stale module-header claim is settled deliberately and stated.**
- [x] **`run` is unchanged.** I checked this by diff because the criterion is explicit: `packages/cli/src/index.ts` is touched, but only to re-export `DEFAULT_REBUILD_INTERVAL_SECONDS`. `run`'s drive loop lives in `run.js` and is untouched.

**The scheduling decision — which the task named as the thing most likely to be got wrong — is argued rather than picked.** A timer, not a turn after each batch, and the reason is ADR-0022: the rebuild is a call the HOST schedules and never a side effect of a write. Since ingest is the only other thing happening in this process, "one chunk after each batch" would be that side effect under another name, inserting an arbitrarily long catch-up between a sender and its acknowledgement for work that batch did not cause. The cadence is `run`'s own poll interval rather than a new number, so a follower advances at the same rate on both shapes — which is the "prefer the shape `run` already uses if it transfers" instruction, followed with the transfer checked rather than assumed.

**It also answered the part of that criterion about contention, which was easy to skip.** The rebuild and the prune share ONE in-flight guard, because the scarce thing is the single database handle rather than the verb; an overlapping tick is SKIPPED rather than queued, since both passes are resumable by construction while queueing would let a backlog grow behind a slow pass. And `--prune-interval 0` keeps its exact meaning: disabling a disk reclaim is deliberately not a way to disable processor upgrades. That last sentence is the kind of coupling that would otherwise be discovered by an operator.

**The stale claim was CORRECTED rather than made true, and the reasoning is the right one.** No code path ever created a generation from an unseen batch: `add` is reached from `open()` and from a `run`-only re-read, and a batch matching no live wire context is refused with `400 context-mismatch`. Making the header true would have been a FEATURE — a receiver registering generations on a sender's say-so — not a cleanup. Best of all, it found the SAME false sentence in `CONTEXT.md`'s glossary and fixed it there too, noting that "one artifact saying it and the other not is how a false claim survives being fixed". That is the difference between fixing a doc and fixing a claim.

**It noticed a second thing that had quietly become false.** The `--promotion` refusal on `index` justified itself by saying the receiving half "schedules no rebuild, so nothing is ever added beside its live fold" — true when written, false as of this change. The flag is still refused, but the message now says what it is actually about (this command takes no promotion input, and whether it should is unanswered), which keeps ADR-0048's direction where refusal-to-optional is additive. Catching the justification you just invalidated is exactly what I want from a builder.

**One user-visible addition beyond the criteria, and it is the right call.** A receiver that comes up holding a successor now says so in its startup lines, names the canonical generation still answering, and states that the successor advances only from what a sender pushes for its own `{source, config}`. The reason it matters: this half has no fetcher, so a successor it cannot feed would otherwise sit at nothing with the refusal visible only on the SENDER. A process holding no successor stays silent, so the line means something when it appears. It is disclosed in the changeset, which is the ratification surface.

**`deps.rebuildIntervalSeconds` with deliberately no flag** — "turning it off is turning off processor upgrades on this half" — is the right instinct about which knobs belong to operators and which to tests.

Merging. Two remain: `a-restarted-deployment-hands-over-the-write-duty-it-cannot-discharge` and `an-hmr-update-reconfigures-the-tab-it-is-running-in`.

PR #177, opened 2026-09-19
feat(an-hmr-update-reconfigures-the-tab-it-is-running-in): An HMR updat…

wighawag:
Task: `work/tasks/done/an-hmr-update-reconfigures-the-tab-it-is-running-in.md`

All checks green: `pnpm format:check`, `check:adr`, `check:refs`, `check:changesets`, `pnpm build`, `pnpm typecheck`, `pnpm test` (browser 371, core 1185, server 342, whole monorepo clean). `pnpm changeset status --since=main` reports "no changesets found" only because the changeset file is still untracked; it names exactly the three changed packages, which the changeset covers.

## Drift check (done first)

The task is a launch snapshot, and I checked it before building:

- Its superseded paragraph is already superseded *in the task body* by the conductor's 2026-09-17 re-order, so there is no stale premise to build on.
- `blockedBy: a-module-handed-to-a-tab-is-identified-by-its-handler-sources` **has landed** (`work/tasks/done/`), and `moduleProcessorIdentity` (`packages/browser/src/moduleIdentity.ts`) derives a module arrival's identity from its handler sources at `init`, `addGeneration` and `updateProcessor` alike. So a handler edit really does move the identity and `unchanged` is rare and true, exactly as the re-order promised.
- `the-chain-facing-container-holds-its-generations-in-slots` has landed, so the `successor` slot bounds a burst.
- ADR-0086 is consistent with what I built; ADR-0085's "the browser needs none of this" section is the decision implemented.

No drift. Built.

## What I did

**`@etherfold/browser`: `reconfigureFromHotUpdate(indexer, {createState, createProcessor}, processorConfig?)`** (`packages/browser/src/hotUpdate.ts`, exported from `src/index.ts`). A thin adapter in front of `addGeneration`: it registers the handed-over processor as a **successor beside** the live generation (never tears down and rebuilds), and answers `ReconfigureReport`.

**`@etherfold/core`: `ReconfigureReport` moved here** (`packages/core/src/arrival.ts`), re-exported unchanged from `@etherfold/server` (same name, same place in the package's exports, same JSON body on the admin route) and from `@etherfold/browser`. Core is the only package every arrival already depends on.

**No `import.meta.hot` anywhere in the package.** The library exposes the API; the application calls it from its own handler. Asserted over a real minified bundle, and asserted a second way: an app that never calls it tree-shakes it out entirely.

**Tests** (`packages/browser/test/anHmrUpdateReconfiguresTheTab.test.ts`, 7 cases, written before the production code and driving the handover directly): an edited handler registers a successor with the incumbent answering complete reads throughout and **not one further range asked of the chain**; five saves leave the incumbent plus ONE successor; a `createProcessor` that throws answers `failed` and leaves the generations, the pointer and the cursor identical and still folding, and the next good save lands; a save that changed nothing answers `unchanged` naming the generation it landed on; the three outcomes are distinguishable on one field; the bundle carries no HMR global and is tree-shaken out.

**Docs**: a worked `import.meta.hot.accept(...)` call site with all three branches in `docs/guide/indexing-in-a-browser-app/index.md` and in `packages/browser/README.md`; the **generation** entry in `CONTEXT.md` now names the second arrival, why the browser needs none of the endpoint's machinery, and where the one contract lives. Changeset added.

**ADR status lines, checked rather than assumed.** ADR-0084 **already carries plain `status: accepted`** — an earlier task of its family removed the line. ADR-0085 **keeps `accepted, not yet implemented`, and I did not remove it**: its pushed-bytes arrival is decided-and-deferred and is still unbuilt, carried by the untasked `work/specs/ready/a-processor-artifact-is-pushed-to-a-running-deployment.md`. This is not the last piece of that ADR's family.

**Out of scope, noted not fixed**: `work/notes/observations/two-doc-sites-still-describe-the-deleted-declared-identity-or-a-shared-generation-store.md` — the browser README still says an omitted `processorIdentity` falls back to the author-declared one (deleted by ADR-0086), and the guide's pre-existing `addGeneration` snippet opens the successor's store with the default `databaseName`, which is the canonical generation's database.

## Decisions

**The API is a FREE FUNCTION, `reconfigureFromHotUpdate(indexer, {createState, createProcessor}, processorConfig?)`, and not a method on the hook.** A method would be reachable from the indexer object every app holds, so every production bundle would retain it; a free function goes out with the app's own `if (import.meta.hot)` block, which is what makes "a production build carries no dev-only machinery" a fact rather than a promise. I measured both (the test asserts the arrival is absent from a bundle of an app that only calls `createIndexerState`). It also reaches for nothing private, so it is genuinely an adapter over `addGeneration` rather than privileged access. Alternatives: a method on `createIndexerState(...)`'s return (discoverable, but always shipped), or a second factory wrapping the hook (a second way to build an indexer with no rule for choosing). **Touches:** nothing today; a worker-hosted app cannot use it, because a module object cannot cross a `MessagePort` — the arrival belongs wherever the container is built.

**The name.** `reconfigureFromHotUpdate` against the glossary: *reconfigure* already means "register a generation beside the live one, which is not an outage" (`addGeneration`'s own JSDoc, and the **generation** entry), and *hot update* is already the repo's word for this arrival (ADR-0086: "a no-op hot update"). It re-means nothing and duplicates nothing: `updateProcessor` stays the IN-PLACE verb and this is the beside-the-incumbent one. Alternatives considered and rejected: `applyHotUpdate` (says nothing about registering beside), `handOverProcessor` (invents a verb), anything with `hmr` in it (a bundler's acronym in a published API).

**The three outcomes reuse `ReconfigureReport`, and the type MOVED from `@etherfold/server` to `@etherfold/core`.** The AC asks for "the SAME shape the reconfigure endpoint uses", and a structurally-identical duplicate is three contracts that agree on the day they are written. The browser cannot depend on `@etherfold/server` (hono, `remote-sql`, `node:*`), and core is the only package every arrival already depends on — the same argument `core/src/index.ts` already makes for the reorg keys and the emission-stream port. The full rationale moved with it; `registry.ts` keeps a short note saying why it re-exports. Alternative: declare a browser-local twin, rejected as the drift this type exists to prevent. **Touches:** `@etherfold/server` (declaration only; every existing import, including the CLI's and the admin route's, is unchanged) and the future pushed-bytes arrival, which should answer this same type.

**`unchanged` is decided by ADD-THEN-COMPARE, not by derive-then-compare.** The identity of a module arrival is only knowable after `createProcessor` has returned, and `Indexer.add` already resolves a generation it holds rather than building a second engine over its state — so the honest question is "is the record that came back one I already had", which is what the function asks. Deriving the identity first would mean calling `createState` and `createProcessor` myself and then calling `add` with the same factories, i.e. building everything twice. The cost, stated: an `unchanged` still runs the app's `createState`, so a factory that creates a database creates one. **Touches:** nothing; `displacedBySuccessor` already declines to displace anything when the arriving generation occupies a slot, so an `unchanged` drops nothing.

**There is deliberately NO `{force}` on this call, and the message says where force lives.** Forcing means registering a generation beside one of the same name, and the name *is* the generation — it is not expressible on the beside-the-incumbent path. The `unchanged` message therefore names the one case that can still surprise a developer (a change the handler text does not carry: an imported helper, an entity declaration, a captured value) and points at `updateProcessor(next, {force: true})`, which costs the rebuild this call exists to avoid. Alternative: accept `{force}` and quietly ignore it, or synthesise a distinct identity — the second is the author-declared identity ADR-0086 deletes, wearing a flag. **Touches:** `updateProcessor`, which keeps `{force}` and is now documented as the escape hatch for this arrival too.

**The successor's store stays the APPLICATION's obligation, stated at the seam rather than solved.** A successor folds beside an incumbent that goes on writing, so it needs its own storage, and `GenerationContext` carries only `stream` — a processor change keeps the stream, so a `databaseName` keyed on the context alone collides. I did not widen the context or invent a default: the storage decision is the app's (ADR-0077), and `addGeneration` already has this shape and is already used this way by the existing suites. Instead the obligation is stated on `HotUpdateGeneration`, in both doc examples, and in the test fixture. **Touches:** the doc sites in the observation note above, which show the collision and predate this task.

**What the tab reports for each outcome is the app's, and the examples show one rendering.** `registered` names the generation now folding beside the live one; `unchanged` says plainly that the handler sources are the ones already running and that the warm fold was kept; `failed` carries the underlying message verbatim. The library logs the same three through `named-logs` in the words the CLI's re-read uses, so a browser console and a server log read alike.

**The reference app (`examples/browser-reference`) was NOT switched over.** Its axis-one block still calls `updateProcessor`, which remains correct and is the in-place verb; switching it would mean giving it a per-save store and is a change to the reference wiring rather than to this seam. The AC asks for a doc example and there are now two (package README and guide). **Touches:** a later task that wants the reference app to demonstrate the warm path.


wighawag:

The third arrival from ADR-0085, and the last task this family was waiting on.

- [x] **An HMR update carrying a changed processor registers a successor with no page reload, and the incumbent keeps answering throughout.**
- [x] **Repeated updates stay bounded:** a burst of saves leaves the incumbent plus ONE successor, which is the durable-slot behaviour asserted end to end here rather than assumed.
- [x] **An update whose processor THROWS leaves the indexer exactly as it was** — generations, pointer and fold all untouched — and reports the failure. Nothing partial is registered.
- [x] **An update that changes no identity is a successful no-op that SAYS so,** names the generation it landed on, and keeps the warm fold. There is a further case asserting a registration, a no-op and a failure are told apart on ONE field, which is the "three indistinguishable causes" problem this outcome exists to remove.
- [x] **The three outcomes are reported in the SAME shape as the reconfigure endpoint, and it got there the right way.** `ReconfigureReport` is HOISTED into `packages/core/src/arrival.ts` and `@etherfold/server` now imports it instead of declaring its own. That is the criterion's "one contract rather than three" satisfied by de-duplication rather than by a second copy that agrees today. It is why this PR legitimately touches `server`, and it is the only reason it does.
- [x] **The API is usable from an app's own `import.meta.hot.accept(...)` handler, demonstrated in a doc example** — in the browser guide and in `examples/browser-reference`, passing the module that handler receives.

**The no-bundler-global criterion is proved mechanically, not asserted.** Every `import.meta.hot` occurrence in `packages/browser/src` is prose or a JSDoc `@example`; stripping comments leaves zero executable references. That literal reading also has to be the right one, because a sibling criterion REQUIRES documenting the `import.meta.hot.accept` call site — so the two together mean "no code dependency", which is what "unaffected by construction rather than by a guard" is actually about.

Then it went further than the criterion asked, twice, and these are the two tests I would have wanted most:

- it **bundles the package** and asserts the output carries no reference to a bundler HMR global — so the claim is checked against the artifact rather than against the source;
- it asserts the machinery is **tree-shaken out of an application that never calls it** — so the dev-only path demonstrably does not ship to production.

A grep proves the source is clean today; those two prove the property a deployment actually cares about, and they will keep proving it.

**Two documentation defects found in adjacent files and correctly left alone.** `work/notes/observations/two-doc-sites-still-describe-the-deleted-declared-identity-or-a-shared-generation-store.md` records that `packages/browser/README.md` still says leaving `processorIdentity` off keeps "the author-declared identity the processor computes from its `version`" — which no longer exists after #170 — and, more seriously, that a guide snippet passes `createBrowserStateStore(next.entities)` with no `databaseName`, so the successor would share `etherfold-state` with the canonical generation: one store, colliding rows and sync cursor, one of the two demoted by the writer claim. That second one is a correctness bug in a snippet a reader would copy. Both are captured rather than fixed in scope, which is right; I am routing them in the triage below.

Merging. That is every agent-buildable task on the board.

PR #178, opened 2026-09-19
feat(run-and-build-drive-the-folds-they-hold-rather-than-one-captured-r…

wighawag:
Task: `work/tasks/done/run-and-build-drive-the-folds-they-hold-rather-than-one-captured-receiver.md`

## What I built

**Drift check first.** I re-took the measurement rather than trusting the task body, and it held: a re-run `build` over the same handle with changed processor bytes registers a successor into the `successor` slot, folds it to the tip, and exits with `canonical` still naming the *old* generation. `container.ingestion` does *not* throw today (the opening fold is never a follower because `add` reads an in-memory fold list that is empty at `open`), which is exactly the "the assembly rests on a bug" the task describes. Nothing in the task body was contradicted.

**1. `createDirectIngestion` resolves per ask** (`packages/core/src/directIngestion.ts`). Its parameter is widened to `LogIngestion | LiveIngestions`, where `LiveIngestions = () => Promise<readonly LogIngestion[]>` — the same shape `IndexerRegistryEntry.liveIngestions` already has on the HTTP side. On the resolving arm it selects with core's own `sameWireContext`, exactly as the ingest route does. The single-receiver arm is byte-for-byte unchanged.

**2. `NoLiveReceiverError`** (`packages/core/src/errors.ts`), the in-process twin of the route's `400 context-mismatch`, naming every live context rather than picking one.

**3. The CLI assembly stops reading a getter that throws** (`folding.ts`, `index.ts`, `run.ts`, `indexCommand.ts`). `openFolding` reads the receiver off `container.held()[0]`, where it is already optional; `prepareIndexing` feeds `createDirectIngestion(() => container.liveIngestions())`. `streamBuilder` on `FoldingAssembly` / `PreparedIndexing` / `RunningIndexer` / `RunningReceiver` is now optional and is a *report* rather than the thing anything feeds.

**4. `build` settles before it exits** (`driveCycles`). One `container.rebuildMore()` after the loop, beside `pruneHeldUntilComplete` and under the same `stopAtTip && !deps.signal?.aborted` guard. `run`'s `betweenCycles` block is untouched.

**5. Docstrings corrected**, not made true by refusing (see Decisions).

**Tests.** `packages/core/test/directIngestion.test.ts` (+5 cases: selection, a live set that moved, both refusal shapes, the one-receiver arm unchanged); `packages/cli/test/aRerunBuildSettlesItsPointer.test.ts` (new, at the seam the task named: a real handle, a bundle edited on disk, read back through the canonical pointer after exit); `packages/cli/test/theWireFollowsTheFoldsHeld.test.ts` (new). I verified the two settle tests go **red** with the `rebuildMore()` call removed, and the stopped-from-outside test goes **red** with the `!deps.signal?.aborted` guard removed.

Gate: `build` + `typecheck` + `test` (all packages, platforms, examples) green; `format:check`, `check:adr`, `check:refs`, `check:changesets` green. Changeset added.

## Decisions

**The shape of the resolved ingestion target: the existing helper is WIDENED, and it resolves per ask.** `createDirectIngestion` takes `LogIngestion | LiveIngestions` rather than gaining a sibling. Alternatives considered: a sibling `createContainerIngestion(container)`, and pushing the resolution into the CLI. Rejected both on coherence — there is ONE concept here ("the ADR-0004 wire, with no wire") and a second exported name for it is a second thing to keep in step, exactly the drift ADR-0044 cites when it refuses a second read-only implementation of one seam. The resolver type deliberately reuses the server's existing vocabulary (`liveIngestions`) instead of coining a new one. Touches: `@etherfold/core`'s public API (additive), and `prepareIndexing`, which is the only caller that passes the new arm. `index` still passes `container.liveIngestions` to the route and is unaffected.

**What a process with NO live receiver answers: a refusal, never a number — and `retryable` is a fact about the instance.** Live receivers exist and none matches → `retryable: false` (a misconfiguration, the same family as `WireContextMismatchError`). Nothing live at all → `retryable: true`. Alternatives considered: (a) answering a block number, rejected because that is a fetcher fetching ranges into nothing, i.e. the silent stall; (b) always non-retryable, rejected because it would kill a restarted `run` in precisely the window ADR-0044's second amendment says is recoverable ("a survivor that is behind is left following, the stream is unfed for the moment, and a line says so") — and it would make the *next* task's 17 failing CLI tests fail harder rather than green, since a follower-only restart is supposed to rebuild off the stored stream and finish the upgrade. Touches: the `follows`-derivation task, which is what first makes this state reachable.

**A `build` does not wait out that retryable refusal; it re-throws.** `run` retries it forever, correctly, because its gap between cycles holds the rebuild that closes the state. `build` has no such gap (its rebuild is a single step *after* the loop), so retrying is a guaranteed hang. It is re-thrown beside the `fatal` re-throw, ahead of the settle and the prune, so a one-shot that fetched nothing exits non-zero rather than dressing up an empty run. Alternative considered: abort and exit 0, rejected as a false success. This deliberately does **not** pre-decide what a follower-only `build` *should* eventually do (probably rebuild to completion and exit 0) — that belongs to the task that makes the state reachable, and changing this is one branch.

**What `build`'s settle promises, and what it refuses.** It advances every follower it holds by ONE bounded chunk and settles the pointer ONCE. It does **not** promise a successor reaches level, and must not: a settle that waited would make the one-shot unbounded, which is the one thing a command whose exit is the point may never be. It is also skipped entirely when the build is stopped from outside, on the same rule the retention pass already states. In the case this task is about — a re-run with changed bytes — the successor is fed by the wire and is level by the time the loop ends, so the settle finds it.

**The docstring is CORRECTED, not made true by refusing.** Refusing to register the successor would be a new refusal an operator meets on the ordinary redeploy, where the honest behaviour (fold the new generation and publish it) costs nothing and is what the generation model exists for. This touches one adjacent user-visible string: `NEVER_PROMOTES_BUILD` in `config.ts` justified refusing `--promotion` with "it never promotes and never moves the canonical pointer on its own", which is now false. I kept the refusal (accepting the flag is a config change outside this task) and rewrote its *reason* in the exact shape `NEVER_PROMOTES_INDEX` was rewritten for the same reason: it now says the command takes no input and settles its one successor under the default policy. `configuration.test.ts` pinned the phrase "never promotes"; that assertion now pins "takes no promotion input".


wighawag:

Read the diff against the task's nine acceptance criteria. All met, and the two that were most likely to be fudged were not.

**The forward-pointer was honoured.** The task said "nothing here changes which generation is a FOLLOWER or which one WRITES; a diff that touches `follows` or `writesStream` derivation is out of scope". `git diff` against `packages/core/src/{receivingContainer,container,indexer}.ts` is **empty**. That was the one boundary that keeps this task an expand step and the next one a migrate step, and it held.

**The drift check was re-taken rather than trusted.** The report confirms `container.ingestion` does not throw *today* and says why (the in-memory fold list is empty at `open`), which is exactly the "the assembly rests on a bug" framing rather than a restatement of it.

Per criterion:

1. Assembly no longer reads a getter that can throw — `openFolding` reads the receiver off `container.held()[0]` where it is already optional; `streamBuilder` becomes an optional REPORT on all four assembly types rather than the thing anything feeds. ✅
2. A live ingestion that changes is followed — `createDirectIngestion` widened to `LogIngestion | LiveIngestions`, resolving per ask with core's own `sameWireContext`, and `theWireFollowsTheFoldsHeld.test.ts` asserts it rather than leaving it true by construction. ✅
3. / 4. `build` settles before exit — one `container.rebuildMore()` after the loop, and `aRerunBuildSettlesItsPointer.test.ts` asserts it end to end over a real handle with a bundle edited on disk, read back through the canonical pointer after exit. The builder verified both settle tests go **red** with the call removed; that is the check that separates a real test from a passing one. ✅
5. Still stops at the tip, retention pass intact, and a `build` stopped from outside skips the settle under the same `!deps.signal?.aborted` guard as the prune — also verified red with the guard removed. ✅
6. `run` unchanged — `betweenCycles` untouched; the settle is in the `stopAtTip` exit path only. ✅
7. Docstring CORRECTED rather than made true by refusing, with the reasoning stated, and the one adjacent user-visible string it falsified (`NEVER_PROMOTES_BUILD`) was caught and rewritten rather than left. Catching that was not asked for. ✅
9. / 10. Tests mirror repo style; changeset present and argues the change rather than announcing it. ✅

The `NoLiveReceiverError` retryable split (misconfiguration → not retryable; nothing live at all → retryable) is the judgement call in here. It is decided deliberately, argued from ADR-0044's second amendment, and explicitly scoped so it does not pre-decide what a follower-only `build` should eventually do. That is the right disposition for a state the NEXT task makes reachable.

Gate-1 green across all packages, platforms and examples. Merging.

PR #179, opened 2026-09-19
feat(whoever-fetches-a-stream-writes-it-and-the-stream-outlives-every-f…

wighawag:
Task: `work/tasks/done/whoever-fetches-a-stream-writes-it-and-the-stream-outlives-every-fold.md`

Everything is green: build, typecheck, format, all 18 test packages, and the four `check:*` gates. The working tree holds only this task's changes.

## What I built

**Half one: the write duty and the fetch come off the generation.**

`StreamWriter` (`packages/core/src/stream/writer.ts`) is new and is the one writer of one stream. `ReceivingIndexer` holds one per stream it holds a fold on; it is what `{source, config}` resolves to on the wire, and it answers `expectedFromBlock` from the **stream's own coverage claim** read back through a new `StreamCursorSource` port (`streamCursorSourceOn`, `@etherfold/server`), with the reorg window *walked* out of the stream's own tail (`windowOfStoredStream`). `writesStream`, the per-fold `ingestion`, `follows`, `reconcileWriters` and `handOverTheWire` are deleted; a fold is one shape (a `GenerationRebuild`) and never holds an appender. After the append the writer *offers* the delta to each fold on that stream (`GenerationRebuild.follow`), which is one code path with the scheduled chunk.

**Half two: a stream outlives every fold over it.** `deleteGeneration` takes `{reapStream}`, false by default; `reclaim` is the one caller that asks. The registry now keeps a durable set of stream digests (`keptStreams`, `_generation_streams` on SQL, a `kept` level on IndexedDB) that the open-time sweep compares against, so a kept stream survives a restart while a pre-generation orphan is still collected.

Two corrections ride along: the append-only path gained the hole guard (`StreamHoleError`, permissive on an absent stream), and a latent replay bug — a catch-up re-applying a branch it had already retracted, measured as `UNIQUE constraint failed: _blocks.number` — is fixed without changing what a fresh rebuild replays.

## Decisions

**What drives the fetch, and where its position comes from.** The `ReceivingIndexer` holds one `StreamWriter` per stream it holds a fold on, built with the first fold on that stream and never rebuilt. Its position is `getFromBlock` over the stream's stored coverage claim, and its unconfirmed window is walked out of the stream's own tail rather than borrowed from a fold — so ADR-0038's arbiter moved *with* the duty and an empty-state successor cannot drag the fetch backwards. Alternative considered and rejected: keeping the fetch on a fold and only moving the appender, which is the hand-over ADR-0087 measured as duplicating or holing. Touches: `ReceivingIndexerOptions` now needs `streamCursor` beside `appendEmissions` and `replay`, and a container missing any of the three is refused at `open` (`refuseContainerThatCannotFetch` / `refuseFoldWithNoStream`) rather than folding for ever on stale history — which is the failure ADR-0087's second amendment measured.

**A fold advances two ways through one implementation.** After the append the writer offers the delta to every fold on that stream; a fold that is level applies it there and then, one that is behind declines and is carried by its bounded rebuild. I did this rather than making `rebuildMore` the only path because *every* host would otherwise have to schedule a rebuild to fold anything at all, and the server package schedules none — an HTTP-only deployment would have ingested and never folded. Per-fold advances are serialised through one promise chain (`ReceivingIndexer.advance`), because a live delta and a scheduled chunk both decide from the same durable checkpoint and an overlap would meet the storage seam's duplicate-height refusal.

**`writerOf` survives, renamed.** It is now `fetcherOf` on the module and on `GenerationRegistry`, with the derivation unchanged (oldest surviving generation on a stream, registration order, stored nowhere). The rename is what makes "no caller reads it as a duty" structural rather than a comment. Alternative: keep the name with corrected docs — rejected, because the chain-facing container still derives `follows` from it and a reader would have had to hold two meanings of one word.

**The chain-facing `Indexer` (browser) is deliberately NOT restructured, and it still derives `follows` from `fetcherOf`.** There the thing that fetches a stream *is* a generation (`IndexerGeneration` opens `load()` with `eth_chainId`; a batch is a side effect of it advancing), so "which generation fetched this" and "which generation writes it" are one fact, which is ADR-0044's follower rule and is untouched. ADR-0087's Consequences name `run` and `build` as the restructure and say nothing about the browser engine. **A reviewer should know the cost:** a browser tab reloaded with a changed processor is in the same shape the receiving side was — the elected fetcher is a generation the tab holds no fold for — so the same class of stall is reachable there. Splitting `IndexerGeneration`'s fetch from its fold is a separate change nothing has asked for, and I did not smuggle it in.

**How a kept stream is told from a pre-generation orphan across a restart.** A durable record in the registry's own records (`GenerationRegistryState.keptStreams`; `_generation_streams` on SQL, a `kept` key level on IndexedDB, a set in the memory port), written when a generation is registered on a stream and removed only by an asked-for deletion. The sweep compares a stored subtree against the generations *and* that set. Alternatives considered: inferring it from the presence of a `_stream_coverage` row (substrate-specific, implicit, and it would make the sweep collect nothing on SQL), and dropping the sweep (trades away its own stated purpose). Deliberately **not** counted by `maxStreams`, which goes on counting distinct streams among registered generations, so keeping a stream brings no deployment closer to a refusal. Touches: the schema (a new fixed table in `db.sql`, no version bump — nothing is published) and every `GenerationRegistryPort` implementation.

**Every deletion path, enumerated and decided.**
1. `replaceTheSuccessor` → `dropReplaced` (registering into an occupied `successor` slot): row + state namespace, **no reap**. This is the path the ADR names.
2. `dropSuperseded` (drop-on-promotion): row + state namespace, **no reap**. Also automatic, also nobody asked.
3. `reclaim` (operator verb): row + state namespace + **reap** when no generation is left folding the stream. The one caller that asks.
4. `registry.deleteStream` (operator verb): unchanged, guard intact — **widened** to accept a stream the registry records with no generation on it, because otherwise exactly the streams the keep exists for would be undeletable. `UnknownStreamError` now means "this indexer has no such stream at all".
5. The **sweep on registry open**: keeps its purpose, now compares against the stream records too.
6. Generation caps: refuse, never evict — untouched.
7. The chain-facing `Indexer`'s own drops (`dropSuperseded`, `dropReplaced`): they call the same `deleteGeneration`, so they stopped reaping too, which is consistent with the ADR and costs a browser tab disk it can reclaim by asking.

**A behaviour change on an opt-in flag, worth a reviewer's eye.** With no duty to strand and no reap, `dropOnPromotion` now really does drop the superseded generation on an ordinary processor upgrade, where the strand clause used to decline it. The flag does what it says; what the deployment keeps instead is the stream, which is the expensive half. `DeclinedReclaim` loses its `writes-a-followed-stream` reason (the per-generation `deletion-failed` decline is intact).

**A latent replay bug the change made systematic, fixed at the source.** `generateStreamFromReplay` de-duplicates a re-offered block by window membership, which cannot answer "did this fold already hold it and take it back" — so a catch-up reaching back over its own reorg window re-applied the dead branch at a height its replacement occupies (`UNIQUE constraint failed: _blocks.number`). A block applied *and* retracted within one walk, at or below where the fold has already folded through, is now delivered neither way; the `lastToBlock` half of that condition is what keeps a rebuild from a fresh cursor reproducing the live run exactly (ADR-0042, asserted by `replay.test.ts`). A second short-circuit in `GenerationRebuild.more` skips a chunk when the stream has not moved.

**One narrow tolerance in `GenerationRebuild.follow`.** A live delta whose events carry no raw half is folded as delivered instead of being refused, because `reparse` needs `topics`/`data`/`address` and a capture predating the stream-seam narrowing has none (`BASE_ABANDONED` in `@etherfold/conformance-workload-stratagems`). The delta a deployment just derived is already decoded under the source it is running now. The asymmetry is deliberate and loud: the same rows read back off *disk* have no decoded half to fall back on, so `more()` still reports `undecodable`.

**The eleven tests, each re-scoped.** Five in `receivingContainer.test.ts`: the container now gets a stored stream and the fixture gets real topics (a decoded-only log is one no fold can replay); the `writesStream`/`follows` assertions became "neither fold has either, and the live context is the stream's writer"; the "second fold with no `replay`" refusal became a refusal at `open` (the asymmetry went with the election) plus a new case that a container which could fetch nothing is refused; and **`expectedFromBlock` on a successor over a written database changed from `START_BLOCK` to the stream's resume point — that number is the change**. Six elsewhere: `aChangedContextCreatesASuccessor` (3 cases — the successor no longer asks for a range the stream holds, and it catches up by re-folding; its fixture gained real topics and a `replay` source), `aRerunBuildSettlesItsPointer` (1 — a build stopped from outside no longer leaves a level successor, because the exit work is what carries it; the claim it exists for, that the pointer did not move, is unchanged; the other two cases stayed green), `aSuccessorLandsInADurableSlot` (2 — the replaced successor's stream is kept, and the strand decline became "it is dropped and the stream survives"), `anIndexProcessAdvancesItsSuccessor` (1 — the case that relied on the wire to advance a restarted successor now asserts the stronger thing: with no tick the successor does not advance *at all*, and the tick is what does both). Beyond the eleven: `writerSuccession.test.ts` is re-scoped whole and kept at its path, because it asserts the same *situation* with the opposite answer — deleting a generation hands over nothing, because nothing was held; `aGenerationNoSlotNamesIsReclaimed`, `theRebuildReplaysTheLocalStream`, `aRebuildInProgressIsVisible`, `equivalence`, `entityStore`, `theCanonicalPointerMovesBack`, `theReceivingContainerPublishesWhatItApplied` and `rebuild` were adjusted where `rebuildMore` now reports one entry per fold rather than per follower. New: `packages/cli/test/aRestartReFoldsTheStoredStream.test.ts` (the methods asked, the ranges asked for, the row count, and two folds at once) and `packages/core/test/streamWriter.test.ts` (the position and the hole guard), plus keep/sweep cases on both durable registries.

**`ADR-0087` is marked implemented and carries a fourth amendment** recording the four things it did not say: the sweep, `deleteStream`'s widening, the declines that dissolve with `follows`, and the replay bug. No new ADR: every decision above is this ADR's substance or a correction to it, and a second number would fork the record.


wighawag:
BLOCK — Gate-3 (conductor's diff-vs-criteria pass)

One blocking finding, narrow and precisely fixable. The rest of this PR is excellent and I am not asking for any of it to change.

## The finding: ADR-0087's status line is not this task's to expire, and the value invented is not a status this repo has

The diff changes ADR-0087's frontmatter from `status: accepted, not yet implemented` to `status: accepted, implemented`. Three things are wrong with that, and the third is the one the gate could never catch.

**1. It belongs to the closing task, which is blocked behind two others precisely so that it is last.** `the-stream-writer-vocabulary-is-documented-and-adr-0087-expires` says "THIS IS THAT TASK" and carries the criterion. It is `blockedBy` every other leaf of the family for the reason `work/protocol/ADR-FORMAT.md` argues at length: "every task in a chain can see that it is not the last one while the actual last one has no way to know that it is". ADR-0006 and ADR-0073 both went wrong that way. This is the same hand-off failure in the opposite direction — a task that is not last acting as though it were.

**2. The claim is not true yet.** Two tasks remain. `a-restarted-deployment-hands-over-the-write-duty-it-cannot-discharge` still has to assert the original defect end to end on its own measured numbers, and the vocabulary pass has not happened: `CONTEXT.md` still teaches the retired rule outside the entries this PR touched. The status line is a claim about the CODE and the corpus, and right now the ADR would say "implemented" while the glossary teaches the rule it replaced.

**3. `accepted, implemented` is not in the vocabulary at all.** `ADR-FORMAT.md` lists exactly seven status values and that is not one of them. Its rule is the opposite: "**ABSENT means accepted and current**, which is what most ADRs are, so add one only when the plain reading would MISLEAD." So the correct end state, when the family really is done, is **no status line** — not a new one. `check:adr` only checks for duplicate numbers, so nothing in the gate can see this; it would have quietly established an eighth status value that nobody legislated and that every future reader would have to reconcile against the table.

**The fix is to restore `status: accepted, not yet implemented`** and leave its removal to the closing task, which will delete the line outright rather than replace it.

## Everything else checks out

Verified against the diff rather than the report:

- `writesStream`, `reconcileWriters` and `handOverTheWire` are gone from `packages/*/src` — zero occurrences.
- The fetch genuinely moved: `StreamWriter` answers `expectedFromBlock` from the stream's own coverage claim with the reorg window walked out of the stream's tail, so ADR-0038's arbiter moved *with* the duty and an empty-state successor cannot drag the fetch backwards. That is the thing the previous attempt proved cannot be skipped.
- A container that cannot fetch is refused at `open` rather than folding for ever on stale history — which is exactly the `["eth_chainId"]` failure the ADR's second amendment records, closed structurally.
- All seven deletion paths enumerated and decided individually, including the open-time sweep the ADR never mentioned, with `keptStreams` making a kept stream durable across a restart and leaving the sweep's own purpose intact. `maxStreams` deliberately not charged for a kept stream.
- `deleteStream` widened so the streams the keep exists for are not undeletable — a consequence the task did not anticipate and that would have been a trap.
- All eleven tests re-scoped individually with the reasoning given, plus `writerSuccession.test.ts` kept at its path asserting the same situation with the opposite answer.
- The `expectedFromBlock` change from `START_BLOCK` to the stream's resume point is correctly identified as *the* change rather than a test edit.
- `CONTEXT.md` updated for this task's own concepts; changeset present.

Two things I want to credit explicitly. Declining to restructure the chain-facing browser `Indexer`, **and saying out loud that the same class of stall is reachable there**, is the right call stated the right way — the honest scope boundary plus the cost, rather than a silent omission or a smuggled fix. And the latent replay bug (`UNIQUE constraint failed: _blocks.number`) was found, measured, and fixed at its source with the `lastToBlock` condition that keeps ADR-0042's exact-reproduction property intact.

Restore the two frontmatter lines and this is an approve.

wighawag:
APPROVE ✅ — Gate-3, block cleared

The single blocking finding is resolved on this branch (`6580ddc2`): ADR-0087's frontmatter is back to `status: accepted, not yet implemented`, and its removal is left to `the-stream-writer-vocabulary-is-documented-and-adr-0087-expires`, which will delete the line outright rather than replace it — ABSENT being what `ADR-FORMAT.md` means by implemented.

I checked the amendment prose as well: nothing else in the document claimed implementation, so the one-line revert leaves it self-consistent. `check:adr`, `check:refs`, `check:graph` and `format:check` all green on the branch head; the change is one markdown frontmatter line and cannot reach build, typecheck or test.

Everything else stands as verified in the block comment above — the fetch genuinely moved off the generation, the `["eth_chainId"]` death is closed structurally by refusing at `open`, all seven deletion paths were decided individually including the sweep the ADR never named, `keptStreams` makes the keep survive a restart, and all eleven tests were re-scoped with reasoning rather than softened.

Merging.

PR #180, opened 2026-09-20
feat(a-restarted-deployment-hands-over-the-write-duty-it-cannot-dischar…

wighawag:
Task: `work/tasks/done/a-restarted-deployment-hands-over-the-write-duty-it-cannot-discharge.md`

Everything is green: `pnpm format:check`, `check:adr`, `check:refs`, `check:graph`, `check:changesets`, `changeset status --since=main` (verified through a throwaway index so the real one was untouched), `pnpm build`, `pnpm typecheck`, and the full `pnpm test` (18 suites, exit 0).

## What I found first

The task is a fan-in and it had drifted in exactly one direction: **its three blockers landed, and the closing work is smaller than the body implies but is NOT vacuous.** I checked each criterion against the tree rather than rebuilding it.

Already covered before I touched anything:

- **Re-fold instead of re-fetch, and the wire assertion** — `packages/cli/test/aRestartReFoldsTheStoredStream.test.ts` (methods asked, ranges asked for, two folds at once). That file's restart runs against a chain that has NOT moved on, so it can assert "nothing twice" but structurally cannot assert **growth**. That is the gap this task owns.
- **The dissolution as behaviour** — `packages/core/test/rebuild.test.ts` "DROPS the superseded generation on promotion and KEEPS its stream", plus `aSuccessorLandsInADurableSlot` and `aGenerationNoSlotNamesIsReclaimed`.
- **The citations** (criterion 8) — `docs/adr/0087-…md:11` and `work/specs/ready/a-generation-retains-the-code-that-folds-it.md:88` both name this task **by slug** with the reason stated inline. The trap is still removed; no new path citation has appeared (`check:refs` green, 281 artifacts indexed). The only path citation to this item anywhere is to `work/questions/task-…md`, which does not move on a done-move.

## What I built

**The acceptance test** (`packages/cli/test/aRestartedDeploymentGoesOnAppending.test.ts`, 8 cases). The prior attempt's branch already carried it and it is good work; I kept it, removed what must not ship, and strengthened two places. It stands a `run` up over a real libSQL handle, stops it, rewrites the bundle and re-runs over the **same** handle, and reads the stored **rows** (block, log index, transaction, `removed`/`alive`) rather than a count, because the rejected option also appended. It lands on the family's three numbers: `2` and frozen (the defect), `4 where 2 are correct` (the hand-over), and what it asserts, a stream that grows `2 → 4` with no log stored twice, the first run's rows still byte-identical, and the coverage claim's `startBlock` unmoved. It also covers the same-processor restart, the reconfigure path, an **unheld** canonical generation still answering from its table namespace and on the feed, `/status` reporting one entry per generation **held**, and the stream surviving a second restart that replaces the `successor` slot's occupant, down to the replacing fold re-folding bytes it never fetched.

**Two deletions from the prior attempt's branch.** It carried `NEGCTRL === 'no-append'` and `NEGCTRL === 'twice'` sabotage hooks in `packages/core/src/stream/writer.ts`. Those are reverted; `git diff main -- packages/core/src/stream/writer.ts` is empty. The negative control is expressed as the measurement instead: the rows **and** the ranges, in the same case.

**The residue, which is a real finding.** Criterion 6 says a residue of transferable duty means the criterion is not met. There is one, and it is precisely something the previous release note claimed: `.changeset/whoever-fetches-a-stream-writes-it.md` says "`DeclinedReclaim` loses `writes-a-followed-stream`", and the union still carried it, with JSDoc explaining the retired keep. Nothing can produce it (`reclaim` pushes only `deletion-failed`). Removed, and the operator-facing text that explained the vanished behaviour corrected with it: the reclaim endpoint's documented answers (`@etherfold/server/src/api/admin.ts`), the server README's description of the verb, `movePointer`'s claim that drop-on-promotion still declines the ordinary upgrade, `unslottedGenerations`' strand clause (receiving-side callers only), `/status`'s `follows` field on both sides of the seam, and the two comments still saying a deletion reaps the stream with the generation or that "the generation that INDEXES a stream appends to it". `CONTEXT.md`'s drop-on-promotion bullet contradicted its own generation-slot bullet ten lines later; it now states the rule per runtime.

A changeset accompanies it (`@etherfold/core` minor, `@etherfold/server` and `etherfold` patch).

## Decisions

**Which criteria were already covered, and by what.** Criterion 1's wire half and criterion 2's no-duplication half are covered by `packages/cli/test/aRestartReFoldsTheStoredStream.test.ts`; I did not copy them. What is NOT covered there is **growth**, because that file's restarted chain serves the same logs and the same tip, so its stream cannot grow and "appends" is unassertable in it. Criterion 6's behavioural half is covered by `packages/core/test/rebuild.test.ts` ("DROPS the superseded generation on promotion and KEEPS its stream") and `aSuccessorLandsInADurableSlot`; what was NOT covered was the residue in the type and the prose, which is this task's finding. Criterion 8 needed verification only, and it holds. Everything else (criteria 3, 4, 5, 7's `/status` half) had no end-to-end CLI assertion and is what the new file adds.

**The negative control is not re-run, and no sabotage hook ships.** The prior attempt proved the test's sensitivity by suppressing the append in `packages/core/src/stream/writer.ts`; that configuration grows without bound and OOM-killed two builds. I deleted both hooks rather than bounding them: a `globalThis`-reachable sabotage switch in a published package is a defect on its own. The alternative considered was an out-of-tree patch harness under `docs/spikes/`; rejected as manufacturing work, because the measurement the task actually asks for (the stored rows plus the recorded `eth_getLogs` ranges) is stronger than a suppression and the test already takes it. This touches `work/notes/observations/a-stream-writer-whose-append-silently-does-nothing-grows-without-bound.md`, which stays an open, unreachable-on-today's-code robustness finding and is not mine to chase.

**`DeclinedReclaim.reason` is narrowed to one member rather than left unreachable, and it stays a union.** Deleting `'writes-a-followed-stream'` is a breaking type change on a published package; keeping it would have been a reason an operator can read in the type and never receive, which is the class of thing criterion 7 exists to remove, and the previous release note already told readers it was gone. I kept the field a one-member union rather than collapsing it to a string literal in the report shape, so a second decline reason can be told apart without a shape change. Touches `@etherfold/server`'s `reclaim-generations` response, which carries `reason` verbatim, and the `whoever-fetches-a-stream-writes-it` changeset, whose claim this makes true.

**`unslottedGenerations`' JSDoc was corrected rather than left ambiguous, on a checked fact.** It lives in shared `@etherfold/core` and reads as though the strand clause still applies somewhere; its only callers are `ReceivingIndexer.reclaim` and the server's admin listing, both receiving-side, so the clause has no subject at all there. The chain-facing `Indexer`'s own strand clause in `container.ts` is UNTOUCHED and still live, because ADR-0087 deliberately did not restructure the browser engine (a generation still fetches there). `CONTEXT.md` now states that difference explicitly instead of stating the retired rule flatly.

**No new ADR.** Every correction above is ADR-0087's own stated substance catching up with the code; a second number would fork the record, which is the reasoning the blocker's builder gave and I see no reason to depart from it.

**One process note, not a design decision:** reverting the sabotage hook with `git checkout origin/main -- packages/core/src/stream/writer.ts` staged that one path. The file's content is identical to `main`, so the net diff is empty and the runner's `git add -A` sees the same tree either way; I did not stage, commit, move or push anything else.


wighawag:

Merging. Every acceptance criterion is met, the two claims most worth distrusting were checked against the tree rather than taken on the report's word, and the one red signal on this PR is inherited from `main` and demonstrably not this branch's.

## The red CI is not yours

`verify` fails on a **dead link**, in a file this branch does not touch:

```
Found dead link ./../../../packages/cli/README
  in docs/spikes/the-build-command-and-its-pinning-rule-are-documented/README.md
```

Traced rather than assumed. `382421fe` is the last green CI run on `main`; the next one, `a05ba3d5` — PR #173, which introduced that spike README — is the first failure, and every `main` commit since has been red. That is **two days and six merged PRs** before this one, and it reproduces on `main` at `a70e15bb`, which contains none of this diff. This PR touches no file under `docs/spikes/` and no `packages/cli/README.md`.

It went unnoticed because `dorfl.json`'s `verify` does not run `docs:build`, while the CI workflow does, so the acceptance gate is green on a tree CI rejects. Filed as an observation with the bisect; it is a pre-existing defect and a gate gap, not a reason to hold this task.

## Criteria

1. **Appends, end to end through the CLI over a real handle** ✅ — `aRestartedDeploymentGoesOnAppending.test.ts`, "APPENDS: the stored stream GROWS from where the previous deployment left it". A real libSQL handle, stopped, re-run over the *same* handle with a rewritten bundle. Asserts the stream moves off `2`, which is the number the defect froze.

2. **Nothing stored twice, on the rows, both halves together** ✅ — and this is the criterion the task warned could be faked, since the rejected hand-over also appends. It is asserted in **one case**, four ways: row identity is unique on the log's own identity rather than on a count; `after.slice(0, stored.length)` still equals the first run's rows *exactly*, so this is not a re-fetch from the start block that happens to agree; nothing is a retraction (`removed === 0 && alive === 1`), which catches the duplicate wearing a reorg's clothes; and on the **wire**, every recorded `eth_getLogs` range is `> START_BLOCK`, so no range was bought twice. That is the measurement this family holds itself to, not an "it appended" assertion.

3. **Incumbent-holding deployment and the reconfigure path unaffected** ✅ — a `describe` each, both asserting appends *and* no duplication.

4. **An unheld generation still answers reads (ADR-0053)** ✅ — asserts `canonicalOf` is the incumbent while `heldBy` does *not* contain it. Pinning `promotion: 'manual'` so the subject cannot stop being canonical mid-assertion is the right call and is reasoned in the test.

5. **The stream survives restart-and-replace** ✅ — down to the replacing fold re-folding bytes it never fetched, which is the half that actually proves the keep.

6. **The dissolved hand-over stated, not dropped** ✅ — and this is the best thing in the PR. The criterion says a residue of transferable duty means the criterion is *not* met. There was one: `DeclinedReclaim.reason` still carried `'writes-a-followed-stream'`, which the previous release note had already told readers was gone. **I verified the removal is provably behaviour-neutral** rather than trusting it: on `main`, `git grep 'writes-a-followed-stream' -- packages/` returns the type declaration and *nothing else* — no producer anywhere — and on this branch `reclaim` pushes only `deletion-failed`. So the member was a promise the type made and could not keep.

7. **Operator-facing text and `/status`** ✅ — the admin endpoint's documented answers, the server README, `movePointer`, `unslottedGenerations`, `streamReader`, both `follows` fields and `CONTEXT.md`'s drop-on-promotion bullet, which contradicted its own generation-slot bullet ten lines below it. `/status` reporting one entry per generation **held** is asserted in the new suite.

8. **Citations resolve across the done-move** ✅ — verified independently: `docs/adr/0087-…:11` and `work/specs/ready/a-generation-retains-the-code-that-folds-it.md:88` both name this task **by slug**, the trap is still removed, no new path citation appeared, and `check:refs` is green at 281 artifacts.

9–10. **Tests and changeset** ✅ — 8 new cases; `@etherfold/core` minor, `@etherfold/server` and `etherfold` patch.

## The judgement calls, and why they pass

**Narrowing a published union is breaking, and `minor` is the right bump.** `@etherfold/core` is `0.7.0`, so under 0.x a minor *is* the breaking bump. Keeping the member would have left exactly what criterion 7 exists to remove: a reason an operator can read in the type and never receive. Keeping it a one-member union rather than collapsing it to a bare literal is the better call, since a second reason can later be told apart without a shape change.

**The sabotage hooks are gone, and that was verified, not assumed.** The prior attempt's branch left two `NEGCTRL` switches in `packages/core/src/stream/writer.ts`. `git diff main -- packages/core/src/stream/writer.ts` is **empty**. A `globalThis`-reachable sabotage switch in a published package would have been a defect shipped to prove a test, and deleting it rather than bounding it is right: the measurement the task asks for (rows plus ranges) is a stronger sensitivity argument than suppressing the write.

**ADR-0087's `status:` line is untouched, which is correct.** `git diff main -- docs/adr/` is empty. That line belongs to `the-stream-writer-vocabulary-is-documented-and-adr-0087-expires`, which is blocked behind this task precisely so it is unambiguously last. Removing it here would have been the error two earlier ADRs made.

**The fan-in did not pad.** The task's stated failure mode was manufacturing a second copy of an existing test. Instead the report names what was already covered and by which file, and identifies the one thing `aRestartReFoldsTheStoredStream.test.ts` structurally *cannot* assert — growth, because its restarted chain has not moved on — which is precisely the gap this task owned.

**`unslottedGenerations`' JSDoc distinguishes the two runtimes** rather than flatly asserting the new rule: the chain-facing `Indexer`'s strand clause is still live and still correct, because a generation does still fetch there. Getting that distinction right is the difference between documentation and a future bug report.

PR #181, opened 2026-09-20
feat(the-stream-writer-vocabulary-is-documented-and-adr-0087-expires):…

wighawag:
Task: `work/tasks/done/the-stream-writer-vocabulary-is-documented-and-adr-0087-expires.md`

Done. Here is what I did and what I verified.

## The family, checked rather than assumed

All three blockers are in `work/tasks/done/`: `run-and-build-drive-the-folds-they-hold-rather-than-one-captured-receiver.md`, `whoever-fetches-a-stream-writes-it-and-the-stream-outlives-every-fold.md`, `a-restarted-deployment-hands-over-the-write-duty-it-cannot-discharge.md`. I read all three records (the middle one in full, including its `## Decisions`).

The two claims the prompt asks to verify in the code:

- **Nothing derives a write duty from registration order.** True on the receiving runtime: `writesStream` and `reconcileWriters` exist nowhere in `packages/*/src` (only in tests asserting their absence), `ReceivingIndexer` holds one `StreamWriter` per stream (`receivingContainer.ts:604`, `1118-1125`) positioned from `StreamCursorSource` over the stream's coverage claim, and no fold is handed an appender. On the chain-facing `Indexer` the fetch was deliberately not restructured: `fetcherOf` still decides which generation fetches, and that generation appends through the keeper it holds (`container.ts:839`). That is ADR-0044's follower rule, which ADR-0087 leaves standing, and it is recorded as a deliberate scope decision in the blocker's own done record. See the Decisions block for why I judged the status line removable on that basis rather than stopping.
- **A stream survives having no folds across a restart.** True. `GenerationRegistryState.keptStreams` is durable (`_generation_streams` on SQL, a `kept` key level on IndexedDB, a set in the memory port), the sweep on open compares a subtree against the generations *and* that set (`registry.ts:775`), and `packages/server/test/generationRegistry.test.ts` asserts it by re-opening over the same database ("SURVIVES A RESTART, because the keep is a ROW and not a decision in memory") while a pre-generation orphan is still collected.

## What changed

- **`CONTEXT.md`**: one rule now, stated the same way in every entry that touches it. The rule is *a stored stream has exactly one writer and it is whoever FETCHES it*; the only thing that varies is what does the fetching (the deployment on a receiving runtime, the fetching generation on a chain-facing one). I fixed a contradiction the four patches had left: the **writer token** entry said "no generation is ever handed the thing that appends" while the **follower** entry said a generation fetches and writes on the chain-facing container, and the **generation slot** entry said the strand clause "is GONE" flatly while the **drop-on-promotion** entry correctly said it survives on one runtime. The code agrees with the latter in both cases (`wouldStrandAFollower` is live in `container.ts:1521`, absent from `receivingContainer.ts:1637`). Also touched for one account: **read-only stream view**, **stream writer**, **stream identity**, **pause/draining**, **indexer**, and the two-containers paragraph.
- **`docs/adr/0087-…`**: the `status: accepted, not yet implemented` frontmatter is removed, and one dated amendment is added saying what "implemented" covers and what it deliberately does not.
- **`docs/guide/indexing-in-a-browser-app/index.md`**: removed a parenthetical citing "a known hazard, a follower generation that can clear the writer's stream" through a dead GitHub link. That note was deleted by the ADR-0069 commit that fixed the hazard (`391dbf89`); the read-only view's `clear` is a no-op.
- **`work/notes/observations/drop-superseded-still-documents-a-reap-it-can-no-longer-perform.md`**: new capture note (out-of-scope drift, not fixed).

## Doc surfaces checked

All 18 package/platform READMEs, `README.md`, `docs/guide/`, `docs/index.md`, `docs/design/`. Only `packages/server/README.md` describes the rule and it is already correct (the `reclaim` decline "no longer where dropping would leave a fold folding a stream nothing appends to"). No other guide or README teaches the retired rule; stating that is the answer to that criterion.

## Gate

`check:refs`, `check:adr`, `check:graph`, `check:changesets` green; `check:refs` needed no fixes (the family's earlier tasks fixed their own citations, and observation notes are exempt by the script's own design, which is why the one surviving `tasks/ready/` citation of a landed task does not fail it). Prettier (pinned 3.8.3) clean on every file I touched. No changeset: nothing under `packages/` or `platforms/` changed, and the repo's rule is that a changeset is for a published package's public API. `build`/`typecheck`/`test` not run: no source file changed (`node_modules` is not installed in this worktree).

## Decisions

**How I restated the writer contrast (the writable-store entry).** The old contrast was a POLARITY: a stream's writer is the *oldest surviving* generation, a state store's is the *last claimant*, earliest-wins against latest-wins. That polarity is gone, so restating it as "the deployment fetches it" left a sentence that was true and taught nothing. I rewrote it as ROLE against ACT: a stream's writer is decided by what a thing DOES (whoever fetches it), so the duty is not assignable, cannot be given to something that is not fetching, and a second fold arriving changes nothing; a state store's writer is decided by an ACT and its recency, so arrival is exactly what decides and `openForWriting` takes the store from whoever held it. I then made the consequence explicit, because that is what makes the contrast worth keeping: a **demotion** exists on the store axis and has no counterpart on the stream one. Alternative considered and rejected: dropping the contrast to the `StreamWriter`-versus-`writerClaim` mechanical difference, which is true but would stop teaching why the two must not be conflated. Touches: the **writer token** entry, which carries the same distinction from the other side and is now worded to match.

**Removing the status line, given that the chain-facing container still derives the fetch (and so the append) from registration order.** This is the judgement call the task turns on, and it is the one a reviewer should check. Read literally, my task says to leave the line if "a write duty [is] still derivable from registration order", and on the browser engine it is. I removed it anyway, because what ADR-0087 retired is an ELECTION that hands the pen to a fold that is not fetching, and that shape is gone: on the receiving runtime the deployment fetches and appends, on the chain-facing one the fetcher and the writer are one object, so the invariant "a stream's writer is the thing fetching it" holds on both. Registration order there answers *which generation fetches* (ADR-0044, untouched by this ADR and strengthened by it), not *which non-fetching fold holds the pen*. The alternative was to STOP and leave the line; I rejected it because the line means "no code implements this", which is now flatly false and misleads in the more expensive direction, and because the scope boundary is not an under-build discovered by me: the blocker's builder recorded it as a deliberate decision with its cost stated, the repo carries an observation about the residue, and a task for it sits in `work/tasks/ready/` (`the-reloaded-tab-stall-is-measured-on-the-configuration-a-tab-actually-has`). The change is one line and trivially reversible if the human disagrees. What it touches: that ready task, and ADR-0044, whose follower rule is what the chain-facing derivation now rests on alone.

**I amended ADR-0087 rather than only deleting the line.** With the status gone the ADR reads as current and unqualified, and its own headline says "never by a GENERATION", which is false on the chain-facing runtime; leaving that would recreate the drift this task exists to remove, one layer up. So the ADR carries a fifth dated amendment saying what is implemented, what is deliberately out of scope and why, and naming the follow-on task BY SLUG rather than by path (the ADR's own convention, and what keeps `check:refs` green when that task lands). Alternatives: delete the line only (rejected, above), or retitle the ADR (rejected: a title is the citation handle, and thirty-odd references would rot).

**What I left alone, deliberately.** (1) `CONTEXT.md` still says registration order picks the fetching generation on the chain-facing container, because the code does; deleting it to satisfy "no entry deriving the duty from registration order" would have made the glossary false, which is the outcome the prompt ranks worst. It is framed as the FETCH derivation, with one sentence saying plainly that the retired rule was the other thing. (2) The residue itself is not written into `CONTEXT.md`: the glossary teaches the rule in force, and an open, unmeasured hypothesis belongs in the observation and the task that own it (both exist). (3) ADR-0044, 0055, 0071 and 0072 keep their bodies and their absent statuses. They are frozen records, ADR-0087 already names what it supersedes, and no earlier task in the family annotated them, so adding a `superseded in part` line is a convention call for a human rather than something to smuggle into the closing task. I noted the cost (a reader arriving from a `CONTEXT.md` ADR-0044 citation meets the retired rule with no signpost) in the observation note. (4) `Indexer.dropSuperseded` and `dropReplaced` in `packages/core/src/container.ts` still document and log a stream reap that `deleteGeneration` can no longer perform (`reapStream` defaults off, so `deletion.reaped` is always `undefined` there); that is inside this task's `packages/*/src` fence, so it is a capture note, not an edit.


wighawag:

Merging. This is the closing task of the ADR-0087 family and it gets the thing the family was most likely to get wrong exactly right.

## The status line

`git diff main -- docs/adr/0087-…` removes the frontmatter block entirely. **No status line at all** is the correct end state, because `ADR-FORMAT.md` says an ABSENT status means accepted and current. It is not replaced with an invented value; a previous build in this repo coined `accepted, implemented`, which is not one of the seven, and had to be reverted. Not repeated here.

## The judgement call, checked against the protocol rather than against the argument

The builder flagged it themselves, which is the right instinct: the task says to leave the line if "a write duty [is] still derivable from registration order", and on the chain-facing `Indexer` (a browser tab) `fetcherOf` still decides by registration order which generation fetches, and that generation appends. Read literally, that is the stop condition.

I checked it independently and **the removal is right**, on `ADR-FORMAT.md`'s own terms rather than on pragmatics:

- **There is no third option.** The status vocabulary has seven values and none of them means "partially implemented". So this is binary: keep `accepted, not yet implemented`, or remove it.
- **The protocol's test is which reading MISLEADS** ("add one only when the plain reading would MISLEAD"). `accepted, not yet implemented` is defined in the table as "decided, and the code is kept compatible with it" — i.e. *no code implements this*. After four merged PRs restructuring the receiving runtime, that is flatly false, and false in the more expensive direction. Absence claims "accepted and current", which the ADR is.
- **The scope boundary is not an under-build discovered here.** It was recorded as a deliberate decision in the blocker's own done record, the repo carries an observation about the residue, and a task to measure it sits in `work/tasks/ready/`. The stop condition exists to catch a family that quietly landed less than it decided; this is a family that decided its scope out loud.

**And the ADR was AMENDED rather than merely stripped**, which is what makes the removal honest. Without it the document would read as current and unqualified while its own headline says "never by a GENERATION", which is false on the chain-facing runtime — recreating the drift this task exists to remove, one layer up. The fifth amendment states what is implemented, what is deliberately out of scope and why, and names the follow-on task **by slug**, so it survives that task's done-move. Retitling was considered and rejected for the right reason: a title is the citation handle.

## Criteria

1. **One rule in `CONTEXT.md`** ✅ — *a stored stream has exactly one writer and it is whoever FETCHES it*, with the only variation being what fetches. The builder found and fixed a contradiction the four earlier patches had left between each other: the **writer token** entry said no generation is ever handed the appender while the **follower** entry said a generation fetches and writes on the chain-facing container, and the **generation slot** entry said the strand clause "is GONE" flatly while **drop-on-promotion** correctly said it survives on one runtime. **I verified which is true:** `wouldStrandAFollower` appears in `packages/core/src/container.ts` (1521, 1556) and **nowhere in `receivingContainer.ts`**. The per-runtime wording matches the code; the flat wording did not.

2. **The writer/store contrast survives as a teaching point** ✅ — and this was the criterion most likely to be satisfied vacuously. It is not: the old contrast was a POLARITY (earliest registration wins the stream, latest claim wins the store) and that polarity is genuinely gone, so restating it under the new rule would have left a sentence that is true and teaches nothing. It is rewritten as **ROLE against ACT** — a stream's writer is decided by what a thing *does*, so the duty is not assignable at all; a store's is decided by an act and its recency, so arrival is exactly what decides — and then made to pay for itself by drawing the consequence: **a demotion exists on the store axis and has no counterpart on the stream one.** It also says explicitly that the polarity is gone rather than reversed, which is the misreading it was most at risk of.

3. **Kept when nothing folds it, deleted only when asked, verbs named** ✅ — "every AUTOMATIC reap is gone (registering into an occupied `successor` slot, and drop-on-promotion), deletion is a VERB (`reclaim`, `deleteStream`), and `deleteGeneration` reaps only when a caller ASKS."

4. **Doc surfaces** ✅ — 18 package READMEs, the root README, `docs/guide/`, `docs/index.md`, `docs/design/`, with the absence stated after checking, which is what the criterion asks for. One real correction: the browser guide cited "a known hazard, a follower generation that can clear the writer's stream" through a **dead** GitHub link. **I verified the deletion is not hiding a live hazard**, since the browser runtime is precisely the one ADR-0087 does *not* cover: the note was deleted by `391dbf89`, the ADR-0069 commit that FIXED it, and `packages/core/src/stream/segments.ts:223` documents the fix in terms — the read-only view's `clear` "is a no-op, so it can no longer destroy the stream its writer is still appending to". Stale text about a fixed hazard, behind a broken link. Right to remove.

5. **Family confirmed done, and which was checked stated** ✅ — all three named and verified in `work/tasks/done/`, plus the two code-level claims re-checked rather than inferred from a green board: no `writesStream`/`reconcileWriters` in `packages/*/src`, and `keptStreams` durable with a test asserting stream survival across a re-open.

6. **Status line removed, with the check stated** ✅ — see above.

7. **Citations resolve** ✅ — `check:refs` green, and the new ADR amendment cites the follow-on task by slug.

8. **No behaviour change** ✅ — verified: the diff touches `CONTEXT.md`, one ADR, one guide page, one observation note and the done-move. **Nothing under `packages/`.**

9. **Changeset** ✅ — correctly none, doc-only, with the repo's rule checked rather than guessed.

## Nits, none blocking

**The red `verify` is inherited, again.** Same single dead link in `docs/spikes/the-build-command-and-its-pinning-rule-are-documented/README.md`, from PR #173 two days ago; this branch touches no file under `docs/spikes/`. There is a mild irony in a docs-coherence task merging over a broken docs build, but it is genuinely outside this task's remit: that link does not teach the retired rule, which is what criterion 4 scopes. Both the link and the gate gap that hid it are filed as an observation with the bisect.

**The capture note is correctly scoped.** `drop-superseded-still-documents-a-reap-it-can-no-longer-perform` records that `Indexer.dropSuperseded`/`dropReplaced` still document and log a stream reap `deleteGeneration` can no longer perform. That is inside the task's own `packages/*/src` fence, so capturing rather than fixing is exactly right, and the note says so. Kept as an observation.

**Leaving ADR-0044/0055/0071/0072 unannotated** is the right call and is stated rather than silent. Adding `superseded in part` lines to four frozen records is a convention decision for a human, not something to smuggle into a closing task, and the cost (a reader arriving from a `CONTEXT.md` ADR-0044 citation meets the retired rule with no signpost) is written down in the note rather than swallowed.

PR #182, opened 2026-09-20
feat(the-reloaded-tab-stall-is-measured-on-the-configuration-a-tab-actu…

wighawag:
Task: `work/tasks/done/the-reloaded-tab-stall-is-measured-on-the-configuration-a-tab-actually-has.md`

Everything is green and the tree holds only the intended changes.

## What I did

I checked the task's premise against the tree first, then **ran** the mechanism the observation only traced. The stall is **real**, but only under one trigger and one configuration, and both differ from the note's headline.

**Which registry a real tab has: a memory one, and the stall cannot be reached on it.** Both entry points default to `openMemoryGenerationRegistry` (`IndexerState.ts:1202`, `host/serve.ts:374`), and nothing in `packages/*/src`, `examples/`, `platforms/` or `docs/` passes `openGenerationRegistryOnIndexedDB`. Measured: on the default, a changed-handler reload fetches `102..107` and lands on the tip. **The stall requires the opt-in durable registry** — which is the documented way to keep a revert target, not a test-only shape.

**Trigger A (changed handler) is a REFUSAL, not a stall**: `CanonicalGenerationNotHeldError`, before any chain read. Recorded as a different bug class.

**Trigger B (reload after a promotion) IS the silent stall**, reproduced with numbers: `{eth_chainId: 1}` and nothing else — zero `eth_blockNumber`, zero `eth_getLogs` — tab cursor 105 against a node tip of 107, stored stream frozen at 105, and the host's own pacing rule reports `at-tip` because a follower never asks for a block number. Identical whether the tab's state survives the reload or not.

**The existing test's two-generation configuration is not reachable** from either entry point (both build one-element `generations` lists), and it is a constraint rather than an oversight: with one fold that container throws.

**I also built and measured the obvious narrow fix.** It removes the stall, keeps core + browser green, and leaves no duplicate and no hole — because on this runtime the fold replays the stored stream before it fetches, so ADR-0087's duplicate-or-hole result does not transfer as written. But it reads `this.held` mid-`open`, so listing the same two folds the other way round gives **2 fetchers, 2 identical `eth_getLogs`, and block 106's log stored twice** (7 rows where 6 are correct). The patch was reverted; it lives in the spike.

Evidence: `docs/spikes/the-reloaded-tab-stall-is-measured-on-the-configuration-a-tab-actually-has/` (harness, both raw outputs, the patch, a README). Decision: `docs/adr/0088-...`. An `## Update` was appended to the observation. No behaviour change: no file under `packages/*` differs from `main`; no changeset. Full gate green.

## Decisions

- **The ADR names the NARROW direction, not extending ADR-0087 to the browser engine — and the specific narrow form is "the fetcher is the oldest generation the container HOLDS", not "the registered fetcher must be held".** Chosen because the measurement removed the argument that forced the broad move on the receiving side: there the restarted successor's state was empty and it re-appended covered history through a guardless appender, while here `promiseToLoad` replays the stored stream first, so the fold lands exactly on the coverage (measured: 6 events, blocks 100/102/104/106, none twice, covers to 107). Alternatives considered and recorded as rejected in the ADR: extending ADR-0087 (rejected on proportion — a browser-engine restructure against a defect reachable only on an opt-in configuration nothing ships — and explicitly left available); refusing at open symmetrically with `CanonicalGenerationNotHeldError` (rejected as the answer, kept as the fallback: it would refuse a tab that merely promoted and reloaded); defaulting `dropOnPromotion` to true (rejected outright — it destroys the revert window ADR-0084 exists for, to work around a wrong derivation). Touches: the follow-on FIX task, ADR-0044's registration-order clause on the chain-facing runtime only, and the re-scope of `aTabHoldsItsGenerationsInSlots.test.ts`.
- **The ADR carries `status: accepted, not yet implemented`.** Absence means accepted and current, and no code implements this, so absence would mislead. Given this family just paid for that line surviving ten tasks, the ADR says in its own body that the line comes off in the same change that lands the fix. Alternative considered: no status line (rejected as a false claim about the code).
- **Coherence check on the concept: no new term is introduced.** `fetcherOf` keeps its meaning, its home and its purity; only the record set the chain-facing container asks it about narrows from "registered" to "held". `CONTEXT.md`'s glossary sentence ("a stored stream has one writer and it is whoever fetches it... the fetching generation on a chain-facing one") stays true word for word, so no glossary edit is needed. The one genuinely new consequence — that "which generation writes this stream" becomes a fact about the PROCESS on this runtime — is named in the ADR's Consequences, together with the fact that it is not a regression (two tabs already both fetch and both write today) and that its answer is the single-indexer lease, not this rule.
- **The harness is a `tsx` SCRIPT under `docs/spikes/`, not a vitest test, and it imports by relative path.** The task forbids changes under `packages/*/test` and forbids leaving anything red, so the measurement must not be collected by `packages/browser`'s vitest config. Being outside every workspace package, bare specifiers cannot resolve from there, so each import names the file the browser package's own imports resolve to (core/state-store from `dist`, browser from `src`), keeping exactly one copy of each package in the process. Documented at the top of the file. Alternative considered: a vitest file plus a bespoke config (rejected — same resolution problem, plus a second test config to keep in step).
- **An `## Update` was appended to the observation even though the defect DID reproduce.** The acceptance criterion only mandates one on refutation, but the note's own `## Update` is explicit that its live hypothesis was never run, and leaving that standing invites the next reader to re-derive it. It is append-only capture of "this signal was run, here is the result, here is the evidence", not builder rationale — which is here, per WORK-CONTRACT. Deliberately NOT written as a `work/notes/findings/` entry: measuring our own code's internals is excluded from that bucket by name.
- **I did not re-scope the reload test, and I state in both the ADR and the spike why the note's proposed re-scope is wrong.** Opening that container with one generation does not go red, it THROWS; the honest re-scope is the promotion-then-reload shape, and it lands green with the fix.


wighawag:

Merging. This is a measure-and-decide task and it did the hard version of both halves: it refuted its own task's stated trigger, found the real one, then patched the obvious fix and measured it breaking something else.

## The scope fence, checked first

This task is explicitly NOT a fix, and its criterion says no file under `packages/*/src` **or** `packages/*/test` changes. Verified:

```
$ git diff --name-only main...HEAD | grep ^packages/
(nothing)
```

The diff is one ADR, a spike directory, an `## Update` on the observation, and the done-move. The narrow-fix patch is kept as `the-narrow-fix-candidate.patch` under the spike, labelled `NOT SHIPPED -- reverted before this task landed`, which is exactly where a measured-but-unshipped patch belongs.

## Criteria

1. **Which registry a real tab has** ✅ — and **I checked this independently rather than taking it**, because the whole reachability answer rests on it. Both entry points read `spec.registry ?? await openMemoryGenerationRegistry(...)` (`IndexerState.ts`, `host/serve.ts`), and a repo-wide grep for `openGenerationRegistryOnIndexedDB` across `packages/*/src`, `examples/`, `platforms/` and `docs/` returns **only two doc-comment mentions and no call site**. The durable registry is opt-in and nothing ships it. That is recorded as the reachability answer, which is what the criterion asks for.

2. **Both triggers, on one generation over a durable registry** ✅ — and the task's *named* trigger is refuted while a different one is confirmed. Trigger A (changed-handler reload) is a loud `CanonicalGenerationNotHeldError` before any chain read. Trigger B (reload after a promotion) is the silent stall, measured in both sub-shapes: fresh state and surviving state, identical result.

3. **A refusal recorded as a refusal** ✅ — stated explicitly as "a different bug class from a stall", not folded into the stall finding. This is the criterion most likely to be quietly blurred and it is not.

4. **The measurement is chain reads, and separates the handshake** ✅ — this is the criterion the task was most specific about, and the numbers land on it exactly:

   ```
   chain reads:        {eth_chainId: 1}   <- the load-time handshake, and nothing else
   eth_getLogs ranges: []                 <- none, ever
   node tip:           107
   tab cursor:         lastToBlock 105
   reported phase:     "at-tip"
   ```

   Methods and ranges, not a `follows` flag and not a slot listing. And the last line is the finding that makes this worse than a stall: a follower never calls `eth_blockNumber`, so `latestBlock` stays at 105, `someGenerationBehind` compares 105 with 105, and **the host reports itself healthy and at-tip while the chain is two blocks ahead.** That was not asked for; it is the sharpest thing in the report.

5. **The existing test's configuration, judged** ✅ — unreachable, with the entry points named and the reason given as a CONSTRAINT rather than an oversight: it holds both folds because with one fold that container *throws*. That is the judgement the task asked for, in the terms it asked for.

6. **An ADR naming the direction and the trade-off** ✅ — ADR-0088, and it clears the bar the task set for a narrow proposal. The task said a narrow family "carries a known hazard ... a narrow proposal must say how it escapes that, or say that it does not and be rejected on those grounds." ADR-0088 says how, and **measures it**: on this runtime the fold replays the stored stream at `load` before anything is fetched, so it lands exactly on the coverage — six events over blocks 100, 102, 104, 106, none twice, covering to 107. So the duplicate-or-hole result ADR-0087 measured does not arise here, and the reason it was forced there (an empty state fetching from `defaultFromBlock` through an unguarded append-only appender) is named and shown not to apply. That is the difference between escaping a hazard and asserting you have.

8. **Evidence kept under `docs/spikes/<slug>/`** ✅ — a re-runnable harness, both raw outputs, and the candidate patch.

9. **No behaviour change, nothing red** ✅ — see the fence check.

10. **Changeset** ✅ — correctly none.

## The best thing in this PR

The narrow fix was not merely proposed, it was **applied and measured**, and the measurement found the trap:

```
without the fix   scenario 5 (same two folds, listed edited-first):  ONE fetcher
with the fix      scenario 5:                                        2 FETCHERS on one stream
```

Two identical `eth_getLogs`, one log stored twice. The cause is that the one-line predicate reads `this.held` **mid-`open`**, which populates incrementally per spec, so the answer depends on the order a caller listed its specs in. The ADR promotes that from a footnote to the largest piece of the follow-on work — `open` must decide `follows` once it knows every fold it will hold — which is precisely the kind of thing that would otherwise be discovered by the fix task the expensive way.

This is the discipline this family has been strongest at: reproduce, patch the obvious fix, measure that it makes something worse, report the numbers.

## Smaller things, all correct

**ADR-0088 carries `status: accepted, not yet implemented`, and that is right.** The fix is a follow-on task and no code implements the decision, which is exactly the status's definition in `ADR-FORMAT.md`. The ADR also states who removes it and when. Note the contrast with ADR-0087 in this same drive, whose line came off because its code *had* landed: both calls are right, in opposite directions, for the same reason.

**The observation was APPENDED to, not rewritten** — one `## Update` section, no prior text removed, per `WORK-CONTRACT.md`'s append-only rule for capture buckets. It records that the note's own originally-proposed remedy was measured and found order-dependent, which is a note correcting itself in public rather than quietly.

**The two-tabs hazard is named and disclaimed correctly** — "NOT introduced here", since two tabs holding the same canonical fold already both fetch and both write today, each with its own registry instance. Flagging it so a later reader does not mistake it for a regression is the right move, and pointing at the single-indexer lease as its actual owner rather than widening this decision is better still.

**`dropOnPromotion: true` rejected outright** for the right reason: it would destroy the revert window that slots, the promotion policy and `predecessor` exist for, to work around a derivation that is wrong on its own terms.

## Nit, not blocking

The red `verify` is the **same inherited dead link** from PR #173, in `docs/spikes/the-build-command-and-its-pinning-rule-are-documented/README.md`. Worth noting that this PR adds a new spike README of its own and the dead-link count is **still exactly one**, so it introduced none. Both the link and the gate gap that hid it for seven PRs are filed as an observation with the bisect.

PR #183, opened 2026-09-20
feat(a-reloaded-tab-fetches-because-the-fetcher-is-a-fold-the-tab-actua…

wighawag:
Task: `work/tasks/done/a-reloaded-tab-fetches-because-the-fetcher-is-a-fold-the-tab-actually-holds.md`

Everything is green. Final verification summary:

| check | result |
| --- | --- |
| `pnpm build` | Done (all packages/platforms) |
| `pnpm typecheck` | Done (all packages, platforms, examples) |
| `pnpm test` | core 1208 passed (+4), browser 378 passed (+5), all other packages green |
| `pnpm format:check` / `check:adr` / `check:refs` / `check:graph` / `check:changesets` | all pass |
| `pnpm docs:build` | build complete (the ADR now has no frontmatter) |
| spike harness re-run | scenarios 2a/2b flip STALLED→FETCHES; **every other line identical**, scenario 5 included |

## What I did

**Reality check first.** `packages/core/src/container.ts` and `packages/browser/` are byte-identical to the commit the spike measured (`8ca52576`), the sibling prose task has not landed, and I re-ran the harness before touching anything: it reproduced `measurement-today.txt` line for line (`{eth_chainId: 1}`, zero ranges, `at-tip` at 105 while the node was at 107). The task's premises hold, with two corrections noted below.

**The fix** (`packages/core/src/container.ts`). `Indexer.open` is now two phases: `registerGeneration` (build state → processor → identity → registry record) runs for every spec, then `holdGeneration` builds each engine knowing the complete fold set. `follows` is derived with `fetcherOf(willHold, stream)` — the same pure function, unchanged, asked about the folds the container HOLDS instead of every record the registry carries. `add` keeps its public signature and passes `held + the new record`, which is already complete once the container is open. The long comment block that argued FOR the registered-set reading is rewritten to state ADR-0088's rule and why reading the held set is safe now (the answer is taken once the set is complete), keeping the ADR-0071/0072 history. `wouldStrandAFollower` and `dropSuperseded`'s strand clause are untouched.

**Tests, written before the code and asserting chain reads.** `packages/browser/test/aTabHoldsItsGenerationsInSlots.test.ts` gains the promotion-then-reload shape (parameterised over a fresh state and the tab's own surviving state): methods `{eth_chainId: 2, eth_blockNumber: 1, eth_getLogs: 1}`, range `102..107`, cursor and `latestBlock` both at the node's tip, six events in the stream with none stored twice, the revert target still in `predecessor`. Beside it, an order probe in both spec orders and the still-loud `CanonicalGenerationNotHeldError` for the changed-handler reload. `packages/core/test/follower.test.ts` gains the same two properties over a registry that outlived its process. `fakeChain` now records every method, not just `eth_getLogs` ranges, which is what makes "it asked the node for nothing" assertable.

**Each test fails against exactly one wrong rule**, checked by putting each back in turn: the registered-set rule fails the not-held case; the half-built-`held` rule (the rejected candidate patch) fails both order probes with 2 fetchers.

**Also:** ADR-0088's `status` line removed entirely (no status line at all); the fix measurement kept at `docs/spikes/a-reloaded-tab-fetches-because-the-fetcher-is-a-fold-the-tab-actually-holds/` with a README reading it against its two neighbours; a changeset; one observation note for a latent oddity I noticed and did not chase.

## Decisions

**The shape for deciding `follows` once the fold set is known: two phases inside `open`, with the set passed as a parameter of records.** `open` calls `registerGeneration` for every spec, collects the `GenerationRecord`s it gets back, and passes that array to `holdGeneration` for each. Records were chosen as the carrier because they already hold the `createdAt` the ranking uses, so `fetcherOf` is called with exactly the set the ADR names and needs no second registry read. Alternatives considered: (a) a mutable `this.openingFoldSet` field read by `add` — rejected, it is hidden state and `add` would silently behave differently depending on whether `open` is on the stack; (b) an optional second parameter on the public `add` — rejected, it puts an internal phase in the published API; (c) computing the fold set from the SPECS before building anything, which is the obvious reading of "open must decide follows once it knows every fold it will hold" — **not available**, see the next entry. Touches: `Indexer.open` and `Indexer.add` only; `add`'s public signature is unchanged, and the two new methods are `protected`.

**The phase boundary is AFTER registration, not before the factories, and the ADR does not say why that is forced.** A generation cannot be named before it is built: a fold arriving as a module has no bytes to hash, so `GenerationSpec.processorIdentity` is filled in from inside `createProcessor` (ADR-0086, and `@etherfold/browser`'s `generationSpecFor` does exactly this). So the cheapest-looking implementation — derive every id from the specs up front, then run `add` unchanged — would throw `requireProcessorIdentity` on the browser's own entry point. Recorded because a later reader will otherwise ask why phase one does the expensive half.

**What the ADR got wrong: `CONTEXT.md` did NOT stay true word for word.** ADR-0088's Consequences say it does. That is right about the one-writer sentence it quotes and wrong about the clause after it: the **follower** glossary entry also said "WHICH generation fetches is the OLDEST SURVIVING one registered on the stream", which is the sentence this ADR exists to narrow. I updated that entry to state the present-set rule, why it is present rather than registered, and that the answer is taken once; and I added a sentence saying the DROP question (`wouldStrandAFollower`) is deliberately still asked over the registered records, so the next reader does not "harmonise" it. I deliberately did NOT touch the drop-on-promotion entry's parenthetical, which describes `fetcherOf` in the context of the drop clause and is still accurate there. ADR-0088's closing paragraph now records both the implementation and this correction.

**The changeset names `@etherfold/browser` as well as `@etherfold/core`, both `patch`.** No file under `packages/browser/src` changed — only its test fixture and its tests — but the behaviour change a browser app SEES is this one, and the package is where it is observable. The alternative was core alone; I took the wider one because it costs an unpublished version bump and the narrower one risks a land-time bounce.


wighawag:

Merging. This implements ADR-0088, avoids the trap that ADR measured, and is the **first PR in this repo to merge on a green `verify`** since 2026-09-18 — through the stricter gate that now includes `docs:build` and `build:examples`.

## The measurement, on the same harness

Re-run over the existing spike's seven scenarios, so it is directly comparable to both prior measurements:

```
                          today        obvious patch      this change
2a reload-after-promotion  STALLED   -> FETCHES        -> FETCHES
2b  (state survives)       STALLED   -> FETCHES        -> FETCHES
5  order probe             1 fetcher -> 2 FETCHERS     -> 1 fetcher
0, 1, 3, 4                 unchanged    unchanged         unchanged
```

That middle column is the whole point. The one-line fix removed the stall and silently introduced two fetchers on one stream; this does not. Criteria 1 and 2 are met together, which is what the task demanded and what the rejected patch could not do.

## Criteria

1. **The tab fetches, asserted on chain reads** ✅ — methods and `eth_getLogs` ranges, not the `follows` flag.
2. **Order-independent** ✅ — scenario 5 with the same two folds listed edited-first still names exactly one fetcher.
3. **`follows` decided only once the fold set is known** ✅ — and this is the substance. `open` is now genuinely two-phase: `registerGeneration` for every spec first, then `holdGeneration` for each against `willHold`, the complete record set. The derivation changed from `fetcherOf(await this.registry.list(), …)` to `fetcherOf(willHold, …)`. Not a predicate tweak.
4. **The reload test re-scoped to promotion-then-reload** ✅ — `describe('a tab that reloads AFTER A PROMOTION goes on FETCHING')`, session 1 saves and promotes, the reload supplies exactly ONE fold. This is the fix I added to the task after review, because the original criterion said "open with ONE generation", which the spike itself records as throwing. Honoured exactly.
5. **Changed-handler reload still refuses loudly** ✅ — scenario 1 unchanged.
6. **Default memory-registry path unchanged** ✅ — scenario 0 unchanged.
7. **`fetcherOf` still pure, nothing persisted** ✅ — same function, same home; only the argument set changed. No pointer, no stand-down mark.
8. **ADR-0088's status line removed** ✅ — `git diff` shows the whole frontmatter block gone. No invented value.
9–10. **Tests and changeset** ✅ — `@etherfold/core` and `@etherfold/browser` patch.

Plus the criterion I added post-review: **runtime `add` unaffected** ✅ — handled explicitly, with the reasoning that the set is already complete there because the container is open, so one call suffices.

## The two guard rails I added after review, both honoured

**The 50-line comment that argued the opposite is rewritten, not left to rot.** `container.ts` previously justified reading the registry over `this.held` in terms — "which is whatever order the caller passed its specs in". That sentence was about to become the best argument against the code it sat on. It now states ADR-0088's rule, says why the held set is safe to read (the decision is taken after the set is complete), cites the measured two-fetcher result, and keeps the ADR-0071/0072 history that is still true.

**The two record-based derivations were left alone.** `git diff` on `container.ts` shows **no change** to `wouldStrandAFollower` or its `fetcherOf(registered, …)` call. Those ask a different question — what would deleting this record strand — over a deliberately different set, and harmonising them would have been a defect dressed as consistency. The temptation was real and it was resisted.

## The builder contradicted its own ADR, and was right

ADR-0088 claimed `CONTEXT.md` "stays true word for word" under this change. It does not: the **follower** entry also derived which generation fetches from the oldest *registered* on the stream — the exact sentence the ADR exists to narrow. The builder found it, fixed the entry, and **amended the ADR to record that it had been wrong**. That is the sixth time in this family a builder has contradicted its source document and been correct to; it is also precisely why the ADR is the better record for it.

The phase boundary is also justified rather than asserted: it sits after the factories because a fold arriving as a module has no bytes to hash, so identity is filled in inside `createProcessor` (ADR-0086) and a generation cannot be named any earlier. That is the non-obvious constraint that makes this restructure the shape it is.

## The new observation is correctly scoped

`opening-with-three-specs-can-hold-an-engine-whose-record-was-just-displaced` — three specs on one stream, the third displaces the second out of the `successor` slot, and `open` still builds an engine for the displaced one. It is explicit that it was **not verified by running it**, that it **predates this change and is unchanged by it**, and that it is likely unreachable in a browser tab because `BROWSER_GENERATION_CAPS` is two. Captured, not chased, and not smuggled into this diff. Right call on all three counts.

PR #184, opened 2026-09-20
feat(the-chain-facing-drop-stops-documenting-a-reap-it-cannot-perform):…

wighawag:
Task: `work/tasks/done/the-chain-facing-drop-stops-documenting-a-reap-it-cannot-perform.md`

All checks green. Here is what I did.

## What I verified first (the premise)

The task's central factual claim holds today:

- `packages/core/src/generation/registry.ts:1053` — `const reapStream = options?.reapStream === true;`, and `reaped` is set only under `reapStream && ...` (line 1066). So the default is FALSE.
- The only implementation of `deleteGeneration` in any `src/` is that one (server/browser only touch it from tests).
- Both chain-facing call sites passed **no options**: `container.ts:1560` (`dropSuperseded`) and `container.ts:1701` (`dropReplaced`). So `reaped` was provably always `undefined` on both paths and the two log branches were unreachable.
- Log strings: nothing asserts on them. `reaping the stream` / `reaped with it` appear only in `packages/core/src/container.ts`, `receivingContainer.ts`, `registry.ts`, `docs/adr/0046-...` (frozen record), two old changesets, and the work/ note. No test file matches.
- `CONTEXT.md` line 61 still carries the sentence the task quotes verbatim ("every AUTOMATIC reap is gone ... `deleteGeneration` reaps only when a caller ASKS"), so it needed no edit and got none.

## Changes (`packages/core/src/container.ts` only)

1. `dropSuperseded`'s JSDoc opener now says the drop takes its **registry row and its state store, and NOT its stream**, which outlives every fold over it and is deleted only where a caller ASKS (ADR-0087).
2. `wouldStrandAFollower`'s JSDoc keeps the strand argument intact and loses only the reap clause: the follower hazard is now stated as the missing appender "and only that", with the stored stream explicitly not at risk. The predicate itself is byte-for-byte unchanged, as is the `drop-on-promotion DECLINED` log.
3. `dropReplaced`'s JSDoc drops the "and a reap of its stream where no registered generation is left folding it" clause and gains a short **"The STREAM is NOT reaped with it" (ADR-0087)** paragraph, mirroring its receiving-side twin.
4. The two unreachable log branches are gone, and with them the `reaped` local in `dropReplaced` and the now-unused `deletion` binding in `dropSuperseded`. Each call site carries a one-line `// NO REAP` comment.
5. A changeset: `.changeset/the-chain-facing-drop-stops-documenting-a-reap-it-cannot-perform.md` (`@etherfold/core: patch`).

No behaviour change: the drops call exactly the same registry method with exactly the same (absent) options, and `reclaim` / `deleteStream` / `deleteGeneration`'s `reapStream` are untouched.

## Verification run

`pnpm format:check` clean, `pnpm build` green, `pnpm typecheck` green, `@etherfold/core` 1209 passed / 1 skipped (67 files), `@etherfold/browser` 378 passed (50 files), `node scripts/check-changesets.mjs .` and `pnpm check:adr` / `check:refs` / `check:graph` all pass. `changeset status --since=main` still reports "no changesets found" purely because the new `.changeset/*.md` is untracked (the helper uses `git diff` against the merge base, which cannot see untracked files); it resolves as soon as the runner's `git add -A` lands. No new tests: this changes no behaviour and nothing asserted on the removed strings. Working tree holds only the two intended files.

## Decisions

- **The two success logs now positively state the stream is KEPT, rather than just losing the dead clause.** `dropSuperseded` ends "Its state store is gone and the stream `<digest>` is KEPT: a stream outlives every fold over it and is deleted only when asked", and `dropReplaced` the same. The alternative was the strictly minimal edit: delete the ternary and leave the sentence silent about the stream. I chose the positive statement because it is exactly what the already-landed receiving twin logs (`receivingContainer.ts:1556`), so an operator reading either runtime's drop line gets the same fact, and because a silent log invites the next reader to re-ask the question this task exists to answer. It touches operator-visible log text beyond pure removal, which is why it is recorded; nothing asserts on these strings, and the changeset says so.
- **I also corrected two trailing clauses inside `dropSuperseded`'s JSDoc that the task did not enumerate but that carried the same stale reap.** The old text argued the decline was "worse than keeping the bytes" and that the case it serves is one "where the retired generation owns a whole stream of its own that goes with it". Under ADR-0087 no stream goes with it, so that was the identical false claim one paragraph lower; leaving it would have left the observation's subject half present in the very method the task names. The wording is now "worse than keeping a row and a state store nobody reads" and "the retired generation is alone on its stream and dropping it strands nothing". Alternative considered: fix only the opening sentence the observation quotes. Touches nothing outside this JSDoc.
- **Vocabulary: "state store", not "state namespace".** The acceptance criterion names "the registry row and the state namespace"; `container.ts` says "state store" six times against "state namespace" once, because the chain-facing container drops a key-value store rather than ADR-0053's SQL table namespace. I used the file's own term so the chain-facing prose stays internally consistent, while `receivingContainer.ts` keeps "state namespace". Same concept, per-runtime word already established in each file.


wighawag:

Merging. Small task, done exactly, and it found the better version of its own brief.

## Criteria

1. **All three stale reap claims corrected** ✅ — `dropSuperseded`'s JSDoc, `dropReplaced`'s JSDoc, and the reap clause inside `wouldStrandAFollower`'s JSDoc. That third site was added to the task after review and is the one that mattered most, because there the retired reap was doing real work as the *justification for the predicate*. It now names the missing appender as the hazard "and only that", which keeps the argument while dropping the false half.

2. **The dead branches and the dead local are gone** ✅ — verified rather than assumed. `reaped` now appears exactly once in the whole file, in new prose saying the stream is **not** reaped. The unreachable ternaries at the old 1458 and 1611 are gone with it.

3. **No behaviour change** ✅ — verified by filtering the diff down to non-comment, non-string lines. The complete set of logic changes is:

   ```
   - const deletion = await this.registry.deleteGeneration(superseded.record);
   + await this.registry.deleteGeneration(superseded.record);
   - let reaped: string | undefined;
   - reaped = (await this.registry.deleteGeneration(record)).reaped;
   + await this.registry.deleteGeneration(record);
   ```

   Two unused bindings dropped. Same calls, same arguments, same order. The operator-asked reap paths are untouched.

4. **`CONTEXT.md` correctly left alone** ✅ — the task told the builder it was already right and to confirm rather than re-edit; the diff touches three files and `CONTEXT.md` is not one of them.

5. **Tests** ✅ — nothing asserts on the removed strings, which the task asked to be checked rather than guessed.

6. **Changeset** ✅ — present, as the task stated flatly rather than leaving it to be discovered at the gate.

## Better than asked

The brief was to stop documenting a reap that cannot happen. The obvious reading is "delete the false clause", which would have left the logs silent about the stream. Instead both messages now make a **positive** statement — `the stream <digest> is KEPT: a stream outlives every fold over it and is deleted only when asked` — so an operator reading a drop in the logs learns the invariant rather than merely not being misled. That is strictly more useful than the silence the task would have accepted.

It also caught an adjacent staleness in the same JSDoc that the task did not name: the claim that the filter-change case is served because "the retired generation owns a whole stream of its own that goes with it". The stream no longer goes with it either, so that is now "is alone on its stream and dropping it strands nothing" — same conclusion, true premise.

`verify` green, including the `docs:build` and `build:examples` steps added to the gate earlier in this drive.

PR #185, opened 2026-09-20
feat(the-guide-s-add-generation-recipe-gives-the-new-fold-its-own-store…

wighawag:
Task: `work/tasks/done/the-guide-s-add-generation-recipe-gives-the-new-fold-its-own-store.md`

Work is done. Report follows.

## What I did

**First, reality check.** The defect is still present: `docs/guide/indexing-in-a-browser-app/index.md` (the `addGeneration` snippet under "The same edit, without the blank app") still passed `createBrowserStateStore(next.entities)` with no `databaseName` and returned the unclaimed store. Confirmed against `packages/state-store-indexeddb/src/store.ts:61` (`DEFAULT_DATABASE_NAME = 'etherfold-state'`, the database the live generation folds into) and `WritableStateStore` in `packages/state-store/src/store.ts:344` (a `token` only `openForWriting` can produce).

**The fix** (`docs/guide/indexing-in-a-browser-app/index.md`): the snippet now names the store per generation with the hot-update example's own scheme (`app-${context.stream}-${++saves}`) and wraps it in `openForWriting`, and a short paragraph under it says why (shared `databaseName` is one store, collision on rows and cursor, writer-claim demotion, ADR-0075/ADR-0077; and `createState` hands back a claimed store because folding is writing). The page now says one thing in page order: the `addGeneration` recipe and the `reconfigureFromHotUpdate` example below it are the same two lines.

**The sweep (criterion 4).** Checked every `createState` / `createBrowserStateStore` snippet on the guide, in the package READMEs, in the root README, in `examples/`, and the ADR mentions:

- Guide, "Wiring it" (snapshot boot): had an explicit `databaseName` but returned `openAndBootstrap`'s `SnapshotAwareStateStore` straight out of `createState`. That is not a `WritableStateStore` (no `token`), so the snippet did not typecheck and taught an unclaimed fold. **Corrected**: `return openForWriting(store, {signal})`, with `{signal}` taken as the patience argument and a comment that it bounds the claim and not the download (matching the `createState` JSDoc and `packages/browser/test/snapshotOnlyMode.test.ts:253`).
- Guide, `reconfigureFromHotUpdate`: already correct. Untouched.
- `packages/browser/README.md` (worker host, main-thread hook, hot update): all three already `openForWriting` + explicit `databaseName`. Correct; untouched.
- `packages/processor-entities/README.md` (lines 30, 82) and `packages/state-store-indexeddb/README.md` (24-25): store-construction/backend-selection demos, not generation factories, and the `openAndBootstrap` result there is not handed to a generation, so no claim is owed at those call sites. Correct as written.
- Root `README.md`: `createState: () => createBrowserStateStore(greetings.entities, {databaseName: 'greetings'})` had defect two (no claim). **Corrected** (see Decisions).
- `examples/browser-reference` and `examples/event-processor-nfts`: real compiled code, already `openForWriting` over a `databaseName` keyed on `context.stream`. Correct.
- ADR-0024 / 0043 / 0079 mention `createBrowserStateStore` in prose only, not as copyable recipes.

**Conventions (criterion 7).** `CONTEXT.md` `## Conventions` requires a changeset for "changes that affect a published package's public API". This change touches no published package at all: the diff is `docs/`, the root `README.md` (the root package is `private: true`, so changesets ignores it) and `work/notes/`. So no changeset, and `pnpm changeset status --since=main` has no changed published package to complain about. This is also why nothing under `packages/` was edited: any file there, README included, would make `changeset status` demand one.

**Checks run locally.** `prettier --check .` on the whole repo (green, via a prettier 3.8.3 installed outside the repo, since `node_modules/` is absent here), `check:adr`, `check:refs`, `check:changesets`, `check:graph` (all green). I did not run `pnpm docs:build`: it needs a full monorepo install plus typedoc. The change is markdown only, and the two ADR links I added use the exact relative form already used twice in the same file (lines 208 and 220), so there is no new link target.

**Working tree**: only the four intended files, no scratch.

## Drift in the task body (reported, not a blocker)

The task says the other half of the source observation, "the browser package README's account of processor identity, is already corrected -- confirm that and do not re-edit it." **It is not corrected.** `packages/browser/README.md:136` still ends "Leave it off and the generation keeps the author-declared identity the processor computes from its `version`, exactly as before", and contradicts itself at line 233, which says an author-declared identity is what ADR-0086 deletes. I did not fix it (out of this task's acceptance criteria, and any edit under `packages/` would force a changeset that criterion 7 forbids). I appended a dated line to `work/notes/observations/two-doc-sites-still-describe-the-deleted-declared-identity-or-a-shared-generation-store.md` so that record is accurate for whoever picks it up. This did not block the build: the premise my work rests on (the guide snippet is broken) was verified true independently.

I also left a new observation, `work/notes/observations/the-seam-calls-one-store-instance-for-every-generation-the-shipped-pattern.md`: the state-store seam docs call `createState: () => store` "the shipped generation pattern", which a reader can take as licence for exactly the collision the guide now warns against, and the only place that does it is a benchmark harness.

## Decisions

**1. Reused the hot-update example's naming scheme verbatim instead of designing one.** The corrected snippet uses `app-${context.stream}-${++saves}` with `let saves = 0`, character for character the scheme the `reconfigureFromHotUpdate` example 40 lines below already uses, so the page states one convention twice rather than two. Alternatives considered: a fixed suffix (`app-${context.stream}-next`), rejected because a replaced successor's database can still be open when the next one is built, which is precisely why the neighbouring example uses a counter; and deriving the name from the processor identity, rejected because that identity does not exist when `createState` runs (ADR-0043's state-then-fold order). Touches: nothing outside this page; the prose now names `context.stream` and the counter as the two halves of the name, so the reason travels with the scheme.

**2. Forwarded the claim `{signal}` in the snapshot-boot factory but not in the `addGeneration` snippet.** The boot factory is an app's startup path and both package-README boot factories forward it, so it gets `openForWriting(store, {signal})`. The `addGeneration` snippet deliberately does not, so that it is identical to the hot-update example it sits above; criterion 2 asks those two to agree, and adding the signal to one of them would be the page teaching two shapes again. The `createState` JSDoc calls forwarding a convention rather than a guarantee, so the shorter form is incomplete, not wrong. Alternative: add `{signal}` to both, which would have meant editing the already-correct hot-update example. Touches: the hot-update example, which is why I left it alone.

**3. Corrected the root `README.md` snippet, which is outside criterion 4's literal list.** Criterion 4 names "the guide and the package READMEs"; the root README is neither, but it carries the identical defect (`createState` returning an unclaimed store), it is the project's front page, and it is the first snippet a new reader copies. Alternative considered: state its incorrectness and leave it, rejected because the criterion exists precisely because nothing in the gate catches an uncompiled snippet. It costs no changeset (the root package is private). Touches: nothing else; the wording is copied from `packages/browser/README.md`'s equivalent factory so the two agree.

**4. Corrected the guide's snapshot-boot factory as part of the sweep rather than only reporting it.** It is a second instance of defect two, found by the sweep, and criterion 4 permits correction. It is a slightly larger edit than the sweep's "check" verb implies (it adds the patience parameter and an import), so it is recorded here rather than buried. Touches: nothing outside the page.


wighawag:

Merging. It fixed the recipe, swept the neighbours as asked, and — best of all — **caught my own task body stating something false and refused to repeat it.**

## Criteria

1. **The recipe gives the new fold its own store** ✅ — `databaseName: \`app-${context.stream}-${++saves}\`` wrapped in `openForWriting`. Both defects closed: the collision and the unclaimed store.

2. **The page agrees with itself and with the contract** ✅ — the snippet now matches the `reconfigureFromHotUpdate` example forty lines below, and the guide says so explicitly ("The hot-update recipe below is these same two lines, driven by your bundler's saves"). No third convention invented, which was the stated way to get this wrong.

3. **The prose says WHY** ✅ — and says it well: two generations sharing a `databaseName` are one store by IndexedDB's own definition, so they collide on the rows and the sync cursor until the writer claim demotes one, with ADR-0075/0077 cited. It also explains the naming scheme rather than leaving it to be copied superstitiously: `context.stream` keeps two apps on one origin apart, the counter keeps this fold apart from the one it replaces.

4. **Neighbours swept** ✅ — found and fixed a second instance in the root `README.md`'s `openAndBootstrap` snippet, which had the same missing-claim defect.

5. **Fence held** ✅ — verified: no file under `packages/*/src`. Docs and READMEs only.

6–7. **`docs:build` green; no changeset** ✅ — correct for documentation-only.

**I checked the import myself**, since a fix to a broken snippet that is itself broken would be the worst outcome here. `openForWriting` really is re-exported from `@etherfold/processor-entities` (`src/index.ts`), and deliberately: "an app that opens one beside `openAndBootstrap` should not need a second import to say which of the two it is doing." The snippet is copyable as written.

## It contradicted the task body, and it was right

My task said the other half of the `two-doc-sites-…` observation — the browser README's account of processor identity — "is already corrected; confirm that and do not re-edit it." I took that from a review that had verified it by reading. **It is not corrected.** `packages/browser/README.md:136` still ends: "Leave it off and the generation keeps the author-declared identity the processor computes from its `version`, exactly as before" — while line 233 of the same file says "a value nobody can check against the code it claims to name is the author-declared identity ADR-0086 deletes." A direct self-contradiction, 97 lines apart.

The builder checked the claim instead of taking it, found it false, appended the finding to the observation rather than rewriting it (append-only, correctly), and did **not** quietly widen its scope to fix it. That is the right move on all four counts, and it is the seventh time in this family a builder has been right to contradict its own task text. The remaining half stays open and I am tasking it separately.

## The new observation is a good catch

`the-seam-calls-one-store-instance-for-every-generation-the-shipped-pattern` — `packages/state-store`'s `openForWriting` docstring and two READMEs call `createState: () => store`, one instance shared by every generation, "the shipped generation pattern", while the browser docs and now the guide say a generation beside the live one needs its OWN `databaseName`. The observation is careful about what is actually wrong: **both statements are true** (the seam's point is only that a second claim on one instance is not a second claim), but a reader who meets "the shipped pattern" first has been told the opposite of what the side-by-side case needs. And it notes the phrase names a pattern **nothing ships** — the only use in the tree is a benchmark harness. Captured, correctly fenced, not chased.

PR #186, opened 2026-09-20
feat(the-browser-readme-stops-promising-an-author-declared-identity-adr…

wighawag:
Task: `work/tasks/done/the-browser-readme-stops-promising-an-author-declared-identity-adr-0086-deleted.md`

Done. The task's premise held: I verified the false sentence was still present, verbatim, before changing anything.

## What I did

**Reality check first.** `packages/browser/README.md:136` still ended "Leave it off and the generation keeps the author-declared identity the processor computes from its `version`, exactly as before." Not already fixed, so the task was live.

**`packages/browser/README.md`** — replaced the false clause with the true answer rather than deleting it. It now says the identity is still DERIVED, that a fold arriving as a MODULE is named by a digest of its HANDLER SOURCES, that the edit moves it and a no-op save does not (`{stateDiscarded: false}`), that there is no declared `version` to fall back on, and that an unnameable processor is REFUSED. Borrowed from the two paragraphs that were already correct (the `updateProcessor` bullet at 198, the derivation paragraph at 233); all three now agree.

**The `version` question, checked in code, not assumed.** `getVersionHash()` and the `version` field are gone from `packages/*/src` (only historical mentions survive in docstrings). Nothing reads a processor `version` for any purpose, so I said that plainly rather than dropping or softening the word.

**Other doc surfaces checked.** `docs/guide`, `docs/design`, `docs/index.md`: clean. `packages/cli/README.md:150` already correct. One more site carried the same retired rule and I corrected it: `packages/utils/README.md:68` claimed an arrival with no `identity` left "the author's declared one" naming the fold; such a deployment is in fact refused (`refuseUnbundledProcessor` at config resolution, `requireArrivalIdentity` as backstop).

**No source change.** Nothing under `packages/*/src` touched.

**Gate:** `format:check`, `check:adr`, `check:refs`, `check:graph`, `check:changesets`, `docs:build` all pass; `changeset status --since=main` passes once the changeset is tracked (verified with a throwaway index; the repo's real index was not touched). `pendingChangesets.test.ts` green (188/188).

## Decisions

**A changeset IS required, contrary to the task's acceptance criterion.** Criterion 8 asserted "No changeset" but told me to confirm rather than assume. I confirmed, and it is false: `pnpm changeset status --since=main` is in the `verify` gate and fails with "Some packages have been changed but no changesets were found" for a README-only change, because changesets attributes any file under a package directory to that package. Shipping without one would have bounced the gate at land time. I added `.changeset/the-browser-readme-stops-promising-an-author-declared-identity.md` as `patch` on `@etherfold/browser` and `@etherfold/utils`. Alternative considered and rejected: `changeset add --empty`, which `packages/core/test/pendingChangesets.test.ts` explicitly fails ("names no package, so releasing it would discard it"). Precedent agrees: `the-build-command-and-its-pinning-rule-are-documented` was also documentation-only, carried a real patch changeset on `etherfold`, and closed with "Documentation only; no published behaviour changes" — the line I reused. This touches nothing but the release notes; a README ships in the npm tarball, so a patch is honest. Whoever wrote criterion 8 should know the repo convention is the opposite of what it assumed.

**Corrected a third doc site (`packages/utils/README.md`) rather than only reporting it.** Criterion 5 says other package READMEs are "corrected or [their] absence stated after checking", so this is inside the fence, but it is a second package and therefore a second changeset entry. Alternative considered: capture it as an observation and leave it, which I rejected because the criterion authorises the correction and because leaving one of two instances of the same retired rule is what produced this task in the first place.

**Appended a closing line to the observation note** (`work/notes/observations/two-doc-sites-still-describe-the-deleted-declared-identity-or-a-shared-generation-store.md`) recording that both halves are now closed and that a third site turned up. The note currently ends "this half is still open", and a stale open-state line is precisely the drift that caused this task's predecessor to report a fix that had not happened. Observations are the append-only bucket anyone may add to, and I did not move or re-status the file. Alternative: leave it to the done record, rejected because the next agent reads the note, not the done record.


wighawag:

Merging. This closes the last open item of the drive, and it did the part that is easy to skip.

## Criteria

1. **The false sentence is gone** ✅ — no "author-declared identity … computes from its `version`", no "exactly as before".

2. **Replaced with the true answer, positively stated** ✅ — this was the stated way to get the task wrong (delete the clause, answer nothing, let the reader guess the retired rule). Instead: the identity is still DERIVED, a module-delivered fold is named by a digest of its handler sources, an edit moves it and a no-op save does not and is answered `{stateDiscarded: false}`.

3. **The file agrees with itself** ✅ — the `processorIdentity` paragraph, the `updateProcessor` bullet and the derivation paragraph now say one thing.

4. **The `version` question was actually investigated** ✅ — and this is the criterion most likely to have been waved through. The answer is stronger than "it is not used": ADR-0086 deleted **the field itself**, along with the `getVersionHash()` that composed it with the entity and config declarations, so nothing in the tree reads a processor `version` for anything. **I verified it** — `getVersionHash` survives only in two historical JSDoc references in `packages/core/src/internal/`, both describing what used to happen. Stating that explicitly is what stops the next reader wondering whether `version` was merely forgotten.

5. **The sweep found a third site** ✅ — `packages/utils/README.md` said an arrival with no `identity` left "the author's declared one" naming the fold. It does not: such a deployment is REFUSED (`refuseUnbundledProcessor`, with `requireArrivalIdentity` as the structural backstop). That was a second live falsehood, found only because the criterion asked for the neighbours to be checked rather than assumed clean.

6. **Fence held** ✅ — verified: nothing under `packages/*/src`. READMEs and `work/` only.

7. **`docs:build` green** ✅ — through the gate that now includes it.

8. **Changeset** ✅ — and the builder was right to overrule my task text here. I wrote "No changeset: nothing shipped in a package changes", and asked for it to be confirmed rather than assumed. It checked, and a README **is** shipped with its npm package, so `@etherfold/browser` and `@etherfold/utils` take a patch. Correct, and the reason it is correct is exactly why the criterion said to confirm.

## The bonus the task did not ask for

The new paragraph also names the refusal path: a processor whose handlers have no readable source — all `bind`-ed or behind a proxy — has nothing to derive from and is refused before anything is registered, with both ways out in the message. That is the question a reader hits immediately after learning the identity is derived from source text, and it was previously unanswered anywhere on that page.

## Ledger

`two-doc-sites-still-describe-the-deleted-declared-identity-or-a-shared-generation-store` is now **fully closed**, with the closure appended rather than the note rewritten. Its history is worth keeping: the guide half was fixed in #185, this half was twice reported as already-fixed and twice was not, and the third site was never in the original note at all.

